PR #214, opened 2014-09-10
Add read() method to ReadableByteStream

tyoshino:


yutakahirano:
read() is not added to the IDL in BinaryExtension.md.

tyoshino:
> read() is not added to the IDL in BinaryExtension.md.

Created PR #273

Issue #258, opened 2015-01-06
[Reader] Should resolve readyPromise on getReader() call?


tyoshino:
@yutakahirano pointed out that `readableStream.getReader()` must resolve the current promise stored in `[[readyPromise]]` and replace it with a new pending promise.

After `getReader()` call, the `ready` getter starts masking (starts returning `[[lockReleased]]`). But if `[[readyPromise]]` is already obtained like the following example, it stays to be working as `[[readyPromise]]` (i.e. receives and delivers resolution signal to `[[readyPromise]]`).

```
readableStream.ready.then( ... );
reader = readableStream.getReader();
```

It should be ok even if it gets resolved while the reader is active as we just see `readableStream.state` being `"waiting"` and get `[[lockReleased]]`.

If it doesn't get resolved while the reader is active, then, we cannot get notified of release of the lock. But I think in such a case, one who was watching the stream is not interested in release of lock but transition to readable/closed. So, this case is also not a problem?

Do you think this needs to be taken care of? @yutakahirano's suggestion was to resolve it on `getReader()` call.

domenic:
Gah, I thought I had a test case for this. I guess I missed it though.

Right now the contract for `ready` is that it only fulfills if the state is no longer "waiting". I would prefer to keep that contract if possible. So I don't think resolving it on the `getReader()` call is very good.

I think this might be fixable via something like `return this._readyPromise.then(() => this._reader !== undefined ? this._reader.ready : undefined)` but last time I tried something like that I got stuck in an infinite ping-pong loop between the two `ready`s. I think it is still doable I just need to be careful.

> If it doesn't get resolved while the reader is active, then, we cannot get notified of release of the lock. But I think in such a case, one who was watching the stream is not interested in release of lock but transition to readable/closed. So, this case is also not a problem?

Yes, that is my feeling as well, since people asking for `ready` are really interested in transition to readable/closed/errored (i.e. anything but waiting).

yutakahirano:
> Right now the contract for ready is that it only fulfills if the state is no longer "waiting". I would prefer to keep that contract if possible. So I don't think resolving it on the getReader() call is very good.

IIUC we already have waiting => waiting transition.

```
var stream = ...;
var reader = stream.getReader();
stream.ready.then(() => console.log('state becomes ' + stream.state));
reader.releaseLock();
```

The above code prints 'state becomes waiting', right?

domenic:
[...]

PR #262, opened 2015-01-14
Reader fixups

### LINKED ISSUE #260: Make stream readers look closed when released
In https://github.com/whatwg/streams/issues/241#issuecomment-68834017 @tyoshino proposed making a stream reader look "closed" once it is released. Right now it just passes through `ready`/`state`/`closed` to the underlying stream, although `read()` and `cancel()` fail.

There's no immediately-convincing argument for one over the other, but after some thought I think the "closed" is better. It means that more invariants that apply to streams (e.g., if state is "readable", you can call read() successfully) will apply to readers. And it allows you to pass the reader object to someone without them having to think about whether it's been released or not, as long as they are equipped to handle closed streams.

domenic:
Ready for review @tyoshino @yutakahirano. Fixes #260 and both waiting -> waiting transitions discovered in #258.

Preview branch at https://streams.spec.whatwg.org/branch-snapshots/reader-fixups/

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -561,7 +561,11 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
   <li> If <b>this</b>@\[[reader]] is not <b>undefined</b>, return the result of transforming
     <b>this</b>@\[[reader]]@\[[lockReleased]] by a fulfillment handler that returns Get(<b>this</b>,
     <code>"ready"</code>).
-  <li> Return <b>this</b>@\[[readyPromise]].
+  <li> Return the result of transforming <b>this</b>@\[[readyPromise]] by the following fulfillment handler:
+    <ol>
+      <li> If <b>this</b>@\[[reader]] is <b>undefined</b>, return <b>undefined</b>.
+      <li> Otherwise, return <b>this</b>@\[[reader]]@\[[lockReleased]].
+    </ol>
 </ol>
 
 <h5 id="rs-state">get state</h5>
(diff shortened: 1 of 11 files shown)


index.bs
```diff
@@ -558,8 +558,13 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
 </div>
 
 <ol>
-  <li> If <b>this</b>@\[[reader]] is not <b>undefined</b>, return <b>this</b>@\[[reader]]@\[[lockReleased]].
-  <li> Return <b>this</b>@\[[readyPromise]].
+  <li> If <b>this</b>@\[[readableStreamReader]] is not <b>undefined</b>, return the result of transforming
+    <b>this</b>@\[[readableStreamReader]]@\[[lockReleased]] by a fulfillment handler that returns <b>this</b>@\[[readyPromise]].
```
yutakahirano: Shouldn't this@[[readyPromise]] be Get(this, "ready")?
The current rule exposes the resolution that should be hidden in the following sequence.

1. The stream is locked to a reader.
1. `.ready` is accessed and some handlers are attached on it.
1. The reader is released.
1. The stream is locked to another reader.
1. Data is enqueued and stream@[[readyPromise]] is resolved.

index.bs
```diff
@@ -558,8 +558,13 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
...
+  <li> Return the result of transforming <b>this</b>@\[[readyPromise]] by the following fulfillment handler:
+    <ol>
+      <li> If <b>this</b>@\[[readableStreamReader]] is <b>undefined</b>, return <b>undefined</b>.
+      <li> Otherwise, return <b>this</b>@\[[readableStreamReader]]@\[[lockReleased]].
+    </ol>
```
yutakahirano: ditto

index.bs
domenic: Thank you for catching that!!! You are totally right.

index.bs
domenic: I think this is the base case, and so should directly use [[readyPromise]]?

Regardless, I'll be able to use your scenario to write a test case and prove one way or the other.

yutakahirano: Oh, I see. I thought that introduces waiting -> waiting transition, but it is a harmless one, because some data are enqueued and a reader pulls all of them. Once I talked with @tyoshino about it and agreed we cannot avoid that kind of waiting -> waiting transition.

Thank you!

```
// stream is not locked.
stream.ready.then(() => console.log(stream.state));
reader = stream.getReader();
...
// Some items are enqueued, and reader pulls all of them.
...
reader.releaseLock();
```

domenic: Ah, I see. I think this is analogous to the case without any reader involved:

```js
stream.ready.then(() => console.log(stream.state));

// enqueue a chunk
stream.read();

// will log "waiting"
```

This is somewhat of a more general problem. (If it is a problem.) I will open a new bug to discuss it, although it might be fine to leave as-is.

domenic:
Updated :)

yutakahirano:
lgtm

index.bs
tyoshino: #268 may improve this point a little. Please take a look.

PR #265, opened 2015-01-17
First class source sink objects

### LINKED ISSUE #256: Allow object with methods to be passed as (writeable) sink
Sorry if something similar was already discussed, I searched in the issues and couldn't find anything.

One pattern node uses[1] is to attach internal fields (e.g. pointers to native structures) to objects. When methods are called on instances later on, their native implementations can access these internal fields. The problem is now that these constructs only work if the methods are actually called on the instances, e.g.:

```js
var TTYSink = getNativeBinding('TTY'); // returns a class created via a FunctionTemplate
var sink = new TTYSink(1); // tty stream for stdout
sink.write("Hello World").then(() => { /* written */ }); // will directly call the native implementation of `write`
```

It would be great if it would be possible to pass such an object into the `Writeable` constructor. But unfortunately right now this will fail because `Writeable` will try to call `write` as a function instead of as a method, which forces ugly hacks like binding all methods explicitly:

```js
var stdout = new Writeable({
  /* ... */
  close: sink.close.bind(sink),
  write: sink.write.bind(sink)
});
```

By preserving the this-context, the following would "just work":

```js
var stdout = new Writeable(new TTYSink(1));
```

I'm not sure if this is a V8-specific concern or if there's a better way of solving this when implementing the sink (e.g. if V8 already has an option to attach user data to all methods of an object). Still wanted to bring it up for discussion. :)

[1] According to v8-users, this is a common pattern: https://groups.google.com/d/msg/v8-users/Axf4hF_RfZo/hA6Mvo78AqAJ

domenic:
Fixes #256 (/cc @jkrems).

Could use some careful review as the changes were pretty wide-ranging. I am not too worried about the algorithms---the reference implementation and its extensive test suite help with my confidence there---but the editorial changes always leave the potential for pointers, "dead code", etc.

Preview at https://streams.spec.whatwg.org/branch-snapshots/first-class-source-sink-objects/

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -20,7 +20,7 @@ Logo: https://resources.whatwg.org/logo-streams.svg
 
 <style>
   ol > li { margin: 0; }
-  .note + .example { margin-top: 1em; }
+  .note + .example, .note + .note { margin-top: 1em; }
 </style>
 <script src="https://resources.whatwg.org/file-issue.js" async></script>
 
@@ -1201,13 +1201,7 @@ would look like
 
 <pre><code class="lang-javascript">
   class WritableStream {
-    constructor({
-      start = (error) => {},
-      write = (chunk) => {},
-      close = () => {},
-      abort = (reason) => close(),
-      strategy = %DefaultWritableStreamStrategy%
-    } = {})
+    constructor(underlyingSink = {})
 
     get closed()
     get ready()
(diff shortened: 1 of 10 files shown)


index.bs
```diff
@@ -1308,52 +1274,47 @@ Instances of <code>WritableStream</code> are created with the internal slots des
 <h4 id="ws-constructor">new WritableStream({ start, write, close, abort, strategy } = {})</h4>
```
yutakahirano: Should the title be also modified, for the symmetry with ReadableStream?

domenic: Yes, good catch, thank you! This actually has normative consequences, as we no longer do destructuring into functions, but instead preserve the whole object.

Issue #185, opened 2014-08-28
Promises based transform()


tyoshino:
Branched from https://github.com/whatwg/streams/pull/180/files#r16783645

tyoshino

> Do we want to also change transform() to use Promises to notify completion?

domenic

> No, because transform() is more like ReadableStream: you can enqueue multiple chunks before calling done().

> Although in theory, for both ReadableStream and for Transform, we could use a combination of enqueue + promises. I think it would be more awkward though? Unsure, might write up some sample code.

domenic:
Consider the example at https://streams.spec.whatwg.org/#example-rs-pull, in particular:

```js
    pull(enqueue, close, error) {
      const buffer = new ArrayBuffer(CHUNK_SIZE);

      fs.read(fd, buffer, 0, CHUNK_SIZE, position).then(bytesRead => {
        if (bytesRead === 0) {
          return fs.close(fd).then(close);
        } else {
          position += bytesRead;
          enqueue(buffer);
        }
      })
      .catch(error);
    },
```

with promises, this would become

```js
    pull(enqueue, close) {
      const buffer = new ArrayBuffer(CHUNK_SIZE);

      return fs.read(fd, buffer, 0, CHUNK_SIZE, position).then(bytesRead => {
        if (bytesRead === 0) {
          return fs.close(fd).then(close);
        } else {
          position += bytesRead;
          enqueue(buffer);
        }
      });
    },
```

That is, it allows you to return a promise for error propagation. That's kind of nice. Not revolutionary. And I think it introduces a one-turn delay?

However, consider the case of a simple duplicater transform:

```js
transform(chunk, enqueue, done, close, error) {
  enqueue(chunk);
  enqueue(chunk);
  done();
}
```

could become

```js
transform(chunk, enqueue, close) {
  enqueue(chunk);
  enqueue(chunk);
}
```

which is a lot nicer. (But this time you definitely get a one-turn delay before we can call `transform` again.) Async transforms too:

```js
transform(chunk, enqueue, done, close, error) {
  getAsyncData(chunk).then(d => { enqueue(d); done(); }, error);
}
```

becomes

```js
transform(chunk, enqueue, close) {
  return getAsyncData(chunk).then(enqueue);
}
```

One thing that has me confused is that the current transform implementation doesn't pass `close` and `error` params like I do above. Perhaps because we expect the creator to just call `transform.writable.close()` and `transform.writable.abort(er)`? That seems lame. Maybe it was just an oversight.

---

[...]

PR #272, opened 2015-01-27
Convert readable stream pull() to promise-returning

domenic:
See discussion in #185 for background.

This allows more convenient integration with promise-returning underlying source functions, removing the need for `.catch(error)` as seen in the modified example.

It also rationalizes the slightly-strange way in which pull() is mutexed based on whether enqueue() has been called: instead, it is now mutexed so as to not be called again until the previous iteration's promise has fulfilled. This gives more direct control to the underlying source, and allows them to deal more easily with cases where there are just no chunks to enqueue.

Although superficially it seems this may reduce performance for cases where underlying sources can enqueue many chunks synchronously, this is not actually the case, as can be seen from the modified test illustrating such a scenario. If an underlying source can enqueue multiple chunks synchronously, then it should just do so! It shouldn't wait to be pull()ed, as the previous test was doing.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -275,8 +275,9 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
       <code>"errored"</code> state
   </tr>
   <tr>
-    <td>\[[pulling]]
-    <td>A boolean flag indicating whether data is currently being pulled from the underlying sink
+    <td>\[[pullingPromise]]
+    <td>A promise returned by the <a>underlying source</a>'s <code>pull</code> method, stored so that the stream can
+      re-pull when it fulfills
   </tr>
   <tr>
     <td>\[[queue]]
@@ -320,9 +321,10 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
     <li> <code>start(enqueue, close, error)</code> is called immediately, and is typically used to adapt a <a>push
       source</a> by setting up relevant event listeners, or to acquire access to a <a>pull source</a>. If this process
       is asynchronous, it can return a promise to signal success or failure.
-    <li> <code>pull(enqueue, close, error)</code> is called when the stream's internal queue of chunks is depleted, and
-      the consumer has signaled that they wish to consume more data. Once it is called, it will not be called again
-      until the passed <code>enqueue</code> callback is called.
+    <li> <code>pull(enqueue, close)</code> is called when the stream's internal queue of chunks is depleted, and the
+      consumer has signaled that they wish to consume more data. If <code>pull</code> returns a promise, then it will
+      not be called again until the promise returned by the previous call has fulfilled; if the promise rejects, the
+      stream will become errored.
     <li> <code>cancel(reason)</code> is called when the consumer signals that they are no longer interested in the
       stream. It should perform any actions necessary to release access to the <a>underlying source</a>. If this
       process is asynchronous, it can return a promise to signal success or failure.
@@ -344,7 +346,7 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
   <li> Set <b>this</b>@\[[readyPromise]] and <b>this</b>@\[[closedPromise]] to new promises.
   <li> Set <b>this</b>@\[[queue]] to a new empty List.
   <li> Set <b>this</b>@\[[state]] to <code>"waiting"</code>.
-  <li> Set <b>this</b>@\[[started]], <b>this</b>@\[[draining]], and <b>this</b>@\[[pulling]] to <b>false</b>.
+  <li> Set <b>this</b>@\[[started]] and <b>this</b>@\[[draining]] to <b>false</b>.
   <li> Set <b>this</b>@\[[readableStreamReader]] to <b>undefined</b>.
   <li> Set <b>this</b>@\[[enqueue]] to CreateReadableStreamEnqueueFunction(<b>this</b>).
   <li> Set <b>this</b>@\[[close]] to CreateReadableStreamCloseFunction(<b>this</b>).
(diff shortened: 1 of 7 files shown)


index.bs
```diff
@@ -779,20 +781,24 @@ The <code>length</code> property of the <code>cancel</code> method is <b>1</b>.
...
-  <li> If <var>pullResult</var> is an abrupt completion,
+  <li> If <var>stream</var>@\[[draining]] is <b>true</b> or <var>stream</var>@\[[started]] is <b>false</b> or
+    <var>stream</var>@\[[state]] is <code>"closed"</code> or <var>stream</var>@\[[state]] is <code>"errored"</code>,
+    return <b>undefined</b>.
+  <li> If <var>stream</var>@\[[pullingPromise]] is not <b>undefined</b>,
     <ol>
-      <li> Call-with-rethrow Call(<var>stream</var>@\[[error]], <b>undefined</b>, «‍<var>pullResult</var>.\[[value]]»).
-      <li> Return <var>pullResult</var>.
+      <li> Upon fulfillment of <var>stream</var>@\[[pullingPromise]], call-with-rethrow
+        CallReadableStreamPull(<var>stream</var>).
```
tyoshino: I suggest a design like WritableStreamAdvanceQueue, i.e. set only one fulfillment callback on creation of pullingPromise, and in the callback, check whether CallReadableStreamPull was ever called since the last time we invoked underlyingSource.pull.

CallReadableStreamPull is invoked on any event that may cause shouldApplyBackpressure to be flipped. We shouldn't set a fulfillment callback every time when CallReadableStreamPull is called.

domenic:
@tyoshino updated to be more conservative about calling pull() per your feedback. PTAL.

index.bs
```diff
@@ -344,7 +351,7 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
   <li> Set <b>this</b>@\[[readyPromise]] and <b>this</b>@\[[closedPromise]] to new promises.
   <li> Set <b>this</b>@\[[queue]] to a new empty List.
   <li> Set <b>this</b>@\[[state]] to <code>"waiting"</code>.
-  <li> Set <b>this</b>@\[[started]], <b>this</b>@\[[draining]], and <b>this</b>@\[[pulling]] to <b>false</b>.
+  <li> Set <b>this</b>@\[[started]], <b>this</b>@\[[draining]], and <b>this</b>@\[[pulLScheduled]] to <b>false</b>.
```
tyoshino: L -> l

index.bs
```diff
@@ -781,20 +788,28 @@ The <code>length</code> property of the <code>cancel</code> method is <b>1</b>.
...
+    <var>stream</var>@\[[state]] is <code>"closed"</code> or <var>stream</var>@\[[state]] is <code>"errored"</code> or
+    <var>stream</var>@\[[pullScheduled]] is <b>true</b>, return <b>undefined</b>.
+  <li> If <var>stream</var>@\[[pullingPromise]] is not <b>undefined</b>,
     <ol>
-      <li> Call-with-rethrow Call(<var>stream</var>@\[[error]], <b>undefined</b>, «‍<var>pullResult</var>.\[[value]]»).
-      <li> Return <var>pullResult</var>.
+      <li> Set <var>stream</var>@\[[pullScheduled]] to <b>true</b>.
+      <li> Upon fulfillment of <var>stream</var>@\[[pullingPromise]],
+        <ol>
+          <li> Set <var>stream</var>@\[[pulLScheduled]] to <b>false</b>.
```
tyoshino: L -> l

reference-implementation/test/readable-stream.js
```diff
@@ -290,6 +289,96 @@ test('ReadableStream is able to pull data repeatedly if it\'s available synchron
...
+    t.equal(rs.state, 'readable', 'stream becomes readable (even before promise fulfills)');
+    t.equal(timesCalled, 1, 'pull is not yet called a second time');
+    t.equal(rs.read(), 1, 'read() returns enqueued value');
+
+    setTimeout(() => {
+      t.equal(timesCalled, 1, 'after 30 ms, pull has still only been called once');
+
+      resolve();
+
+      returnedPromise.then(() => {
```
tyoshino: looks a little misleading. can't we just use Promise.resolve(undefined).then(...) or setTimeout(..., 0)?

reference-implementation/test/readable-stream.js
```diff
@@ -290,6 +289,96 @@ test('ReadableStream is able to pull data repeatedly if it\'s available synchron
...
+    }
+  });
+
+  t.equal(rs.state, 'readable', 'since start() synchronously enqueued chunks, the stream is readable');
+
+  // Wait for start to finish
+  rs.ready.then(() => {
+    t.equal(rs.read(), 'a', 'first chunk should be as expected');
+    t.equal(rs.read(), 'b', 'second chunk should be as expected');
+    t.equal(rs.read(), 'c', 'second chunk should be as expected');
```
tyoshino: third

tyoshino:
580327931c95c50adbdfaaaffdf53064ee486b2d looks good.

Please think of this scenario:

1. `rs.read()` is called
1. `pull()` is called to refill the queue
1. `pull()` called `enqueue()` to fill up the queue and returned
1. `rs.read()` is called
1. `rs.read()` is called
1. `rs.read()` is called ...

Even if the `read()` calls drain the queue, `pull()` won't be called until the next microtask is run. Are you aware of this? I'm fine with this but want to confirm.

reference-implementation/test/readable-stream.js
domenic: I think it is accurate. After the `returnedPromise` fulfills, the stream calls `pull()` again. So we should also subscribe to `returnedPromise` to test if `pull()` was called.

domenic:
> Even if the read() calls drain the queue, pull() won't be called until the next microtask is run. Are you aware of this? I'm fine with this but want to confirm.

Yeah. I tried to address this with the commit message. If the underlying source has more chunks available synchronously, it should just enqueue them. Even if the backpressure signal says the queue is full, it should use that to stop any flow, but still enqueue any chunks it already has. Whereas, if the underlying source can't get any more chunks without an asynchronous delay, there will be no problem anyway. Make sense?

Going to merge but happy to make any updates if e.g. I am misunderstanding https://github.com/whatwg/streams/pull/272#discussion_r24150104

reference-implementation/test/readable-stream.js
tyoshino: Yes, it works. But we don't have to use `returnedPromise`. As we're fulfilling `returnedPromise` on L315, we're sure that `returnedPromise` is fulfilled at L317. So, L317 is just putting the closure into the microtask queue. We can also use Promise.resolve(undefined).then(...) there.

Here what we want to test is

> `pull()` will be called once the returnedPromise is resolved.

`Once ...` corresponds to L315. Yes, it's relevant to `returnedPromise`.

`pull() ...` is what we want to verify on L317. This is irrelevant to `returnedPromise`.

Use of `returnedPromise` at L317 (though it works) makes the verification step look depending on `returnedPromise`. So, it looked misleading to me.

domenic: I agree that Promise.resolve().then(...) would work. But don't you think a good way to verify "pull() will be called once the returnedPromise is resolved" is `returnedPromise.then(() => checkThatPullWasCalled())`?

One way of looking at it is that I could move the resolve() call on L315 somewhere else randomly (say, in a top-level `setTimeout(resolve, 100)`). L317 is meant to stand on its own independent of L315.

tyoshino: Hmm, I see. It's representing kinda story that we're observing relation between returnedPromise and pull() count and somebody else does resolve() at some point. OK. I understand.

tyoshino:
Looks good. Thanks!

> but still enqueue any chunks it already has.

OK

PR #274, opened 2015-01-28
Fixed cancel test on errored stream

calvaris:
I think the description of that test is confusing if you read what it is really checking, that the promise is rejected with the passed error.


domenic:
Thank you! Presumably the test was changed at some point but not the title.

domenic:
Merged as 87990202ca263171f78a5e8a931da93080809bea.

PR #277, opened 2015-02-04
Have each of the stream and reader hold promises

tyoshino:


reference-implementation/lib/readable-stream-abstract-ops.js
```diff
@@ -118,19 +182,11 @@ export function CreateReadableStreamEnqueueFunction(stream) {
 
 export function CreateReadableStreamErrorFunction(stream) {
   return e => {
-    if (stream._state === 'waiting') {
-      stream._resolveReadyPromise(undefined);
-    }
     if (stream._state === 'readable') {
       stream._queue = [];
```
domenic: Any reason not to move this to ErrorReadableStream?

reference-implementation/lib/readable-stream-abstract-ops.js
```diff
@@ -37,22 +37,87 @@ export function CallReadableStreamPull(stream) {
...
+    stream._resolveReadyPromise(undefined);
+  }
   stream._resolveClosedPromise(undefined);
 
+  stream._state = 'closed';
+
+  return undefined;
+}
+
+export function MarkReadableStreamReadable(stream) {
```
domenic: Nit: keep things alphabetical in this file to match the spec.

reference-implementation/lib/readable-stream.js
```diff
@@ -1,7 +1,8 @@
 var assert = require('assert');
 import * as helpers from './helpers';
-import { AcquireExclusiveStreamReader, CallReadableStreamPull, CloseReadableStream, CreateReadableStreamCloseFunction,
-  CreateReadableStreamEnqueueFunction, CreateReadableStreamErrorFunction, IsReadableStream, ReadFromReadableStream,
+import { AcquireExclusiveStreamReader, CallReadableStreamPull,
```
domenic: Nit: alphabetize and wrap to 120 chars

reference-implementation/lib/readable-stream-abstract-ops.js
```diff
@@ -37,22 +37,87 @@ export function CallReadableStreamPull(stream) {
...
+
 export function CloseReadableStream(stream) {
-  stream._state = 'closed';
+  if (stream._readableStreamReader !== undefined) {
+    if (stream._state === 'waiting') {
+      stream._readableStreamReader._resolveReadyPromise(undefined);
+    }
+
+    // rs.ready() was pending because there was a reader.
+    stream._resolveReadyPromise(undefined);
```
domenic: I think there is (or if there isn't, should be) some duplicate code here with `releaseLock()`. Might be good to extract them out into a single abstract op?

reference-implementation/test/readable-stream.js
```diff
@@ -122,26 +122,6 @@ test('ReadableStream reading a stream makes ready and closed return a promise fu
   );
 });
 
-test('ReadableStream avoids waiting-to-waiting transitions when immediately reading', t => {
```
domenic: A bit of a shame to lose this property, but I guess it was never terribly robust anyway (e.g. if multiple people were calling `read()` in their `ready` handler).

domenic:
This looks really excellent. Thank you for taking the time. Much better than the mess I was getting myself in to.

I left a few minor comments but overall am very happy with this. Do you want to do the spec portion or shall I?

reference-implementation/lib/readable-stream-abstract-ops.js
tyoshino: Good point. Moved everything to ErrorReadableStream.

BTW, I'm trying to have some border between the code that is specific to the ReadableStream subclass implementation and one that is common to ReadableStream and all of its subclasses. Do you have any ideas about that?

reference-implementation/lib/readable-stream.js
tyoshino: Done

reference-implementation/lib/readable-stream-abstract-ops.js
tyoshino: OK. Will do.

reference-implementation/lib/readable-stream-abstract-ops.js
tyoshino: Thanks. Done. And this revealed one bug in `releaseLock()` that is calling _resolveReadyPromise() only when the state is `"readable"`. Fixed.

tyoshino: NVM. It was not a bug...

reference-implementation/test/readable-stream.js
tyoshino: Yeah. We were determining the right `rs.ready` value to expose to the user looking at `rs` by looking at it in a fulfillment callback (which is run asynchronously to actual operation such as `reader.read()`. So, we could check the value while the stream is locked). This property is kinda side-effect of that approach, I guess.

tyoshino:
Done addressing Domenic's comments.

I also removed export from some functions in the -abstract-ops file which are not used by any other file. And, added some identity tests. May I copy the tests from your patch https://github.com/whatwg/streams/compare/cache-promises? I'm also fine if you could just commit them after mine.

> spec portion

I'll do but may not be able to finish today. I'll push wip one then. Please take over it.

tyoshino:
OK. Done updating the spec text. Ready for review.

reference-implementation/lib/exclusive-stream-reader.js
```diff
@@ -11,17 +12,17 @@ export default class ExclusiveStreamReader {
       throw new TypeError('This stream has already been locked for exclusive reading by another reader');
     }
 
-    stream._readableStreamReader = this;
+    AttachExclusiveStreamReader(stream, this);
```
domenic: Any reason not to inline this? It seems to be only used in one place. Same with detach. I don't feel strongly though so if you think it's clearer we can keep it.

reference-implementation/lib/exclusive-stream-reader.js
```diff
@@ -89,11 +79,11 @@ export default class ExclusiveStreamReader {
...
 
     if (this._encapsulatedReadableStream._readableStreamReader !== this) {
-      return this._closedAfterRelease;
+      return this._closedPromise;
     }
 
-    this.releaseLock();
-    return this._encapsulatedReadableStream.cancel(reason, ...args);
+    // Bypass lock check.
+    return CancelReadableStream(this._encapsulatedReadableStream, reason, ...args);
```
domenic: I think it is OK to stop passing any args now since it is not a direct delegation to a possibly-overridden `.cancel` method.

reference-implementation/lib/readable-stream-abstract-ops.js
```diff
@@ -117,22 +153,47 @@ export function CreateReadableStreamEnqueueFunction(stream) {
 }
 
 export function CreateReadableStreamErrorFunction(stream) {
-  return e => {
+  return e => ErrorReadableStream(stream, e);
```
domenic: I don't think this change is necessary. We can leave it as inline (both here and in the spec).

index.bs
```diff
@@ -797,16 +763,44 @@ The <code>length</code> property of the <code>cancel</code> method is <b>1</b>.
...
 
+<h4 id="cancel-readable-stream">CancelReadableStream ( stream )</h4>
+
+<ol>
+  <li> If <b>this</b>@\[[state]] is <code>"closed"</code> or <code>"errored"</code>, return
+    <b>this</b>@\[[closedPromise]].
+  <li> Let <b>this</b>@\[[queue]] be a new empty List.
+  <li> Call-with-rethrow CloseReadableStream(<var>stream</var>).
+  <li> Let <var>sourceCancelPromise</var> be PromiseInvokeOrNoop(<b>this</b>@\[[underlyingSource]],
+    <code>"cancel"</code>, (<var>reason</var>)).
```
domenic: It looks like I missed a spot in 14bbebcfa9f0e4d8776d33f7b26342124f6bec37; should be «‍_reason_» instead of (_reason_).

index.bs
```diff
@@ -605,18 +592,12 @@ Instances of <code>ExclusiveStreamReader</code> are created with the internal sl
...
   </tr>
   <tr>
-    <td>\[[lockReleased]]
-    <td>A promise that becomes fulfilled when the reader releases its lock on the stream
-  </tr>
-  <tr>
-    <td>\[[readyAfterRelease]]
-    <td>A promise returned by the reader's <code>ready</code> getter after the reader's lock is released, or
-      <b>undefined</b> before then
+    <td>\[[state]]
```
domenic: Alphabetize these ([[[state]] should come after [[readyPromise]]).

reference-implementation/lib/exclusive-stream-reader.js
domenic: This will effect: spec's class definition block; spec header; spec step 3; and CancelReadableStream definition. They should all just take a _reason_ argument.

domenic:
Mostly looks good. Just a few comments, and also need to rebase on master. And yeah feel free to pull the tests, no need to keep them in a separate commit.

reference-implementation/lib/exclusive-stream-reader.js
tyoshino: I gave it its own operation name for readability. But maybe just adding comments is fine. Reverted.

reference-implementation/lib/exclusive-stream-reader.js
tyoshino: I see. Removed.

reference-implementation/lib/readable-stream-abstract-ops.js
tyoshino: Reverted.

index.bs
tyoshino: Done. Do you prefer using a raw character than using a character reference?

tyoshino:
OK. Addressed the comments and merged the tests. Please check diff between 1fbebb5 b488805.

index.bs
```diff
@@ -446,14 +442,7 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
...
   <li> If <b>this</b>@\[[readableStreamReader]] is not <b>undefined</b>, return a new promise rejected with a <b>TypeError</b>.
-  <li> If <b>this</b>@\[[state]] is <code>"closed"</code> or <code>"errored"</code>, return
-    <b>this</b>@\[[closedPromise]].
-  <li> If <b>this</b>@\[[state]] is <code>"waiting"</code>, resolve <b>this</b>@\[[readyPromise]] with <b>undefined</b>.
-  <li> Let <b>this</b>@\[[queue]] be a new empty List.
-  <li> Call-with-rethrow CloseReadableStream(<b>this</b>).
-  <li> Let <var>sourceCancelPromise</var> be PromiseInvokeOrNoop(<b>this</b>@\[[underlyingSource]],
-    <code>"cancel"</code>, (<var>reason</var>)).
-  <li> Return the result of transforming <var>sourceCancelPromise</var> by a fulfillment handler that returns <b>undefined</b>.
+  <li> Call-with-rethrow CancelReadableStream(<b>this</b>, <var>reason</var>).
```
domenic: s/Call-with-rethrow/return

index.bs
```diff
@@ -812,16 +772,44 @@ The <code>length</code> property of the <code>cancel</code> method is <b>1</b>.
...
 
+<h4 id="cancel-readable-stream">CancelReadableStream ( stream )</h4>
+
+<ol>
+  <li> If <b>this</b>@\[[state]] is <code>"closed"</code> or <code>"errored"</code>, return
+    <b>this</b>@\[[closedPromise]].
+  <li> Let <b>this</b>@\[[queue]] be a new empty List.
+  <li> Call-with-rethrow CloseReadableStream(<var>stream</var>).
+  <li> Let <var>sourceCancelPromise</var> be PromiseInvokeOrNoop(<b>this</b>@\[[underlyingSource]],
+    <code>"cancel"</code>, &laquo;<var>reason</var>&raquo;).
```
domenic: I prefer using the Unicode characters directly

domenic:
Great! Fixed a few nits and merged as 9ec484b50d5fb949b0fe3e7fd1b11813be4b4cb7.

tyoshino:
Thanks for committing.

Issue #266, opened 2015-01-19
Requirement on promises equality


tyoshino:
In the spec, we remember and return the same promises for efficiency if it's semantically correct. The algorithms we have in the spec are representing this. This may look like a requirement to the spec readers. And, we have some unit test items where we actually check this kind of equality.

Do you think this really should be a requirement?

I'm asking this since:
- When implementing the Streams in Blink, we encountered some difficulties with ensuring some of the equality we can infer from the algorithm of the methods/getters.
- There seems to be no real practical needs for such a requirement.

domenic:
I think it should definitely be a requirement. It's very strange to have JavaScript properties changing all the time such that e.g. `rs.ready !== rs.ready`. They should only change when they need to.

tyoshino:
@domenic OK. So, it means that `rs.ready` should return the same instance until we're asked to replace `[[readyPromise]]` on e.g. `"readable"` to `"waiting"` transition?

How about the result of `cancel()` call on a `"closed"` or "`errored`" stream? It's now specified to return `[[closedPromise]]` than creating a new one resolved/rejected for efficiency. Do you think it also should be a requirement?

BTW, `writableStream.close()` can also be simplified to return its `[[closedPromise]]` when the stream is in the `"closed"` or `"errored"` state.

/cc @yutakahirano

domenic:
> So, it means that rs.ready should return the same instance until we're asked to replace [[readyPromise]] on e.g. "readable" to "waiting" transition?

Yeah, definitely.

> How about the result of cancel() call on a "closed" or "errored" stream? It's now specified to return [[closedPromise]] than creating a new one resolved/rejected for efficiency. Do you think it also should be a requirement?

We should make sure to normatively specify and test one way or another. But since it's a method instead of property I would be OK either way. I think the way it is now makes some sense but I'd be OK either way. We just need to make sure all implementations agree.

domenic:
@tyoshino @yhirano as implementers, any thoughts one way or the other, for cancel(), close(), abort(), or any other methods? I am OK specifying them as new promises in all cases if that seems desirable. It's a bit nicer if the methods don't share return values with other places in the system, but it is a minor efficiency loss, so I could go either way.

yutakahirano:
I'm concerned about the spec complexity. `.ready` and `.closed` changed in recent two weeks. Currently they are defined very cleanly, but can we keep the readability when we start caching?

I don't see any reason to keep promise identity on "cancel()" return value.

[...]

PR #279, opened 2015-02-10
Add readableAmount getter to ReadableByteStream

tyoshino:


tyoshino:
I'm making a PR to update the ReadableByteStream to address #266. I'll make ref impl change once it's done.

tyoshino:
#280 is the PR to update the ReadableByteStream to address #266.

domenic:
I wonder if this should be a getter or a method. Right now it is a method on the underlying source but a getter on the stream.

I think either:

- We should make it a getter on both, with the assumption that the underlying source implementer will respect the vague idea that "getters should not do very much"
- Or we should make it a method on both, so that we don't have to assume anything.

Alternately we could come up with more complicated schemes like making it a method on the underlying source that we only call at certain times and then cache the result of, and then a getter on the stream returns the cached result.

domenic:
Oops. I accidentally merged this as 127df9ed02c66057d87e3c8807a032d6b3471fc8. Anyway let's continue the discussion (here I guess?). I'll leave it as a getter externally/method on underlying source for now.

acolwell:
FWIW if the solution for the MSE case is being moved to only being addressed by ReadableByteStream, then I don't think readableAmount is needed. If I understand the ReadableByteStream correctly, I think MSE should simply be able to call readInto() with the size it needs and everything would be fine. It looks like readInto() essentially acts like POSIX read() so that would be sufficient for MSE's needs.

I suppose this is a long way of saying, don't add readableAmount just for MSE. :)

tyoshino:
Even in the new async read() byte stream Domenic is working on at https://github.com/whatwg/streams/pull/296, you can call `reader.read(view)` with view whose size equals to the number of bytes you want to get to read only what you want.

Kindly hold!

tyoshino:
`readableAmount getter` was proposed for users of hypothetical underlying sources which can tell how many bytes can be read synchronously to pass a single contiguous buffer of sufficient enough size to `readInto()` call while keeping it able to consume only part of the synchronously readable data not to replenish back pressure too much.

But when I revisited the issue, I think that it doesn't make much sense.
- If the total size to read into a single contiguous buffer is known, they can just allocate a buffer of the size and repeat invoking `readInto()` until it's filled.
- If the total size is unknown, we may still need to repeat `readInto()` calls if parts of data arrive little by little.

If reading into a single contiguous buffer doesn't really matter, we can just read into some small buffer and process bytes in it repeatedly. It just works.

`readableAmount getter` just reduces the number of `readInto()` call. That's not big benefit. Closing for now. Can be opened if necessary.

Issue #217, opened 2014-09-12
Run ReadableStream tests against ReadableByteStream


domenic:
At some point we probably want to figure out a way of running the ReadableStream tests against ReadableByteStream to cover their common surface area in terms of usage.

Will need to figure out some way of factoring out the creation logic, since that differs between each.

We might want to leave this for later when I am figuring out #216.

isonmad:
Is this issue obsoleted by e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393?

domenic:
Maybe it should be renamed, but running the templated tests against a BYOB-mode stream is still a good idea.

ricea:
I have a tentative plan for this:

1. Move tests that are relevant to both types to a new directory, readable-streams-bikeshed. "relevant" usually means that the test result doesn't depend on the chunk type. For example, move the test `'ReadableStream instances should have the correct list of properties'` from readable-stream/general.js to a new file named readable-streams-bikeshed/general.js.
2. Modify the moved tests to use Uint8Array chunks. Probably defining chunks named `a` and `b` at the top of the file and replacing `"a"` with `a`, etc. would work in most cases.
3. Wrap all the tests in the file in a loop like `for (const type of [undefined, 'bytes'])`
4. [optional] Check if readable-byte-streams/ contains duplicates of the copied tests and if so remove them.

This is important to make sure byte streams are rock-solid, but is a tremendous amount of work. Volunteers welcome.

domenic:
That plan sounds great. An alternative to step 3 that might work nicely would be to wrap the test JS file into a function, e.g.

```js
self.runReadableStreamTests = type => {
  // ... lots of tests here
};
```

and then have two small wrapper test files. This would separate out the test results into two files which can be nicer for test expectation files etc.

ricea:
So maybe something like
* `resources/common/foo.js`: _definitions of tests for the foo feature_
* `readable-streams/common/foo.js`: _wrapper for `type: undefined`_
* `readable-byte-streams/common/foo.js`: _wrapper for `type: 'bytes'`_

?

PR #280, opened 2015-02-10
Byte stream promises update

tyoshino:


tyoshino:
This PR applies bunch of changed recently made on ReadableStream to ReadableByteStream. Each of the commits in the PR applies the changes step-by-step.

domenic:
Wow, heroic work! Merged.

It would be nice to have tests but then again many of them probably would fall out of #217.

I wonder if we should consider moving ReadableByteStream into the main spec soon, since we want it for Fetch and will probably be driving it toward completion soon.

PR #281, opened 2015-02-10
Use angle quotes for lists

tyoshino:


index.bs
```diff
@@ -1406,7 +1406,7 @@ Instances of <code>WritableStream</code> are created with the internal slots des
 <ol>
   <li> Assert: <var>stream</var>@\[[state]] is <code>"closing"</code>.
   <li> Let <var>sinkClosePromise</var> be PromiseInvokeOrNoop(<var>stream</var>@\[[underlyingSink]],
-    <code>"close"</code>).
+    <code>"close"</code>, «‍»).
```
domenic: No need for empty, at least, not in ES Invoke... I'll make sure PromiseInvokeOrNoop also doesn't need empty lists.

index.bs
```diff
@@ -1499,7 +1499,7 @@ a variable <var>stream</var>, that performs the following steps:
     </ol>
   <li> Set <var>stream</var>@\[[writing]] to <b>true</b>.
   <li> Let <var>writeResult</var> be PromiseInvokeOrNoop(<var>stream</var>@\[[underlyingSink]], <code>"write"</code>,
-    «‍<var>writeRecord</var>.\[[chunk]]»).
+    « <var>writeRecord</var>.\[[chunk]]»).
```
domenic: Seems wrong?

domenic: Oh weird I was using some multibyte left angle quote here... you inserted a space but you also fixed that. So strange.

index.bs
domenic: Hmm it seems ES conventions don't include passing through missing parameters, e.g. [Invoke](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-invoke) explicitly sets the arguments list to empty even though Call already does. I'll add those steps I guess.

domenic:
Merged as dbfac2bb5132ee7a196c482a57c790c13ba58c42 with follow-up 0ebdd9ae707f56725463e37f941d170feb02f9de to allow omitting args.

index.bs
tyoshino: I added this since I saw it in the algorithm of WritableStream.abort(reason). Shall we remove it too and make PromiseInvokeOrFallbackOrNoop's argument omittable but only args2?

index.bs
tyoshino: Oh, I see. I copied your U+200D while replaced it with SP at writeRecord. I wasn't aware that. Anyway, thanks for fixing and committing.

tyoshino:
lgtm. thanks

PR #282, opened 2015-02-10
Add BufferFeedableReadableByteStream

tyoshino:


domenic:
Per many many discussions this is no longer quite the direction we're planning, so closing to clean up a bit :)

tyoshino:
Yeah. If anyone wants to see the discussion, please look at #253.

PR #283, opened 2015-02-10
Update ReadableByteStream to follow the style rule employed for the main...

tyoshino:
... spec

### DIFF
--- a/BinaryExtension.md
+++ b/BinaryExtension.md
@@ -29,6 +29,8 @@ By admitting returning the ArrayBuffer with no data written on it, `ReadableByte
 
 ### ReadableByteStream
 
+#### Class Definition
+
 ```
 class ReadableByteStream {
     constructor(underlyingByteSource = {})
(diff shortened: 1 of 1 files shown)


BinaryExtension.md
```diff
@@ -152,39 +148,36 @@ When an error function _F_ is called with argument _error_, the following steps
...
-##### ReadableByteStream.prototype.cancel ( reason )
+##### cancel(reason)
 
-1. If `this.[[state]]` is `"closed"`, return a new promise resolved with **undefined**.
-1. If `this.[[state]]` is `"errored"`, return a new promise rejected with `this.[[storedError]]`.
-1. If `this.[[state]]` is `"waiting"`, resolve `this.[[readyPromise]]` with **undefined**.
-1. Set `this.[[state]]` to `"closed"`.
-1. Resolve `this.[[closedPromise]]` with **undefined**.
+1. If **this**@[[state]] is `"closed"`, return a new promise resolved with **undefined**.
+1. If **this**@[[state]] is `"errored"`, return a new promise rejected with **this**@[[storedError]]`.
```
domenic: Leftover backtick at end of line

domenic:
I think this should update to use the underlyingByteSource object, like in #265. Right now the internal slots table is updated but the algorithms are not.

There's a bit of merge conflict since I accidentally merged #279.

Also let me know if you think we should move this into the main spec. We could either convert it to HTML like the rest of the spec or I could spend some time adding an [Ecmarkdown](https://github.com/domenic/ecmarkdown) post-processing step so that you don't have to convert as much and the source format remains fairly readable. (Eventually I probably want to use Ecmarkdown for the main spec but it's not a great use of time ([or is it?](https://xkcd.com/1319/)).)

domenic:
Can close this and roll up into #300.

PR #287, opened 2015-02-19
Prototyping several changes for better support for reading bytes

tyoshino:


tyoshino:
This is an extended version of WritableStream with the following functionality:
- single API to return a connected pair of writable side and readable side
- drained semantics on the readable side
- readable side can acknowledge read chunks
- writable side can receive acknowledgement info

See the example in the test file to see how this works for ReadFile/fread style source with buffer allocator residing in the source side.

write() returns a structure representing the status of the operation and read() returns an object to acknowledge the operation. These are for synchronous pumping, but can be removed if it's considered to be too complex.

async readInto() can be realized by using the writable side of the operation stream. You'll receive eof via write()'s return value. This serves for needs to allow users to bring their own buffer than allocating buffers inside a source. If write() being used for reading is confusing, we could rename write() of the operation stream to e.g. issue() and read() to accept(), take(), etc.

tyoshino:
I meant that I guess this single API can serve for various use cases.

tyoshino:
The following stuffs proposed in recent discussion are also incorporated
- readableAmount
- window

tyoshino:
- Added waitSpaceChange() method and made pipeOperationStreams() respect/sync window and space.
- Merged sinks and sources in the unit test to realize the concept explained in https://github.com/tyoshino/streams_integration/blob/master/ByteSource.md

domenic:
Sorry for the delay on this.

I'm having a hard time understanding all of what's going on here just from the code... What is the total API surface area added? What does this replace? What use cases does it solve that are not currently covered? You're putting a lot of work into this, so I want to understand more :).

My guess is that since it has all these new APIs it gives us more control? I know we've talked about use cases that require more control scattered throughout many different issues, but could you outline them in one place and show how they map to new APIs?

Based on the the tests, these look really complicated to create. How are they to use? How big is the public API for most authors? How does this map to the three classes of authors I outlined in https://github.com/whatwg/streams/issues/253#issuecomment-74315547 ? Do they interop with the existing API at all?

As I explored in [this gist](https://gist.github.com/domenic/197ae6fab8adc6890c7a#part-3-streams-as-two-sided-pipes), streams as two-sided pipes can be pretty tricky to get right. But I guess the problems there are mostly for readable streams, whereas this is meant to be a way of creating WritableStreams? For writable streams it just means manually pumping the readable side, which is a bit more awkward than the underlying source write() callback IMO, but maybe it is necessary to get more control?

tyoshino:
Thanks Domenic.

This PR includes two big proposals:

- merge of underlying sink/source, writable/readable stream into OperationStream
- precise flow control stuff

----

`WritableOperationStream` is based on a little modified version of the `WritableStream`. It has `.write()` that returns an object that represents a value that becomes available in the future. I used `OperationStatus` and `Operation` instead of a promise that will be fulfilled on arrival of the written value to the sink experimentally, but this can be simplified to return a promise.

```
const p = wos.write();
...
const op = ros.read();
op.complete();
...
at write side, we see p gets fulfilled.
```

And, `ReadableOperationStream` is based on `ReadableStream` but modified to have power to represent acknowledgement to the read data. With this change, ReadableOperationStream gained similar surface as the `underlying sink` concept for the `WritableStream`.

The motivation behind this was that:

- the async read() approach we wanted to add for supporting ReadFile/fread looked similar to `WritableStream`. I thought we may reuse it for implementing async read().
- I thought the buffer pool approach explored in the issue initiated by Ben could be realized by giving the reading API to have acknowledgement functionality. I.e. making ack mean return of the obtained ArrayBuffer. This lead to ReadableStream to ReadableOperationStream change of addition of `.complete()` and `.error()` to the return value (Operation object).
    - this resembles the `underlying sink` concept. underlying sink's write() accepts value from the queue of the WriteableStream, processes it and returns a promise to acknowledge it. It'll propagated to the original `write()` call on the WritableStream and fulfills/rejects the returned promise.
- in thinking about the dual for the async read() API (which is realized as WritableOperationStream in this PR), I found that ReadableOperationStream (see `pipeOperationStreams()`) can be easily connected to.

Based on the observations above, I tried to merge underlying source/sink, ReadableStream and WritableStream into two surfaces (`WritableOperationStream` and `ReadableOperationStream`) but backed by a single object named `OperationStream`. The stream creation methods on FakeFileBackedByteSource and BytesSetToOneExpectingByteSink explain how the API is used for construction of a stream object to give to the user. Either of WritableOperationStream or ReadableOperationStream is given to the object that processes actions to the I/O and the other is returned to the user. Unlike the current WritableStream/ReadableStream constructor that take UnderlyingSource/Sink object, the underlying source/sink sees the same API as the user sees and accesses it using promises. `_loop()` methods are optimized by having a single loop to forward output from the I/O to the stream synchronously. So, it may look more complicated (also due to the addition of the OperationStatus object instead of a promise). But it can be simplified if the users are fine with a little less efficiency.

----

I also designed and implemented some precise flow control stuff using this opportunity. Sorry for confusing but they're independent parts.

- get space()
- waitForSpace()
- get window()
- set window()

----

> Based on the the tests, these look really complicated to create. How are they to use? How big is the public API for most authors? How does this map to the three classes of authors I outlined in #253 (comment) ? Do they interop with the existing API at all?

- (1): Obtains a ReadableOperationStream from a source, and calls pipeOperationStreams in the simplest case. Or use read(), state and ready on ReadableOperationStream manually. If the source is allocate-and-forget type, the read()'s complete() and error() don't need to be invoked. GC takes care of the buffers' lifetime.
- (2) and (3): Use automatic buffer allocation described by FakeFileBackedByteSource.createBufferProducingStreamWithPool(). Or use bring-your-own-buffer type reading described by FakeFileBackedByteSource.createBufferFillingStream().
    - FakeFileBackedByteSource.createBufferProducingStreamWithPool() returns a ReadableOperationStream that gives us allocated buffers filled with the contents read from the encapsulated source I/O together with acknowledgement functions `.complete()` and `.error()`. The acknowledgement is propagated to the code behind the FakeFileBackedByteSource instance and used to ok-to-return-to-the-pool signal.
    - FakeFileBackedByteSource.createBufferFillingStream() returns a WritableOperationStream that accepts buffers, fills them and returns (or just signals completion) via the returned OperationStatus. This is almost equivalent to your async read() API.

tyoshino:
> My guess is that since it has all these new APIs it gives us more control? I know we've talked about use cases that require more control scattered throughout many different issues, but could you outline them in one place and show how they map to new APIs?

I think so. Sorry that it's all-in-one PR.

tyoshino:
> As I explored in this gist, streams ...

Yes! I remember that we've visited this before and concluded that we shouldn't. I forgot the address of the discussion you recorded. I'll take a look at the gist again and check if this PR resolves the issues there or not.

tyoshino:
byteCountingPipe() is an example of the transform stream concept realized by the operation stream. Using the helper function `selectOperationStreams()`, it polls events on the streams and do some transformation.

It can be modified to have .pipeThrough() style chainable functionality, I think.

tyoshino:
When you read the comments above, please take a look at this commit :)
https://github.com/whatwg/streams/blob/815649bb91325e1dbd4c26bd95fbc13ac8383ddf/reference-implementation/lib/experimental/operation-stream.js

I may make more commits today and they might make the references in the comments invalid.

domenic:
> merge of underlying sink/source, writable/readable stream into OperationStream

This is an interesting one. I am still pretty sure this will not be good for the user-facing API. That is, I think the current user-facing API, both for stream creators and stream users, is pretty optimal. In particular the dual start() and pull() hooks for readable streams, and the write() and close() hooks for writable streams, are quite nice for stream creators.

I think we should consider the API separately from the implementation, however. If it turns out that a two-sided stream implementation is conceptually simple and makes the spec and implementation easier to do, then that is good, as long as the API niceness is preserved. Conversely, if there are API improvements to be made, we can discuss them separately from discussing the strategy used to implement them.

Another thought: it might be possible that the work here is going to be most valuable when considering how to spec and implement TransformStream. Currently TransformStream is kind of a hacked-together thing where I reuse most of the ReadableStream and WritableStream logic. However it is probably more efficient to implement it as a single primitive, which happens to expose the ReadableStream and WritableStream APIs. OperationStream seems kind of like that approach. (Or, rather, a special-case of it, where the transform is the identity.)

> precise flow control stuff

This is the kind of API improvements stuff that I think we need to consider on its own merits, outside of the "implementation detail" of two-sided pipe streams. More below :)

> ...
> Based on the observations above, I tried to merge underlying source/sink
> ...

These are really interesting observations. Can you explain more why a readable stream would need acknowledgement/error for each chunk read? Currently we kind of have acknowledgement---when you call read(), you implicitly acknowledge the data. And when you cancel the stream, you indicate some kind of error. (I am not sure why you would need to indicate error on an individual chunk.)

> I also designed and implemented some precise flow control stuff using this opportunity. Sorry for confusing but they're independent parts.

This is the stuff I really want to learn about!!

space/waitForSpaceChange are on writable stream. I guess they are more sophisticated versions of the current backpressure signal, of "writable" or "waiting", that can be used to communicate _how much_ can be written. This makes sense.

I think it mostly makes sense for WritableByteStream, because then when piping you could ask for as many bytes from the readable byte stream as `space` indicates on the target writable stream. Maybe it could make sense for general readable streams using their `size()` strategy method, but that is kind of complicated. E.g. it assumes compatible notions of size between the source stream and the target. Hmm.

I don't quite understand window though. I imagine it maps to TCP window stuff but I don't understand that perfectly either :(. Could you help me understand it? And, how general is such an idea---just byte streams, or anything with a size, or...? Does it depend on the capabilities of the underlying source, so e.g. it works for sockets but not for files?

tyoshino:
> This is an interesting one. I am still pretty sure this will not be good for the user-facing API. That is, I think the current user-facing API, both for stream creators and stream users, is pretty optimal. In particular the dual start() and pull() hooks for readable streams, and the write() and close() hooks for writable streams, are quite nice for stream creators.

I finally factored out underlying source and sink part but not like what we have in the spec now. The streams don't have a queue but just have code for managing promises and writer/reader.

- https://github.com/whatwg/streams/blob/aac7c91caa4fea6320c806c2c515e09e8e638af9/reference-implementation/lib/experimental/readable-operation-stream.js
- https://github.com/whatwg/streams/blob/aac7c91caa4fea6320c806c2c515e09e8e638af9/reference-implementation/lib/experimental/writable-operation-stream.js

> I think we should consider the API separately from the implementation, however. If it turns out that a two-sided stream implementation is conceptually simple and makes the spec and implementation easier to do, then that is good, as long as the API niceness is preserved. Conversely, if there are API improvements to be made, we can discuss them separately from discussing the strategy used to implement them.

Agreed. Now it's placed in https://github.com/whatwg/streams/blob/aac7c91caa4fea6320c806c2c515e09e8e638af9/reference-implementation/lib/experimental/operation-queue.js as just a helper.

tyoshino:
Oh, sorry. I'm committing changes using wrong (private) account info recently...

domenic:
I was wondering about that :)

Can you show samples of the underlying source and sink APIs in use? E.g. maybe samples from https://streams.spec.whatwg.org/#creating-examples ?

tyoshino:
OperationQueue is actually one example of both underlying source and sink. I'll try to see how/if this works for other cases soon.
https://github.com/whatwg/streams/blob/aac7c91caa4fea6320c806c2c515e09e8e638af9/reference-implementation/lib/experimental/operation-queue.js

domenic:
I see, that is pretty helpful (although some concrete examples with websockets or whatever would also be helpful). I think we'd want to use revealing constructor pattern parameters instead of having the underlying source/sink reach into the readable/writable stream directly. But that is a mostly cosmetic thing I imagine. I am starting to get really intrigued.

tyoshino:
Now, queue and promises are separated.

See this file for ReadbleStream and WritableStream interface definition.
https://github.com/whatwg/streams/blob/operationStream/ThinStream.md

- Regarding locking, I aggressively disallowed all methods on inactive stream/reader/writer. This relieves us from the task to give them some possibly convenient (but maybe no one would use) semantics to them while inactive.
- Separated "operation" concept from the ReadableStream and WritableStream again. They now lives in OperationQueue library.
- Methods are grouped based on their purpose.
- Employed revealing constructor pattern.
- Added unstoppable push source example for demonstration.

----

I haven't demonstrated, but:

- I guess the new queue-free API can be used for the offline discussion regarding detaching @yutakahirano
- I guess WritableStream with ExclusiveStreamWriter can be used for defining async read interface. Beyond locking, written buffer is not taken over by the writer. If you return a promise (or OperationStatus) from write(), you can return the filled buffer to the user of the stream, not writer.
  - write() corresponds to async readInto()
  - ExclusiveStreamWriter corresponds to a reader infrastructure for the async readInto()

tyoshino:
I have also been trying to keep queue handling class useful for source/sink implementor. But not yet put in good form.

tyoshino:
Sorry for not responding to your questions. I'll respond to them from the office tomorrow...

tyoshino:
Domenic: Response to https://github.com/whatwg/streams/pull/287#issuecomment-76333828 cont'd

> These are really interesting observations. Can you explain more why a readable stream would need acknowledgement/error for each chunk read?

When we were discussing the idea of having a buffer pool inside a readable stream and giving them to the user on reading, we found that we need to explicitly tell the source when the buffer becomes ok to reuse. As you said, `ReadableStream`'s `read()` implicitly acknowledges the data, it's not always that we finish consuming contents on the read buffer synchronously. So, I attempted to add explicit and delay-able acknowledgement feature as that read() returns an object on which methods for acknowledgement are placed.

```
const dataAndAckMethod = s.read();
// Use data part.
...
// When data is comsumed (possibly in another task), call the ack method
```

In addition to that, I thought that similarly to bring-your-own-buffer model readable stream, we may want to use please-write-to-this-buffer model writable stream. I.e. suppose that we have a sink that prepares and exposes some memory region as an ArrayBuffer and ask the producer code to write bytes into it. To wrap this sink with some streams variant API, we need a method which returns a region for writing data and also returns some methods to tell the sink completion of the writing.

```
const regionAndCompleteMethod = s.write();
// Write data to the region
...
// When done (possibly in another task), call the complete method.
```

This reduces the number of ArrayBuffer allocation compared to please-give-me-a-filled-array-buffer model writable stream. This `s.write()` is logically equivalent to current WHATWG ReadableStream if the return value type is changed to the Operation object.

----

Anyway they're just attempts to address each of the semantics separately. Now, I'd like to go back to investigate which (set of) semantics we should support

Source of bytes

- Choice: who allocates buffers
  - Code prepares and passes an ArrayBuffer to source. Source writes bytes to it and notifies code of completion.
    - writing bytes completes synchronously (when read() returns, writing is finished).
    - Source notifies code of completion asynchronously.
  - Source prepares and passes a filled ArrayBuffer to code.
    - Code returns the ArrayBuffer to source when done consuming.
    - GC collects consumed ArrayBuffers.

Sink of bytes

- Choice: when write is allowed
  - Sink automatically starts accepting write (w/ backpressure hint)
  - Some preceding method call makes sink accept write
- Choice: who allocates buffers
  - Sink prepares and passes an ArrayBuffer to code. Code writes bytes to it and notifies sink of completion.
    - Choice: allocation is async? sync?
      - write method returns an ArrayBuffer synchronously
      - write method returns an ArrayBuffer asynchronously
  - Code prepares and passes a filled ArrayBuffer to sink.
    - Choice: recycle the buffer?
      - Sink consumes contents synchronously, so code can start reusing the buffer immediately.
      - Sink notifies code once I finish consuming bytes from it to allow code to recycle it.
      - Sink takes ownership of the ArrayBuffer and have the GC collect it once consumed.

tyoshino:
Again, reply to Domenic's comment https://github.com/whatwg/streams/pull/287#issuecomment-76333828

> space/waitForSpaceChange are on writable stream. I guess they are more sophisticated versions of the current backpressure signal, of "writable" or "waiting", that can be used to communicate how much can be written. This makes sense.

Right.

> I think it mostly makes sense for WritableByteStream, because then when piping you could ask for as many bytes from the readable byte stream as space indicates on the target writable stream. Maybe it could make sense for general readable streams using their size() strategy method, but that is kind of complicated. E.g. it assumes compatible notions of size between the source stream and the target. Hmm.

Yes. There could be some translation between each pair of source type and target type, but not always so simple and worth to be abstracted in the streams library, I guess. For byte streams, yes, I think it would be useful.

> I don't quite understand window though. I imagine it maps to TCP window stuff but I don't understand that perfectly either :(. Could you help me understand it? And, how general is such an idea---just byte streams, or anything with a size, or...? Does it depend on the capabilities of the underlying source, so e.g. it works for sockets but not for files?

Right. It's inspired by TCP. For byte streams, we can specify how much data we want to receive from the remote host, read from kernel, read from other process, etc. using window. In some old thread, I said that I guessed it's not rare that we want to control the high water mark of the strategy from user code based on current situation e.g. back pressure from the sink the data read from the readable stream finally reaches. Basically, it's fine to understand this as a knob for adjusting hwm.

tyoshino:
Addition of window and space are just a bonus. I felt that they might be connected with the issue we're trying to solve, but it's almost turned out to be unrelated. So, I think we can post pone it.

domenic:
Continued from https://github.com/whatwg/streams/issues/253#issuecomment-77001963

> From operationStream, I'd like to incorporate:
>
> - queue / stream separation

Does this affect public API (either for stream creators or users)?

I notice the new comments at the top of the file (which are extremely helpful) have slightly different underlying sink/source designs. Instead of being based on strategy they give more direct control---which is probably a good change. Is that part of queue / stream separation?

> - revised set of promises

These are interesting. To summarize:

- Readable stream:
  - `readable` which is fulfilled when the stream enters the readable state
  - `errored` which is fulfilled when the stream enters the errored state
  - No more `closed` (seems bad); no more `ready` (possibly can be replaced by above two)
  - How does this fit with the async pull() + sync read() idea?
- Writable stream:
  - Similar to above: `writable` and `errored`; no more `closed`
  - Also has `waitSpaceChange()` promise-returning method.

I am not sure these are an improvement, personally. What was the motivation?

> revised preconditions for getReader() / getWriter() and release()

The addition of the "locked" state is hopefully not too annoying. Prototyping will help confirm. But in general (I know I'm repeating myself, sorry) I am in favor of changing the preconditions in ways that make things more straightforward.

tyoshino:
> Does this affect public API (either for stream creators or users)?

Creators are affected. Users are not affected.

> I notice the new comments at the top of the file (which are extremely helpful) have slightly different underlying sink/source designs. Instead of being based on strategy they give more direct control---which is probably a good change. Is that part of queue / stream separation?

Yes, that's exactly what I meant by queue / stream separation. Maybe there're still sources / sinks that really want to use a queue. We can provide some additional helper classes for them. I.e.

- ThinWritableStream + QueueSink === Current WritableStream
  - QueueSink would have API similar to current WritableStream's underlying sink
- ThinReadableStream + QueueSource === Current ReadableStream
  - QueueSource would have API similar to current ReadableStream's underlying source

Thin.* are used by sources / sink which needs freedom.

> > revised set of promises

> I am not sure these are an improvement, personally. What was the motivation?

After bunch of changes, I'm not so inclined to push this. They were for some simplification. I don't remember so much... I've almost reverted them for now.

> The addition of the "locked" state is hopefully not too annoying. Prototyping will help confirm. But in general (I know I'm repeating myself, sorry) I am in favor of changing the preconditions in ways that make things more straightforward.

We've almost reached agreement that it's fine as far as we can hide internal state and forbid operations on the stream while it's locked. Then, prohibiting everything should be fine. Yes, we need to be careful not to decrease usability. But maybe ok...

On the latest commit, I've prototyped byte source / stream (reader) separation as discussed around https://github.com/whatwg/streams/issues/253#issuecomment-76998452.

tyoshino:
Please ignore files with top comment "WIP WIP WIP". They're incomplete snippets.

tyoshino:
Interfaces
- thin-readable-byte-stream.js: manual pull readable stream
- thin-readable-stream.js: auto pull readable stream
- thin-writable-byte-stream.js: manual disposing writable stream
- thin-writable-stream.js: auto disposing (no garbage port) writable stream

- thin-stream-base.js: pipeStreams is defined and some helpers are exposed for building transform stream

Examples
- fake-file-backed-byte-source.js: Sample implementation of a file backed byte source from which we can get either of manual pull readable stream or auto pull readable stream.
- mock-byte-sink.js: Sample implementation of a file backed byte sink from which we can get manual disposing writable stream or auto disposing readable stream.
- stream-queue.js: A queue with readable stream side and writable stream side.
- thin-stream-utils.js: Helpers for examples

Test
- test/experimental/thin-stream.js

tyoshino:
Almost all of the ideas prototyped in this branch has been implemented / abandoned / sublimed into somethings else and landed till https://github.com/whatwg/streams/commit/e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393.

The idea of ReadableStream with acknowledgement is well described in some issues e.g. https://github.com/whatwg/streams/issues/324. The reference implementation has been drastically modified. Because of these reasons, we don't need to keep this PR any more.

Closing.

PR #296, opened 2015-03-12
Change the model for ReadableStream to have async read()

### LINKED ISSUE #253: Support reading bytes into buffers allocated by user code on platforms where only async read is available
Had a thought-provoking conversation with @piscisaureus today. He works on libuv and was giving me their perspective on how they do I/O and how it interfaces with Node.js's streams.

He pointed out that resources like files and pipes shared with other processes (seekable resources, he called them) don't have the epoll + read interface that ReadableByteStream's current design is based on. In Node.js, they do blocking I/O in a thread pool with pre-allocated buffers.

This works for a `ready` + `read()` JS interface, since then the implementation can pre-allocate (say) a 1 MiB buffer, read into it off-thread using blocking I/O, then fulfill the `ready` promise. When JS calls `read()`, it just returns the pre-allocated buffer.

It doesn't work as well with ReadableByteStream's `ready` + `readInto(ab, offset, sizeDesired)` interface, since we don't know what size buffer to allocate until too late. If we pre-allocate something too small, `readInto` will always be returning a number below the desired `size`, which is a bit user-hostile. If we pre-allocate too large, we are at the least wasting memory, and at the worst getting ourselves into a bunch of trouble as we need to figure out how to merge the "leftovers" with other chunks we read in the future.

(I am assuming here it is easy in V8 to map the [offset, offset + max(sizeRead, sizeDesired)] portion of `ab` onto a given C++ backing buffer.)

A `readIntoAsync(ab, offset, desiredSize)` model would work a bit better, since then we'd know to pre-allocate a buffer of size `desiredSize`. But, I was curious if you guys had any other thoughts? Did you think of this issue when designing the API, @tyoshino? I imagine @willchan could be helpful here too.

### LINKED ISSUE #264: Factor out portions of a test suite that can be run against already-constructed streams
This is tricky, but important. The idea is that in the browser there will be various APIs that return an already-constructed stream. (E.g., `fetch`.) We would like to be able to test that these browser-constructed streams behave per spec. However, all of our current tests just create streams and then test them.

One way to do this would be to perhaps annotate the tests as having certain "prerequisite" streams passed in. For example:

- Empty readable stream
- Readable stream with 3 chunks
- Errored readable stream
- Readable stream with 1 chunk that becomes errored after 20 seconds

Then you could imagine that in the pure-JS test suite, we just manually create such streams, and then run the matching tests. But for the `fetch` tests, we use `fetch` + a local HTTP server with endpoints that respond in various ways to create the required stream.

Related a bit to #217.

### LINKED ISSUE #266: Requirement on promises equality

### LINKED ISSUE #291: Test ReadableStream does not call pull multiple times after previous pull finishes is incorrectly specified.
You can find the test currently in readable-stream.js:304.

I think the test should not have the timeout in line 334 because according to the spec, after reading we pull asynchronously, but if we leave the timeout where it is and let it work, the pull from the second read could end up happening because it is waiting to be run.

Am I missing anything?

domenic:
This replaces the dual ready + read() approach previously, which was derived from the epoll(7) + read(2) paradigm. In #253 we did a deep dive on real-world implementation strategies for byte streams, and discovered that the ready + read() model causes a conflict with the semantics we want for byte streams. Briefly, because some byte streams will demand to know the size of the buffer they must fill before doing any I/O (the fread(3) model), the byte stream method that reads into a given typed array must be asynchronous. If such byte streams are then to conform to the readable stream interface, with a no-argument read() method that is an auto-allocating version of the read-into method, then read() must also be async, across all readable streams.

This is a usability upgrade for consumers, in many cases. However, for non-byte streams, it potentially costs more microtasks when multiple chunks of data would be available synchronously. We are hopeful that even if current JS engine microtask queues are not as fast as they could be, they will improve over time until this overhead is negligible (which in theory it should be).

Alongside this change, we moved the read() method from the readable stream to the reader (now called "readable stream reader" instead of "exclusive stream reader"). This drastically simplifies the stream/reader interaction, and also allows the possibility of different types of readers which have different reading behavior---again, very helpful for byte streams. The usability impact is also positive, as it guides consumers toward using piping instead of directly reading chunks from the stream.

Note that read() promises fulfill with { value, done } instead of using an EOS sentinel value. This avoids a number of problems (see extensive discussion in #253), and also provides a mechanism by which readable byte streams can smuggle out "unused" buffers given to them, by using { value: zeroLengthViewOntoBuffer, done: true }.

Finally, the state property is now removed (from readable stream), since there is no longer a "waiting" vs. "readable" distinction.

This commit also adds some new infrastructure for _templated tests_, and ports some portion of the existing tests there. This is our solution for #217 and #264.

Note that we re-merged all related code into a single readable-stream.js file, as the setup with the three separate files (readable-stream.js, exclusive-stream-reader.js, and readable-stream-abstract-ops.js) was problematic in causing circular dependencies.

- Closes #253!
- Closes #266 by simplifying reader-related stuff, removing the problematic `ready` promise, and ensuring that cancel() always returns a new promise instead of reusing [[closedPromise]].
- Closes #291 since the relevant tests have been re-written and the related infrastructure around pull simplified.
- Closes #264 by introducing templated tests.

---

We apologize for this large of a change, this late in the game! We really got bit by not considering the zero-copy byte stream use case carefully enough. We spent a lot of concentrated time working out how this change will enable byte streams, and @tyoshino is working on a prototype implementation over in #287 which will allow us to bring byte streams into the main spec/reference implementation/test suite soon. In the end, it's painful to see this large of a diff, but the result will be a lot better from many angles---not only byte stream compatibility, but general developer-facing ergonomics as well.

---

Branch snapshot at https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/. Sections that are good for at-a-glance understanding of the change, or sections that have seen particular rework, include:

- [New discussion of locking](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#locking)
- [New examples of using readable streams](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#rs-intro)
- New [ReadableStream](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#rs-class-definition) and [ReadableStreamReader](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#reader-class-definition) class definitions
- New [ReadableStreamReader.prototype.read()](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#reader-read) method definition, including non-normative description
- [New section on "other streams"](https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/#other-streams), replacing the crusty "subclassing streams" section, describing how different types of streams (e.g. ReadableStreams and putative ReadableByteStreams) share common idioms and can all interoperate with anything obeying the writable stream contract.


reference-implementation/lib/readable-stream.js
```diff
@@ -156,67 +131,349 @@ export default class ReadableStream {
...
+
+  cancel(reason) {
+    if (IsReadableStreamReader(this) === false) {
+      return Promise.reject(
+        new TypeError('ReadableStreamReader.prototype.cancel can only be used on a ReadableStreamReader'));
+    }
+
+    if (this._ownerReadableStream === undefined) {
+      return Promise.resolve(undefined);
+    }
```
tyoshino: Return a promise rejected with the error if this was released when it got errored?

domenic: Good catch, will update

domenic: Amended last commit in the PR with fix

reference-implementation/lib/readable-stream.js
```diff
@@ -156,67 +131,349 @@ export default class ReadableStream {
...
-    this._closedPromise_reject(reason);
-    this._closedPromise_resolve = null;
-    this._closedPromise_reject = null;
+  return undefined;
+}
+
+function CreateReadableStreamCloseFunction(stream) {
+  return () => {
+    if (stream._state !== 'readable') {
+      return undefined;
```
tyoshino: It's out of the main topic of the PR, but don't we want to say "you're making a mistake" when close()-ed twice or more?

domenic: Hmm. I think you are right. Underlying source should be more careful than this. Open a new issue?

reference-implementation/lib/readable-stream.js
```diff
@@ -98,57 +70,60 @@ export default class ReadableStream {
...
-          } else if (source.state === 'closed') {
-            closeDest();
-          }
-        } else if (ds === 'errored' && (source.state === 'readable' || source.state === 'waiting')) {
-          dest.closed.catch(cancelSource);
-        } else if ((ds === 'closing' || ds === 'closed') &&
-            (source.state === 'readable' || source.state === 'waiting')) {
-          cancelSource(new TypeError('destination is closing or closed and cannot be piped to anymore'));
+      lastRead = reader.read();
+      Promise.all([lastRead, dest.ready]).then(([{ value, done }]) => {
```
tyoshino: Need to address https://github.com/whatwg/streams/pull/288#discussion_r25886680 ?

reference-implementation/lib/readable-stream.js
tyoshino: Created https://github.com/whatwg/streams/issues/298

tyoshino:
Took a look at readable-stream.js. Overall looks good.

reference-implementation/lib/readable-stream.js
domenic: Fixed by adding ` if (dest.state === 'writable')` below (line 93). All tests still pass as expected.

Might be a nicer way to do this algorithm that avoids closedPurposefully by checking dest.state more thoroughly. We will see if that is possible later, I think. When move pipeTo into the main spec seems like a good opportunity.

domenic:
Awesome. index.bs review would be good too (especially to the non-algorithm text). Tests might not be worth it.

tyoshino:
https://streams.spec.whatwg.org/branch-snapshots/async-read-take-3/ ?

domenic:
Yep!

PR #306, opened 2015-03-24
Made "Multiple readers can access the stream in sequence" test robust

calvaris:
It seems that the test should ensure that the promises are resolved before ending the test

### DIFF
--- a/reference-implementation/test/readable-stream-reader.js
+++ b/reference-implementation/test/readable-stream-reader.js
@@ -139,6 +139,8 @@ test('closed should be fulfilled after reader releases its lock (multiple stream
 });
 
 test('Multiple readers can access the stream in sequence', t => {
+  t.plan(2);
+
   const rs = new ReadableStream({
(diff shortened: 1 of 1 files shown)


reference-implementation/test/readable-stream-reader.js
```diff
@@ -155,7 +155,8 @@ test('Multiple readers can access the stream in sequence', t => {
   reader2.read().then(r => t.deepEqual(r, { value: 'b', done: false }, 'reading the second chunk from reader2 works'));
   reader2.releaseLock();
 
-  t.end();
+  // Let promises run.
+  setTimeout(() => t.end(), 50);
```
domenic: Good catch! It would be better to use t.plan(2) at the top of the test though, I think.

domenic:
Oh, you updated this, thanks! Merged as 8549ce0db83cb83cb918007a7742b12824a8cbe8.

Issue #301, opened 2015-03-18
Generalized and precise flow/buffer control


tyoshino:
Before recent redesign, we always had a queue in the JS world behind a readable stream, and we calculated how to exert backpressure to the "underlying source" based on the size of the queue.

We're now separating the streams API semantics visible to the consumer (mainly implemented in the reader) and a queue behind it which doesn't always exist (depending on who (JS, C++ code, kernel, device, etc.) buffers data). This made me think more generally, that is, there're many points of buffering behind streams. The number of such points depends on which API we're using, which platform we're using, etc.

It sounds like both the high water mark & strategy concept, and the precise flow control stuff I proposed only make sense for streams with a queue in JS world. It seems it's hard to well-define `queueSize`, `highWaterMark`, `pullAmount`, etc. as cross-platform, cross-API feature.

To put it differently, for non-JS-queue-backed APIs, what do we guarantee for the consumer when HWM is set to X?

I still guess some kind of best-effort approach is still useful. The UA tries to keep the number of bytes buffered on the path between the ultimate data source (disk, the peer server, etc.) and the JS world. As Domenic originally suggested, the strategy itself should be chosen intelligently by the UA considering memory size, CPU power, browser architecture, etc. etc. and it's responsible only for computing how much data is buffered in **the JS world**. Combining buffer size info (C++, kernel, JS, ...), the UA would intelligently limits the amount of data to pull.

And, the `pullAmount` or something in the proposed precise flow control APIs will just give a hint about how quickly the consumer can consume data in the form of the number of bytes. It is not directly connected with such as the size of the queue in JS. The stream utilizes the info using some algorithm to control amount of data to pull.

tyoshino:
As Yutaka has been advocating, we should stop attempting to specify direct connection between the "underlying source" which the JS code faces and the readable (byte) stream operation. Not directly, but just have a good API surface that **could** be utilized for optimization if the implementor of the readable (byte) stream wants. Bring your own buffer style reader allows the implementor to directly connect `read(view)` with ReadFile, etc, but it's not required. As we discussed offline, the JS library may be requested to drain data from underlying source into an ArrayBuffer in JS given some consuming pressure. In such case, we don't pass the view provided by the consumer to ReadFile, but copies data already read out to the provided view.

domenic:
[...]

PR #310, opened 2015-03-31
Introduce readable stream controller class

### LINKED ISSUE #309: Making the "queue manipulator" its own class
To deal with #301 per the plan in https://github.com/whatwg/streams/issues/301#issuecomment-83939205, where in the current set of functions (enqueue, close, error) grows an additional getter "spaceLeftGetter", I'm planning a bit of a change the signature and wiring. Would love thoughts.

The naive extension is of course something like (enqueue, close, error, getSpaceLeft). But at this point the argument list is starting to feel overloaded. I think there's a better solution.

The basic idea is: Instead of creating separate functions stream@[[enqueue]], stream@[[close]], and stream@[[error]] for each stream (plus a stream@[[getSpaceLeft]]), we create a _class_, called `ReadableStreamQueueManipulator`, which has `enqueue`, `close`, and `error` methods (plus a `spaceLeft` getter). We then create an instance of this class when creating the stream, and hand it out in calls to the underlying source's `start` and `pull` methods.

This has a number of advantages:

- Saves on memory, as we don't need to create four separate closures per class. Instead there's just the three methods/one getter, shared on the prototype of `ReadableStreamQueueManipulator`.
- Makes the signature a bit less crazy. Instead of writing stuff like

  ```js
    start(enqueue, close, error, getSpaceLeft) {
      enqueue('a');
      enqueue('b');

      if (getSpaceLeft() > 0) {
        enqueue('c');
      }

      setTimeout(close, 100);
    }
  });
  ```

  we can do

  ```js
    start(queue) {
      queue.enqueue('a');
      queue.enqueue('b');

      if (queue.spaceLeft > 0) {
        queue.enqueue('c');
      }

      setTimeout(() => queue.close(), 100);
    }
  });
  ```

- Better allows for future extensibility. (Even if we never add new methods or properties to `ReadableStreamQueueManipulator`, any other stream classes that want to follow this pattern would probably benefit from their interface also being a single queue manipulator object with custom methods. Maybe `ReadableByteStream` would benefit, indeed!)

It also has a couple drawbacks:

- A bit more typing. `setTimeout(close, 100)` vs. `setTimeout(() => queue.close(), 100)`. Or `ws.onend = close` vs. `ws.onend = queue.close.bind(queue)`.
- They're now methods, instead of closures, which is less convenient. `ws.onend = queue.close` will fail. As will

  ```js
    start({ enqueue, close }) {
      // oops, destructuring un-bound the methods, so enqueue() will fail.
    }
  });
  ```

It invites a few naming decisions:

- I named it `ReadableStreamQueueManipulator` because it exposes an interface for manipulating the stream's queue. You could also try to conceptualize it as the queue itself---and I think that's what most authors will do, as I illustrate in the naming of my variables in the above sample code. But at least the current spec doesn't give a very clean separation. `enqueue`/`close`/`error` manipulate a lot of state that belongs to the stream, not to any queue. I think `ReadableStreamQueueManipulator` is fine for now (especially since it's barely author-exposed)...
- On the other hand, you don't really "close" or "error" a queue. You do that to the stream. Maybe it's just `ReadableStreamManipulator`? `ReadableStreamController`?
- We _could_ pun on the writable stream interface with names like `write`/`close`/`abort` instead of `enqueue`/`close`/`error`. However, I think this is probably not a good idea, because of all the stuff @tyoshino explored in #287. The readable stream's queue is actually quite different from a writable stream, e.g. there's no notion of an enqueue succeeding (whereas a write might succeed or fail, depending on the hardware/network/etc.); all operations are sync instead of async; etc.

domenic:
Instead of creating enqueue, close, and error functions per-stream, and passing (some subset of them) to the underlying source's start() and pull() methods, we instead create an instance of the ReadableStreamController class, which has methods enqueue(), close(), and error().

This closes #309, which contains more discussion and justification. It sets the stage for #301 (which will work by adding another property to the ReadableStreamController class).

Spec should be available for review at https://streams.spec.whatwg.org/branch-snapshots/rs-controller when the build completes.

### DIFF
--- a/reference-implementation/test/brand-checks.js
+++ b/reference-implementation/test/brand-checks.js
@@ -1,11 +1,24 @@
 const test = require('tape-catch');
 
 let ReadableStreamReader;
+let ReadableStreamController;
 
 test('Can get the ReadableStreamReader constructor indirectly', t => {
   t.doesNotThrow(() => {
     // It's not exposed globally, but we test a few of its properties here.
     ReadableStreamReader = (new ReadableStream()).getReader().constructor;
+  });
+  t.end();
+});
+
+test('Can get the ReadableStreamController constructor indirectly', t => {
+  t.doesNotThrow(() => {
+    // It's not exposed globally, but we test a few of its properties here.
+    new ReadableStream({
+      start(c) {
+        ReadableStreamController = c.constructor;
+      }
+    });
   });
   t.end();
 });
@@ -44,6 +57,14 @@ function fakeReadableStreamReader() {
     cancel(reason) { return Promise.resolve(); },
     read() { return Promise.resolve({ value: undefined, done: true }); },
     releaseLock() { return; }
+  };
+}
+
+function fakeReadableStreamController() {
+  return {
+    close() { },
+    enqueue(chunk) { },
+    error(e) { }
   };
 }
 
@@ -150,7 +171,7 @@ test('ReadableStream.prototype.pipeTo works generically on its this and its argu
 
 test('ReadableStreamReader enforces a brand check on its argument', t => {
   t.plan(1);
-  t.throws(() => new ReadableStreamReader(fakeReadableStream()), /TypeError/, 'Contructing a ReadableStreamReader ' +
+  t.throws(() => new ReadableStreamReader(fakeReadableStream()), /TypeError/, 'Constructing a ReadableStreamReader ' +
     'should throw');
 });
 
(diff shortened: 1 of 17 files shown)


reference-implementation/test/brand-checks.js
```diff
@@ -47,6 +60,14 @@ function fakeReadableStreamReader() {
   };
 }
 
+function fakeReadableStreamController() {
+  return {
+    close() {},
+    enqueue(chunk) { },
+    error(e) { }
```
tyoshino: style nitpicking: use either of {} or { }?

tyoshino:
lgtm a62b4f3

tyoshino:
Took a look at all the diff lines. I'm not sure if there some `t.throws` based tests remain not-converted since even if you forget to update, the function may throw a **TypeError** for example if the function is set to **undefined** by mistake.

domenic:
Great idea. I did a search for `/TypeError/` and it looks like we're clear. But much appreciated.

PR #311, opened 2015-03-31
Implement teeing, plus a bunch of stage-setting tweaks

domenic:
This PR supercedes #302. It is based on the same algorithm, but fully formalized and moved into the spec.

To get it formalized, I had to do a lot of cleanup around other abstract operations---mostly factoring out abstract operations from places that used to be accessible only through the public API. This should help create a better version of https://github.com/yutakahirano/fetch-with-streams/pull/28, for example. Each of the commits should be pretty simple to review individually.

The only real decisions that I want to highlight are:

- From https://github.com/whatwg/streams/pull/302#issuecomment-83484519, I chose "both cancels reject".
- The public `rs.tee()` method does not do any cloning; the TeeReadableStream abstract operation takes a second parameter that lets you decide whether to do structured cloning or not. Presumably for uses that will share across threads, like much of fetch, we'll need cloning.
- Should we expose `rs.tee()` at all? I mostly exposed it because otherwise my unit testing setup was not able to test it :P. But I can work around that, certainly. Maybe it is best to leave it off for now? I am a bit afraid that having non-cloning `rs.tee()` available would be a footgun for fetch body streams, given that many of the things you do with fetch body streams involve shuffling them off to another thread. Or do we assume that if you're going to do such things, you'll use the whole `response` object, and if you do `response.body.tee()`, you know what you're getting in to? That also seems plausible.

Branch snapshot at https://streams.spec.whatwg.org/branch-snapshots/tee-again/; sections of note:

- https://streams.spec.whatwg.org/branch-snapshots/tee-again/#tee-a-readable-stream
- https://streams.spec.whatwg.org/branch-snapshots/tee-again/#rs-tee
- https://streams.spec.whatwg.org/branch-snapshots/tee-again/#tee-readable-stream

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -693,20 +693,8 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
 </div>
 
 <pre is="emu-alg">
-  1. If IsReadableStreamReader(*this*) is *false*, throw a *TypeError* exception.
-  1. If *this*@[[state]] is "closed", return a new promise resolved with CreateIterResultObject(*undefined*, *true*).
-  1. If *this*@[[state]] is "errored", return a new promise rejected with *this*@[[storedError]].
-  1. Assert: *this*@[[ownerReadableStream]] is not *undefined*.
-  1. Assert: *this*@[[ownerReadableStream]]@[[state]] is "readable".
-  1. If *this*@[[ownerReadableStream]]@[[queue]] is not empty,
-    1. Let _chunk_ be DequeueValue(*this*@[[ownerReadableStream]]@[[queue]]).
-    1. If *this*@[[ownerReadableStream]]@[[closeRequested]] is *true* and *this*@[[ownerReadableStream]]@[[queue]] is now empty, call-with-rethrow FinishClosingReadableStream(*this*@[[ownerReadableStream]]).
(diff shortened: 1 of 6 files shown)


domenic:
@tyoshino hoping to merge this tomorrow, would be great if you had review time today :). They're not too bad, I promise.

tyoshino:
Sorry, Domenic. I was OOO on Friday. I've done reviewing till the spec side change of b326a6ebb0e67e5a57faecac0285dc003bce07bf. I think you can go ahead committing it. I'll leave comments and we can discuss them later.

domenic:
Thanks for the review! Updating and merging. Links to outstanding discussions:

- https://github.com/whatwg/streams/pull/311#commitcomment-10582068
- https://github.com/whatwg/streams/pull/311#commitcomment-10583524
- https://github.com/whatwg/streams/pull/311#commitcomment-10583567 (maybe resolved with new paragraph)

tyoshino:
Wow, all the links you've kindly listed are revoked as they were placed on this issue... Sad.

- https://github.com/whatwg/streams/commit/f6d78821a8e4142cf8e9b6c42700020111bb17ed#commitcomment-10582068

annevk:
I would be in favor of not exposing `tee()` if we don't know exactly how it needs to work yet. Especially if what we need for Fetch is available through other means.

PR #315, opened 2015-04-07
Added some more readable stream tests

calvaris:


calvaris:
The last two commits can be more controversial, because the function to force gc can be weird, but so far so good.

calvaris:
I rebased the branch and updated it with the remaining tests that could be (or not included). Let me know if there is any problem with them.

domenic:
All merged!!

calvaris:
Awesome, thanks!

PR #317, opened 2015-04-07
Precise flow control for readable streams

domenic:
Please read the commit messages for more detail.

Review welcome, of course, but also soliciting feedback on:

- desiredSize name
- Should we move strategy to a second constructor parameter, instead of putting it on the underlying source and sink?

Link to notably-changed spec sections:

- https://streams.spec.whatwg.org/branch-snapshots/precise-flow-control/#queuing-strategies
- https://streams.spec.whatwg.org/branch-snapshots/precise-flow-control/#rs-constructor
- https://streams.spec.whatwg.org/branch-snapshots/precise-flow-control/#other-stuff (both built-in queuing strategies changed significantly)

PR #326, opened 2015-04-10
Allow enqueue after cancel, plus fix #314

### LINKED ISSUE #314: pipeTo with preventClose: true returns a promise that is probably fulfilled too early
In IRC @tyoshino pointed out a potential issue with our pipeTo implementation. Related issues were previously discussed in https://github.com/whatwg/streams/issues/236#issue-46428456.

If you have `preventClose: false` (the default), then the returned promise will be fulfilled only once all the writes have successfully completed and the close succeeds.

However if you have `preventClose: true`, the returned promise will fulfill immediately once `source` is closed (empty) and `dest.ready` fulfills. This is probably earlier than the point in time when the last write succeeds.

I think the expected behavior for `preventClose: true` is for the promise to fulfill once the last write to `dest` succeeds.

This should be pretty easy to fix.

domenic:
On top of #317.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -291,14 +291,15 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
       state and queue of this stream
   </tr>
   <tr>
-    <td>\[[pullScheduled]]
-    <td>A boolean flag set to <b>true</b> when the underlying source's <code>pull</code> method is scheduled to be
-      called again after the current call to it finishes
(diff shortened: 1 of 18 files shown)


domenic:
@tyoshino fixed the redundant line. Will merge once #317 is LGTMed.

tyoshino:
thanks. lgtm!

PR #337, opened 2015-04-18
Avoid using binary

annevk:
Binary is 0 or 1 which is not exactly what we’re dealing with here.

(Somewhat unfortunate WebSocket used it as a term for a property.)


domenic:
I guess I was using it as an alternative to "textual data" e.g. in the sense of http://stackoverflow.com/questions/6855712/why-does-git-treat-this-text-file-as-a-binary-file. But yeah probably best to take this to ensure nobody is confused.

Issue #312, opened 2015-04-01
Extract the minimal algorithm from which we can easily infer the public behavior of the interfaces


tyoshino:
We've been sporadically saying that we need a definition of the interfaces not by JS code but by describing publicly visible behavior of each property.

I'm trying to build it at:
https://github.com/whatwg/streams/blob/asyncbytestream/BinaryExtension.md

IIRC, @domenic said that the current spec also includes some texts doing this. But I couldn't find it.

domenic:
Hmm, there might be some confusion here.

The current spec does define the interface in terms of its publicly visible behavior. It does so by giving one algorithm that will produce the normatively-required results. But if you have another algorithm that would do so, and is black-box indistinguishable, that is also a fine way of implementing the spec.

For example, the queue-with-sizes data structure is fantastically inefficient as specced. Presumably one would instead keep track of the sizes as things are added and removed from the queue. The key thing is that that behavior is black-box indistinguishable from the behavior specced (which sums up all the sizes every once in a while).

The danger of something like https://github.com/whatwg/streams/blob/asyncbytestream/BinaryExtension.md is that it does not give precision to interactions between different parts of the API, especially things like interaction between the underlying source and the stream's methods, or the sequencing of different events (calling read() twice, etc.). I think the better path is to produce a full algorithmic specification of the methods (and validate the algorithm with a reference implementation), as we have been doing.

From another angle, perhaps the main part of this issue is the idea of how to state invariants and contracts that must apply to all readable streams. That's what I think the templated tests accomplish. We could expand those a bit (to allow porting more existing tests) if we allowed templated "actions", e.g., create a stream, run some test code, then run an action like erroring the stream, then run some more test code.

tyoshino:
The text in BinaryExtension.md is still premature. We of course need to describe that. Actually, the text already includes the state concept to guarantee some invariants, preconditions. So, already not fully blackboxed. In prototyping the thin streams in the operationStream branch, I tried to describe the constraints using thin layer of algorithm. I want to do that in natural language..

It's possible to infer the normatively-required results from what we have in the spec now (with a queue), but I want to try to do more decoupling.

tyoshino:
By "natural language", I mean writing spec texts. I don't mean that we shouldn't write algorithm.

domenic:
Awesome, yeah :). Then the text above it is similar to the notes, e.g. in https://streams.spec.whatwg.org/#rs-get-reader ?

[...]

Issue #300, opened 2015-03-17
Byte streams


tyoshino:
This is a tracking bug for progress on byte stream standardization.

Status:
- We've recently switched to adopt async read() (see #253).
- #176 is now obsolete.
- Requirements are summarized in the "Byte Streams" section in [Requirements.md](https://github.com/whatwg/streams/blob/master/Requirements.md).

Sub issues:
- [ ] Naming the reader getter methods #294

tyoshino:
Updating BinaryExtension.md at https://github.com/whatwg/streams/tree/asyncbytestream

tyoshino:
Added definitions of methods on the BYOB reader https://github.com/whatwg/streams/blob/asyncbytestream/BinaryExtension.md

I tried to write the definitions without introducing detailed internal algorithm. @yutakahirano, could you please take a quick look?

yutakahirano:
looks good

tyoshino:
Thanks. Also added ReadableByteStreamReader.

tyoshino:
I was trying to fulfill #312 in BinaryExtension.md in asyncbytestream branch. But I've moved the place to experiment that to PR #343. The BinaryExtension.md file in asyncbytestream branch still holds some draft and explainer texts.

domenic:
The explainer texts are really nice IMO. We should consider maintaining something like them (probably for all the other classes too) in a docs/ directory or something. Or maybe just work on them in MDN, haha.

tyoshino:
> The explainer texts are really nice IMO. We should consider maintaining something like them (probably for all the other classes too) in a docs/ directory or something. Or maybe just work on them in MDN, haha.

Yeah. As a first step, done moving in PR #351

tyoshino:
Spec text update just started at https://github.com/whatwg/streams/tree/rbscontroller

tyoshino:
Added an item to discuss whether enqueue() should throw or not. Moved from https://code.google.com/p/chromium/issues/detail?id=577097#c4

tyoshino:
> enqueue() should check type or not

Now `ReadableStreamBYOBController.enqueue()` is specified to perform brand checking by checking presence of the `[[ViewedArrayBuffer]]` internal slot. Filled the checkbox in the OP.

PR #343, opened 2015-04-22
Add ReadableByteStream definition

tyoshino:


tyoshino:
WIP

domenic:
https://streams.spec.whatwg.org/branch-snapshots/readablebytestreamspec/ for preview. Has some minor errors. I will push an extra commit to fix those (but won't force-push in case you have work locally).

domenic:
In terms of the MVP for helping MSE, I think we want:

- Complete class definitions
- AcquireReadableByteStreamReader(rbs) and ReadFromReadableByteStreamReader(reader, destView)
- The text sections filled in ("Readable Byte Streams" and "Using Readable Byte Streams").

I think it'd also be ideal to not have any known-inaccurate algorithms, and for the ones we're unsure on have a warning `<div>`.

tyoshino:
Thanks for the advice. I've pushed one more commit. Yes, I found that dfn was missing for underlying byte stream but couldn't push the fixed one yesterday.

I tried to define ReadableByteStream without the default built-in queue. If it turns out to be not promising, I'll just revert them to just refer to ReadableStream's algorithms.

BTW, when I run ecmarkupify, npm complains like "... node_modules/ecmarkupify/package.json not found". Do you have any ideas what's happening?

domenic:
> BTW, when I run ecmarkupify, npm complains like "... node_modules/ecmarkupify/package.json not found". Do you have any ideas what's happening?

The only thing I can think of is make sure you are on a recent io.js and then do `rm -rf node_modules && npm install`. This should be done in the root, not in reference_implementation (which is a separate package).

tyoshino:
Thanks! I didn't know I needed io.js. It works now.

tyoshino:
Ready for review. This PR should also satisfy #312.

domenic:
Will try to do review tonight or tomorrow morning.

What are your thoughts on a reference implementation/tests?

index.bs
```diff
@@ -1874,6 +1882,585 @@ A few abstract operations are used in this specification for utility purposes. W
   1. Return Record{[[size]]: size, [[highWaterMark]]: highWaterMark}.
 </pre>
 
+<h2 id="rbs">Readable Byte Streams</h2>
+
+<h3 id="rbs-intro">Using Readable Byte Streams</h3>
+
+<div class="example">
```
domenic: These informative sections don't seem quite complete. But, I can try my hand at expanding them.

index.bs
```diff
@@ -1874,6 +1882,585 @@ A few abstract operations are used in this specification for utility purposes. W
...
+
+If one were to write the <code>ReadableByteStream</code> class in something close to the syntax of [[!ECMASCRIPT]], it
+would look like
+
+<pre><code class="lang-javascript">
+  class ReadableByteStream {
+    constructor(underlyingByteSource = {})
+
+    cancel(reason)
+    getReader()
```
domenic: Without getByobReader() or similar, I don't think we can satisfy MSE :-/.

index.bs
```diff
@@ -1874,6 +1882,585 @@ A few abstract operations are used in this specification for utility purposes. W
...
+
+<table>
+  <thead>
+    <tr>
+      <th>Internal Slot</th>
+      <th>Description (<em>non-normative</em>)</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>\[[closedPromise]]
```
domenic: Hmm, why is [[closedPromise]] here instead of on the reader? Sorry if the answer becomes obvious after reading further.

domenic:
OK, I guess I couldn't resist at least a high-level review.

I am a bit confused by the controller + underlying source model here. From what I can see, the idea is that pull() is called on each read, and each time it is called, you can call c.resolve(chunk) once. Is that the idea?

I thought though that ReadableByteStream was going to need an internal queue as well. Because what if someone doesn't read for a long time, and the kernel buffer starts getting full? Do we make underlying byte source implementers do their own buffering to handle that case?

Also I know we had a discussion in person which concluded that `c.resolve(chunk)` was better than `return chunk` or `return promiseForChunk`, mainly having to do with cancelation. Can you help remind me why that was important?

The `resolve` name is probably not the best but we can figure it out later. (Maybe it is fine but it feels like it's too tied to promise terminology, whereas promises are kind of more an implementation detail, that the underlying byte source should not need to know about.)

index.bs
```diff
@@ -1874,6 +1882,585 @@ A few abstract operations are used in this specification for utility purposes. W
...
+    1. Set _reader_@[[state]] to "closed".
+    1. Resolve _reader_@[[closedPromise]] with *undefined*.
+    1. Repeat for each _readRequestPromise_ that is an element of _reader_@[[readRequests]],
+      1. Resolve _readRequestPromise_ with CreateIterResultObject(*undefined*, *true*).
+  1. Set _reader_@[[readRequests]] to a new empty List.
+  1. Set _reader_@[[ownerReadableByteStream]]@[[reader]] to *undefined*.
+  1. Set _reader_@[[ownerReadableByteStream]] to *undefined*.
+</pre>
+
+<h4 id="tee-readable-byte-stream" aoid="TeeReadableByteStream">TeeReadableByteStream ( stream, shouldClone )</h4>
```
domenic: Hmm, this is complicated. With the queues and such. Could you summarize for me how it works? It looks similar to TeeReadableStream but I am betting there are important differences.

tyoshino:
> I am a bit confused by the controller + underlying source model here. From what I can see, the idea is that pull() is called on each read, and each time it is called, you can call c.resolve(chunk) once. Is that the idea?

Right. `pull()` is just a notification of `read()`. The user can call `c.resolve(chunk)` at any time.

The readable byte stream is kept as minimum as possible but able to represent invariants, constraints, semantics of the API surface completely. The controller is not designed considering usability. Maybe, I should not have added the controller API but only define abstract operations to manipulate the readable byte stream's state.

> I thought though that ReadableByteStream was going to need an internal queue as well. Because what if someone doesn't read for a long time, and the kernel buffer starts getting full? Do we make underlying byte source implementers do their own buffering to handle that case?

The latter.

> Also I know we had a discussion in person which concluded that c.resolve(chunk) was better than return chunk or return promiseForChunk, mainly having to do with cancelation. Can you help remind me why that was important?

You were imagining some example implementation of the underlying source where you're trying to encapsulate an async API that returns a promise. And you said that we basically want to return the promise (as-is or after some transform) than converting it to call to `c.resolve()`, etc.

> The resolve name is probably not the best but we can figure it out later. (Maybe it is fine but it feels like it's too tied to promise terminology, whereas promises are kind of more an implementation detail, that the underlying byte source should not need to know about.)

Was using fulfill. Fulfill might be better as the streams are supposed to be enqueued non-promise values?

index.bs
tyoshino: I copy&pasted this from ReadableStream. It looks ReadableStream should also be fixed by removing this from ReadableStream. I'll do it.

tyoshino: PR #352

domenic:
> The readable byte stream is kept as minimum as possible but able to represent invariants, constraints, semantics of the API surface completely. The controller is not designed considering usability. Maybe, I should not have added the controller API but only define abstract operations to manipulate the readable byte stream's state.

I see. So we may want to make it clear that this is still undergoing revision. But the public API will be OK.

>> Do we make underlying byte source implementers do their own buffering to handle that case?

> The latter.

This seems sad. Also I think we wanted to support buffering (in the non-BYOB case) to avoid "jitter"? Hmm.

> You were imagining some example implementation of the underlying source where you're trying to encapsulate an async API that returns a promise. And you said that we basically want to return the promise (as-is or after some transform) than converting it to call to c.resolve(), etc.

Right, I remember the argument for using promises instead of `c.resolve` :). But I forget the argument for using `c.resolve` instead of promises. Sorry, I guess I was a bit unclear what my question was there.

> Was using fulfill. Fulfill might be better as the streams are supposed to be enqueued non-promise values?

I meant more pick a verb that has nothing to do with promises. Looking through a thesaurus, here are some random options: answer, conclude, fill, finish, satisfy, produce, complete, finalize, respond.

---

Anyway, let's start merging stuff and iterating on it in further PRs and issues, using #300 as the central tracking issue once the dust settles. I think we should squash everything into a single commit that does the following:

- Remove ReadableByteStream reference implementation and tests
- Contains the contents of both this PR and #351

If this sounds good to you, let me know. If you are ready to go to sleep then I will do it (and file a few follow-up issues/PRs). Or if you want to take care of some other stuff first before we merge then that's good too :).

tyoshino:
> Right, I remember the argument for using promises instead of c.resolve :). But I forget the argument for using c.resolve instead of promises. Sorry, I guess I was a bit unclear what my question was there.

IIRC, it was something like race between:
- the stream getting updated by the fulfillment/rejection callback set to the returned promise
- controller.close()

We were drawing timeline on a whiteboard to discuss this.

domenic:
Yeah. I just wish we had captured it somewhere digital, haha.

tyoshino:
> If you are ready to go to sleep then I will do it (and file a few follow-up issues/PRs).

Thanks. I'm adding BYOB stuff in local copy but that doesn't block the work. I'm still in the office but wouldn't get time to do squashing, etc. before last train.

So, please :)

domenic:
Merged as 94a7fc13e5916494322787af0be8651547d81c60 per plan at https://github.com/whatwg/streams/pull/343#issuecomment-98135291.

PR #349, opened 2015-04-30
Add tests for checking that tee() branches are errored/closed when the o...

tyoshino:
...riginal stream is errored/closed


tyoshino:
I suspected that the close test of this PR might not work since TeeReadableStream() doesn't set fulfillment callback to reader@[[closedPromise]]. But I realized that since the branches have highWaterMark of 1, it's guaranteed that they have active reader.read() call with which they can get notified of closure as far as their queue is empty. So, no problem. Right?

domenic:
Yeah, this LGTM. Feel free to merge if I don't get to it soon. The intent was indeed that you use read() to get notifed of closure, and it's great to have tests to validate.

tyoshino:
Thanks. Merged.

PR #351, opened 2015-05-01
Update byte stream spec

tyoshino:
- Moved all the effort to normatively defining the byte stream components to #343 
- Put casual explainer to docs/Explainer.md (as suggested at https://github.com/whatwg/streams/issues/300#issuecomment-97831102)


domenic:
Merged as 94a7fc13e5916494322787af0be8651547d81c60 per plan at https://github.com/whatwg/streams/pull/343#issuecomment-98135291.

PR #356, opened 2015-05-06
Make the context be the transform object when calling transform and flush

edhager:
I know the TransformStream is a work in progress...

With the parameter destructuring used in the constructor of TransformStream class, the global context ends up being used when transform and flush are called.  That makes it difficult to use a transform object that wants to use its own state.

Here is a sample test and a change to transform-stream that allows my test to pass.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -3,7 +3,7 @@ import WritableStream from './writable-stream';
 
 export default class TransformStream {
   constructor(transformer) {
-    if (transformer.flush == null) {
+    if (transformer.flush === undefined) {
       transformer.flush = (enqueue, close) => close();
     }
 
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -2,8 +2,12 @@ import ReadableStream from './readable-stream';
 import WritableStream from './writable-stream';
 
 export default class TransformStream {
-  constructor({ transform, flush = (enqueue, close) => close(), writableStrategy, readableStrategy }) {
-    if (typeof transform !== 'function') {
+  constructor(transformer) {
+    if (transformer.flush == null) {
+      transformer.flush = (enqueue, close) => close();
+    }
+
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -2,8 +2,12 @@ import ReadableStream from './readable-stream';
 import WritableStream from './writable-stream';
 
 export default class TransformStream {
-  constructor({ transform, flush = (enqueue, close) => close(), writableStrategy, readableStrategy }) {
-    if (typeof transform !== 'function') {
+  constructor(transformer) {
+    if (transformer.flush == null) {
```
domenic: === undefined, please

reference-implementation/test/transform-stream.js
```diff
@@ -367,3 +367,25 @@ test('TransformStream flush gets a chance to enqueue more into the readable, and
   })
   .catch(e => t.error(e));
 });
+
+test('Transform context', t => {
```
domenic: Test name should be "Transform stream should call transformer methods as methods"

reference-implementation/test/transform-stream.js
```diff
@@ -367,3 +367,25 @@ test('TransformStream flush gets a chance to enqueue more into the readable, and
...
+    },
+
+    flush(enqueue, close) {
+      close();
+    }
+  });
+
+  ts.writable.write('a');
+  ts.writable.close();
+  ts.readable.getReader().read().then((result) => {
```
domenic: No extra parens around result

reference-implementation/test/transform-stream.js
domenic: Needs to test flush as well.

domenic:
Thanks for this! Some things to touch up in the code and test but this is definitely the intent (see https://github.com/whatwg/streams/pull/265 for the earlier change to readable and writable streams).

edhager:
I made the suggested adjustments.

domenic:
Thanks! Merged as 2baa51e7de6e345102fb551d38395a151172a2aa

Issue #355, opened 2015-05-03
ReadableByteStream needs reference implementation/tests


domenic:
Can't really count on these APIs being ready until we've validated them :)

- [x] Add RBS and basic tests for it: #361
- [ ] Add RBS to the templated tests
- [ ] Sync spec and reference implementation

tyoshino:
This is happening at https://github.com/whatwg/streams/tree/bytestream

tyoshino:
Domenic created a PR https://github.com/whatwg/streams/pull/361

tyoshino:
Added TODO list to OP

tyoshino:
Now #361 is done. Updated the OP by adding it.

tyoshino:
Merged the tasks into https://github.com/whatwg/streams/issues/300. Let's close this.

Issue #353, opened 2015-05-03
Readable byte streams should support an internal queue


domenic:
The current draft does not include such a queue; there is a one-to-one map of calls to `rbsReader.read()` and `rbsUnderlyingSource.pull()`. I think this is potentially problematic, for two reasons:

- Kernel buffers can become full, if I understand correctly. Thus, the code for `rbsUnderlyingSource.pull()` will not be a simple call to `read(2)` or similar. Instead, the underlying source author will need to do their own buffering. That seems bad.
- We want readable byte streams to be resistant to uneven flow: e.g., if the consumer slows down for a bit, then gets much faster, then slows down for a bit, it would be better to have some chunks read ahead of time to so that we can match their pace (up to a high water mark).

This does make things more complicated though for the BYOB case, as we have to figure out what happens if the internal queue has things in it and then we switch to a BYOB reader. This was discussed at https://github.com/whatwg/streams/issues/177#issuecomment-83984473

@tyoshino, have your thoughts changed since that discussion? Or is the current draft lacking a queue just as a first-draft thing?

tyoshino:
The boundary between the underlying byte source and the readable byte stream doesn't need to correspond to any specific layer boundary. It just depends on the philosophy the author of the spec gave to the spec. When we want to drain some data to one world (storage, kernel, process) from another (another storage, userland, another process), they're always allowed to do that in various means.

Having some strategy-controlled buffer example in the spec is good. But the spec shouldn't impress the readers that they must implement the buffer in JavaScript. My suggestion is having two separate classes. One is just describing all the publicly visible requirements (including invariants, interaction between methods, properties, etc.), and the other is a strategy-controlled buffer which is convenient for push source implementor. The buffer class is good example for explaining how to interpret backpressure and controlling buffer size, data generation in general. That can be referred to when writing non-JavaScript buffering components. And of course, it's also useful as a built-in JS library.

I want the separation especially for byte stream. If the source is push source, then it's convenient for the implementor of the source if the stream has a built-in buffer. In that case, the source doesn't understand BYOB style pulling, so we need to take care of the gap by copying the contents to the view provided by the consumer. But this is not expected for a source that understands BYOB. For BYOB read requests, the source just wants to get the view and fulfill the corresponding pending request with the view (possibly detached) filled with generated contents as-is. There shouldn't any automatic copy. Having a queue even for BYOB-capable source may be useful. But the queue system for BYOB-capable source and one for push source would be very different.

[...]

PR #361, opened 2015-06-11
Byte stream update, including reference implementation

domenic:
Doing a PR of @tyoshino's changes so I can do line comments.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -187,12 +187,11 @@ similarly, avoiding writes that would cause the desired size to go negative.
 
 <!-- TODO: writable streams too, probably -->
 
-A <dfn>readable stream reader</dfn> (<dfn>readable byte stream reader</dfn> or
-<dfn>readable byte stream byob reader</dfn> for <a>readable byte stream</a>) or simply reader is an object that allows
-direct reading of <a>chunks</a> from a <a>readable stream</a>). Without a reader, a <a>consumer</a> can only perform
-high-level operations on the readable stream: waiting for the stream to become closed or errored,
-<a lt="cancel a readable stream">canceling</a> the stream, or <a>piping</a> the readable stream to a writable stream.
-Many of those high-level operations actually use a reader themselves.
+A <dfn>readable stream reader</dfn> (<dfn>readable byte stream reader</dfn> for <a>readable byte stream</a>) or simply
+reader is an object that allows direct reading of <a>chunks</a> from a <a>readable stream</a>). Without a reader, a
+<a>consumer</a> can only perform high-level operations on the readable stream: waiting for the stream to become closed
+or errored, <a lt="cancel a readable stream">canceling</a> the stream, or <a>piping</a> the readable stream to a
+writable stream. Many of those high-level operations actually use a reader themselves.
 
 A given readable stream only has at most one reader at a time. We say in this case the stream is
 <dfn lt="locked to a reader">locked to the reader</dfn>, and that the reader is <dfn lt="active reader">active</dfn>.
@@ -1270,7 +1269,7 @@ Instances of <code>WritableStream</code> are created with the internal slots des
   <tr>
     <td>\[[queue]]
     <td>A List representing the stream's internal queue of pending writes
-.  </tr>
+  </tr>
   <tr>
     <td>\[[started]]
     <td>A boolean flag indicating whether the <a>underlying sink</a> has finished starting
@@ -1826,6 +1825,11 @@ A few abstract operations are used in this specification for utility purposes. W
   1. Return the result of calling the get accessor function of %TypedArray%.prototype.buffer on _O_ passing no arguments.
 </pre>
 
+<div class="note">
+  Ideally, this operation and its counterparts <a href="https://bugs.ecmascript.org/show_bug.cgi?id=4369">would be
+  defined by ECMAScript</a>.
+</div>
+
 <h4 id="get-byte-length" aoid="GetByteLength">GetByteLength ( O )</h4>
 
 <pre is="emu-alg">
@@ -1987,7 +1991,7 @@ Instances of <code>ReadableByteStream</code> are created with the internal slots
 
 <pre is="emu-alg">
   1. Set *this*@[[state]] to "readable".
-  1. Set *this*@[[reader]] and *this*@[[storedError]] to *undefined*.
+  1. Set *this*@[[reader]], *this*@[[storedError]] and *this*@[[controller]] to *undefined
   1. Set *this*@[[controller]] to Construct(`ReadableByteStreamController`, «*this*, _underlyingByteSource_»).
 </pre>
 
@@ -2005,46 +2009,33 @@ Instances of <code>ReadableByteStream</code> are created with the internal slots
 
 <pre is="emu-alg">
   1. If IsReadableByteStream(*this*) is *false*, throw a *TypeError* exception.
-  1. Return AcquireReadableByteStreamByobReader(*this*).
+  1. Return Construct(`ReadableByteStreamByobReader`, «*this*»).
 </pre>
 
 <h5 id="rbs-get-reader">getReader()</h5>
 
 <pre is="emu-alg">
   1. If IsReadableByteStream(*this*) is *false*, throw a *TypeError* exception.
-  1. Return AcquireReadableByteStreamReader(*this*).
+  1. Return Construct(`ReadableByteStreamReader`, «*this*»).
 </pre>
 
 <h5 id="rbs-pipe-through">pipeThrough({ writable, readable }, options)</h5>
 
-<pre is="emu-alg">
-  1. Call-with-rethrow Invoke(*this*, "pipeTo", «_writable_, _options_»).
-  1. Return _readable_.
-</pre>
+TBA
 
 <h5 id="rbs-pipe-to">pipeTo(dest, { preventClose, preventAbort, preventCancel } = {})</h5>
 
-The <code>pipeTo</code> method is one of the more complex methods, and is undergoing some revision and edge-case
-bulletproofing before we write it up in prose.
-
-For now, please consider the reference implementation normative:
-<a href="https://github.com/whatwg/streams/blob/master/reference-implementation/lib/readable-stream.js">reference-implementation/lib/readable-stream.js</a>,
-look for the <code>pipeTo</code> method.
+TBA
 
 <h5 id="rbs-tee">tee()</h5>
 
-<pre is="emu-alg">
-  1. If IsReadableByteStream(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _branches_ be TeeReadableByteStream(*this*, *false*).
-  1. ReturnIfAbrupt(_branches_).
-  1. Return CreateArrayFromList(_branches_).
-</pre>
+TBA
 
 <h3 id="rbs-controller-class" lt="ReadableByteStreamController">Class <code>ReadableByteStreamController</code></h3>
 
-The <code>ReadableByteStreamController</code> class has methods that allow control of a <code>ReadableStream</code>'s
-state. When constructing a <code>ReadableByteStream</code>, the <a>underlying byte source</a> is given a
-corresponding <code>ReadableByteStreamController</code> instance to manipulate.
+The <code>ReadableByteStreamController</code> class has methods that allow control of a
+<code>ReadableByteStream</code>'s state. When constructing a <code>ReadableByteStream</code>, the
+<a>underlying byte source</a> is given a corresponding <code>ReadableByteStreamController</code> instance to manipulate.
 
 <h4 id="rbs-controller-class-definition">Class Definition</h4>
 
@@ -2060,132 +2051,21 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
     close()
     enqueue(chunk)
     error(e)
-    respond(chunk)
+    respond(bytesWritten, buffer)
   }
 </code></pre>
 
 <h4 id="rbs-controller-internal-slots">Internal Slots</h4>
 
-Instances of <code>ReadableByteStreamController</code> are created with the internal slots described in the following
-table:
-
-<table>
-  <thead>
-    <tr>
-      <th>Internal Slot</th>
-      <th>Description (<em>non-normative</em>)</th>
-    </tr>
-  </thead>
-  <tr>
-    <td>\[[consumedBytesOfHeadOfQueue]]
-    <td>The number of bytes of the first element in \[[queue]] which are already consumed
-  </tr>
-  <tr>
-    <td>\[[controlledReadableByteStream]]
-    <td>The <code>ReadableByteStream</code> instance controlled
-  </tr>
-  <tr>
-    <td>\[[pendingViews]]
-    <td>A List of objects implementing the ArrayBufferView type provided by the consumer via
-      <code>ReadableByteStreamByobReader</code>; used only when the provided underlying byte source doesn't have the
-      "pullInto" method.
-  </tr>
-  <tr>
-    <td>\[[queue]]
-    <td>A List representing the stream's internal queue of pending writes
-  </tr>
-  <tr>
-    <td>\[[underlyingByteSource]]
-    <td>An object representation of the stream's <a>underlying byte source</a>
-  </tr>
-</table>
+TBA
 
 <h4 id="rbs-controller-constructor">new ReadableByteStreamController(controlledReadableByteStream, underlyingByteSource)</h4>
 
-<pre is="emu-alg">
-  1. If IsReadableByteStream(_controlledReadableByteStream_) is *false*, throw a *TypeError* exception.
-  1. If _controlledReadableByteStream_@[[controller]] is not *undefined*, throw a *TypeError* exception.
-  1. Let _pullFunction_ be GetV(_underlyingByteSource_, "pull").
-  1. ReturnIfAbrupt(_pullFunction_).
-  1. If _pullFunction_ is not *undefined*,
-    1. If IsCallable(_pullFunction_) is *false*, throw a *TypeError* exception.
-  1. Let _pullIntoFunction_ be GetV(_underlyingByteSource_, "pullInto").
-  1. ReturnIfAbrupt(_pullIntoFunction_).
-  1. If _pullIntoFunction_ is not *undefined*,
-    1. If IsCallable(_pullIntoFunction_) is *false*, throw a *TypeError* exception.
-  1. Set *this*@[[controlledReadableByteStream]] to _controlledReadableByteStream_.
-  1. Set *this*@[[underlyingByteSource]] to _underlyingByteSource_.
-  1. Set *this*@[[pendingViews]] be a new empty List.
-  1. Set *this*@[[queue]] be a new empty List.
-  1. Set *this*@[[consumedBytesOfHeadOfQueue]] to *0*;
-  1. Let _startResult_ be InvokeOrNoop(_underlyingByteSource_, "start", «*this*»).
-  1. ReturnIfAbrupt(startResult).
-</pre>
+TBA
 
 <h4 id="rbs-controller-prototype">Properties of the <code>ReadableByteStreamController</code> Prototype</h4>
 
-<h5 id="rbs-controller-close">close()</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamController(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _stream_ be *this*@[[controlledReadableByteStream]].
-  1. If *this*@[[closeRequested]] is *true*, throw a *TypeError* exception.
-  1. If _stream_@[[state]] is not "readable", throw a *TypeError* exception.
-  1. If *this*@[[queue]] is empty,
-    1. CloseReadableByteStream(_stream_).
-    1. Return *undefined*.
-  1. Set *this*@[[closeRequested]] to *true*.
-  1. Return *undefined*.
-</pre>
-
-<h5 id="rbs-controller-enqueue">enqueue(chunk)</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamController(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _stream_ be *this*@[[controlledReadableByteStream]].
-  1. If *this*@[[closeRequested]] is *true*, throw a *TypeError* exception.
-  1. If _stream_@[[state]] is not "readable", throw a *TypeError* exception.
-  1. Let _reader_ be _stream_@[[reader]].
-  1. If _reader_ is *undefined*,
-    1. EnqueueInReadableByteStreamController(*this*, _chunk_).
-  1. Otherwise,
-    1. If _reader_ has a [[readRequests]] internal slot,
-      1. If _reader_@[[readRequests]] is empty,
-        1. EnqueueInReadableByteStreamController(*this*, _chunk_).
-      1. Otherwise,
-        1. Assert: *this*@[[queue]] is empty.
-        1. RespondToReadableByteStreamReaderReadRequest(_stream_, _chunk_).
-    1. Otherwise, _reader_ must have a [[readIntoRequests]] internal slot so,
-      1. If _reader_@[[readIntoRequests]] is empty,
-        1. EnqueueInReadableByteStreamController(*this*, _chunk_).
-      1. Otherwise, throw a *TypeError* exception.
-  1. Return *undefined*.
-</pre>
-
-<h5 id="rbs-controller-error">error(e)</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamController(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _stream_ be *this*@[[controlledReadableByteStream]].
-  1. If _stream_@[[state]] is not "readable", throw a *TypeError* exception.
-  1. Let *this*@[[queue]] be a new empty List.
-  1. ErrorReadableByteStream(_stream_, e).
-  1. Return *undefined*.
-</pre>
-
-<h5 id="rbs-controller-respond">respond(chunk)</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamController(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _stream_ be *this*@[[controlledReadableByteStream]].
-  1. Let _reader_ be _stream_@[[reader]].
-  1. If _reader_ is *undefined*, throw a *TypeError* exception.
-  1. If _reader_ has a [[readRequests]] internal slot, throw a *TypeError* exception.
-  1. Assert: _reader_ has a [[readIntoRequests]] internal slot.
-  1. If _reader_@[[readIntoRequests]] is empty, throw a *TypeError* exception.
-  1. RespondToReadableByteStreamByobReaderReadIntoRequest(_stream_, _chunk_).
-  1. Return *undefined*.
-</pre>
+TBA
 
 <h3 id="byte-reader-class">Class <code>ReadableByteStreamReader</code></h3>
 
@@ -2210,97 +2090,15 @@ it would look like
 
 <h4 id="byte-reader-internal-slots">Internal Slots</h4>
 
-Instances of <code>ReadableByteStreamReader</code> are created with the internal slots described in the following table:
-
-<table>
-  <thead>
-    <tr>
-      <th>Internal Slot</th>
-      <th>Description (<em>non-normative</em>)</th>
-    </tr>
-  </thead>
-  <tr>
-    <td>\[[closedPromise]]
-    <td>A promise returned by the reader's <code>closed</code> getter
-  </tr>
-  <tr>
-    <td>\[[ownerReadableByteStream]]
-    <td>A <code>ReadableByteStream</code> instance that owns this reader
-  </tr>
-  <tr>
-    <td>\[[readRequests]]
-    <td>A List of promises returned by calls to the reader's <code>read()</code> method that have not yet been resolved,
-      due to the <a>consumer</a> requesting <a>chunks</a> sooner than they are available; also used for the
-      <a href="#is-readable-byte-stream-reader">IsReadableByteStreamReader</a> brand check
-  </tr>
-  <tr>
-    <td>\[[state]]
-    <td>A string containing the reader's current state, used internally; one of <code>"readable"</code>,
-      <code>"closed"</code>, or <code>"errored"</code>
-  </tr>
-  <tr>
-    <td>\[[storedError]]
-    <td>A value indicating how the reader's stream failed, to be given as a failure reason or exception when trying to
-      operate on the reader; applicable only when \[[state]] is <code>"errored"</code>
-  </tr>
-</table>
+TBA
 
 <h4 id="byte-reader-constructor">new ReadableByteStreamReader(stream)</h4>
 
-<div class="note">
-  The <code>ReadableByteStreamReader</code> constructor is generally not meant to be used directly; instead, a stream's
-  <code>getReader()</code> method should be used. This allows different classes of readable streams to vend different
-  classes of readers without the <a>consumer</a> needing to know which goes with which.
-</div>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStream(_stream_) is *false*, throw a *TypeError* exception.
-  1. If IsReadableByteStreamLocked(_stream_) is *true*, throw a *TypeError* exception.
-  1. Set _stream_@[[reader]] to *this*.
-  1. Set *this*@[[ownerReadableByteStream]] to _stream_.
-  1. Set *this*@[[state]] to "readable".
-  1. Set *this*@[[storedError]] to *undefined*.
-  1. Set *this*@[[readRequests]] to a new empty List.
-  1. Set *this*@[[closedPromise]] to a new promise.
-  1. If _stream_@[[state]] is "closed" or "errored", ReleaseReadableByteStreamReader(*this*).
-</pre>
+TBA
 
 <h4 id="byte-reader-prototype">Properties of the <code>ReadableByteStreamReader</code> Prototype</h4>
 
-<h5 id="byte-reader-closed">get closed</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamReader(*this*) is *false*, return a promise rejected with a *TypeError* exception.
-  1. Return *this*@[[closedPromise]].
-</pre>
-
-<h5 id="byte-reader-cancel">cancel(reason)</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamReader(*this*) is *false*, return a promise rejected with a *TypeError* exception.
-  1. If *this*@[[state]] is "closed", return a new promise resolved with *undefined*.
-  1. If *this*@[[state]] is "errored", return a new promise rejected with *this*@[[storedError]].
-  1. Assert: *this*@[[ownerReadableByteStream]] is not *undefined*.
-  1. Return CancelReadableByteStream(*this*@[[ownerReadableByteStream]], _reason_).
-</pre>
-
-<h5 id="byte-reader-read">read()</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamReader(*this*) is *false*, return a promise rejected with a *TypeError* exception.
-  1. Return ReadFromReadableByteStreamReader(*this*).
-</pre>
-
-<h5 id="byte-reader-release-lock">releaseLock()</h5>
-
-<pre is="emu-alg">
-  1. If IsReadableByteStreamReader(*this*) is *false*, throw a *TypeError* exception.
-  1. If *this*@[[ownerReadableByteStream]] is *undefined*, return *undefined*.
-  1. If *this*@[[readRequests]] is not empty, throw a *TypeError* exception.
-  1. SyncReadableByteStreamReaderStateWithOwner(*this*@[[ownerReadableByteStream]]).
-  1. DetachReadableByteStreamReader(*this*).
-  1. Return *undefined*.
-</pre>
+TBA
 
 <h3 id="byte-byob-reader-class">Class <code>ReadableByteStreamByobReader</code></h3>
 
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -187,11 +187,12 @@ similarly, avoiding writes that would cause the desired size to go negative.
 
 <!-- TODO: writable streams too, probably -->
 
-A <dfn>readable stream reader</dfn> (<dfn>readable byte stream reader</dfn> for <a>readable byte stream</a>) or simply
-reader is an object that allows direct reading of <a>chunks</a> from a <a>readable stream</a>). Without a reader, a
-<a>consumer</a> can only perform high-level operations on the readable stream: waiting for the stream to become closed
-or errored, <a lt="cancel a readable stream">canceling</a> the stream, or <a>piping</a> the readable stream to a
-writable stream. Many of those high-level operations actually use a reader themselves.
+A <dfn>readable stream reader</dfn> (<dfn>readable byte stream reader</dfn> or
```
domenic: I think on the "model" level we don't want to differentiate between different types of readers. So no need for the parenthetical at all, IMO.

index.bs
```diff
@@ -1269,7 +1270,7 @@ Instances of <code>WritableStream</code> are created with the internal slots des
   <tr>
     <td>\[[queue]]
     <td>A List representing the stream's internal queue of pending writes
-  </tr>
+.  </tr>
```
domenic: Typo

index.bs
```diff
@@ -1819,6 +1820,24 @@ throughout the rest of this standard.
 
 A few abstract operations are used in this specification for utility purposes. We define them here.
 
+<h4 id="get-viewed-array-buffer" aoid="GetViewedArrayBuffer">GetViewedArrayBuffer ( O )</h4>
+
```
domenic: Maybe add a link to https://bugs.ecmascript.org/show_bug.cgi?id=4369 in this first one. Something like

```
<div id="note">Ideally, this operation and its counterparts <a href="https://bugs.ecmascript.org/show_bug.cgi?id=4369">would be defined by ECMAScript</a>.</div>
```

index.bs
```diff
@@ -1960,22 +1981,14 @@ Instances of <code>ReadableByteStream</code> are created with the internal slots
     <td>A value indicating how the stream failed, to be given as a failure reason or exception when trying to operate
       on an errored stream
   </tr>
-  <tr>
-    <td>\[[underlyingByteSource]]
```
domenic: I think this change, and the accompanying ones where we extract functions from the UBS, is not great. We should keep a reference to it and call its methods as methods. See https://github.com/whatwg/streams/issues/256 and https://github.com/whatwg/streams/pull/265.

index.bs
```diff
@@ -2056,17 +2077,49 @@ table:
...
+    <td>The number of bytes of the first element in \[[queue]] which are already consumed
+  </tr>
+  <tr>
     <td>\[[controlledReadableByteStream]]
     <td>The <code>ReadableByteStream</code> instance controlled
   </tr>
+  <tr>
+    <td>\[[pendingViews]]
+    <td>A List of objects implementing the ArrayBufferView type provided by the consumer via
+      <code>ReadableByteStreamByobReader</code>; used only when the provided underlying byte source doesn't have the
```
domenic: This isn't quite accurate, right? It's used whenever enqueue() is used.

index.bs
```diff
@@ -2056,17 +2077,49 @@ table:
...
+      "pullInto" method.
+  </tr>
+  <tr>
+    <td>\[[queue]]
+    <td>A List representing the stream's internal queue of pending writes
```
domenic: Also not quite accurate

index.bs
domenic: Oh, I see. You just moved it to the controller. Interesting! I wonder if we can/should do that for ReadableStream...

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,948 @@
...
+    if (IsReadableByteStream(controlledReadableByteStream) === false) {
+      throw new TypeError('ReadableByteStreamController can only be constructed with a ReadableByteStream instance');
+    }
+
+    if (controlledReadableByteStream._controller !== undefined) {
+      throw new TypeError(
+          'ReadableByteStreamController instances can only be created by the ReadableByteStream constructor');
+    }
+
+    const pullFunction = underlyingByteSource['pull'];
```
domenic: I prefer letting these fail at call time instead of immediately. The values could change over time and this kind of aggressive type checking is unnecessary.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,948 @@
...
+    }
+
+    const pullIntoFunction = underlyingByteSource['pullInto'];
+    if (pullIntoFunction !== undefined && typeof pullIntoFunction !== 'function') {
+      throw new TypeError('pullInto property of an underlying byte source must be a function');
+    }
+
+    this._controlledReadableByteStream = controlledReadableByteStream;
+
+    this._underlyingByteSource = underlyingByteSource;
```
domenic: Hmm, all these being on the controller seems a bit weird... we should probably be consistent between RBS and RS.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,948 @@
...
+        new TypeError(
+            'ReadableByteStreamByobReader.prototype.read can only be used on a ReadableByteStreamByobReader'));
+    }
+
+    if (view === undefined) {
+      return Promise.reject(new TypeError('Valid view must be provided'));
+    }
+
+    const ctor = view.constructor;
+    let elementSize = 1;
```
domenic: Can't we use `ctor.BYTES_PER_ELEMENT`? In spec-speak it looks like https://people.mozilla.org/~jorendorff/es6-draft.html#sec-typedarray.bytes_per_element

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,948 @@
...
+    if (this._state === 'errored') {
+      return Promise.reject(this._storedError);
+    }
+
+    assert(this._ownerReadableByteStream !== undefined, 'This stream must be attached to a stream');
+
+    return CancelReadableByteStream(this._ownerReadableByteStream, reason);
+  }
+
+  read(view) {
```
domenic: The spec has this factored out into an abstract operation

domenic: Or maybe the spec is just not updated yet, sorry.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,948 @@
...
+    if (!IsReadableByteStreamByobReader(this)) {
+      return Promise.reject(
+        new TypeError(
+            'ReadableByteStreamByobReader.prototype.read can only be used on a ReadableByteStreamByobReader'));
+    }
+
+    if (view === undefined) {
```
domenic: I think we want to use a stronger brand check. Check if it's Object, then check if it has a [[ViewedArrayBuffer]], like in https://people.mozilla.org/~jorendorff/es6-draft.html#sec-arraybuffer.isview. Once you do that you can probably actually use `view@[[ByteLength]]` etc directly, actually, without needing the abstract operations that throw if they're not present.

domenic: I realize the above may be confusing since I am talking mostly about how to write the spec while commenting on the code. For the code, I think just use `ArrayBuffer.isView` here, and then continue using `.byteLength` etc. as below.

reference-implementation/run-tests.js
```diff
@@ -6,16 +6,18 @@ import WritableStream from './lib/writable-stream';
...
 global.WritableStream = WritableStream;
 global.ByteLengthQueuingStrategy = ByteLengthQueuingStrategy;
 global.CountQueuingStrategy = CountQueuingStrategy;
 global.TransformStream = TransformStream;
+global.ReadableByteStream = ReadableByteStream;
 
 
 if (process.argv.length === 2) {
-  const tests = glob.sync(path.resolve(__dirname, 'test/*.js'));
+  const tests = glob.sync(path.resolve(__dirname, 'test/readable-byte-stream.js'));
```
domenic: Don't want to commit this.

reference-implementation/test/readable-byte-stream.js
```diff
@@ -0,0 +1,1415 @@
...
+    t.equals(view.constructor, Uint8Array);
+    t.equals(view.buffer.byteLength, 32);
+    t.equals(view.byteOffset, 0);
+    t.equals(view.byteLength, 32);
+
+    t.end();
+  });
+});
+
+test('ReadableByteStream: enqueue(), read(view) partially, then read()', t => {
```
domenic: Wow, these tests are awesome! It actually works!!

reference-implementation/test/readable-byte-stream.js
```diff
@@ -0,0 +1,1415 @@
...
+test('ReadableByteStream: read(), then respond()', t => {
+  let controller;
+
+  const rbs = new ReadableByteStream({
+    start(c) {
+      controller = c;
+    },
+    pull() {
+      try {
+        controller.respond(1);
```
domenic: This is throwing because it's respond()ing but not to a pullInto(), right? I think the test title should say that, and also you should respond with a Uint8Array or something, because otherwise I'd expect a TypeError anyway when respond()ing with a number.

domenic: Oh, I forgot that respond() takes a number as the first argument. Well, we should include both arguments, right?

reference-implementation/test/readable-byte-stream.js
```diff
@@ -0,0 +1,1415 @@
...
+    start(c) {
+      controller = c;
+    },
+    pull() {
+      t.fail('pull must not be called');
+      t.end();
+    },
+    pullInto(buffer, offset, length) {
+      try {
+        controller.respond(2);
```
domenic: Again, including both arguments is important to make it clearer what the bad behavior is that's causing the error.

reference-implementation/test/readable-byte-stream.js
domenic: Oh gah I now realize that the second buffer argument is optional. Hmm, we'll need to work through the transfer story in a bit more detail, but we can leave that for later.

reference-implementation/run-tests.js
```diff
@@ -6,16 +6,18 @@ import WritableStream from './lib/writable-stream';
...
 
   tests.forEach(require);
```
domenic: We should add RBS to the templated tests!

domenic:
Very exciting! First pass review done. The respond(bytesRead, view) signature is a bit strange, but maybe fine. And we still need to figure out the whole transfer and detaching story. But we can do that later.

The tests are really great. Amazing to see this in action, actually working.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/commit/6bd7b93bd92ca5bd257c3c27eeda40ffc5c466b1

reference-implementation/lib/readable-byte-stream.js
tyoshino: This is attempt to satisfy https://github.com/whatwg/streams/issues/312. I've put all the helper stuff (buffering, fragmentation, coalescing) into `ReadableByteStreamController` experimentally. `ReadableByteStream` and `ReadableByteStream.*Reader` doesn't have any variable for these work.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done https://github.com/whatwg/streams/commit/afeede5f436a7b1ddc162818090cfb78f8d744f3

reference-implementation/lib/readable-byte-stream.js
tyoshino: Sorry. Right. The spec is not sync with the ref impl yet.

reference-implementation/lib/readable-byte-stream.js
tyoshino: - Re: spec:
    - Added a TODO to https://github.com/whatwg/streams/issues/300
    - Shall we also check `IsDetachedBuffer()` on [[ViewedArrayBuffer]] as well as `ValidateTypedArray()` (which is not applicable for `DataView`)?
- Re: code: Done https://github.com/whatwg/streams/commit/422795d9e11e8a869e3eafd59d97895da5ccde36

reference-implementation/run-tests.js
tyoshino: Reverted. I forgot that we can just pass the path to the test file to run to `npm test`.

tyoshino: https://github.com/whatwg/streams/commit/4c87a4d5dd9113f43fe6b103a78851cefd1cf60d

reference-implementation/test/readable-byte-stream.js
tyoshino: :)

reference-implementation/test/readable-byte-stream.js
tyoshino: > This is throwing because it's respond()ing but not to a pullInto(), right? I think the test title should say that, and also you should respond with a Uint8Array or something, because otherwise I'd expect a TypeError anyway when respond()ing with a number.

Sorry. I'll update the test titles to describe what's being tested more clearly.

`pull()` must not responded by `respond()` as there's no existing `pullInto()` i.e. no pending buffer to fill.

https://github.com/whatwg/streams/commit/ffaea95c7ea3acb55fdc371895a7c9476049fc85

tyoshino: > Oh gah I now realize that the second buffer argument is optional. Hmm, we'll need to work through the transfer story in a bit more detail, but we can leave that for later.

This is a side effect of https://github.com/whatwg/streams/issues/353#issuecomment-111045309. I stored the buffer inside the controller to get prepared for `enqueue()` call in response to `pullInto()`. So, we don't have to receive the buffer from the underlying byte source (only if the underlying byte source hasn't detached it. in such cases, it must pass the active buffer to `respond()`).

reference-implementation/run-tests.js
tyoshino: Added a TODO item to https://github.com/whatwg/streams/issues/355

tyoshino:
Domenic, just to make sure, I'm not updating the spec text as it's better to get the algorithm and API surface reviewed through the reference implementation first. Not urging you but FYI.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
+const assert = require('assert');
+import { CreateIterResultObject, InvokeOrNoop, PromiseInvokeOrNoop, typeIsObject } from './helpers';
+
+export default class ReadableByteStream {
+  constructor(underlyingByteSource = {}) {
+    this._state = 'readable';
+    this._reader = undefined;
+    this._storedError = undefined;
+
+    this._controller = undefined;
```
domenic: Extra line

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    this._controller = new ReadableByteStreamController(this, underlyingByteSource);
+  }
+
+  cacnel(reason) {
```
domenic: Misnamed... probably would be caught by templated tests :)

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+
+  close() {
+    if (!IsReadableByteStreamController(this)) {
+      throw new TypeError(
+          'ReadableByteStreamController.prototype.close can only be used on a ReadableByteStreamController');
+    }
+
+    const stream = this._controlledReadableByteStream;
+
+    if (stream._closeRequested) {
```
domenic: is _closeRequested on the stream or on the controller? If it's on the stream, then it should be initialized in the stream constructor to false, and the below this._closeRequested should be stream._closeRequested. If it's on the controller (probably better given your idea of moving stuff there), then this should be `this`, and it should be initialized in the controller constructor to false.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+
+    if (this._totalQueuedBytes > 0) {
+      this._closeRequested = true;
+
+      return;
+    }
+
+    const reader = stream._reader;
+
+    if (reader === undefined || IsReadableByteStreamReader(reader)) {
```
domenic: I think the naming of IsReadableByteStreamReader is a bit confusing, even though it mirrors the getReader/getByobReader methods. Maybe IsReadableByteStreamDefaultReader?

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+      CloseReadableByteStream(stream);
+      return;
+    }
+
+    assert(IsReadableByteStreamByobReader(reader), 'reader must be ReadableByteStreamByobReader');
+
+    if (this._pendingPullIntos.length > 0 && this._pendingPullIntos[0].bytesFilled > 0) {
+      DestroyReadableByteStreamController(this);
```
domenic: In all but one case, DestroyReadableByteStreamController is followed by ErrorReadableByteStream. It might be worth thinking about how to combine them into one function. If not possible, then that's fine.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    }
+    if (stream._state !== 'readable') {
+      throw new TypeError('The stream is not in the readable state and cannot be enqueued to');
+    }
+
+    const reader = stream._reader;
+
+    if (reader === undefined) {
+      EnqueueInReadableByteStreamController(this, chunk);
+      return;
```
domenic: IMO most of the early returns in this function would actually be simpler as if/elses. "If there's no reader, do X. Otherwise, do Y." Instead of "If there's no reader, do X and return. Do Y."

Just a style thing though, not important.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+      }
+
+      assert(this._queue.length === 0);
+
+      const req = reader._readRequests.shift();
+      // TODO: Detach chunk.
+      req.resolve(CreateIterResultObject(chunk, false));
+
+      if (this._closeRequested === true) {
+        CloseReadableByteStream(stream);
```
domenic: This is really confusing for me because of the _closeRequested issue. It seems now like there are two different _closeRequested members---one for the stream, and one for the controller? They should probably be renamed...

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+
+    if (pullIntoDescriptor.bytesFilled >= pullIntoDescriptor.elementSize) {
+      this._pendingPullIntos.shift();
+
+      let remainder;
+      if (pullIntoDescriptor.bytesFilled % pullIntoDescriptor.elementSize > 0) {
+        remainder = buffer.slice(result.bytesUsed, pullIntoDescriptor.bytesFilled);
+      }
+
+      RespondToReadIntoRequest(reader, pullIntoDescriptor.buffer, pullIntoDescriptor.bytesFilled);
```
domenic: Can we move this before `let remainder`? If so we can make that code simpler.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    if (this._ownerReadableByteStream === undefined) {
+      return;
+    }
+
+    if (this._readRequests.length > 0) {
+      throw new TypeError('Tried to release a reader lock when that reader has pending read() calls un-settled');
+    }
+
+    if (this._ownerReadableByteStream._state === 'errored') {
+      MarkReadableByteStreamReaderErrored(this, this._ownerReadableByteStream._storedError)
```
domenic: This function seems to be missing? I can't find it with Ctrl+F...

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+
+    if (this._ownerReadableByteStream === undefined) {
+      return;
+    }
+
+    if (this._readIntoRequests.length > 0) {
+      throw new TypeError('Tried to release a reader lock when that reader has pending read() calls un-settled');
+    }
+
+    if (this._ownerReadableByteStream._state === 'readable') {
```
domenic: Doesn't handle error case?

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+      CloseReadableByteStreamReader(this);
+    }
+  }
+}
+
+function CallPull(controller) {
```
domenic: Name is too generic, sadly.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+  const source = controller._underlyingByteSource;
+
+  const pullFunction = source['pull'];
+  if (pullFunction === undefined) {
+    return;
+  }
+
+  const stream = controller._controlledReadableByteStream;
+
+  if (typeof pullFunction !== 'function') {
```
domenic: I think these can be omitted. If it's not a function trying to call it will fail, so you can just use the catch block.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+      throw new TypeError(
+          'ReadableByteStreamController instances can only be created by the ReadableByteStream constructor');
+    }
+
+    this._controlledReadableByteStream = controlledReadableByteStream;
+
+    this._underlyingByteSource = underlyingByteSource;
+
+    this._considerReissueUnderlyingByteSourcePullOrPullInto = false;
+    this._insideUnderlyingByteSource = false;
```
domenic: In ReadableStream these are called, respectively, pullAgain and pulling. Those names seems shorter and accurate enough to maybe use here.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    DestroyReadableByteStreamController(controller);
+    ErrorReadableByteStream(stream, new TypeError('pull property of an underlying byte source must be a function'));
+    return;
+  }
+
+  controller._callPullOrPullIntoLaterIfNeeded = false;
+  controller._insideUnderlyingByteSource = true;
+
+  try {
+    pullFunction.apply(source);
```
domenic: Nit: more idiomatic to use `.call`

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    DestoryReadableByteStreamController(controller);
+    if (stream._state === 'readable') {
+      ErrorReadableByteStream(stream, e);
+    }
+  }
+
+  controller._insideUnderlyingByteSource = false;
+}
+
+function CallPullInto(controller) {
```
domenic: Name too generic again

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+  const source = controller._underlyingByteSource;
+
+  const pullIntoFunction = source['pullInto'];
+  if (pullIntoFunction === undefined) {
+    return;
+  }
+
+  const stream = controller._controlledReadableByteStream;
+
+  if (typeof pullIntoFunction !== 'function') {
```
domenic: Same as before, can leave for the catch block.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+  }
+
+  assert(controller._pendingPullIntos.length > 0);
+  const pullIntoDescriptor = controller._pendingPullIntos[0];
+
+  controller._callPullOrPullIntoLaterIfNeeded = false;
+  controller._insideUnderlyingByteSource = true;
+
+  try {
+    pullIntoFunction.apply(source, [pullIntoDescriptor.buffer,
```
domenic: Hmm, so the signature is (buffer, offsetToWriteAt, ????) the third param confuses me; should it perhaps be `pullIntoDescriptor.byteLength - pullIntoDescriptor.byteOffset - pullIntoDescriptor.bytesFilled`?

I am also suspicious that this kind of (buffer, int, int) structure maybe should be a Uint8Array instead?

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+  controller._insideUnderlyingByteSource = false;
+}
+
+function CallPullOrPullIntoLaterIfNeeded(controller) {
+  if (controller._insideUnderlyingByteSource) {
+    controller._callPullOrPullIntoLaterIfNeeded = true;
+    return;
+  }
+
+  Promise.resolve().then(CallPullOrPullIntoRepeatedlyIfNeeded.bind(undefined, controller));
```
domenic: Use `process.nextTick(() => ...)` to enqueue a microtask.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+}
+
+function CallPullOrPullIntoRepeatedlyIfNeeded(controller) {
+  const stream = controller._controlledReadableByteStream;
+
+  while (true) {
```
domenic: What if they respond asynchronously? I am confused when this loops.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+    for (const req of reader._readRequests) {
+      req.resolve(CreateIterResultObject(undefined, true));
+    }
+
+    reader._readRequests = [];
+  } else {
+    assert(IsReadableByteStreamByobReader(reader), 'reader must be ReadableByteStreamByobReader');
+
+    if (reader._readIntoRequests.length > 0) {
+      return;
```
domenic: This needs a comment or something ... when you close a BYOB reader that has readInto requests, it doesn't get detached? Would it be better to not close it yet, and e.g. set closeRequested or something?

Apologies since this is probably something we discussed previously. It is complicated >_<

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+  controller._pendingPullIntos = []
+  controller._queue = [];
+}
+
+function DetachReadableByteStreamReader(reader) {
+  reader._ownerReadableByteStream._reader = undefined;
+  reader._ownerReadableByteStream = undefined;
+}
+
+function EnqueueInReadableByteStreamController(controller, chunk) {
```
domenic: Maybe just reuse the queue-with-sizes machinery already in the spec? Then you wouldn't need _totalQueuedBytes. Obviously it is less efficient, and in a real implementation we would use something like _totalQueuedBytes. But it is perhaps simpler and less work.

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+
+    pullIntoDescriptor.bytesFilled += bytesToCopy;
+
+    totalBytesToCopyRemaining -= bytesToCopy
+  }
+
+  return ready;
+}
+
+function InitializeReadableByteStreamReader(reader, stream) {
```
domenic: Again the function names are a bit confusing. Maybe name the ones that are shared e.g. InitializeReadableByteStreamReaderShared or something?

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+      return;
+    }
+
+    CallPullInto(controller);
+    CallPullOrPullIntoRepeatedlyIfNeeded(controller);
+
+    return;
+  }
+
+  const ready = FillPullIntoDescriptorFromQueue(controller, pullIntoDescriptor);
```
domenic: It might be possible to refactor this section to reuse RespondToReadIntoRequestsFromQueue. E.g. push pullIntoDescriptor into _pendingPullIntos and then call it. I am not sure though.

reference-implementation/run-tests.js
domenic: ffaea95c7ea3acb55fdc371895a7c9476049fc85 added it back haha

domenic:
Review done!

reference-implementation/lib/readable-byte-stream.js
tyoshino: Fixed 2bf0357a7fe4ccc2a78cfc1b0bb78a190a69cee4

reference-implementation/lib/readable-byte-stream.js
tyoshino: Fixed 2bf0357a7fe4ccc2a78cfc1b0bb78a190a69cee4

reference-implementation/lib/readable-byte-stream.js
tyoshino: AAh! This is required. Reverted with a comment. a1745016eae6ec0dd8e49c938e43c89744d37499

reference-implementation/lib/readable-byte-stream.js
tyoshino: You're right! It's a typo. I fixed this and also the bug in enqueue(), and added tests for them. f2a9f0721871be0531bf6db33f556187c318d0d6

tyoshino: And initialize. 186a2626e0b9f5825ccc10a769dc0e4f558c3f11

reference-implementation/lib/readable-byte-stream.js
tyoshino: Or shall we prefix operations which are generic to all readers by "Generic"?

domenic: I like suffix, but yeah, "Shared" or "Generic" makes sense.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Actually two. Maybe you searched for Destroy. There's one with typo (Destory). Fixed by 186a2626e0b9f5825ccc10a769dc0e4f558c3f11.

We could have a helper to run both, but maybe leads to increase of texts and a little longer time to trace call path in reviewing?

reference-implementation/lib/readable-byte-stream.js
tyoshino: Yeah, agreed. I changed them to use if-elses with leaving error handling ones as-is. 14e3bbc707d21f8243bdd49d4a5bc74dae26164a

reference-implementation/lib/readable-byte-stream.js
tyoshino: Sorry about having you spend much time to understand it. It's just a mistake!!

reference-implementation/lib/readable-byte-stream.js
domenic: I was just kind of hoping there was a way to refactor so that they always happen together, instead of happening together usually except in 2 cases. But I haven't thought about it too hard so probably there is a good reason for it to be separate :)

reference-implementation/lib/readable-byte-stream.js
```diff
@@ -0,0 +1,965 @@
...
+          EnqueueInReadableByteStreamController(this, chunk);
+        } else {
+          assert(this._queue.length === 0);
+
+          const req = reader._readRequests.shift();
+          // TODO: Detach chunk.
+          req.resolve(CreateIterResultObject(chunk, false));
+
+          if (this._closeRequested === true) {
+            CloseReadableByteStream(stream);
```
domenic: I am still a little unsure why enqueue() would ever cause a close to happen but I will try to read through it again tomorrow with fresh eyes.

reference-implementation/lib/readable-byte-stream.js
tyoshino: This code is not tested and actually doesn't work. I've fixed it and made it tested by 28cbdc8bd875ce5a1800fa9a38e0e2178fef887c.

tyoshino: and, addressed your comment. 7b559f5f99709daa003a573d7d34c8e257a2f3d8

reference-implementation/lib/readable-byte-stream.js
tyoshino: Since we have auto release, actually this line is unreached. I just removed this and replaced the if with assertion. f3be413272042e204083abbf52aa875e9cd753bc

reference-implementation/lib/readable-byte-stream.js
tyoshino: We have the auto release feature. So, this should be assert(). Fixed by f3be413272042e204083abbf52aa875e9cd753bc.

tyoshino: Also added some auto release tests f174dd56021e124aa055566a53bd95145a85f180

reference-implementation/lib/readable-byte-stream.js
tyoshino: Prefixed f174dd56021e124aa055566a53bd95145a85f180

reference-implementation/lib/readable-byte-stream.js
tyoshino: I just tried, but the exception it throws is "undefined is not a function" since it first evaluates pullFunction.apply (or pullFunction.call) and the result is undefined, and then undefined(source) is evaluated. So, it's not appropriate to set to storedError. I'd like to keep this. WDYT?

tyoshino: Added tests for non-callable pull/pullInto a8e4945b852621befee784f544f45c3eb4657fdc

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done 95cb1d5e456771b090547a68bbe122f11a272c29

reference-implementation/lib/readable-byte-stream.js
tyoshino: Prefixed f174dd56021e124aa055566a53bd95145a85f180

reference-implementation/lib/readable-byte-stream.js
tyoshino: See the reply to the previous comment.

reference-implementation/lib/readable-byte-stream.js
tyoshino: It's (buffer, offsetToWriteAt, sizeAllowedToWrite). The buffer we received from the consumer is `byteLength` long. It's already filled for `bytesFilled` bytes. So, byteLength - bytesFilled bytes are available for writing.

I'm fine with changing this to Uint8Array. I kept this ArrayBufferView free since it seems a little complex interface is acceptable for the controller. But yeah, I realized that this is just inconsistent with enqueue() which accepts ArrayBufferViews.

reference-implementation/lib/readable-byte-stream.js
domenic: Hmm, but that is really just an artifact of how we are implementing. In the spec we would use Call(...) and not .call(...), and of course in the spec error texts don't matter. I think it is better to just consolidate. A real implementation would probably work to have better error messages (and it's quite possible, see e.g. line 694 of ReadableStream2.js in https://codereview.chromium.org/1167343002) but it's not for the spec/reference impl.

In fact, maybe using CallOrNoop here would help clean up? Not sure.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Changed to Uint8Array on e187dbf9d506fad2f9094f0710fe9045c3b82f00

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done 24d5faf412bbf208b9f5b27b902d6f427f00ed83

reference-implementation/lib/readable-byte-stream.js
tyoshino: respond() and enqueue() calls ReadableByteStreamControllerCallPullOrPullIntoLaterIfNeeded() to call pull/pullInto in the next tick if needed. In the cases where pull/pullInto doesn't fulfill all the pending requests, we invoke it again. This allows implementing a pull/pullInto that calls respond()/enqueue() for each pull/pullInto. They'll be called repeatedly and synchronously by this while loop.

I found a bug in ReadableByteStreamControllerCallPullOrPullIntoLaterIfNeeded. Fixed and added tests by bf35ce30e4b1824b684617ad6605748454b66d0a

tyoshino: We can also choose to enqueue tasks to do that instead of while loop. But I just tried to repeat it synchronously for performance.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Yeah, this is complicated.

We have 3 options:

1. Wait for the buffer to return. Once returned, `controller.respond(0, returnedBuffer)` and `controller.close()`
1. `controller.close()` and wait for the buffer to return. Once returned `controller.respond(0, returnedBuffer)`
1. Accept the same sequence (at controller side) as (2) but make the reader look not closed until the `respond()` is made

I chose (2) to thinking of a case like this:

At the time when an underlying byte source notices that no more data comes (or on `reader.cancel()` call), the buffer passed via `pullInto()` may still owned by some other thread and we need to wait for the buffer to be returned. (*)

(2) allows faster delivery of closure signal to the consumer.

(1) and (3) are equivalent from the view point of the consumer. So, I'd just choose (1) from them.

I'm not sure (*) is so realistic. But it seems `reader.cancel()` should be able to mark the stream closed synchronously. This requires the same state transition sequence as (2).

reference-implementation/lib/readable-byte-stream.js
tyoshino: Yeah, I also wondered why I didn't use it, but just remembered. The BYOB reader may consume only a part of a chunk in a queue. Since the existing queue-with-sizes machinery is designed to work chunk-wise, we need to shift, update (partially consume), and then unshift.

See `FillPullIntoDescriptorFromQueue()`.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Sorry. For some reason, I had to push --force. Please comment on 6c55dbf instead of bf35ce3 if you want.

reference-implementation/lib/readable-byte-stream.js
tyoshino: Addressed by 7db54d1c0dcb0edc0419371cfac1d5b169b52c25

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done by 7db54d1c0dcb0edc0419371cfac1d5b169b52c25

reference-implementation/run-tests.js
tyoshino: Aaah! Fixed again 719ba77f76c04458f9139b52aed33abda1762b89

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done 48a23cdb61248cfce57834503a184d54b8fe2980

reference-implementation/lib/readable-byte-stream.js
tyoshino: Done by 7db54d1c0dcb0edc0419371cfac1d5b169b52c25

reference-implementation/lib/readable-byte-stream.js
tyoshino: We could. I've written these functions separately for each context than an all-in-one function that switches its behavior by checking various conditions, and it turned out it's easier to code.

I've moved asserts to simplify the code a little on a4f9a2848ce5d70725a56a7d61ec2e521c903336

reference-implementation/lib/readable-byte-stream.js
tyoshino: > I think it is better to just consolidate.

OK.

> In fact, maybe using CallOrNoop here would help clean up?

Good point. Since we now allow `enqueue()` anytime, it's reasonable to allow not implementing `pull()`. An underlying byte source that doesn't understand BYOB would only use `enqueue()`, and it's totally fine. So, it's also reasonable to allow leaving `pullInto()` undefined.

Fixed by 76bf9ecb54d6ea990f7803c2f8b482dc2bbd548d

reference-implementation/lib/readable-byte-stream.js
tyoshino: Ya. For now I'll leave them as is.

tyoshino:
> And we still need to figure out the whole transfer and detaching story. But we can do that later.

Done f697ea8e59368185ba501ff59559b06da73ff1d2

tyoshino:
Renamed e5b1c9d70f73242ebcf4ceb512a7d62b2b43915b

tyoshino:
> The respond(bytesRead, view) signature is a bit strange, but maybe fine.

We could choose the Uint8Array, but here the minimal set of information that `respond()` want to know are these two values.

domenic:
I think we should consider merging this soon. But of course the spec needs updates first :-S. But yeah I would like to get this into master so that we can start building on top of it, e.g. expanding the non-normative and introductory text and usage examples.

tyoshino:
Made 8 more commits (both bug and readability fixes)

tyoshino:
Made some more commits mainly in order to make it parallel with ReadableStream.

domenic:
@tyoshino still would like to merge this if possible given how it is a big change/improvement from the current situation (in test/implementation coverage alone, if nothing else). To do that I guess we need to either update the spec to parallel the reference implementation, or remove most of the spec and say "please see reference implementation; still under development."

tyoshino:
@domenic okay. maybe latter and also start writing a patch to add the spec text.

index.bs
tyoshino: OK. Removed

index.bs

index.bs
tyoshino: Added!

tyoshino:
Reduced the change on the spec side. Ready for review.

domenic:
Looks good. Needs a rebase though.

tyoshino:
Merged as 482d14b16d7bc8b2d88355baa55d618b1c82485d.

PR #365, opened 2015-06-18
Update FAQ with some extra answers

### LINKED ISSUE #126: FAQs that need answers
- [ ] Are you supporting object streams or only binary/string streams?
- [ ] What are the main differences from Node.js streams?
- [ ] How does this relate to "observables" or "async iterators"?
- [ ] What's the deal with this `wait()` pattern?

### LINKED ISSUE #360: Add "when are streams appropriate"
Not sure if this belongs in the spec or in something more dedicated like http://www.w3.org/2001/tag/doc/promises-guide#when-to-use

Candidates, for readable streams:

- Consumers care about every value, and do not want to miss any
- Errors are either final and non-recoverable (socket closed, stream goes into errored state), or just part of the protocol (404, just part of the data stream)
- Single consumer by default (you tee a stream and hand out the two branches for multi-consumer)
- Best when connected to some underlying I/O source, especially one for which backpressure has meaning.
- Best when concatenating all the values in the stream would give something meaningful. (Is this actually important, or is it just true in the cases we've been considering?)

This came up when considering whether streams were a good fit for https://webbluetoothcg.github.io/web-bluetooth/#bluetoothgattcharacteristic. After talking with @jyasskin I eventually concluded that they're not, since in that case, even though there is an underlying I/O source, most of the other points mentioned don't quite fit.

This might be part of #126.

domenic:
- Explain when streams are appropriate
- Explain how streams work with any chunk type
- Relate readable stream to other async plural constructs
- Compare with Node.js streams

Closes #360. Closes #126.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -2,6 +2,81 @@
 
 This is a work in progress, documenting some design decisions that were made and that were non-obvious enough that we feel the need to explain them. We'll probably be adding to it as we go. If there's something you feel belongs here, please file an issue or pull request and we'll add it!
 
+## Are readable streams appropriate for my use case?
+
+Readable streams fit best in situations where:
+
+- You are trying to represent some underlying I/O source of data, especially one for which backpressure is meaningful.
+- Consumers care about the logical concatenation of all chunks of the stream, such that every value produced is important.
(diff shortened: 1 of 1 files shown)


tyoshino:
lgtm

PR #366, opened 2015-06-18
Use "perform" for abstract ops that cannot throw

### LINKED ISSUE #342: Make sure promise returning methods don't throw
I see some call-with-rethrows in promise returning methods' algorithm. Check that they're correctly implemented so that we don't throw but return a rejected promise to conform to http://www.w3.org/2001/tag/doc/promises-guide#always-return-promises.

domenic:
Closes #342, by making it clear that the promise-returning methods do not throw since they always call out to "perform"s or similar.

Also found one case of unnecessary exception guarding, and a few unnecessary ReturnIfAbrupts.

Leaves out ReadableByteStream for now, pending #361.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -394,9 +394,9 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
   1. Resolve _startResult_ as a promise:
     1. Upon fulfillment,
       1. Set *this*@[[started]] to *true*.
-      1. Return RequestReadableStreamPull(*this*).
+      1. Perform RequestReadableStreamPull(*this*).
     1. Upon rejection with reason _r_,
-      1. If *this*@[[state]] is "readable", return ErrorReadableStream(*this*, _r_).
+      1. If *this*@[[state]] is "readable", perform ErrorReadableStream(*this*, _r_).
 </pre>
 
 <h4 id="rs-prototype">Properties of the <code>ReadableStream</code> Prototype</h4>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1623,9 +1626,10 @@ a \[[stream]] internal slot. When a <code>WritableStream</code> error function <
     1. Resolve _writeRecord_.[[promise]] with *undefined*.
     1. DequeueValue(_stream_@[[queue]]).
     1. Let _syncResult_ be SyncWritableStreamStateWithQueue(_stream_).
-    1. If _syncResult_ is an abrupt completion, then return ErrorWritableStream(_stream_, ‍_syncResult_.[[value]]).
-    1. Otherwise, return WritableStreamAdvanceQueue(_stream_).
-  1. Upon rejection of _writeResult_ with reason _r_, return ErrorWritableStream(_stream_, _r_).
+    1. If _syncResult_ is an abrupt completion, then perform ErrorWritableStream(_stream_, ‍_syncResult_.[[value]]).
```
tyoshino: Perform SyncWritableStreamStateWithQueue?

tyoshino: GetTotalQueueSize() at L1598 can also be changed to Perform?

tyoshino:
Done.

Great change! Error handling paths are very clear now.

In ReadableByteStream change, I was doing similar but by writing only `AbstractOperation()` without any verb when it's designed to never throw. I just need to add "perform" to them.

index.bs
domenic: Good catches both. For GetTotalQueueSize() and other operations with return values that we need to use, I am just going to omit the ReturnIfAbrupt. Maybe it would be better to do something like AssertNotAbrupt, but I think it is OK to just start using the value.

Fixing and pushing.

Issue #187, opened 2014-08-28
Valid value for strategy.size()


tyoshino:
The only check about strategy.size() is NaN check inside EnqueueValueWithSize().
- Do we want to accept non-integer value?
- Do we want to accept 0?

PR #370, opened 2015-06-29
Don't allow negative-sized chunks

### LINKED ISSUE #368: Should negative chunk size be allowed?
https://streams.spec.whatwg.org/#enqueue-value-with-size allows negative chunk sizes.
Is there a reason behind that?
Also, highWaterMark cannot be negative.

domenic:
Closes #368. See previously #187 and f1a7e78d50274b34b556dff0d80b96e2a9207c60. We can always relax this later if someone has a use case.

### DIFF
--- a/reference-implementation/lib/queue-with-sizes.js
+++ b/reference-implementation/lib/queue-with-sizes.js
@@ -8,8 +8,8 @@ export function DequeueValue(queue) {
 
 export function EnqueueValueWithSize(queue, value, size) {
   size = Number(size);
-  if (Number.isNaN(size) || size === +Infinity || size === -Infinity) {
-    throw new RangeError('Size must be a finite, non-NaN number.');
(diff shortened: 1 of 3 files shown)


reference-implementation/lib/queue-with-sizes.js
```diff
@@ -8,8 +8,8 @@ export function DequeueValue(queue) {
 
 export function EnqueueValueWithSize(queue, value, size) {
   size = Number(size);
-  if (Number.isNaN(size) || size === +Infinity || size === -Infinity) {
-    throw new RangeError('Size must be a finite, non-NaN number.');
+  if (Number.isNaN(size) || size === +Infinity || size < 0) {
+    throw new RangeError('Size must be a finite, non-negative number.');
```
tyoshino: non-NaN, non-negative?

tyoshino:
lgtm

PR #373, opened 2015-07-02
Initialize variables of ReadableStreamReader to the final value direc…

tyoshino:
…tly when obtained from a closed/errored stream

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -757,13 +757,22 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
 <emu-alg>
   1. If IsReadableStream(_stream_) is *false*, throw a *TypeError* exception.
   1. If IsReadableStreamLocked(_stream_) is *true*, throw a *TypeError* exception.
-  1. Set _stream_@[[reader]] to *this*.
-  1. Set *this*@[[ownerReadableStream]] to _stream_.
-  1. Set *this*@[[state]] to "readable".
-  1. Set *this*@[[storedError]] to *undefined*.
+  1. Set *this*@[[state]] to _stream_@[[state]].
+  1. If _stream_@[[state]] is "readable",
+    1. Set _stream_@[[reader]] to *this*.
+    1. Set *this*@[[ownerReadableStream]] to _stream_.
+    1. Set *this*@[[storedError]] to *undefined*.
+    1. Set *this*@[[closedPromise]] to a new promise.
+  1. Otherwise,
+    1. Set *this*@[[ownerReadableStream]] to *undefined*.
+    1. If _stream_@[[state]] is "closed",
+      1. Set *this*@[[storedError]] to *undefined*.
+      1. Set *this*@[[closedPromise]] to a new promise resolved with *undefined*.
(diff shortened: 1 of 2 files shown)


domenic:
The point of this was that now we could simplify ReleaseReadableStreamReader, right?

tyoshino:
Right. Updated ReleaseReadableStreamReader. Now, the if in it is gone. Each of ErrorReadableStream and FinishClosingReadableStream has the corresponding reader updating code. This looks better in terms of simplicity of code path, but increased LoC. The initialization process now sets the variables to the final value instead of making it readable and then immediately transition to errored/closed. I like this but this also increased LoC.

domenic:
OK, this does not look so bad after I re-read it a few times. A lot of the "code bloat" comes from the `_resolve` and `_reject` members, which are just book-keeping devices.

If you are happy with this, then first please make sure that it parallels your ReadableByteStream PR---I kind of remember the ReadableByteStream having more factored-out functions? But I prefer fewer, so this looks pretty good. If they match up, then yeah, update the spec and rebase and I will merge.

tyoshino:
ReadableByteStream has more factored-out functions basically because it has two readers and the code common to both are factored out. Also, I have some functions split to make it clear boundary between the controller and reader.

reference-implementation/lib/readable-stream.js
```diff
@@ -397,9 +417,44 @@ function ErrorReadableStream(stream, e) {
...
+  if (reader === undefined) {
+    return undefined;
   }
+
+  for (const { _reject } of reader._readRequests) {
+    _reject(e);
+  }
+  reader._readRequests = [];
+
+  ReleaseReadableStreamReader(reader);
```
tyoshino: This call is placed here so that this would be parallel with RBS. In RBS I'm calling this right after flushing read requests because whether we release the reader or not depends on emptiness of read requests queue.

tyoshino:
PTAL. If looks good, I'll sync the spec text with the code.

domenic:
This LGTM. Very sorry for the delay.

tyoshino:
Thanks for review, Domenic. Now the spec text is also ready for review.

index.bs
```diff
@@ -884,12 +893,24 @@ this to streams they did not create, and must ensure they have obeyed the precon
   control of the underlying source.
 </div>
 
+<h4 id="close-readable-stream-reader" aoid="CloseReadableStreamReader">CloseReadableStreamReader ( reader )</h4>
+
+<emu-alg>
+  1. Repeat for each _readRequestPromise_ that is an element of _reader_@[[readRequests]],
```
domenic: Forgot to set to an empty list after this.

domenic:
OK, merged as 7b034bbf9c3a87c01349a4b806b99ff3aab8dc53 with some minor tweaks. In particular I edited the ReadableStreamReader constructor to make the parallelism and grouping a bit more tight. I think it is OK.

index.bs
tyoshino: Thanks for catching it!

PR #377, opened 2015-07-22
Strategies parameters

calvaris:
I added a couple of tests for the parameters of the constructor and the size method that I am planning to add in WebKit and might be useful here too.


calvaris:
I'll rewrite this because I think some changes are needed.

PR #380, opened 2015-07-23
Strategies parameters

calvaris:
Sorry for the noise yesterday but I improved a bit the size tests as they were not checking return values and they are now. I think these test that check the outcome depending on the parameters type could be  useful to be integrated in the spec ones as I'll include them for WebKit.


domenic:
These tests are great. Will do some minor style fixups and merge tomorrow.

domenic:
Merged as 8dca04e41804e70240ce050874238fa828f47fcb with purely stylistic tweaks. Thank you!!

Issue #378, opened 2015-07-22
Add IsDisturbed predicate


domenic:
Should return true if stream is ever .cancel()ed or .read() from. See more details in https://etherpad.mozilla.org/streams-f2f-july

tyoshino:
Thanks for getting this discussed and summarizing it. FTR, could you please list the participants of the f2f (or put a link to some page with such a list) if no problem?

wanderview:
I added the attendees to the etherpad.  Basically, me, @Domenic, and @annevk since we were all in town for other meetings that week.

tyoshino:
Thanks, Ben.

Reading the note. @domenic , you objected to the always-chunked approach at https://lists.w3.org/Archives/Public/public-webapps/2014OctDec/0420.html Has it been resolved?

domenic:
Yes. We can do chunked for now. In the future I would like to explore allowing setting content-length (and thus not auto-setting chunked) but we can leave that for another time.

tyoshino:
Created. Sorry for OT comments.

tyoshino:
Implemented as IsReadableStreamDisturbed() in #385.

PR #385, opened 2015-08-03
Add IsReadableStreamDisturbed predicate

tyoshino:
See the suggestion in the OP of #378.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -750,16 +750,6 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
     <td>A List of promises returned by calls to the reader's <code>read()</code> method that have not yet been resolved,
       due to the <a>consumer</a> requesting <a>chunks</a> sooner than they are available
   </tr>
-  <tr>
-    <td>\[[state]]
-    <td>A string containing the reader's current state, used internally; one of <code>"readable"</code>,
-      <code>"closed"</code>, or <code>"errored"</code>
-  </tr>
-  <tr>
-    <td>\[[storedError]]
-    <td>A value indicating how the reader's stream failed, to be given as a failure reason or exception when trying to
-      operate on the reader; applicable only when \[[state]] is <code>"errored"</code>
-  </tr>
 </table>
 
 <h4 id="reader-constructor">new ReadableStreamReader(stream)</h4>
@@ -773,19 +763,15 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
 <emu-alg>
   1. If IsReadableStream(_stream_) is *false*, throw a *TypeError* exception.
   1. If IsReadableStreamLocked(_stream_) is *true*, throw a *TypeError* exception.
-  1. Set *this*@[[state]] to _stream_@[[state]].
   1. Set *this*@[[readRequests]] to a new empty List.
   1. Set *this*@[[ownerReadableStream]] to _stream_.
   1. Set _stream_@[[reader]] to *this*.
   1. If _stream_@[[state]] is "readable",
-    1. Set *this*@[[storedError]] to *undefined*.
     1. Set *this*@[[closedPromise]] to a new promise.
   1. Otherwise, if _stream_@[[state]] is "closed",
-    1. Set *this*@[[storedError]] to *undefined*.
     1. Set *this*@[[closedPromise]] to a new promise resolved with *undefined*.
   1. Otherwise,
     1. Assert: _stream_@[[state]] is "errored".
-    1. Set *this*@[[storedError]] to _stream_@[[storedError]].
     1. Set *this*@[[closedPromise]] to a new promise rejected with _stream_@[[storedError]].
 </emu-alg>
 
@@ -812,11 +798,7 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
 
 <emu-alg>
   1. If IsReadableStreamReader(*this*) is *false*, return a promise rejected with a *TypeError* exception.
-  1. If *this*@[[ownerReadableStream]] is not *undefined*, set *this*@[[ownerReadableStream]]@[[disturbed]] to *true*.
-  1. If *this*@[[state]] is "closed", return a new promise resolved with *undefined*.
-  1. If *this*@[[state]] is "errored", return a new promise rejected with *this*@[[storedError]].
-  1. Assert: *this*@[[ownerReadableStream]] is not *undefined*.
-  1. Assert: *this*@[[ownerReadableStream]]@[[state]] is "readable".
+  1. If *this*@[[ownerReadableStream]] is *undefined*, return a promise rejected with a *TypeError* exception.
   1. Return CancelReadableStream(*this*@[[ownerReadableStream]], _reason_).
 </emu-alg>
 
@@ -839,6 +821,7 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
 
 <emu-alg>
   1. If IsReadableStreamReader(*this*) is *false*, return a promise rejected with a *TypeError* exception.
+  1. If *this*@[[ownerReadableStream]] is *undefined*, return a promise rejected with a *TypeError* exception.
   1. Return ReadFromReadableStreamReader(*this*).
 </emu-alg>
 
@@ -859,7 +842,8 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
   1. If IsReadableStreamReader(*this*) is *false*, throw a *TypeError* exception.
   1. If *this*@[[ownerReadableStream]] is *undefined*, return *undefined*.
   1. If *this*@[[readRequests]] is not empty, throw a *TypeError* exception.
-  1. If *this*@[[state]] is `"readable"`, perform CloseReadableStreamReader(*this*).
+  1. If *this*@[[ownerReadableStream]]@[[state]] is `"readable"`, reject *this*@[[closedPromise]] with a *TypeError* exception.
+  1. Otherwise, set *this*@[[closedPromise]] be a new promise rejected with a *TypeError* exception.
   1. Set *this*@[[ownerReadableStream]]@[[reader]] to *undefined*.
   1. Set *this*@[[ownerReadableStream]] to *undefined*.
   1. Return *undefined*.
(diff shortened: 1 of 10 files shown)


tyoshino:
The disturbed flag is never changed when the streams is closed. Is it assumed that this is implemented after removing the auto-release feature?

yutakahirano:
FYI: https://github.com/yutakahirano/fetch-with-streams/pull/48#issuecomment-125172550

index.bs
```diff
@@ -300,6 +300,11 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
       state and queue of this stream
   </tr>
   <tr>
+    <td>\[[disturbed]]
+    <td>A boolean flag set to <b>true</b> when either <code>read()</code> or <code>cancel()</code> has been called on
```
domenic: better phrased as "when the stream has been read from or canceled"

index.bs
```diff
@@ -971,6 +978,16 @@ an assert).
   1. Return *true*.
 </emu-alg>
 
+<h4 id="is-readable-stream-disturbed" aoid="IsReadableStreamDisturbed">IsReadableStreamDisturbed ( stream )</h4>
+
+This abstract operation is meant to be called from other specifications that may wish to query whether or not a
```
domenic: "...readable stream has ever been read from or canceled"

reference-implementation/run-tests.js
```diff
@@ -2,12 +2,14 @@ const glob = require('glob');
...
 
 import ReadableStream from './lib/readable-stream';
+import { IsReadableStreamDisturbed } from './lib/readable-stream';
 import WritableStream from './lib/writable-stream';
 import ByteLengthQueuingStrategy from './lib/byte-length-queuing-strategy';
 import CountQueuingStrategy from './lib/count-queuing-strategy';
 import TransformStream from './lib/transform-stream';
 
 global.ReadableStream = ReadableStream;
+global.IsReadableStreamDisturbed = IsReadableStreamDisturbed;
```
domenic: I don't like this testing strategy because the global modifications are meant to be based on https://streams.spec.whatwg.org/#globals. I am not sure what the best strategy for testing it is though. I will try to work on something.

domenic: For example, such tests could never be run in browsers. Maybe they just should be left for other specs' test suites.

domenic:
> The disturbed flag is never changed when the streams is closed. Is it assumed that this is implemented after removing the auto-release feature?

I don't think we need to remove auto-release. However, as @yutakahirano notes, I think it was important (for reasons I forget) to have the disturbed flag change even on closed streams. As such it should probably move up a few steps in the algorithms.

reference-implementation/run-tests.js
tyoshino: So, should we remove this from run-tests.js and ... import in the test file?

index.bs
tyoshino: Done

index.bs
tyoshino: Done

tyoshino:
Maybe one of the origins of the requirement is https://github.com/yutakahirano/fetch-with-streams/issues/37#issuecomment-108116997 ? Is it still active? https://github.com/whatwg/fetch/issues/61#issuecomment-123854971 Or even if it's active, it'll be covered by used flag?

domenic:
Reviewing https://etherpad.mozilla.org/streams-f2f-july "Even more conservative strawman" I think the requirement comes from wanting .json() etc. to automatically make the stream disturbed, even for empty streams. The used flag is only for `new Request(otherRequest)`. This is important because it means that author code that acts similar to json() will also make the stream disturbed.

reference-implementation/run-tests.js
domenic: Yes, that sounds good. And maybe move them to a separate abstract-ops.js or from-other-specs.js test file.

tyoshino: Done

tyoshino:
Addressed the comment. Thanks.

> This is important because it means that author code that acts similar to json() will also make the stream disturbed.

OK. But an operation made on a released reader making the ex-associated stream disturbed sounds strange.

domenic:
> But an operation made on a released reader making the ex-associated stream disturbed sounds strange.

Oh, you are right. It would mean that getting a reader for a closed/error stream would still have to hold on to the stream in order to maybe mark it disturbed.

This seems a bit silly. It means you cannot GC the stream if the reader is still alive, even though the stream is closed or errored.

But that is maybe not so bad since closed or errored streams do not have any data in their internal queue.

Also it would eliminate some internal complexity from the reader (the [[state]] and [[storedError]] fields).

And the silliness seems less bad than debating bodyUsed all over again.

What do you think?

domenic:
Ping. What are your thoughts?

yutakahirano:
I prefer dropping auto-release to introducing released-but-still-connected state.

domenic:
That seems like a good compromise to me. @tyoshino do you want to do the PR or shall I?

domenic:
Actually, I can do that; there is other work I want to do on top of it. I will send a PR later today.

domenic:
OK, I have pushed to this PR with a base commit that removes automatic unlocking on close/error.

Two important questions:

- Should we now reject read() and closed when used on released readers, instead of treating as a closed/errored stream? That seems simpler.

- **Are we OK with breaking compat on this?** It would be a breaking change to Chrome's current behavior, and a change in an area of the spec we've called stable. I personally think it is unlikely to break things, but what do you think? /cc @yutakahirano @calvaris

calvaris:
I think it is also unlikely that we break things.

Our current implementation is also based on closed/errored so it would mean to change our current implementation, which I'm not a big fan of but it wouldn't be a big change either, so I'm against but only barely and based on laziness :)

What I would do is securing the change of behavior if it happens with some tests ensuring that everything is as it should be and not as it was.

domenic:
@calvaris thanks for weighing in! I made a bunch of test updates in https://github.com/whatwg/streams/commit/9945ae13eea9c1156d1feae3bfe665d9f7a88fc4 but if you can think of others we should add them. E.g. if you come up with test descriptions I'll write the tests :)

yutakahirano:
I believe we cannot introduce IsDisturbed with satisfying all proposed requirements and all existing properties. I think dropping the auto-release is a tolerable way.

domenic:
@yutakahirano well, we can introduce it without any backward-incompatible changes if we follow https://github.com/whatwg/streams/pull/385#issuecomment-128149814 and allow readers to retain references to their owner stream even after release.

But I take it you think that is bad?

yutakahirano:
> But I take it you think that is bad?

Yes.

domenic:
OK. Does anyone have any opinion on

> Should we now reject read() and closed when used on released readers, instead of treating as a closed/errored stream?

domenic:
My opinion is that they should now reject. That will allow us to get rid of the internal slots [[state]] and [[storedError]]. The downside is that you would no longer get to treat released readers as closed/errored streams depending on their originating stream. Instead they would effectively behave as if they always came from errored streams. But I don't think that's a big deal; you shouldn't be messing with the reader after it's been released.

tyoshino:
Sorry for delay. I'm fine with the plan. I don't come up with any reasonable use case to keep a reference to a reader while releasing the parent stream. Some code may be accidentally doing so, but could be easily fixed and as Domenic said it's not so harmful to hold an errored/closed stream (which is small) in addition to a reader.

I'm also fine with making read on released reader reject. I think currently people who are unaware of the auto release mechanism are not calling release(), and basically people who are aware of wouldn't read() on released one. Some may be using release() and also using released reader as a representation of closed stream that accepts read() and tell us that the stream is closed repeatedly. But ... maybe the number of users with such a use case is small, I guess.

domenic:
OK. Updated with what I think should be final. Cannot build the spec until https://github.com/tabatkins/bikeshed/issues/466 is fixed, but ready for review.

index.bs
```diff
@@ -855,7 +848,10 @@ Instances of <code>ReadableStreamReader</code> are created with the internal slo
   1. If IsReadableStreamReader(*this*) is *false*, throw a *TypeError* exception.
   1. If *this*@[[ownerReadableStream]] is *undefined*, return *undefined*.
   1. If *this*@[[readRequests]] is not empty, throw a *TypeError* exception.
-  1. Perform CloseReadableStreamReader(*this*).
+  1. If *this*@[[ownerReadableStream]]@[[state]] is `"readable"`, reject *this*@[[closedPromise]] with a *TypeError* exception.
+  1. Otherwise, let *this*@[[closedPromise]] be a new promise rejected with a *TypeError* exception.
```
tyoshino: I thought "let" is used for declaring/substituting to local variables while set is used for substituting new value to a slot. I'm just misunderstanding your style rule or this "let" should be "set"?

tyoshino: Do we really need to set a new rejected promise to [[closedPromise]] here?

index.bs
```diff
@@ -973,8 +958,6 @@ an assert).
     1. Reject _readRequestPromise_ with _e_.
   1. Set _reader_@[[readRequests]] to a new empty List.
   1. Perform ReleaseReadableStreamReader(_reader_).
```
tyoshino: Was the plan to keep auto-release for error?

tyoshino: Oh, the impl is fixed. So, please just delete this line in the spec text :)

reference-implementation/lib/readable-stream.js
```diff
@@ -562,13 +559,6 @@ function ReadFromReadableStreamReader(reader) {
   }
 }
 
-function ReleaseReadableStreamReader(reader) {
```
tyoshino: Let's remove ReleaseReadableStreamReader from the spec text too.

index.bs
tyoshino: Ah, I see. This new rejected promise and the TypeError at L851 are representing "already released". Then, looks good.

tyoshino:
Review done!

domenic:
Thanks for the review! I turns out I failed to do spec changes for the "Remove automatic unlocking on close/error" commit. Your let vs. set catch was correct as well. Fixed and will merge now.

yutakahirano:
Just a confirmation: @wanderview, are you OK with this change? I am planning to implement this behavior change to Chrome, and before that I would like to make sure that everyone is satisfied.

yutakahirano:
@wanderview: ping

wanderview:
I'm sorry, but I haven't been able to keep up with this.  Can you summarize
the differences from the July f2f agreement?

Are we just dropping automatic unlocking on stream close?

yutakahirano:
The f2f agreement suggests a stream should be disturbed  when read even if it is closed (See https://github.com/yutakahirano/fetch-with-streams/pull/48#issuecomment-125172550). In order to achieve that, there are two ways:

1. Auto-released reader makes the stream disturbed.
1. Remove the auto-release mechanism.

My opinion is that the latter is less evil.

> Are we just dropping automatic unlocking on stream close?

And on stream error.

wanderview:
> Remove the auto-release mechanism.

I think this is reasonable.  Thanks for summarizing for me.

PR #386, opened 2015-08-03
Notate throwing vs. non-throwing abstract operations

### LINKED ISSUE #371: Notate abstract ops that can throw
#366 was a good first step in making the flow of potential exceptions clearer in the spec. We should also consider notating whether an abstract operation can throw or not so that callers know whether to use perform or to handle any abrupt completions.

Here's an approach:
Sample screenshot hacked together with web inspector

domenic:
Fixes #371.

@tyoshino, pretty sure this is correct, but if you could do a quick check to make sure my throw/nothrow annotations are OK, that'd be helpful.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -33,6 +33,17 @@ Link Defaults: html (dfn) structured clone
   emu-alg > ol ol, emu-alg > ol ol ol ol ol { list-style-type: lower-alpha; }
   emu-alg > ol ol ol, emu-alg > ol ol ol ol ol ol { list-style-type: lower-roman; }
   emu-alg li { margin: 0; }
+
+  .heading[aoid] .annotation {
+    background-color: beige;
+    border: 1px solid black;
+    border-radius: 3px;
+    cursor: help;
+    display: inline-block;
+    font-size: 70%;
+    font-weight: normal;
+    padding: 1px 2px;
+  }
 </style>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1833,7 +1844,7 @@ throughout the rest of this standard.
   1. Return *undefined*.
 </emu-alg>
 
-<h4 id="get-total-queue-size" aoid="GetTotalQueueSize">GetTotalQueueSize ( queue )</h4>
+<h4 id="get-total-queue-size" aoid="GetTotalQueueSize" throws>GetTotalQueueSize ( queue )</h4>
```
tyoshino: nothrow?

index.bs
```diff
@@ -1843,7 +1854,7 @@ throughout the rest of this standard.
   1. Return _totalSize_.
 </emu-alg>
 
-<h4 id="peek-queue-value" aoid="PeekQueueValue">PeekQueueValue ( queue )</h4>
+<h4 id="peek-queue-value" aoid="PeekQueueValue" throws>PeekQueueValue ( queue )</h4>
```
tyoshino: nothrow?

tyoshino:
Reviewed annotations!

domenic:
Thanks for the careful review; both of your throw -> nothrow finds were correct!

PR #392, opened 2015-08-28
Delayed setting timeous to have better chances on slower systems.

calvaris:
We had to change these tests in similar ways because they timed out in slow systems.


youennf:
Some additional information : the setTimeout callback is kicking too early sometimes on slow WebKit bots, leading to failing assertions.
Proposed changes make the tests more deterministic with that regards.

domenic:
Great change, thank you. Merged as b47d462273abbbb1894f2bb3069acd4d62498643.

PR #396, opened 2015-10-01
More tests

calvaris:


calvaris:
I added the tests we spoke about on IRC.

calvaris:
I updated the commit logs to reflect where those new tests are coming from.

domenic:
OK, everything should have landed! Sorry for making you convert from testharness.js to tape; that should no longer be necessary. Thanks for all the work you did on that though, as it really gave me a jump-start on the process of moving over!

PR #397, opened 2015-10-07
Port a bunch of tests to web-platform-tests format

domenic:
This builds on the work @calvaris has done in in https://github.com/Igalia/streams/tree/webkit-tests/reference-implementation/webkit/reference-implementation, in many cases porting it directly over. Over time I hope that we can move all tests to this format. For now I have focused on those most urgent for Chrome's upcoming implementation, i.e., the readable stream and queuing strategy tests.

I have deleted the original tests which have been ported over, so we can keep better track.

This also adds a bit of new infrastructure to allow us to continue developing and testing the reference implementation inside Node.js, using `npm run wpt`.

--

@tyoshino @calvaris I don't expect you guys to review this entire commit but if you could give it a quick spot check to see if it makes sense I'd appreciate it.

BTW I thought about cleaning up the style of all the tests, to use more ES6, and to use promise_test and promise_rejects and stuff, but in the end I think it's better to just leverage @calvaris's work as-is instead of obsessing over test hygeine :).

I'll probably do the templated and GC tests soon, before I try to get these in to Chrome.

Also, @calvaris, should we be including some kind of license notice for your work? I know it was based on the CC0 tests here but you probably wrote it while employed by someone that requires licensing and stuff. I would imagine adding something like "portions are copyright WebKit MIT license blah blah blah" (or whatever the license for WebKit is) would probably be a good idea.


calvaris:
I had a quick look and everything seems to be in order.

My only concern so far was using any ES6 capabalities that at the moment we began were not fully developed in WebKit, but the status seems to be much better now, with the support at the object syntax, arrow funcions and let and const. So you can continue as you want with the tests hygiene.

reference-implementation/web-platform-tests/readable-streams/bad-strategies.js
```diff
@@ -0,0 +1,117 @@
...
+            });
+        }, 'construction should throw a TypeError for ' + highWaterMark);
+    }
+}, 'Readable stream: invalid strategy.highWaterMark');
+
+var test2 = async_test('Readable stream: invalid strategy.size return value');
+test2.step(function() {
+    var numberOfCalls = 0;
+    var elements = [NaN, -Infinity, +Infinity, -1];
+    var theError = [];
```
tyoshino: theErrors

reference-implementation/web-platform-tests/readable-streams/bad-strategies.js
```diff
@@ -0,0 +1,117 @@
...
+        {
+            size: function() {
+                return elements[i];
+            },
+            highWaterMark: 5
+        });
+
+        var catchFunction = function(i, e) {
+            assert_equals(e, theError[i], 'closed should reject with the error for ' + elements[i]);
+            if (++numberOfCalls, elements.length) {
```
tyoshino: s/,/==/

reference-implementation/web-platform-tests/readable-streams/bad-underlying-sources.js
```diff
@@ -0,0 +1,379 @@
...
+test1.step(function() {
+    var theError = new Error('a unique string');
+    var rs = new ReadableStream({
+        get pull() {
+            throw theError;
+        }
+    });
+
+    rs.getReader().closed.then(
+        test1.step_func(function() { assert_unreached('closed should not fulfill'); }),
```
tyoshino: you can use test1.unreached_func

reference-implementation/web-platform-tests/readable-streams/bad-underlying-sources.js
```diff
@@ -0,0 +1,379 @@
...
+
+    reader.closed.then(
+        test3.step_func(function() { assert_unreached('closed should not fulfill'); }),
+        test3.step_func(function(r) {
+            assert_equals(r, theError, 'closed should reject with the thrown error');
+            test3.done();
+        }));
+});
+
+var test4 = async_test('Underlying source: throwing pull method (second pull)');
```
tyoshino: nitpick: you've changed the description to:
  Underlying source <method>: throwing method/getter
for start but not for pull.

reference-implementation/web-platform-tests/readable-streams/bad-underlying-sources.js
```diff
@@ -0,0 +1,379 @@
...
+test7.step(function() {
+    var controller;
+    var rs = new ReadableStream({
+        start: function(c) {
+            controller = c;
+        }
+    });
+
+    rs.cancel();
+    controller.enqueue('a') // Calling enqueue after canceling should not throw anything.
```
tyoshino: semicolon

reference-implementation/web-platform-tests/readable-streams/bad-underlying-sources.js
```diff
@@ -0,0 +1,379 @@
...
+    var rs = new ReadableStream({
+        start: function(c) {
+            c.enqueue('a');
+            c.enqueue('b');
+            controller = c;
+        }
+    });
+
+    rs.cancel();
+    controller.enqueue('c') // Calling enqueue after canceling should not throw anything.
```
tyoshino: semicolon

reference-implementation/web-platform-tests/readable-streams/bad-underlying-sources.js
```diff
@@ -0,0 +1,379 @@
...
+            c.close();
+            assert_throws(new TypeError(), c.close, 'second call to close should throw a TypeError');
+        }
+    }).getReader().closed.then(test11.step_func(function() { test11.done('closed should fulfill'); }));
+});
+
+var test12 = async_test('Underlying source: calling close twice on a non-empty stream should throw the second time');
+test12.step(function() {
+    var startCalled = false;
+    var readCalled = false;
```
tyoshino: I think we're sure that start() is invoked and c.close() is executed but we wanted to use the t.plan() to detect completion and it just had to count this in.

We can skip readCalled checking, I think. No strong opinion, but we can.

tyoshino: After looking at some test below, I noticed that you've introduced this check not to keep t.plan()'s expectation but really to ensure that we verify the start method is invoked. So, never mind.

tyoshino:
Reviewed bad-strategies.js and bad-underlying-sources.js in detail, and the infra part. Yeah, makes sense. Thanks for the work.

PR #399, opened 2015-10-19
Fix markup

annevk:
I didn’t try to figure out how to build the specification.


annevk:
That looks a lot better.

PR #412, opened 2015-12-04
Made SharedWorker fail if not present

calvaris:
The alternative having the whole test file failing because of it.


youennf:
Can the utility routine handle all three workers so that it gets down to a one line script element (something like: workers_test('bla.js'))?
I guess wpt infrastructure could also handle that in a better way in the future.

calvaris:
I guess there's no problem with that.

domenic:
Awesome, thanks.

Issue #379, opened 2015-07-22
Investigate merging RS and RBS


domenic:
The work in #361 moves a lot of the interesting work into the controller class, making the outer RBS class fairly simple. RBS becomes a shell that delegates to RBSC.

One possibility this opens up is that we could remove RBS, making RS just a shell which delegates to _either_ one of RSC or RBSC. This could have good ecosystem benefits by not splitting the ecosystem into two different types of streams.

To do this I think we need to:

- [x] Finish up #361.
- [x] Port RS to a similar infrastructure.
  - [x] #373
  - [x] #418 (taken over #364)
- [x] See if merging is possible and desirable.
- [x] Naming concern: https://github.com/whatwg/streams/issues/405

Related: #312.

tyoshino:
Added #373. More work can be made after it.

tyoshino:
Added #364, too. #361 is now done.

martinthomson:
Discussed at TPAC2015, the group there felt that there would need to be very good reasons if these were to remain separate.  The extension to support writing into an `ArrayBuffer` (or more generally, an `ArrayBufferView`) seems like a pure extension.

tyoshino:
We need to decide this:

ReadableByteStream has a new pull controlling mechanism. Should we port this to ReadableStream?:
- source.pull doesn't return a promise
- repeats calling pull (see where `_pullAgain` is set to `true`) using a for-loop while necessary
- when pull is requested while running enqueue, schedule run of pull by using process.nextTick

tyoshino:
Currently the merging work is happening at PR https://github.com/whatwg/streams/pull/418.

domenic:
> ReadableByteStream has a new pull controlling mechanism. Should we port this to ReadableStream?:

I don't quite understand the proposal. Can you give a bit more detail? Would it be additive, or replace the existing mechanism? How does some example code differ?

tyoshino:
Sorry for lack of words. Branched into https://github.com/whatwg/streams/issues/423 :)

tyoshino:
There's one more thing to decide regarding return type of the controllers. See https://github.com/whatwg/streams/issues/424.

tyoshino:
Done https://github.com/whatwg/streams/commit/e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393

annevk:
Very exciting, congratulations on finishing this @tyoshino!

Issue #364, opened 2015-06-16
Consider moving stuff into ReadableStreamController


domenic:
See https://github.com/whatwg/streams/pull/361#discussion_r32265033 where @tyoshino does this for ReadableByteStream and it seems to work out well.

domenic:
I started working on this in the [refactor innards branch](https://github.com/whatwg/streams/tree/refactor-innards). However, I got stuck, as there were a few state variables that I wasn't sure where they should go (reader, controller, or stream):

- disturbed is a property of the stream but is modified by stream.cancel(), reader.cancel(), and reader.read()
- state is used for some checks inside the stream, and it seems good to enforce those checks no matter what the controller implementation is. But it is mostly modified and consulted by the controller, sometimes by the reader too.
- storedError is similar to state.

In the end these are not insurmountable problems. We would just need to define some kind of contract for all of the shared properties, e.g. "ReadableStreamController can be implemented in any way as long as it has state and storedError slots" or "ReadableStreamController can have a pointer back to its owner readable stream in order to modify its state and storedError slots." But it was not as straightforward as I'd hoped so I'm probably going to set this aside for a bit longer.

tyoshino:
I was taking some of the abstract operations as representation of the contracts. For example, [[state]] is altered only via ErrorReadableStream and FinishClosingReadableStream. These operations would form an internal interface.

[[disturbed]] should stay in ReadableStream, I agree.

Ah, I didn't mean moving such a lot from Stream to Controller. I thought queuing details would fit int Controller and moved them. state, etc. should fit in Stream, I think.

tyoshino:
I'll take over https://github.com/whatwg/streams/tree/refactor-innards to do what we know we can for now.

tyoshino:
Created: https://github.com/whatwg/streams/pull/418

tyoshino:
https://github.com/whatwg/streams/issues/379 is updated to point #418. Closing this issue as it has taken over this.

PR #417, opened 2015-12-16
Fill the internal slots section of ReadableByteStreamController

tyoshino:


domenic:
Is the intent to put all the stuff in ReadableByteStreamController, then later work on #364 and #379?

Contents LGTM; let me work on figuring out why these builds are not working...

tyoshino:
Yes. It's ideal to get the RS/RBS merge issue done first, but the RBS spec has been TBA for too long time.

index.bs
```diff
@@ -2117,7 +2117,49 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
...
+  <tr>
+    <td>\[[controlledReadableByteStream]]
+    <td>The <a href="#rbs-class"><code>ReadableByteStream</code></a> instance controlled
+  </tr>
+  <tr>
+    <td>\[[underlyingByteSource]]
+    <td>An object representation of the stream's underlying byte source
+  </tr>
+  <tr>
+    <td>\[[pullAgain]]
```
domenic: Can you use the descriptions from ReadableStream for pullAgain and pulling? They are still accurate, right? Or if they mean something different here, could you explain it in a bit more detail like they do?

index.bs
```diff
@@ -2117,7 +2117,49 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
...
+    <td>A boolean flag indicating that pulling from the underlying byte source should be made once again
+  </tr>
+  <tr>
+    <td>\[[pulling]]
+    <td>A boolean flag indicating that the underlying byte source is currently being pulled
+  </tr>
+  <tr>
+    <td>\[[pendingPullIntos]]
+    <td>A List representing the stream's internal queue of pending pullInto operations
```
domenic: Let's call these "pull-into-buffer operations"

index.bs
```diff
@@ -2117,7 +2117,49 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
...
+  </tr>
+  <tr>
+    <td>\[[queue]]
+    <td>A List representing the stream's internal queue of chunks
```
domenic: wrap chunks in `<a>`

index.bs
```diff
@@ -2117,7 +2117,49 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
...
+      <th>Description (<em>non-normative</em>)</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>\[[controlledReadableByteStream]]
+    <td>The <a href="#rbs-class"><code>ReadableByteStream</code></a> instance controlled
+  </tr>
+  <tr>
+    <td>\[[underlyingByteSource]]
+    <td>An object representation of the stream's underlying byte source
```
domenic: underlying byte source needs `<a>` wrappers here and elsewhere

index.bs
```diff
@@ -2117,7 +2117,49 @@ If one were to write the <code>ReadableByteStreamController</code> class in some
...
+    <td>\[[queue]]
+    <td>A List representing the stream's internal queue of chunks
+  </tr>
+  <tr>
+    <td>\[[totalQueuedBytes]]
+    <td>An integer representing the number of bytes queued
+  </tr>
+  <tr>
+    <td>\[[closeRequested]]
+    <td>A boolean flag indicating whether the stream has been closed by its underlying byte source, but still has
```
domenic: Links for underlying byte source and chunks

index.bs
tyoshino: Done

index.bs
tyoshino: Got it! Regarding `pulling`, I made a small change (removing fulfill) since the current pull operation of the readable byte stream doesn't use a promise.

index.bs
tyoshino: Nice suggestion!

index.bs
tyoshino: Done

tyoshino: Is it fine to wrap all `chunk` occurrences in a paragraph(s), or only the first one like dictionary?

index.bs
tyoshino: Done

tyoshino:
Addressed comments: New commit: 97e0c9730040f7504db0a54a1bfe9169b56fdaee

index.bs
domenic: I'm not really sure; maybe let's lean toward only the first one? I don't have a strong opinion.

tyoshino: Got it

Issue #372, opened 2015-07-02
Separate out exported abstract ops from internal ones


domenic:
Ideally, other specs should only refer to a few of our abstract ops, e.g.:
- AcquireReadableStreamReader ( stream )
- CloseReadableStream ( stream )
- EnqueueInReadableStream ( stream, chunk )
- ErrorReadableStream ( stream, e )
- IsReadableStream ( x )
- IsReadableStreamLocked ( stream )
- TeeReadableStream ( stream, shouldClone )

plus maybe a few more. Others are "implementation details", e.g. FinishClosingReadableStream ( stream ) or RequestReadableStreamPull ( stream ) or similar. Additionally, for each of the above, there are some "usage details" that are important. E.g. see the intro of https://streams.spec.whatwg.org/#enqueue-in-readable-stream

A few strategies:
- Separate sections, e.g. exported abstract ops and internal abstract ops
- Some kind of tag, like that proposed for #371 
- A separate "for other specifications" section describing the usage of exported abstract ops
  - Pro: authors of other specs probably don't care to actually read the details of e.g. EnqueueInReadableStream
  - Con: duplication; keeping in sync; is this normative or non-normative?
- Some kind of toggle switch that gives you a "for other spec authors" view, which hides any non-exported abstract ops (and maybe hides the algorithms for the exported ones?)

Thoughts appreciated. Probably especially from @annevk as Fetch editor.

annevk:
I think ideally you can find out from the name which is private and which is public. Would prefixing internal methods with "@" be too weird?

domenic:
That seems pretty weird but maybe it will grow on me. You like it better than the tag idea?

annevk:
A bit, the tags might not make it into other specifications.

domenic:
Yeah, they wouldn't... but if everyone behaves themselves, neither would any @-prefixed internal methods :P

annevk:
This has often confused me with ECMAScript too. And is a problem with most specifications today. A lot of algorithms and not clear which are reusable and which are not. Glad you're leading the way.

tyoshino:
Any of separate section, tag, prefixing work for me. Regardless we put tag/prefix or not, grouping into two sections would improve readability, I guess.

domenic:
One way of doing this is to move stuff like https://fetch.spec.whatwg.org/#streams into the spec. That way we achieve a few things at once:

- Let streams define how to manipulate streams, instead of letting Fetch do it
[...]

PR #418, opened 2015-12-17
Merge ReadableByteStream into ReadableStream

tyoshino:
(This post was initially empty, but to summarize the changes, added a check list)

- [x] sync the spec text with the new reference implementation
- [ ] ~~merge ReadableStreamController and ReadableByteStreamController~~
- [ ] merge ReadableStreamReader and ReadableByteStreamBYOBReader
- [ ] https://github.com/whatwg/streams/issues/423
  - [x] add byobRequest getter to the controller
  - [ ] ~~adopt the while-loop based pulling routine for the non-byte source version~~
  - [ ] ~~stop delaying pull based on the return value of the last pull call~~
- [x] port the asserts in the non-byte source version to the byte source version
- [x] remove auto release feature from the byte source version
- [x] rename Byob to BYOB
- [x] move TransferArrayBuffer to helpers.js
- [x] make the default highWaterMark of the byte source version to 0
- [x] port the functionality that the start method can delay pulling by returning a pending Promise to the byte source version
- [x] add `[[disturbed]]` feature to the byte source version
- [x] port highWaterMark mechanism to ReadableByteStreamController
- [x] merge ReadableStream and ReadableByteStream
  - now it behaves differently based on whether the given source has pullInto or not
  - [x] bunch of renaming for merge
- [x] merge ReadableStreamReader and ReadableByteStreamReader
- [x] https://github.com/whatwg/streams/issues/424
  - [x] make controller.close() and controller.enqueue() fail when the stream is not in the readable state
  - [x] make controller.enqueue() throw a predefined TypeError, not the stored error
- [x] Rename ReadableStreamController to ReadableStreamDefaultController
- [x] Rename ReadableStreamReader to ReadableStreamDefaultReader
- [x] Rename ReadableByteStreamController to ReadableStreamBYOBController


tyoshino:
Since some changes such as auto-release removal are not reflected to RBS yet, they don't looks very close yet. But except for them, they share most of code now.

To make it clear which operation is exposed to controller implementations, I've put some comments.

tyoshino:
Note
- No spec text change made yet
- Tee is not yet updated yet

to get initial high level review first before working on them

tyoshino:
Here's a quick explanation of the interfaces between controllers and the readable byte stream.

Readable byte stream controller:
- knows what kind of reader the stream is locked to by `ReadableByteStreamHasReader()` and `ReadableByteStreamHasByobReader()`
- knows how many requests there are by `GetNumReadRequests()` and `GetNumReadIntoRequests()`
- responds to the requests by `RespondToReadRequest()` and `RespondToReadIntoRequest()`
- manipulates the stream's state by `CloseReadableByteStream()`, `ErrorReadableByteStream()`
- invokes the pull operation on the underlying byte source appropriately based on the signals from the stream (reader) which are `PullFromReadableByteStream()` and `PullFromReadableByteStreamInto()`. Controllers must implement these operations.

domenic:
Awesome, thanks so much for working on this.

So the readers are still owned by the stream? That is probably good, and simpler than what I was originally thinking of.

Let me try to write a summary to help get my head around it. I will use ReadableStream since it is simpler.

ReadableStream has:
  - state, reader, storedError, disturbed, and controller internal slots
  - AddReadRequestToReadableStream, GetNumReadRequests, RespondToReadRequest "spec internal methods" to allow the controller to manipulate its controlled readable stream (and its reader) without manipulating internal state directly.

ReadableStreamController has:
  - controlledReadableStream, underlyingSource, strategySize, strategyHWM, queue, started, closeRequested, pulling, pullAgain internal slots.
  - CancelReadableStreamController, EnqueueInReadableStreamController, ErrorReadableStreamController, GetReadableStreamControllerDesiredSize, ReadableStreamControllerCallPullIfNeeded, ShouldPullReadableStreamController "spec internal methods".

The eventual path forward, then, would be:

- Convert ReadableStreamController's spec-internal methods into actual internal "methods", probably written something like `ReadableStreamController@[[Cancel]](...)`. In source code we can name these `_Close()` and have them be methods in the class, which will be nice.
- Match up all controller internal methods so that both ReadableStreamController and ReadableByteStreamController have the same set of internal methods.
- Convert ReadableStream to call `this@[[controller]]@[[Cancel]](...)`, thus getting polymorphism for when we allow constructing ReadableStreams with ReadableByteStreamControllers, and merging the two classes.

This sounds like it will work really well. There are probably some more complexities for the controller <-> reader interaction, but what you have so far makes sense.

tyoshino:
> So the readers are still owned by the stream? That is probably good, and simpler than what I was originally thinking of.

Yes. Operations on the reader is hidden to the controller.

> ReadableStream has:
>
> - state, reader, storedError, disturbed, and controller internal slots

Right.

> - AddReadRequestToReadableStream, GetNumReadRequests, RespondToReadRequest "spec internal methods" to allow the controller to manipulate its controlled readable stream (and its reader) without manipulating internal state directly.

Right. `[[state]]` and `[[storedError]]` are also visible to the controller but not manipulated directly by the controller.

> ReadableStreamController has:
> 
> - controlledReadableStream, underlyingSource, strategySize, strategyHWM, queue, started, closeRequested, pulling, pullAgain internal slots.
> - CancelReadableStreamController, EnqueueInReadableStreamController, ErrorReadableStreamController, GetReadableStreamControllerDesiredSize, ReadableStreamControllerCallPullIfNeeded, ShouldPullReadableStreamController "spec internal methods".

and `CloseReadableStreamController`. Currently they all contains "ReadableStreamController" in their names and the others don't.

Ah, I just noticed that in the last post I forgot to include `CancelReadable(Byte)StreamController` when listing the interfaces that controllers must implement.

> The eventual path forward, then, would be:
>
> - Convert ReadableStreamController's spec-internal methods into actual internal "methods", probably written something like ReadableStreamController@[[Cancel]](...). In source code we can name these _Close() and have them be methods in the class, which will be nice.

It's really nice! It'll be clear which operation belongs to which object.

> - Match up all controller internal methods so that both ReadableStreamController and ReadableByteStreamController have the same set of internal methods.
> - Convert ReadableStream to call this@[[controller]]@[[Cancel]](...), thus getting polymorphism for when we allow constructing ReadableStreams with ReadableByteStreamControllers, and merging the two classes.

Yes!

> 
> This sounds like it will work really well. There are probably some more complexities for the controller <-> reader interaction, but what you have so far makes sense.

Yeah. The interface adds complication but modularization would make each class easier to understand, I guess.

tyoshino:
Up to 7beb1593dd35097000fb7d68307174272dd2a04c, made some bug fixes, more refactoring, enabled tee

tyoshino:
Made some more commits. Merged 2e7f87e45806d58114592c923aed299798d10161, added distributed property, and reflected auto release feature removal to the readable byte stream.

tyoshino:
2bf449735218b41328f61102e4d3751adfe0aa49 is the last commit. Its diff is very big, sorry.

Finally, I've finished merging RS and RBS.
- we no longer have ReadableByteStream, ReadableByteStreamReader
- ReadableStream has getByobReader which works only when the stream was constructed with a byte source which is a source with the `pullInto` method
- a byte source is given a ReadableByteStreamController
- a non-byte source is given a ReadableStreamController
- ReadableStream and ReadableStreamReader are representing the minimal semantics required. Implementation details are separated into the controllers

tyoshino:
BTW, I guess even ReadableStreamByobReader and ReadableStreamReader can be merged into one reader.

tyoshino:
The next step we need to take is to make the controllers and underlying source API consistent as much as possible.

tyoshino:
Two more commits made today. Now ReadableByteStream has the same queuing strategy mechanism as ReadableStream.

domenic:
I'll try to spend some time reviewing today; this is looking really exciting. Thank you for taking on such a big task!

reference-implementation/web-platform-tests/readable-streams/general.js
```diff
@@ -608,6 +608,51 @@ promise_test(() => {
...
+  let pullCalled = false;
+
+  const theError = new Error('rejected!');
+  const otherError = new Error('other!');
+
+  const rs = new ReadableStream({
+    pull(c) {
+      pullCalled = true;
+      c.error(theError);
+      throw theError;
```
domenic: should be otherError, right? And probably want to use `{ name: 'rejected!' }`, `{ name: 'other' }` so that you can use `promise_rejects` and stuff. See https://github.com/whatwg/streams/commit/7f19175415d9b57af04cb8c2cd355ca38101f0aa

reference-implementation/test/readable-byte-stream.js
```diff
@@ -1538,8 +1509,6 @@ test('ReadableByteStream: Multiple read(view), close() and respond()', t => {
     t.equals(view.byteLength, 0, '1st read: byteLength');
   });
 
-  t.equals(pullIntoCount, 1);
```
domenic: Why aren't we testing pullIntoCount anymore?

reference-implementation/test/readable-byte-stream.js
```diff
@@ -1,25 +1,13 @@
 const test = require('tape-catch');
 
 test('ReadableByteStream can be constructed with no errors', t => {
```
domenic: All the test titles in this file will need updating to not be "ReadableByteStream" but I am not sure what the right substitute is. Maybe "ReadableStream with pullInto".

reference-implementation/test/readable-byte-stream.js
```diff
@@ -1,25 +1,13 @@
...
-  t.doesNotThrow(() => new ReadableByteStream(x),
-                 'ReadableByteStream constructed with an undefined variable as parameter');
+  t.doesNotThrow(() => new ReadableStream({ pullInto() {} }),
+                 'ReadableByteStream constructed with an empty underlying byte source object as parameter');
   t.end();
 });
 
 test('ReadableByteStream: getReader(), then releaseLock()', t => {
-  const rbs = new ReadableByteStream({
-    pull() {
```
domenic: Is pull called now in these sort of cases?

domenic:
Review notes for readable-stream.js since I can't comment inline on such a big diff :). Overall looks quite solid; all my comments are fairly minor. I'm excited.

Per http://w3ctag.github.io/design-principles/#casing-rules we should do s/Byob/BYOB everywhere.

We should come up with a different name than ReadableStreamController/ReadableStreamReader since it is confusing; to me I would think both the BYOB and the normal one are "readable stream controllers/readers". Maybe ReadableStreamDefaultController and ReadableStreamDefaultReader? Also maybe rename ReadableByteStreamController to ReadableStreamBYOBController?

_controller should be _readableStreamController otherwise it's easy for other specs to accidentally collide with an internal slot named [[controller]].

It seems like parts of releaseLock are duplicated and can be moved into ReleaseReadableStreamReaderGeneric?

> ```js
>     this._cancel = CancelReadableStreamController;
>    this._pull = PullFromReadableStreamController;
> ```

Are you planning to move to the internal methods idea discussed in https://github.com/whatwg/streams/pull/418#issuecomment-165598765 ? I think in the reference implementation I would inline the method definitions in the class (`_cancel() { ... }`). In the spec they would be defined under the class as well.

I wonder if we should use methods for everything instead of just the polymorphic operations? Hmm let's stick with just the polymorphic ones for now. But maybe it will make things clearer eventually now that we have this more complex setup. I can do it in an editorial follow-up if we decide to go that route.

> ```js
>  constructor(controlledReadableStream, underlyingByteSource, size, highWaterMark) {
> ```

Size seems unused. I don't think the signatures need to match, so we can probably remove it.

> ```js
> if (Number.isNaN(highWaterMark)) {
> ```

maybe pull out a ValidateAndNormalizeHighWaterMark to share beween ValidateAndNormalizeQueuingStrategy and here.

> ReadableStream API exposed for controllers.

These should probably generally get asserts along the lines of IsReadableStreamReader(stream@[[reader]]) and IsReadableStreamBYOBReader(stream@[[reader]]) to make it clear which apply to which.

> ```js
> // A client of ReadableStreamController may use this function directly to bypass state check.
> ```

The state checks should probably be added as asserts in all cases possible.

> FillPullIntoDescriptorFromQueue

This reminded me that we should consider how we're going to divide these when speccing them, and try to arrange them correspondingly.

I am not sure I have all the categories but maybe something like:

- ReadableStream Abstract Operations
  - Utilities for Other Specifications (includes IsReadableStreamDisturbed but also AcquireReadableStreamReader, TeeReadableStream, ...)
  - Type-checking Helpers
  - Operations on `ReadableStream`s, Used by Controllers
    - will also include a note about Utilities from Other Specifications that can be used here
    - will include a note that ReleaseReadableStreamReaderGeneric can be used here (to avoid the wrapper ReleaseReadableStreamReaderGenericForController)
  - Operations on Readers (used by anyone in particular??)
  - Operations for `ReadableStreamController`s
  - Operations for `ReadableByteStreamController`s
    - this is where we put helpers like FillPullIntoDescriptorFromQueue

> TransferArrayBuffer

should move to helpers.js

reference-implementation/web-platform-tests/readable-streams/general.js
tyoshino: Oh, thanks for the advice. Done.

tyoshino: I've also replaced the assert calls in the promise_test above this one with promise_rejects.

pullCalled has been removed as it's sufficient to expect the exception passed in the pull method.

reference-implementation/test/readable-byte-stream.js
tyoshino: Yes. Before this merge, the RBS had a behavior different from RS. The RBS invoked pull only in response to read() call. Now, it's given the queuing strategy mechanism. The HWM is set to 1, and so it invokes pull automatically once start completes and when the queue size is less than 1.

tyoshino: I'll add some tests to check the automatic pull invocation.

reference-implementation/test/readable-byte-stream.js
tyoshino: How about ReadableStream with byte source? Replaced tentatively. Also replaced `rbs` with `stream`.

reference-implementation/test/readable-byte-stream.js
tyoshino: I removed them because due to the addition of pull blocking by start method the expectations became stale.

I recovered some but realized we need to resolve https://github.com/whatwg/streams/issues/423 first.

tyoshino:
> Per http://w3ctag.github.io/design-principles/#casing-rules we should do s/Byob/BYOB everywhere.

Done

> We should come up with a different name than ReadableStreamController/ReadableStreamReader

Agreed, but this depends on how https://github.com/whatwg/streams/pull/418#issuecomment-171290358 goes.

> _controller should be _readableStreamController otherwise it's easy for other specs to accidentally collide with an internal slot named [[controller]].

Done

> It seems like parts of releaseLock are duplicated and can be moved into ReleaseReadableStreamReaderGeneric?

Done

tyoshino:
> constructor(controlledReadableStream, underlyingByteSource, size, highWaterMark) {
> Size seems unused. I don't think the signatures need to match, so we can probably remove it.

Done

> if (Number.isNaN(highWaterMark)) {
> maybe pull out a ValidateAndNormalizeHighWaterMark to share beween ValidateAndNormalizeQueuingStrategy and here.

Done

> ReadableStream API exposed for controllers.
> These should probably generally get asserts along the lines of IsReadableStreamReader(stream@[[reader]]) and IsReadableStreamBYOBReader(stream@[[reader]]) to make it clear which apply to which.

Done

> // A client of ReadableStreamController may use this function directly to bypass state check.
> The state checks should probably be added as asserts in all cases possible.

Done

tyoshino:
> this._cancel = CancelReadableStreamController;
> this._pull = PullFromReadableStreamController;
> Are you planning to move to the internal methods idea discussed in #418 (comment) ? I think in the reference implementation I would inline the method definitions in the class (_cancel() { ... }). In the spec they would be defined under the class as well.

Sorry! I forgot that. Inlined.

> I wonder if we should use methods for everything instead of just the polymorphic operations? Hmm let's stick with just the polymorphic ones for now. But maybe it will make things clearer eventually now that we have this more complex setup. I can do it in an editorial follow-up if we decide to go that route.

Yeah. Putting the abstract operations into classes just makes the code readable. But except for that point I don't have strong opinion now.

> FillPullIntoDescriptorFromQueue
> This reminded me that we should consider how we're going to divide these when speccing them, and try to arrange them correspondingly.

The list you wrote looks good. I just removed ReleaseReadableStreamReaderGenericForController. It's no longer necessary as automatic release feature is gone. I've moved the position of ReleaseReadableStreamReaderGeneric and renamed it by dropping Generic.

> TransferArrayBuffer
> should move to helpers.js

Done

tyoshino:
Pushed all the commits to address domenic's comments.

tyoshino:
> FillPullIntoDescriptorFromQueue
This reminded me that we should consider how we're going to divide these when speccing them, and try to arrange them correspondingly.

Just noticed that we had an old issue for this (not for all topics): https://github.com/whatwg/streams/issues/372

reference-implementation/test/readable-byte-stream.js
domenic: Works for me!

reference-implementation/test/readable-byte-stream.js
domenic: Should we maybe set the default HWM to 0 if a pullInto is supplied? Always calling pull asking for one byte seems kind of silly.

tyoshino: I see. That makes sense. Then, a ReadableStream with byte source invokes pullInto in response to read(view) and pull only in response to read() by default, but when a non-zero highWaterMark is specified on construction, it pulls more based on the queue size. Fixed.

tyoshino:
Added a list of things done in this PR to the OP.

tyoshino:
Introduced .byobRequest getter and moved the respond() method onto it as discussed at https://github.com/whatwg/streams/issues/423#issuecomment-174425462

tyoshino:
BTW, I made the reference implementation too optimized. E.g. when read() is issued and the source can immediately respond to it, we don't skip pushing the request to the pending request queue but just fulfill it. It's making the code complicated. I plan to rather simplify them than keeping the micro optimization.

tyoshino:
- Added byob flag to underlying source
- Merged pull and pullInto

domenic:
I'm looking through the OP checklist and see

>  merge ReadableStreamController and ReadableByteStreamController

I don't think this is a good idea, or feasible. That is where the main separation is. RBSC has byobRequest on it, after all.

>  merge ReadableStreamReader and ReadableByteStreamBYOBReader

I'm less sure about this, but it still seems reasonable to me that these be separate types... what do you think?

tyoshino:
> I don't think this is a good idea, or feasible. That is where the main separation is. RBSC has byobRequest on it, after all.

Agreed! Yeah, given byobRequest addition, it's no longer good idea.

tyoshino:
As discussed at https://github.com/whatwg/streams/issues/423#issuecomment-174425779, I've removed the synchronous repeated pulling mechanism. Pulling criteria checking code has been simplified to be the same as ReadableStream with non-byte source.

Error propagation was and is still done correctly.

tyoshino:
> I'm less sure about this, but it still seems reasonable to me that these be separate types... what do you think?

Hmm, ok. Having read(view) on a reader obtained by getReader() is just confusing. ReadableStream with non-byte source still has getBYOBReader() but should be less confusing.

tyoshino:
Sorry. The suggestion was not to add read(view) on the reader obtained by getReader(), but make the reader obtained by getBYOBReader() (or could be renamed to getByteReader() or something) provide both read() and read(view).

domenic:
Oh, I see, I didn't realize. So our two options are:

- non-BYOB streams have getReader() that if you do .read(view) it fails; BYOB streams have getReader() that allows .read(view)
- non-BYOB streams have getReader() that if you do .read(view) it fails; BYOB streams have .getReader() that also disallows .read(view), but they have .getByobReader() that allows .read(view) (and disallows .read()?).

Now that you lay this out, I think option 1 is better. But! I still don't think we need to merge the two types, unless that is convenient for spec/implementation. We can just have .getReader() return a normal reader for non-BYOB streams, and a BYOB reader for BYOB streams. Or we can merge them into one type; up to you :)

tyoshino:
Sorry for delay!

Yeah, the option 1 is the plan I imagined. Maybe to make it less confusing, we need to rename `.read(view)` to `.readInto(view)`.

One good thing is that we won't have `.getBYOBReader()` on a ReadableStream which is constructed with the `byob` option set to **false**.

In terms of code cleanness, it's ok not to make the merge happen since we've already simplified it by merging the operations behind the methods enough.

tyoshino:
Maybe it's time to proceed to make this ready for commit. I'll create another branch not to lose the review interaction details, and add text changes to it.

tyoshino:
Created a new PR: https://github.com/whatwg/streams/pull/430

tyoshino:
It's time to do section reorganization previously suggested/investigated by @domenic at https://github.com/whatwg/streams/pull/418#issuecomment-171822826.

In addition to the idea by Domenic, can we separate internal operations and ones can be used directly by clients of each class? Too much?

For example, `InitializeReadableStreamReaderGeneric()` is just common code factored out from the two readers. Not indented for use by clients like `ReadFromReadableStreamReader()`.

tyoshino:
Regarding spec writing, I'd like to get agreement on the following items. Maybe it's good time to establish consistent rules.

- id for headings: Keep using abbreviations like `rs-` or just write down all phrases?
  - e.g. `rs-controller-enqueue` for `enqueue()` in ReadableStreamController, or rename it to `readable-stream-controller-enqueue`?
- operation naming: We've been giving names that can be read roughly as an English phrase. I think. But This led to long moving of functions on renaming as they've been placed in the lexicographical order. As we're introducing more hierarchical sections for the abstract operations, the problem would no longer matter, but that said, common prefix is simpler and some times readable than the English phrase like style. What do you think? E.g. `ReadFromReadableStreamReader` would be renamed to just `ReadableStreamReaderRead`. All the operations on ReadableStream would have the same prefix "ReadableStream".
- Any preference on the ordering of operations implementing the logic of class methods excluding the brand checking? E.g. GetReadableStreamControllerDesiredSize() would be placed after CloseReadableStreamController() when ordered lexicographically, but the corresponding methods are placed in a different order since in the class representation we're placing getters/setters before the other methods.

tyoshino:
Domenic, in the example at https://github.com/whatwg/streams/pull/418#issuecomment-171822826 you've put all the abstract operations in the same section but giving sub sections under it. Does this mean that you think we should use one section parallel to sections explaining each class?

Can we choose to put each operation to the corresponding big section explaining the class the operation operates on?

domenic:
> In addition to the idea by Domenic, can we separate internal operations and ones can be used directly by clients of each class? Too much?

Hmm. My first impression is that it's a bit too much, but I'm not sure. If you want to draft a quick bullet-point version of what the TOC would look like with your idea, we could discuss further.

> Domenic, in the example at #418 (comment) you've put all the abstract operations in the same section but giving sub sections under it. Does this mean that you think we should use one section parallel to sections explaining each class?
>
> Can we choose to put each operation to the corresponding big section explaining the class the operation operates on?

What I meant there was that, similar to the current spec, we want a section underneath "3 ReadableStreams" for abstract operations. Call it "3.5 Readable Stream Abstract Operations". Within that there'd be subsections.

If you think the abstract operations divide up cleanly enough that they could be nested under the class sections (so, e.g., "3.2.5 Abstract Operations for `ReadableStream`", "3.3.5 Abstract Operations for `ReadableStreamController`", etc.) then maybe that would work. I am a bit skeptical though. It seems like a lot of them are kind of hard to decide which class they are associated with. The organization in my comment seems more likely to be useful for readers.

In general, I'm open to suggestions on proposed organization; the best way to do that is to make concrete proposals with full TOCs.

> id for headings

I think we should keep the abbreviations, and we should endeavor to preserve as many links as possible. Nice short and well-curated IDs makes sense to me, for a spec that is of manageable size like this one.

I could imagine moving the abstract ops to a different ID style, like `#ao-IsReadableStream`. But that would break existing incoming links from e.g. Fetch, and I am not sure it is worth it.

> operation naming

I agree that it is probably a good idea to rearrange them with prefixes like you suggest.

> Any preference on the ordering of operations implementing the logic of class methods excluding the brand checking?

I think it's OK for them to not correspond perfectly. Let's keep it lexicographic.

tyoshino:
Done
- Add all new methods, kill all old methods.
- Rename ReadableStreamController to ReadableStreamDefaultController
- Rename ReadableStreamReader to ReadableStreamDefaultReader
- Rename ReadableByteStreamController to ReadableStreamBYOBController
- Reorganize the section based on the idea at https://github.com/whatwg/streams/pull/418#issuecomment-171822826 and discussion at https://github.com/whatwg/streams/pull/418#issuecomment-190276398
- IDs: Given consistent IDs. Please review if they comply your principles, Domenic.
- Renamed operations to have a common prefix, and then reordered them
- Made bunch of bug and typo fixes.

tyoshino:
So, it's done. Sorry but the diff is huge... Please take a look at https://github.com/whatwg/streams/pull/430.

PR #419, opened 2015-12-18
Call ErrorReadableStream only when the state is "readable"

tyoshino:
Invocation of the strategy functions may make the stream errored. Though
such strategies are just broken, we should not throw an assertion error for
it.


domenic:
Nice catch!!

PR #420, opened 2015-12-23
Improved tests abour erroring a RS when enqueing

calvaris:


youennf:
For the first modified test, it might be good to create two different errors.

calvaris:
Good idea, I separated both errors.

reference-implementation/web-platform-tests/readable-streams/bad-strategies.js
```diff
@@ -19,9 +19,10 @@ test(() => {
 
 }, 'Readable stream: throwing strategy.size getter');
 
-test(() => {
+promise_test(() => {
 
-  const theError = new Error('a unique string');
+  const controllerError = new Error('a unique string');
```
domenic: Can you use `{ name: 'error 1' }` and `{ name: 'error 2' }` instead, which would allow us to use `assert_throws` and `promise_rejects`?

calvaris: I did those changes in the new version.

domenic:
Thank you, merged as 7f19175415d9b57af04cb8c2cd355ca38101f0aa!

We should probably move these over to web-platform-tests and not duplicate effort, but I guess there's another thread on that.

PR #422, opened 2016-01-13
Added message to check in bad strategies

calvaris:
Just a quick nit


domenic:
This doesn't work, as promise_rejects doesn't accept a message argument :(

Issue #319, opened 2015-04-07
Port locking infrastructure to writable streams


domenic:
Writable streams need a similar way of being locked for exclusive writing, as readable streams do with their readers. This is notably crucial for off-main-thread piping (#97).

One problem that immediately presents itself, looking at the current list of writable stream properties and methods (closed/ready/state/abort()/close()/write()), is that pretty much all of them belong on the writer. Does that mean writable streams themselves become empty objects with getWriter() methods!?

I guess we could add a few pass-throughs (like ReadableStream.prototype.cancel acquiring, then releasing, a reader). Do we just add pass-through for everything? Well, probably not for ready/state/closed at least, as that causes problems. But if all we have is abort()/close()/write() passthroughs, doesn't that encourage the bad practice of ignoring backpressure signals? Maybe just abort() and close()?

tyoshino:
> become empty objects with getWriter() methods!?

Yeah, pipeTo etc. all belongs to the readable side.

What do you mean by pass-through? Keeping a method also on the WritableStream? As you just said, I also think we should basically limit them to WritableStreamWriter as much as possible.

> just abort() and write()?

Maybe you meant abort() and close().

domenic:
> What do you mean by pass-through? Keeping a method also on the WritableStream?

I mean, like we do for ReadableStream.prototype.cancel, have versions of them on the WritableStream which throw if the WritableStream is locked.

> Maybe you meant abort() and close().

Yes, editing OP.

tyoshino:
OK. I slightly prefer not putting them on WritableStream. The same about `.cancel()`. Basically.

But given that we have `.cancel()` on ReadableStream, maybe it's better to make them consistent. I.e. have `abort()` or both `abort()` and `close()`.

domenic:
Plan:

```js
class WritableStream {
  constructor(underlyingSink = {}, { size, highWaterMark = 0 } = {})
  
  get locked()

  getWriter()
  abort(reason)
  close()
}

class WritableStreamWriter {
  constructor(stream)
  
  get closed()
  get ready() // see #318

  abort(reason)
  close()
  releaseLock()
  write(chunk)
}
```

No need for WritableStreamController. Can eliminate `error` parameter to underlying sink `start()`.

tyoshino:
PR created: https://github.com/whatwg/streams/pull/462

tyoshino:
Fixed. #462 has been succeeded by #488. Only pass-through for abort() has been added.

Issue #318, opened 2015-04-07
Port precise flow control to writable streams


domenic:
After https://github.com/whatwg/streams/pull/317, readable streams will have precise flow control. We should add something similar to writable streams. However, the shape of it will likely depend on the resolution of https://github.com/whatwg/streams/issues/316.

domenic:
Questions:

### What should we expose?

I see two choices:

- `queuedSize` = total size in the queue.
- `desiredSize` = high water mark minus total size in the queue. Can be negative or zero. Represents how much you should write right now.

### How should the `ready` signal work?

A few possibilities:

1. The current `ready` signal, which fulfills when queuedSize dips below the high water mark (or desiredSize goes above zero).
2. A different signal, call it `drained`, which fulfills when queuedSize reaches zero (or desiredSize reaches highWaterMark).
3. A generic `queuedSizeChanged` signal, which fulfills both when you write to the stream and when the underlying sink finishes consuming a chunk.
4. A more monotonic `queueSizeDecreased` signal (probably there is a better name), which fulfills when the underlying sink finishes consuming a chunk. The semantics here are a bit unclear to me honestly.
5. A customizable `.waitForQueueSize(n)` which waits for the queue size to dip below `n`, or `.waitForDesiredSize(n)` which waits for the desired size to go above `n`.

I am leaning toward the simplicity of (1) and/or (2). But (5) also works. I don't really like (3) or (4).

### Who should determine the high water mark?

We could do a few different choices:

1. Get rid of the high water mark entirely. Producers need to monitor `queuedSize` and adjust to their comfort level. Writable streams don't get to signal what they want.
2. As-is now, make it part of the queuing strategy, so that it is set at writable stream creation time by the stream creator.
3. Make it determined by `getWriter({ highWaterMark })`. This is essentially saying that high water mark is a way to save the producer some arithmetic over option (1). The producer would pass in highWaterMark = x and get back desiredSize, instead of getting back queuedSize = y and calculating desiredSize = x - y. See also #375.

I am a bit uncomfortable with (1) and (3) because I think many writable streams will have at the very least a desired chunk size that they would want to signal.

tyoshino:
> What should we expose?

I think `desiredSize` is the right thing to expose. The queue size is a part of the internal state/details of the stream we don't have to expose. I'd just return what the producer need to know, which is desiredSize.

tyoshino:
> How should the ready signal work?

[...]

PR #429, opened 2016-02-09
Implement flow-control in TransformStream fixes #330 maybe

### LINKED ISSUE #330: TransformStream: transform() is never invoked via pull()
When `write()` of the underlying sink of `writable` is called, `maybeDoTransform()` always proceeds to run `transform`. It never stop with `writeChunk` pending but `transform` not started. Maybe what we want to do is not to start `transform` while there's back-pressure on `readable`.

asutherland:
Promise-ify the TransformStream so that it always waits for both a
write() and a pull() to have happened, returning a Promise in both
cases so that callers can know when it's safe to call again.

Previously pull() had no effect and the TransformStream was strictly
pumped by its write() callers.

NB: I may be making a dangerous assumption about pull() always being called.  I will endeavor to better come up-to-speed on the spec and reference implementation and investigate this on my own in the next ~1 week.  But in the meantime, I figured I'd put this here since it does pass the existing tests and the test I added that's only slightly pyramid-of-doomy.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -11,20 +11,15 @@ export default class TransformStream {
       throw new TypeError('transform must be a function');
     }
 
-    let writeChunk, writeDone, errorWritable;
-    let transforming = false;
-    let chunkWrittenButNotYetTransformed = false;
+    let resolveWriteChunk, resolvePullRequested, nextTransformCompleted;
+    let errorWritable;
     this.writable = new WritableStream({
       start(error) {
         errorWritable = error;
       },
       write(chunk) {
-        writeChunk = chunk;
-        chunkWrittenButNotYetTransformed = true;
-
-        const p = new Promise(resolve => writeDone = resolve);
-        maybeDoTransform();
-        return p;
+        resolveWriteChunk(chunk);
+        return nextTransformCompleted;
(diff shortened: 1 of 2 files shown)


asutherland:
It appears I did make a bad assumption about pull() being called whenever the previous pull() promise resolves.  If the transformation function is something cool like a filtering function that does not always enqueue a result, then RequestReadableStreamPull as invoked by ReadFromReadableStreamReader will not reissue the pull, breaking stream flow.

I suspect reissuing the pull may make infinite loops quite likely, suggesting my patch here that tries to be clever with Promise pairing is not the way.  I think I better understand the discussion on #330 now and will look into using `ready` or the propagation of the `write` result if I don't hear better suggestions here or on that bug.

I'm going to leave this PR open for now since I have a test and am about to push my follow-up test that demonstrates why my current PR is no good and plan to try and provide a better patch tomorrow.  But I won't take offense if you close this until I have something that works! :)

domenic:
I think it's fine to leave this open as a reminder that transform streams are not in great shape right now :). First we need to update writable stream (#318 and #319) and then look into fixing all the outstanding transform stream issues. Thanks for working on it though!

domenic:
Going to close this in preparation for a transform stream overhaul :)

Issue #424, opened 2016-01-15
Revisiting what's reasonable return value for each method of the controller


tyoshino:
The return value of the controller methods are kinda inconsistent/strange in the following point.

- `ReadableStreamController.close()` when the stream is already closed
  - I remember in some issue/PR I (or domenic??) said "because a stream may get closed by cancel() which is an action by the consumer, blaming the controller for unsolicited closure (i.e. _closeRequested === false) is not good." or something and decided not to throw for this case.
  - But the controller is notified of cancel. Calling close() after receiving cancel signal should be considered to be ... misbehaving?
- The same for `ReadableStreamController.enqueue()`
- They're inconsistent with `ReadableStreamController.error()`. It fails if the stream is either closed or errored.
- When the stream is errored, `ReadableStreamController.error()` throws the stored error while `close()` throws a TypeError. I think both should throw a TypeError saying "you're misbehaving".

Chromium has shipped it but it's not too late to fix them if necessary. They're "fixed" in RBSController being redesigned in https://github.com/whatwg/streams/issues/379 based on my thoughts. I'd like to get your opinion on this.

domenic:
> Chromium has shipped it

Only behind a flag!! We're safe.

I think I agree with all your proposed fixes. Some of the issues you point out just sound like bad inconsistencies. Your fixes generally move in the direction of adding more errors for potential misbehavior. We can always loosen them if we find out that they are problematic later (e.g. for the close case especially).

tyoshino:
Done in https://github.com/whatwg/streams/pull/418. As noted in https://github.com/whatwg/streams/pull/418#issuecomment-173195730, added a list of things done in the PR to the OP of the PR. The list includes the action items of this issue.

Issue #423, opened 2016-01-15
Finalize pull/pullInto behavior


tyoshino:
(Branched from https://github.com/whatwg/streams/issues/379#issuecomment-171823158)

## Current situation

Based on the result of the discussion at https://github.com/whatwg/streams/issues/363, we've introduced pullInto for ReadableByteStream.

The pull and pullInto in the ReadableByteStream were designed to be invoked in response to read() and read(view) call on the stream. It was temporary design.

Now, in the PR https://github.com/whatwg/streams/pull/418, I'm converging ReadableStream and ReadableByteStream. Based on the conclusion on https://github.com/whatwg/streams/issues/353, ReadableByteStream has a queue in it as well as ReadableStream. I've also ported the strategy based queuing control mechanism to ReadableByteStream for converging. So, now the 1:1 correspondence between read()/read(view) and pull/pullInto is impossible.

I'm trying to give a reasonable semantics to them.

----

## Role of pullInto

Currently, pullInto has a role of notifying the underlying byte source of the outstanding TypedArray (and region on it) to which the underlying byte source should write generated data. So, whenever:

- a new TypedArray comes
- or the outstanding TypedArray is partially filled
- or filled the outstanding TypedArray but there're more TypedArrays waiting to get filled

the controller should invoke a pullInto.

Unlike pullInto, the signal of pull is just that "the amount in the queue has changed". This is probable by looking at `desiredSize`.

----

## Plan taken in the PR

My plan was:

- pull is invoked when:
  - read() is called when the queue is empty
  - read() is called and it consumed all or part of the queue
  - read(view) is called and it consumed all or part of the queue
- pullInto is invoked when:
  - read(view) is called when the queue is empty
  - read(view) is called and it consumed part of the queue but needs to fill more bytes to return to the consumer with at least 1 element (e.g. 2 byte for Uint16Array) filled.

I planned to do nothing on invocation of controller methods. For example, after the source calls controller.enqueue() in pull (or outside pull), the source can probe how much more data is required to fill the queue by accessing controller.desiredSize. But I realized it's more complicated. After calling enqueue() in response to pullInto (either in pullInto or async to pullInto), it's possible there's more read(view)s are waiting.

It seems some more redesign should be made.

----

## Invocation timing and repetition

ReadableByteStream has a new pull/pullInto invocation scheme.

- pull is required when the current call stack contains any pull/pullInto (check by `_pulling`), schedule by setting `_pullAgain`
[...]

PR #430, opened 2016-02-18
Merge ReadableByteStream into ReadableStream

tyoshino:
This commit includes the following design changes:

- sync the spec text with the new reference implementation
- #423
  - add byobRequest getter to the controller
- port the asserts in the non-byte source version to the byte source version
- remove auto release feature from the byte source version
- rename Byob to BYOB
- move TransferArrayBuffer to helpers.js
- make the default highWaterMark of the byte source version to 0
- port the functionality that the start method can delay pulling by returning a pending Promise to the byte source version
- add `[[disturbed]]` feature to the byte source version
- port highWaterMark mechanism to ReadableByteStreamController
- merge ReadableStream and ReadableByteStream
  - now it behaves differently based on whether the given source has pullInto or not
  - bunch of renaming for merge
- merge ReadableStreamReader and ReadableByteStreamReader
- #424
  - make controller.close() and controller.enqueue() fail when the stream is not in the readable state
  - make controller.enqueue() throw a predefined TypeError, not the stored error
  - as a result test/pipe-to-options.js, readable-streams/general.js and test/readable-stream-templated.js have been updated
- rename ReadableStreamController to ReadableStreamDefaultController
- rename ReadableStreamReader to ReadableStreamDefaultReader
- rename ReadableByteStreamController to ReadableStreamBYOBController
- add automatic buffer allocation feature to BYOB controller
- read(view) now checks view.byteLength before setting [[disturbed]]

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -22,7 +22,13 @@ Logo: https://resources.whatwg.org/logo-streams.svg
 !Version History: <a href="https://twitter.com/streamsstandard">@streamsstandard</a>
 
 Opaque Elements: emu-alg
-Link Defaults: html (dfn) structured clone
+</pre>
+
+<pre class="anchors">
+urlPrefix: https://html.spec.whatwg.org/multipage/
+    type: dfn
+        urlPrefix: infrastructure.html
+            text: structured clone
 </pre>
 
 <style>
@@ -118,6 +124,11 @@ Consumers can also <dfn lt="tee a readable stream">tee</dfn> a readable stream. 
 <a lt="locked to a reader">lock</a> the stream, making it no longer directly usable; however, it will create two new
 streams, called <dfn lt="branches of a readable stream tee">branches</dfn>, which can be consumed independently.
 
+For streams representing bytes, an extended version of the <a>readable stream</a> is provided to handle bytes
+efficiently, in particular by minimizing copies. The <a>underlying source</a> for such a readable stream is called
+a <dfn>underlying byte source</dfn>. A readable stream whose underlying source is an underlying byte source is sometimes
+called a <dfn>readable byte stream</dfn>.
+
 <h3 id="ws-model">Writable Streams</h3>
 
 A <dfn>writable stream</dfn> represents a destination for data, into which you can write. In other words, data goes
@@ -202,11 +213,11 @@ similarly, avoiding writes that would cause the desired size to go negative.
 
 <!-- TODO: writable streams too, probably -->
 
-A <dfn>readable stream reader</dfn> (<dfn>readable byte stream reader</dfn> for <a>readable byte stream</a>) or simply
-reader is an object that allows direct reading of <a>chunks</a> from a <a>readable stream</a>). Without a reader, a
-<a>consumer</a> can only perform high-level operations on the readable stream: waiting for the stream to become closed
-or errored, <a lt="cancel a readable stream">canceling</a> the stream, or <a>piping</a> the readable stream to a
-writable stream. Many of those high-level operations actually use a reader themselves.
+A <dfn>readable stream reader</dfn>, or simply reader, is an object that allows direct reading of <a>chunks</a> from a
+<a>readable stream</a>. Without a reader, a <a>consumer</a> can only perform high-level operations on the readable
+stream: waiting for the stream to become closed or errored, <a lt="cancel a readable stream">canceling</a> the stream,
+or <a>piping</a> the readable stream to a writable stream. Many of those high-level operations actually use a reader
+themselves.
 
 A given readable stream only has at most one reader at a time. We say in this case the stream is
 <dfn lt="locked to a reader">locked to the reader</dfn>, and that the reader is <dfn lt="active reader">active</dfn>.
@@ -216,14 +227,9 @@ longer active. At this point another reader can be acquired at will. If the stre
 result of the behavior of its <a>underlying source</a> or via <a lt="cancel a readable stream">cancellation</a>, its
 reader (if one exists) will automatically release its lock.
 
-<h3 id="byte-streams">Readable Byte Streams</h3>
-
-For streams representing bytes, an extended version of the readable stream is provided to handle bytes efficiently.
-
-A <dfn>readable byte stream</dfn> represents a source of bytes, from which you can read.
-
-Although a readable byte stream can be created with arbitrary behavior, most readable byte streams wrap a lower-level
-I/O source, called the <dfn>underlying byte source</dfn>.
+A <a>readable byte stream</a> has the ability to vend two types of readers: <dfn>default readers</dfn> and <dfn>BYOB
+readers</dfn>. BYOB ("bring your own buffer") readers allow reading into a developer-supplied buffer, thus minimizing
+copies.
 
 <h2 id="rs">Readable Streams</h2>
 
@@ -287,6 +293,40 @@ I/O source, called the <dfn>underlying byte source</dfn>.
   on streams, beyond the provided ones of <a>piping</a> and <a lt="tee a readable stream">teeing</a>.
 </div>
 
+<div class="example">
+  The above example showed using the readable stream's <a>default reader</a>. If the stream is a <a>readable byte
+  stream</a>, you can also acquire a <a>BYOB reader</a> for it, which allows more precise control over buffer
+  allocation in order to avoid copies. For example, this code reads the first 1024 bytes from the stream into a single
+  memory buffer:
+
+  <pre><code class="lang-javascript">
+    const reader = readableStream.getBYOBReader();
+
+    let startingAB = new ArrayBuffer(1024);
+    readInto(startingAB, 0)
+      .then(buffer => console.log("The first 1024 bytes:", buffer))
+      .catch(e => console.error("Something went wrong!", e));
+
+    function readInto(buffer, offset) {
+      if (offset === buffer.byteLength) {
+        return buffer;
+      }
+
+      const view = new Uint8Array(buffer, offset, buffer.byteLength - offset);
+      reader.read(view).then(newView => {
+        return readInto(newView.buffer, offset + newView.byteLength);
+      });
+    }
+  </code></pre>
+
+  An important thing to note here is that the final <code>buffer</code> value is different from the
+  <code>startingAB</code>, but it (and all intermediate buffers) shares the same backing memory allocation. At each
+  step, the buffer is transferred <!-- TODO link --> to a new <code>ArrayBuffer</code> object. The <code>newView</code>
+  is a new <code>Uint8Array</code>, with that <code>ArrayBuffer</code> object as its <code>buffer</code> property, the
+  offset that bytes were written to as its <code>byteOffset</code> property, and the number of bytes that were written
+  as its <code>byteLength</code> property.
+</div>
+
 <h3 id="rs-class">Class <code>ReadableStream</code></h3>
 
 The <code>ReadableStream</code> class is a concrete instance of the general <a>readable stream</a> concept. It is
@@ -307,6 +347,7 @@ would look like
     get locked()
 
     cancel(reason)
+    getBYOBReader()
     getReader()
     pipeThrough({ writable, readable }, options)
     pipeTo(dest, { preventClose, preventAbort, preventCancel } = {})
@@ -326,42 +367,20 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
     </tr>
   </thead>
   <tr>
-    <td>\[[closeRequested]]
-    <td>A boolean flag indicating whether the stream has been closed by its <a>underlying source</a>, but still has
-      <a>chunks</a> in its internal queue that have not yet been read
-  </tr>
-  <tr>
-    <td>\[[controller]]
-    <td>A <a href="#rs-controller-class"><code>ReadableStreamController</code></a> created with the ability to control the
-      state and queue of this stream
+    <td>\[[readableStreamController]]
+    <td>A <a href="#rs-default-controller-class"><code>ReadableStreamDefaultController</code></a> or
+      <a href="#rs-byob-controller-class"><code>ReadableStreamBYOBController</code></a> created with the ability to
+      control the state and queue of this stream; also used for the IsReadableStream brand check
   </tr>
   <tr>
     <td>\[[disturbed]]
     <td>A boolean flag set to <b>true</b> when the stream has been read from or canceled
   </tr>
   <tr>
-    <td>\[[pullAgain]]
-    <td>A boolean flag set to <b>true</b> if the stream's mechanisms requested a call to the underlying source's
-      <code>pull</code> method to pull more data, but the pull could not yet be done since a previous call is still
-      executing
-  </tr>
-  <tr>
-    <td>\[[pulling]]
-    <td>A boolean flag set to <b>true</b> while the <a>underlying source</a>'s <code>pull</code> method is executing and has
-      not yet fulfilled, used to prevent reentrant calls
-  </tr>
-  <tr>
-    <td>\[[queue]]
-    <td>A List representing the stream's internal queue of <a>chunks</a>
-  </tr>
-  <tr>
     <td>\[[reader]]
-    <td>A <a href="#reader-class"><code>ReadableStreamReader</code></a> instance, if the stream is <a>locked to a
+    <td>A <a href="#default-reader-class"><code>ReadableStreamDefaultReader</code></a> or
+      <a href="#byob-reader-class"><code>ReadableStreamBYOBReader</code></a> instance, if the stream is <a>locked to a
       reader</a>, or <b>undefined</b> if it is not
-  </tr>
-  <tr>
-    <td>\[[started]]
-    <td>A boolean flag indicating whether the <a>underlying source</a> has finished starting
   </tr>
   <tr>
     <td>\[[state]]
@@ -372,21 +391,6 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
     <td>\[[storedError]]
     <td>A value indicating how the stream failed, to be given as a failure reason or exception when trying to operate
       on an errored stream
-  </tr>
-  <tr>
-    <td>\[[strategySize]]
-    <td>A function supplied to the constructor as part of the stream's <a>queuing strategy</a>, designed to calculate
-      the size of enqueued <a>chunks</a>; can be <b>undefined</b> for the default behavior.
-  </tr>
-  <tr>
-    <td>\[[strategyHWM]]
-    <td>A number supplied to the constructor as part of the stream's <a>queuing strategy</a>, indicating the point at
-      which the stream will apply <a>backpressure</a> to its <a>underlying source</a>.
-  </tr>
-  <tr>
-    <td>\[[underlyingSource]]
-    <td>An object representation of the stream's <a>underlying source</a>, including its <a>queuing strategy</a>; also
-      used for the <a href="#is-readable-stream">IsReadableStream</a> brand check
   </tr>
 </table>
 
@@ -410,9 +414,22 @@ Instances of <code>ReadableStream</code> are created with the internal slots des
   </ul>
 
   Both <code>start</code> and <code>pull</code> are given the ability to manipulate the stream's internal queue and
-  state via the passed <code>controller</code> object, which is an instance of
-  <a href="#rs-controller-class"><code>ReadableStreamController</code></a>. This is an example of the
+  state via the passed <code>controller</code> object. This is an example of the
   <a href="https://blog.domenic.me/the-revealing-constructor-pattern/">revealing constructor pattern</a>.
+
+  If the <var>underlyingSource</var> object contains a truthy property <code>byob</code>, this <a>readable stream</a> is
+  a <a>readable byte stream</a>, and can successfully vend <a>BYOB readers</a>. In that case, the passed
+  <code>controller</code> object will be an instance of <a
+  href="#rs-byob-controller-class"><code>ReadableStreamBYOBController</code></a>. Otherwise, it will be an instance of
+  <a href="#rs-default-controller-class"><code>ReadableStreamDefaultController</code></a>.
+
+  For <a>readable byte streams</a>, <var>underlyingSource</var> can also contain a property
+  <code>autoAllocateChunkSize</code>, which can be set to a positive integer to enable the auto-allocation feature for
+  this stream. In that case, when a <a>consumer</a> uses a <a>default reader</a>, the stream implementation will
+  automatically allocate an <code>ArrayBuffer</code> of the given size, and call the <a>underlying source</a> code
+  as if the <a>consumer</a> was using a <a>BYOB reader</a>. This can cut down on the amount of code needed when writing
+  the <a>underlying source</a> implementation, as can be seen by comparing [[#example-rbs-push]] without auto-allocation
+  to [[#example-rbs-pull]] with auto-allocation.
 
   The constructor also accepts a second argument containing the <a>queuing strategy</a> object with
   two properties: a non-negative number <code>highWaterMark</code>, and a function <code>size(chunk)</code>. The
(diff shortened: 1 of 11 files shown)


tyoshino:
Spec text update WIP. It shouldn't render by bikeshed yet.

domenic:
Yay, this is really exciting!

I added some commits of my own to this branch. I think it'll be better for us to work together than to just do review-ping-pong every day.

Here are some additional potential changes I'd like your opinion on:

- `_invalidate` has to go away as right now it's kind of specced to be a public property (although its call sites access it using `@[[` notation).
  - I moved `_cancel` and `_pull` to be "internal methods".
  - You could do a similar thing for `_invalidate`, but there's no need for polymorphic dispatch, so:
  - Either it should become an abstract operation, or...
  - We should consider moving as many abstract operations as possible to become internal methods.
     - This would be nice because it would remove the giant prefixes like "ReadableStreamDefaultController".
     - However the autolinking would become harder (so it would be broken for a while). In general you can't autolink to a method, but all of our methods except [[Cancel]] and [[Pull]] would be non-polymorphic, so we could in theory write some custom spec processors that makes them link.
     - Maybe it is best to just stick with abstract operations for the non-polymorphic case?
- ReadableStreamBYOBControllerRespondInReadableStat seems unused?

reference-implementation/test/pipe-to-options.js
```diff
@@ -3,21 +3,34 @@ const test = require('tape-catch');
 // Many other pipeTo-with-options tests have been templated.
 
 test('Piping with no options and a destination error', t => {
```
domenic: Why did these tests change? (We'll want to summarize any test changes in the commit message since they reflect potential observable behavior changes.)

domenic: I think I see. It's all because of #424.

reference-implementation/test/readable-stream-templated.js
```diff
@@ -54,13 +54,39 @@ templatedRSTwoChunksClosed('ReadableStream (two chunks enqueued, then closed)',
 );
 
 templatedRSTwoChunksClosed('ReadableStream (two chunks enqueued async, then closed)',
-  () => new ReadableStream({
-    start(c) {
-      setTimeout(() => c.enqueue(chunks[0]), 10);
-      setTimeout(() => c.enqueue(chunks[1]), 20);
-      setTimeout(() => c.close(), 30);
+  () => {
```
domenic: This is also #424?

reference-implementation/web-platform-tests/readable-streams/general.js
```diff
@@ -85,7 +85,7 @@ test(() => {
     start(controller) {
       assert_equals(this, source, 'source is this during start');
 
-      const methods = ['close', 'enqueue', 'error', 'constructor'];
+      const methods = ['_cancel', '_pull', 'close', 'enqueue', 'error', 'constructor'];
```
domenic: These shouldn't be here. I am not sure the best way to test them without breaking our reference implementation :(. Maybe use a `Symbol` for them in the implementation? That should work.

reference-implementation/web-platform-tests/readable-streams/general.js
```diff
@@ -608,23 +608,35 @@ promise_test(() => {
 
 }, 'ReadableStream pull should be able to close a stream.');
 
-test(() => {
+promise_test(t => {
```
domenic: Are all these changes also #424?

domenic:
I added some examples to the appendix. Please review them for correctness. Especially [the socket one](https://streams.spec.whatwg.org/branch-snapshots/rsrbsmergesquash/#example-rbs-push), as I am not sure that `socket.setTCPWindowSize(0)` is the correct way to signal backpressure, or if `socket.setTCPWindowSize` is a realistic API. (I tried to look through tcp(7) man pages but there was no straightforward thing saying "here is how you set the TCP window size", so maybe we should use a more complicated example API.) Is this at all how things might be implemented under the hood in Blink, @yutakahirano?

yutakahirano:
```
let bytesRead;
if (r) {
  bytesRead = socket.readInto(r.view.buffer, r.view.byteOffset, r.view.byteLength);
  r.respond();
} else {
  const buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
  bytesRead = socket.readInto(buffer, 0, DEFAULT_WINDOW_SIZE);
  controller.enqueue(buffer);
}
```

`bytesRead` is not used, is it correct?

domenic:
Ah, it is supposed to be the argument to respond(). So it doesn't need to be assigned in the second branch

yutakahirano:
In the second branch, you need to set the buffer's size to `readBytes`, I think.

yutakahirano:
Regarding backpressure:

Is it possible to *not read* when desiredSize <= 0? I'm not sure how precisely you need to manage the total buffer size, but when you stops reading from the socket the system will stop reading from the network and setTCPWindowSize(0) is not needed, IIUC.

domenic:
> Is it possible to not read when desiredSize <= 0? I'm not sure how precisely you need to manage the total buffer size, but when you stops reading from the socket the system will stop reading from the network and setTCPWindowSize(0) is not needed, IIUC.

The problem is that the kernel queue (aka kernel buffer) could overflow. The point of the enqueue() system is to move things into a userspace queue to avoid that.

yutakahirano:
> The problem is that the kernel queue (aka kernel buffer) could overflow.
So you want to minimize the kernel buffer usage when the consumer doesn't want the data, right?

domenic:
Yeah, that's the idea.

yutakahirano:
I'm sorry I don't have enough knowledge about how Chrome uses a TCP socket. [TCPSocketPosix::SetReceiveBufferSize](https://code.google.com/p/chromium/codesearch#chromium/src/net/socket/tcp_socket_posix.h&l=64) doesn't look to be used for such purpose, but I may be wrong. You will find far better reviewers in net-dev@chromium.org.

tyoshino:
> I added some commits of my own to this branch. I think it'll be better for us to work together than to just do review-ping-pong every day.

OK!

> - `_invalidate` has to go away as right now it's kind of specced to be a public property (although its call sites access it using `@[[` notation).

Fixed

>   - I moved `_cancel` and `_pull` to be "internal methods".

Sounds good. Thanks

>   - You could do a similar thing for `_invalidate`, but there's no need for polymorphic dispatch, so:
>   - Either it should become an abstract operation, or...
>   - We should consider moving as many abstract operations as possible to become internal methods.
>      - This would be nice because it would remove the giant prefixes like "ReadableStreamDefaultController".
>      - However the autolinking would become harder (so it would be broken for a while). In general you can't autolink to a method, but all of our methods except [[Cancel]] and [[Pull]] would be non-polymorphic, so we could in theory write some custom spec processors that makes them link.
>      - Maybe it is best to just stick with abstract operations for the non-polymorphic case?

Ah, I see. OK. I'll make it an abstract operation.

> - ReadableStreamBYOBControllerRespondInReadableStat seems unused?

Fixed

reference-implementation/test/pipe-to-options.js
tyoshino: Yes. I've added a line explaining why this change happened in the OP of the PR

reference-implementation/test/readable-stream-templated.js
tyoshino: Right. As this is expanded into various test cases, I chose to just make this stop enqueuing when cancel() is invoked.

reference-implementation/web-platform-tests/readable-streams/general.js
tyoshino: Done. I don't know if we can use a Symbol directly on a class syntax. Maybe not. So, the choices are to define the methods in the constructor or to place them out of the class block. I've chose the latter. WDYT?

reference-implementation/web-platform-tests/readable-streams/general.js
tyoshino: - IIRC, the following two test are just for increasing test coverage. these pass even on the current master branch
- enqueue() has been changed not to throw the stored error. So, I've removed the last test. This is related to #424.

tyoshino:
In Linux (I'm reading 3.13), the size of window to advertise is determined from sk_rcvbuf (using tcp_adv_win_scale). By default, 3/4 of free space of sk_rcvbuf is advertised. The free space is calculated by deducting sk_rmem_alloc from sk_rcvbuf.

sk_rmem_alloc represents the number of bytes received but not yet consumed by reading out to the user land. It increases when skb_set_owner_r() is called on delivery of data to TCP layer. skb_set_owner_r() also sets sock_rfree() to the sk_buff's destructor which decreases sk_rmem_alloc. tcp_read_sock() invokes this sock_rfree() by calling sk_eat_skb() on finished sk_buff. tcp_read_sock() is invoked on tcp_recvmsg().

When sending data or ack only packet, __tcp_select_window() is called to calculate the up-to-date window size to advertise in the packet. The calculation explained above happens in this function.

TCP_WINDOW_CLAMP socket option bounds window. But it seems only when window scaling is not turned on. See tcp_sock's window_clamp.

SO_RCVBUF socket option sets sk_rcvbuf. See SO_RCVBUF handling code in net/core/sock.c.

These values also cannot be less than SOCK_MIN_RCVBUF / 2. But it should be ok. It's small, and while read(2) doesn't happen the buffer gets filled to make sk_rmem_alloc hit sk_rcvbuf and let __tcp_select_window() decrease the advertised window.

When tcp_moderate_rcvbuf sysctl option is enabled, tcp_rcv_space_adjust() may automatically increase the space every time when read(2) consumes data from the kernel buffer. This is done based on the amount copied to user space in the last RTT and some strategy.

Setting SO_RCVBUF disables this automatic increase (see that SOCK_RCVBUF_LOCK is set to sk->userlocks) and reset the buffer size limit (sk_rcvbuf) to the given value.

So, IIUC, TCPSocketPosix::SetReceiveBufferSize() would help us prevent / recover from unexpected increase of amount of data being buffered in kernel.

tyoshino:
Sorry for spamming. Batching into one comment.

55eb9846851f592249cd7faeda50cf9501f9ccfd lgtm
0aa9a98031582bb9cc6699947329ca2462e9f6c8 lgtm committed a typo fix.
a465cf45e869dade9caccb0a8d581a45072d37d6 lgtm
8a8ba1845920aab60a6fcdf3788e729d437b2f48 lgtm
8a8ba1845920aab60a6fcdf3788e729d437b2f48 lgtm
beb60c19bcddbc05ee09817e99ea64063562600c lgtm

tyoshino:
Self-reply to https://github.com/whatwg/streams/pull/430#issuecomment-193877949

> So, IIUC, TCPSocketPosix::SetReceiveBufferSize() would help us prevent / recover from unexpected increase of amount of data being buffered in kernel.

SO_RCVBUF limits total memory consumption including overhead. So, this value should be set to a slightly larger value (e.g. like tcp_adv_win_scale is reflected to the advertised window size).

tyoshino:
Done reviewing

domenic:
Just noticed: IsReadableStreamBYOBRequest is not defined, and a brand check is missing in the view getter.

domenic:
Sorry for the spam. I will fix the things I commented on except the examples. (That is: adding IsReadableStreamBYOBRequest, adding the missing brand check, moving the BYOB request abstract ops, and the syntax in the .js files.)

domenic:
Another issue: there is a mixture of spec internal types and JS types in the ReadableStreamBYOBRequest constructor. It cannot take a Record as its second parameter.

I'd suggest passing the view directly and letting the caller construct it.

domenic:
Revised examples look really good. I have a tiny tweak I'll make to one for brevity but this system is looking nicer every time I see it :D

tyoshino:
> Another issue: there is a mixture of spec internal types and JS types in the ReadableStreamBYOBRequest constructor. It cannot take a Record as its second parameter.
> 
> I'd suggest passing the view directly and letting the caller construct it.

Ah, yeah! Sorry. Fixed: https://github.com/whatwg/streams/commit/a07ed09a7aeb0d39c8575fc3f081dfe7e279c727

domenic:
OK, I have an idea. Maybe we talked about it before. What if we add some kind of `defaultChunkSize` or `defaultReaderChunkSize` option, which would allow the elimination of the `if` branching we see in both examples. So instead of


```js
        if (controller.byobRequest) {
          const v = controller.byobRequest.view;

          return fs.read(fd, v.buffer, v.byteOffset, v.byteLength, position).then(bytesRead => {
            if (bytesRead === 0) {
              return fs.close(fd).then(() => controller.close());
            } else {
              position += bytesRead;
              controller.byobRequest.respond(bytesRead);
            }
          });
        } else {
          // The consumer is using the default reader.
          const buffer = new ArrayBuffer(DEFAULT_CHUNK_SIZE);
          return fs.read(fd, buffer, 0, DEFAULT_CHUNK_SIZE, position).then(bytesRead => {
            if (bytesRead === 0) {
              return fs.close(fd).then(() => controller.close());
            } else {
              position += bytesRead;
              controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
            }
          });
        }
```

we have the `byobRequest` always present for streams with `byob: true` and `defaultReaderChunkSize: DEFAULT_CHUNK_SIZE`.

The name doesn't need to be super-short since this is an advanced option used by stream creators. It could even be something like `defaultReaderAutoAllocateChunkSize`.

WDYT?

tyoshino:
Just found that I'm accessing [[ViewedArrayBuffer]] by dot notation. This should be wrong. GetViewedArrayBuffer() is ready and I should use it?

domenic:
Oh! Yes. Same for GetByteLength and GetByteOffset.

I will work on upstreaming those into ES tonight.

domenic:
However, if you are in a scenario where you know that the object is a typed array object (i.e. is not user-supplied) you could just access them with @-notation directly. Maybe that is better, not sure.

tyoshino:
> What if we add some kind of defaultChunkSize or defaultReaderChunkSize option, which would allow the elimination of the if branching we see in both examples.

Yes, I remember that we were discussing automatic chunk allocation before in some context.

That should definitely reduce coding for an author of this kind of an underlying byte source (kicks the fs/socket behind it only when a read()/read(view) happens). I'll try.

domenic:
I'll handle the [[ViewedArrayBuffer]] cleanup. There are a few different things I'd like to touch up in that area.

domenic:
Oh, it looks like "Readable Stream BYOB Controller Abstract Operations" is not alphabetized. I will leave that to you in case it interferes with the auto allocation work. I hope https://github.com/whatwg/streams/commit/c5d63deed96e3c197988aa3180626604535ce555 did not interfere much already.

tyoshino:
I'd like to discuss reorganization of the parameters of the ReadableStream constructor.

Now the `size` argument is specific to underlying (non-byte) sources. On the other hand, the `byob` parameter is common to both source types. Not a parameter configuring the sources, but a parameter to tell ReadableStream what kind of source we're passing.

I feel it's strange that `byob` is in `underlyingSource` while `size` is separated from it.

- It's ok that `byob` is in `underlyingSource`, but then `size` should also be included in `underlyingSource`.
- It's also ok that `size` being separated, but then `byob` should also be separated.

domenic:
Yeah, I get that. Maybe we should just merge everything into one blob, but I liked the idea of reusable queuing strategies that could be constructed independently. E.g. you could imagine client code writing

```js
const qs = new CountQueuingStrategy(12);
someLibrary.getMeAStream(qs);
```

and then it just forwards the object along to the stream implementation.

That is, the main separation is that underlyingSource contains properties relevant to the stream creator, whereas queuingStrategy is something the consumer might be interested in setting.

WDYT?

tyoshino:
The automatic buffer allocation feature discussed at https://github.com/whatwg/streams/pull/430#issuecomment-194606911 has been added. See the newly added test.

Before making the change for the feature, I made some refactoring changes. The changes are separated into small commits, so should be easy to review.

tyoshino:
> Oh, it looks like "Readable Stream BYOB Controller Abstract Operations" is not alphabetized.

Fixed all by https://github.com/whatwg/streams/commit/89e2d31ce17a00d5d27aa4cda2bd09a79c1a0734

tyoshino:
> That is, the main separation is that underlyingSource contains properties relevant to the stream creator, whereas queuingStrategy is something the consumer might be interested in setting.

Ah, OK. That makes sense!

tyoshino:
Updated the example to use the automatic buffer allocation feature.

index.bs
```diff
@@ -1729,6 +1739,11 @@ table:
   1. Set *this*@[[totalQueuedBytes]] to *0*.
   1. Set *this*@[[started]], and *this*@[[closeRequested]] to *false*.
   1. Set *this*@[[strategyHWM]] to ValidateAndNormalizeHighWaterMark(_highWaterMark_).
+  1. Let _autoAllocateChunkSize_ be GetV(_underlyingByteSource_, `"autoAllocateChunkSize"`).
+  1. If _autoAllocateChunkSize_ is not *undefined*,
+    1. If Number.isInteger(_autoAllocateChunkSize_) is *true* or _autoAllocateChunkSize_ &lt; 0,
```
domenic: This doesn't look right. (And we shouldn't reference public overridable API functions like Number.isInteger.)

I think instead what we want to do is:

1. Set _autoAllocateChunkSize_ to ToInteger(_autoAllocateChunkSize_).
1. ReturnIfAbrupt(_autoAllocateChunkSize_).
1. If _autoAllocateChunkSize_ ≤ 0, or if _autoAllocateChunkSize_ is +∞ or -∞, throw a **RangeError** exception.

(NaN is automatically converted to 0 so will still throw.)

index.bs
```diff
@@ -1838,13 +1853,21 @@ controllers.
...
+    1. If *this*[[totalQueuedBytes]] &gt; 0,
+      1. Let _entry_ be the first element of *this*@[[queue]].
+      1. Remove _entry_ from *this*@[[queue]], shifting all other elements downward (so that the second becomes the first, and so on).
+      1. Subtract _entry_.[[byteLength]] from *this*@[[totalQueuedBytes]].
+      1. Perform ReadableStreamBYOBControllerHandleQueueDrain(*this*).
+      1. Let _view_ be Construct(%Uint8Array%, «_entry_.[[buffer]], _entry_.[[byteOffset]], _entry_.[[byteLength]]»).
+      1. Return a promise resolved with CreateIterResultObject(_view_, *false*).
+    1. Let _autoAllocateChunkSize_ be *this*@[[autoAllocateChunkSize]].
+    1. If _autoAllocateChunkSize_ is not *undefined*,
+      1. Let _buffer_ be Construct(`ArrayBuffer`, «_autoAllocateChunkSize_»).
```
domenic: %ArrayBuffer%

index.bs
```diff
@@ -1838,13 +1853,21 @@ controllers.
...
+      1. Let _pullIntoDescriptor_ be Record{[[buffer]]: _buffer_, [[byteOffset]]: 0, [[byteLength]]: _autoAllocateChunkSize_, [[bytesFilled]]: 0, [[elementSize]]: 1, [[ctor]]: `Uint8Array`, [[readerType]]: `"default"`}.
```
domenic: %Uint8Array%

index.bs
```diff
@@ -3283,34 +3283,28 @@ size of 1024, it attempts to fill the developer-supplied buffer, allowing full c
...
-          return fs.read(fd, buffer, 0, DEFAULT_CHUNK_SIZE, position).then(bytesRead => {
-            if (bytesRead === 0) {
-              return fs.close(fd).then(() => controller.close());
-            } else {
-              position += bytesRead;
-              controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
-            }
-          });
-        }
+        // Even when the consumer is using the default reader, the auto allocate feature allocate a buffer and pass it
```
domenic: s/allocate/allocates

Also, we don't have a good guideline here, but maybe wrap so that the comment gets split more evenly over two lines.

index.bs
```diff
@@ -3283,34 +3283,28 @@ size of 1024, it attempts to fill the developer-supplied buffer, allowing full c
...
+        // to us via byobRequest.
+
+        assert(controller.byobRequest !== undefined);
```
domenic: Let's remove this line; the comment is sufficient, and examples are easier to read if they're as brief as possible.

index.bs
```diff
@@ -3283,34 +3283,28 @@ size of 1024, it attempts to fill the developer-supplied buffer, allowing full c
...
+        });
       },
 
       cancel() {
         return fs.close(fd);
-      }
+      },
+
+      autoAllocateChunkSize: DEFAULT_CHUNK_SIZE
     });
```
domenic: We should be able to do this for the makeReadableBackpressureByteSocketStream example too, right? Although hmm maybe we want to not do so to show how things work.

domenic:
Sorry for the delay in getting back to you on this. I made some more tweaks to auto-allocate based on my review comments.

I think we should work on merging this soon. Do you think it's ready? Would you mind drafting up a commit message? The main things I want to be sure are included are:

- How this differs from the previous ReadableByteStream separate class model
- Details on the normative changes to existing ReadableStream behavior, including reference to why the tests were changed
- A vague idea of how the under-the-hood changes work (i.e., moving a lot of things to the controllers)

After we get this merged I would like to do an editorial pass on my own through the document. Largely it will consist of adding more cross-links (both within-spec and to other specs) and also updating to use the latest ES conventions (which allow us to use ? and ! prefixes and get rid of ReturnIfAbrupt).

index.bs
tyoshino: Got it. Thanks for fixing!

index.bs
tyoshino: Yeah, I'm just following the 120 col rule for all. I can live and like shorter wrapping. How much limit shall we use?

index.bs
tyoshino: OK

index.bs
tyoshino: We can convert default-reader requests into BYOB, but as onreadable() may be still invoked while there's no outstanding read request, we need the code for accepting received data.

We can implement another kind of auto-allocation feature that is to allocate a buffer when HWM is non-zero. Then, while we're pulling from the network by setting the window to non-zero, we also have buffers ready to accept them to drain data from the kernel.

We can also choose not to drain data from the kernel while there's no outstanding read requests. But it doesn't match the original concept of the example?

tyoshino:
Thanks for fixing. The commit basically looks good. See the replies to comments.

tyoshino:
Here's a draft of the commit log:

----

Merge ReadableByteStream into ReadableStream

Background

We originally designed the ReadableByteStream which includes extended features to handle bytes as a separate class from the ReadableStream which handles a stream of general objects.

While designing the details of ReadableByteStream (see #361), we noticed that we could simplify the ReadableStream and ReadableByteStream by moving variables and logic for handling queuing into their controller class (see #379). This turned out to be also clarifying which part of the logic is representing semantic requirement of readable streams and which part of that is implementing helper logic for easing development of underlying sources.

After the above refactoring, we also noticed that ReadableStream and ReadableByteStream share most of their code. So, we attempted to merge the ReadableByteStream into the ReadableStream.

Change summary

The new ReadableStream has two reader getter methods when BYOB reading is available. Availability of BYOB reading is determined by whether or not the underlying source passed to the ReadableStream had BYOB pulling functionality. This is indicated by the "byob" parameter of the ReadableStream constructor.

The two readers are named the default reader and BYOB reader. The default reader's read() method takes no argument. The resulting promise of the read() method will be fulfilled by a generated chunk. The BYOB reader's read() method takes one argument. It must be an ArrayBufferView. The ReadableStream will fill the passed ArrayBufferView and fulfills the returned promise with it. The ArrayBufferView might be transferred several times.

When the byob parameter is false, the underlying source is given a ReadableStreamDefaultController on start() method call. This class provides methods to enqueue a new chunk and know the status of the queue. Its queuing strategy is configured by the parameters passed to the ReadableStream's constructor. The underlying source can subscribe to events on which chunks are drained from the queue by implementing the "pull" method on it. Only the getReader() method will be available on the ReadableStream.

When the byob parameter is false, the underlying source is given a ReadableStreamBYOBController on start() method call. In addition to the functionalities of the ReadableStreamDefaultController, this controller provides a getter named byobRequest() which exposes the oldest outstanding BYOB reading request into which the underlying source can put bytes directly (see #423). Both the getReader() and the getBYOBReader() method will be available on the ReadableStream.

The ReadableStreamBYOBController can be configured to convert read requests from a default reader into BYOB read request by allocating a buffer and exposing it at the byobRequest getter automatically. This ease implementation of a reactive underlying source.

Changes included in this commit

In addition to the major changes as described above, this commit includes bunch of design / logic / aesthetic changes as follows:

- remove auto release feature from the ReadableByteStream
- rename Byob to BYOB
- move TransferArrayBuffer to helpers.js
- make the default highWaterMark of the byte source version to 0
- port the functionality that the start method can delay pulling by returning a pending Promise to the ReadableStreamBYOBController
- implement `[[disturbed]]` feature to the byte handling logic
- port highWaterMark mechanism to ReadableByteStreamController
- changes on semantics of the controller methods (see #424)
  - make controller.close() and controller.enqueue() fail when the stream is not in the readable state
  - make controller.enqueue() throw a predefined TypeError, not the stored error
  - as a result test/pipe-to-options.js, readable-streams/general.js and test/readable-stream-templated.js have been updated
- rename ReadableStreamController to ReadableStreamDefaultController
- rename ReadableStreamReader to ReadableStreamDefaultReader
- rename ReadableByteStreamController to ReadableStreamBYOBController
- read(view) now checks view.byteLength before setting [[disturbed]]

tyoshino:
Feel free to edit, Domenic.

index.bs
domenic: Oh, I see, that makes sense. I forgot that we were inside onreadable, not inside pull(). I don't think we need another auto-allocation case for this.

> We can also choose not to drain data from the kernel while there's no outstanding read requests. But it doesn't match the original concept of the example?

Right, I am assuming that the kernel buffer might overflow if we do that.

domenic:
OK! I have rebased and squashed and updated the commit message with my tweaks. I'll let you do the honors of any final review, and of merging!

index.bs
domenic: I changed the comment from

```js
          // Note: you can consolidate some of this code using the auto-allocation feature.
          // See "A readable byte stream with an underlying pull source".
          // Here we manually do allocation for example purposes.
```

to

```js
          // Since onreadable can happen even when there's no pending BYOB
          // requests, we need to handle both cases.
```

tyoshino: Looks good. Thanks

tyoshino:
Closed by https://github.com/whatwg/streams/commit/e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393

Thanks!!

PR #431, opened 2016-02-26
Fix constness bug

gwicke:
Make `time` and `pauses` a `let` rather than `const`, as they are modified.

Discovered while compiling with babeljs.


domenic:
Thanks! Merged as 7eee1256484d61343255426b268846790bda4e73.

PR #432, opened 2016-02-27
Es5 distribution

gwicke:
With ServiceWorkers growing in popularity, we would like to start prototyping a server-side execution environment for ServiceWorker code. This includes `ReadableStream`, which is now available in Chrome Canary. The reference implementation would fill this need. However, there does not seem to be any current version in the npm registry.

Therefore, I am proposing to publish a transpiled ES5 version of this reference implementation.

This PR sets up babeljs in a prepublish hook. It is little more than a POC, but is confirmed working by publishing (and unpublishing) to npmjs.org.


domenic:
I appreciate the interest. However, the reference implementation is not intended to be used by people other than the spec editors. It is intentionally slow (in order to ape the spec more closely) and uses techniques that real-world programs should never consider. And, maintaining a distribution for old browsers is not one of its goals. It is explicitly designed only for spec maintenance and test creation.

If you are interested in maintaining a polyfill, I'd suggest you do that in a separate repository.

PR #436, opened 2016-04-06
Merge getBYOBReader() into getReader() and rename class names and switches

### LINKED ISSUE #294: What should we call ReadableByteStream.prototype.getBYOBReader()?

tyoshino:
Fixes #294

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -375,7 +375,7 @@ Instances of {{ReadableStream}} are created with the internal slots described in
   </thead>
   <tr>
     <td>\[[readableStreamController]]
-    <td>A {{ReadableStreamDefaultController}} or {{ReadableStreamBYOBController}} created with the ability to control
+    <td>A {{ReadableStreamDefaultController}} or {{ReadableByteStreamController}} created with the ability to control
     the state and queue of this stream; also used for the IsReadableStream brand check
   </tr>
   <tr>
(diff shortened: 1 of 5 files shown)


domenic:
Thanks! I force-pushed some fixups and will merge. I should have had them as separate commits so you can review; sorry :(. The main changes were:

- Revert `rs-byob-request` -> `rbs-request` ID changes
- Add a few tests (under web-platform tests)
- Tidy up non-normative notes around readers
- Fix up how type is stringified and such in the constructor

tyoshino:
Thanks

> Revert rs-byob-request -> rbs-request ID changes

Nice catch!

PR #439, opened 2016-04-18
Add Contributing.md with various editing guidelines

### LINKED ISSUE #421: [Guideline] Style
I'd like to have a memorandum about the rules we follow when writing this spec.

Here's a draft.

## Spec text

- Put an <a> tag only on the first occurrence of the term in a paragraph
  - https://github.com/whatwg/streams/pull/417#discussion_r48513698
- Use "set" for updating internal slots
- Use "let" for defining local variables
- Add `nothrow` or `throw` attribute to the headings tag of abstract operations to annotate them about exception throwing

## Reference implementation

- Wrap long lines to fit 120 col
- Alphabetize imports
  - https://github.com/whatwg/streams/pull/277#discussion_r24117515

### LINKED ISSUE #437: [Guideline] Where to factor out logic
I'd like to write down a clear well-discussed guidance for at which point we factor out logic and for what purpose.

- make it clear which point is defining semantics and which point is left up to implementors of the API
  - Example: `ReadableStreamAddReadIntoRequest()` (ones listed in the block with the comment "// ReadableStream API exposed for controllers")
- allow optimized invocation with the following logic skipped
  - skip brand checking
    - Example: the `if (IsReadableStream(this) === false)` block in `ReadableStream.getReader()`
  - argument validation
    - Example: `read(view)` with non typed array `view`
  - argument parsing
    - Example: calling `AcquireReadableStreamBYOBReader()` directly than calling `getReader()` (only when the user is sure that it's byte stream)
  - state checking
    - Example: calling `ReadableStreamDefaultReaderRead()` directly than calling `read()`
    - Example (no case): `ReadableStreamDefaultReaderRead()` checks `stream._state` in itself as it's less common that the user is sure in which state the stream is.
    - Example: calling `ReadableStreamDefaultControllerClose()` directly than calling `close()`. The user can know unsolicited closure by listening to cancel() call on the underlying source.
  - result packing
    - Example: calling `ReadableStreamTee()` directly than calling `tee()`
- readability and code simplification

domenic:
Closes #437; closes #421.

@tyoshino can you think of other things we should include?

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,0 +1,66 @@
+# Contribution guidelines for the Streams Standard
+
+## Spec editorial conventions
+
+Wrap lines to 120 columns. (This is not yet consistently followed.)
+
+Use two-space indents.
+
+Do not introduce explicit `<p>` tags. (Bikeshed does this automatically.)
+
+Use "let" for introducing new local variables. Use "set" for updating existing variables or internal slots.
+
+Mark up definitions [appropriately](https://github.com/tabatkins/bikeshed/blob/master/docs/definitions-autolinks.md#definitions). This is mostly applicable when defining new classes or methods; follow the existing examples in the spec.
+
+Mark up abstract operations with `throw` or `nothrow` attributes in their heading tags, according to whether or not they can ever return an abrupt completion.
+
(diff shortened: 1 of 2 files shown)


Contributing.md
```diff
@@ -0,0 +1,66 @@
+# Contribution guidelines for the Streams Standard
+
+## Spec editorial conventions
+
+Wrap lines to 120 columns. (This is not yet consistently followed.)
+
+Use two-spec indents.
```
tyoshino: two space?

tyoshino:
lgtm!

PR #444, opened 2016-04-27
Mostly-editorial tweaks to align with newer ES algorithm styles

domenic:
This explicitly annotates all abstract operation and method calls with ! or ? as appropriate, per the conventions introduced in semi-recent ES spec drafts for dealing with abrupt completions. That subsumes several of our previous idiosyncratic conventions, such as a special meaning for "Perform" or the "Call-with-rethrow" shorthand.

While here, a few other style changes were made, such as spacing in list and record literals, or capping all lines at 120 characters.

The following notable non-editorial changes were made:

- Renamed ReadableStreamHasReader to ReadableStreamHasDefaultReader
- Handled abrupt completions while allocating a new ArrayBuffer for the auto-allocation feature (due to out of memory errors)
- Moved the "this abstract operation can be called by other specifications" from ReadableStreamClose and ReadableStreamError to their counterparts for the ReadableStreamDefaultController, since now those two abstract operations are instead used as part of the internal controller interface
- Formalized tranfers and other byte manipulation via the new SameRealmTransfer abstract operation and the existing CopyDataBlockBytes and CloneArrayBuffer abstract operations
- Made some cleanups to ReadableByteStreamControllerPullInto
- Changed ReadableStreamBYOBRequestInvalidate to ReadableByteStreamControllerInvalidateBYOBRequest, which contains all the logic which previously was redundantly included at every call site

---

@tyoshino, it's OK to rubber-stamp this since it's almost all boring editorial changes, but I wanted to give you a chance to review in case you wanted to.


domenic:
Build failure is https://github.com/whatwg/streams/issues/445

domenic:
I added a couple new commits that update the reference implementation for Node v6. Again, not much interesting to review, but wanted to give you a chance.

reference-implementation/run-web-platform-tests.js
```diff
@@ -19,6 +19,9 @@ wptRunner(testsPath, setup)
   });
 
 function setup(window) {
+  // Necessary so that we can send test-realm promises to the jsdom-realm implementation without causing assimilation.
+  window.Promise = Promise;
```
tyoshino: I lack knowledge to understand this. Should I just learn what wpt-runner is doing more?

tyoshino:
Took a look. reference impl looks good.

Can I see how the exclamations newly introduced into index.bs would be rendered?

reference-implementation/run-web-platform-tests.js
domenic: Mehh. wpt-runner is kind of a giant hack, so it might not be worth your time.

It basically creates a whole nother global environment in which to run its scripts, including new values for Promise, Map, Set, Number, etc. This usually isn't a problem, but for promises it causes problems now that V8 has updated their promise-detection code.

domenic:
Sure! https://streams.spec.whatwg.org/branch-snapshots/update-editorial/

reference-implementation/run-web-platform-tests.js
tyoshino: OK. Understood briefly :) Thanks

tyoshino:
Thanks!

Reviewed from ReadableByteStreamControllerRespond to the bottom. Looks good with some minor issues.

- PromiseInvokeOrNoop in PromiseInvokeOrFallbackOrNoop should be annotated with !, not ?.
- ReadableByteStreamControllerRespondInternal in ReadableByteStreamControllerRespondWithNewView is annotated with ? because of possible abrupt in buffer cloning? Then, return is better than perform? I'm fine with the current combination but wanna make sure that it's intentional.
- The same about ReadableByteStreamControllerRespondInReadableState call in ReadableByteStreamControllerRespondInternal
- CloneArrayBuffer() is taking 4 arguments while even https://tc39.github.io/ecma262/2016/#sec-clonearraybuffer has up to 3 args. What should I look at?

Feel free to land it.

domenic:
> PromiseInvokeOrNoop in PromiseInvokeOrFallbackOrNoop should be annotated with !, not ?.

Good catch, thanks!

> ReadableByteStreamControllerRespondInternal in ReadableByteStreamControllerRespondWithNewView is annotated with ? because of possible abrupt in buffer cloning?

Yes, exactly. Anything that allocates a buffer (not just transfers it) can throw.

> Then, return is better than perform? I'm fine with the current combination but wanna make sure that it's intentional.

Good catch; it was inconsistent. I switched them both to "perform." They are basically equivalent in this case, but I think "perform" is clearer since there is no interesting return value to propagate.

> CloneArrayBuffer() is taking 4 arguments while even https://tc39.github.io/ecma262/2016/#sec-clonearraybuffer has up to 3 args. What should I look at?

https://tc39.github.io/ecma262/#sec-clonearraybuffer is correct. I am trying to fix the cross-linking next to point to the correct latest ES spec. I am currently blocked on https://github.com/bterlson/ecmarkup/issues/91.

PR #456, opened 2016-05-26
Update wpt-runner dependency to require 1.0.1 or later

tyoshino:
The current code doesn't work with wpt-runner 1.0.0.


domenic:
I'll revise this to require the latest version, 1.0.2.

PR #462, opened 2016-05-31
Align WritableStream structure with ReadableStream structure

tyoshino:
- Add WritableStreamDefaultWriter
- Add WritableStreamDefaultController
- Add lock related methods
- Update reader.pipeTo() to use the new WritableStream classes

### DIFF
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -578,6 +578,7 @@ class WritableStreamDefaultController {
     const startResult = InvokeOrNoop(underlyingSink, 'start', [this]);
     Promise.resolve(startResult).then(
       () => {
+        throw new assert.AssertionError('ss');
         controller._started = true;
         WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
       },
@@ -585,7 +586,7 @@ class WritableStreamDefaultController {
         WritableStreamDefaultControllerErrorIfNeeded(controller, r);
       }
     )
-    .catch(e => rethrowAssertionErrorRejection(e));
+    .catch(rethrowAssertionErrorRejection);
   }
 
   error(e) {
@@ -726,7 +727,7 @@ function WritableStreamDefaultControllerProcessClose(controller) {
       WritableStreamDefaultControllerErrorIfNeeded(controller, r);
     }
   )
-  .catch(e => rethrowAssertionErrorRejection(e));
+  .catch(rethrowAssertionErrorRejection);
 }
 
 function WritableStreamDefaultControllerProcessWrite(controller, chunk) {
@@ -756,7 +757,7 @@ function WritableStreamDefaultControllerProcessWrite(controller, chunk) {
       WritableStreamDefaultControllerErrorIfNeeded(controller, r);
     }
   )
-  .catch(e => rethrowAssertionErrorRejection(e));
+  .catch(rethrowAssertionErrorRejection);
 }
 
 function WritableStreamDefaultControllerUpdateBackpressure(controller) {
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -29,7 +29,7 @@ class WritableStream {
 
   get locked() {
     if (IsWritableStream(this) === false) {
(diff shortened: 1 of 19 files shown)


tyoshino:
This is WIP. I'll create some issues to discuss design decisions.

tyoshino:
- Synchronous (or asynchronous) state probing on WritableStream https://github.com/whatwg/streams/issues/463

tyoshino:
- [WritableStream] Shouldn't the state after writer.abort() be closed? https://github.com/whatwg/streams/issues/464

tyoshino:
- Synchronous queue draining for WritableStream https://github.com/whatwg/streams/issues/465

tyoshino:
I also started updating TransformStream. It should reveal issues of WritableStream if any.

reference-implementation/lib/writable-stream.js
```diff
@@ -7,222 +7,462 @@ const { DequeueValue, EnqueueValueWithSize, GetTotalQueueSize, PeekQueueValue }
...
+    return IsWritableStreamLocked(this);
+  }
+
+  getWriter() {
+    if (IsWritableStream(this) === false) {
+      throw new TypeError('WritableStream.prototype.getWriter can only be used on a WritableStream');
+    }
+
+    return AcquireWritableStreamDefaultWriter(this);
+  }
```
domenic: Although I know there is a thread on abort(), for now (not having looked at that thread in detail yet) I think we should keep abort() as a lock-abort-unlock, similar to how cancel() on readable streams is lock-cancel-unlock.

domenic: Oh, I see, you did that in the follow-up commit; nevermind!

reference-implementation/lib/writable-stream.js
```diff
@@ -35,6 +35,18 @@ class WritableStream {
     return IsWritableStreamLocked(this);
   }
 
+  abort(reason) {
+    if (IsWritableStream(this) === false) {
+      throw Promise.reject(new TypeError('WritableStream.prototype.abort can only be used on a WritableStream'));
+    }
+
+    if (IsWritableStreamLocked(this) === true) {
+      throw Promise.reject(new TypeError('Cannot abort a stream that already has a reader'));
```
domenic: return, not throw. Need tests to catch this.

domenic:
Wow, thank you so much for taking this on. This is so great.

I wonder what strategy you have in mind for doing this work. For example, IMO it would be valuable to fix writable streams in the ways we *know* we want to change, and get the spec + tests for them done. All of the stuff in the first two commits seems great to me. Then we could work on transform streams and other less-obvious fixes afterward. What do you think?

Another thing we need to do is convert the existing tests to web platform tests :(. Don't necessarily worry about that for now, but maybe if writing new tests try to keep them in WPT format and directories.

I'll go off to comment on the other issues now!

reference-implementation/lib/writable-stream.js
tyoshino: Thanks. Fixed and added a test.

reference-implementation/lib/writable-stream.js
tyoshino: Yeah!

tyoshino:
> it would be valuable to fix writable streams in the ways we know we want to change,

Yeah. There some experimental features are implemented in this PR, but they'll be reverted and this PR would just include only what we agreed e.g. writer introduction.

tyoshino:
All tests pass except for skipped ones which are related to the locking semantics. I'll fix them and then tweak it / fix style issues.

reference-implementation/lib/writable-stream.js
```diff
@@ -578,14 +578,15 @@ class WritableStreamDefaultController {
     const startResult = InvokeOrNoop(underlyingSink, 'start', [this]);
     Promise.resolve(startResult).then(
       () => {
+        throw new assert.AssertionError('ss');
```
LinusU: Did you mean to commit this line?

tyoshino: Thanks. Removed

tyoshino:
Moved to https://github.com/whatwg/streams/pull/488 to keep all the logs while making the PR rebased and squashed for easier review.

PR #473, opened 2016-06-28
Tweak ReadableByteStreamControllerEnqueue algorithm structure

domenic:
This makes the parallel nature of the algorithm more clear, in that it handles all three cases of default reader, BYOB reader, and no reader. See http://logs.glob.uno/?c=freenode%23whatwg#c999292.

/cc @wanderview


tyoshino:
Merged as 174b42aa30add51f08133443dd0583bc9473cda6

PR #488, opened 2016-08-01
Align WritableStream structure with ReadableStream structure

tyoshino:
- [x] Add WritableStreamDefaultWriter
- [x] Add WritableStreamDefaultController
- [x] Add lock related methods
- [x] Update reader.pipeTo() to use the new WritableStream classes
- [x] Refine reader.pipeTo()'s error handling
- [x] Update TransformStream to use the new WritableStream classes
- [x] Refine TransformStream's error handling
- [x] Add logging to pipeTo() for easier debugging
- [ ] Document the difference between ReadableStream's locking (releaseLock()) and WritableStream's one
  - WritableStream's reader's releaseLock() doesn't check pending writes
- [ ] Add an example for writer.desiredSize

### DIFF
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -342,11 +342,13 @@ class WritableStreamDefaultWriter {
       return Promise.reject(defaultWriterBrandCheckException('write'));
     }
 
-    if (this._ownerWritableStream === undefined) {
+    const stream = this._ownerWritableStream;
+
+    if (stream === undefined) {
       return Promise.reject(defaultWriterLockException('write to'));
     }
 
-    if (this._ownerWritableStream._state === 'closing') {
+    if (stream._state === 'closing') {
       return Promise.reject(new TypeError('Cannot write to an already-closed stream'));
     }
 
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -4,22 +4,565 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
         typeIsObject } = require('./helpers.js');
 const { rethrowAssertionErrorRejection } = require('./utils.js');
 const { DequeueValue, EnqueueValueWithSize, GetTotalQueueSize, PeekQueueValue } = require('./queue-with-sizes.js');
-const CountQueuingStrategy = require('./count-queuing-strategy.js');
 
 class WritableStream {
-  constructor(underlyingSink = {}, { size, highWaterMark = 0 } = {}) {
+  constructor(underlyingSink = {}, { size, highWaterMark } = {}) {
+    // Temporary value. Never used. To be overwritten by the initializer code of the controller.
+    this._state = 'writable';
+    this._storedError = undefined;
+
+    this._writer = undefined;
+
+    this._writeRequests = [];
+
+    // Initialize to undefined first because the constructor of the controller checks this
+    // variable to validate the caller.
+    this._writableStreamController = undefined;
+
+    const type = underlyingSink.type;
+
+    if (type !== undefined) {
+      throw new RangeError('Invalid type is specified');
+    }
+
+    if (highWaterMark === undefined) {
+      highWaterMark = 1;
+    }
+
+    this._writableStreamController = new WritableStreamDefaultController(this, underlyingSink, size, highWaterMark);
+  }
+
+  get locked() {
+    if (IsWritableStream(this) === false) {
+      throw CreateWritableStreamBrandCheckException('locked');
+    }
+
+    return IsWritableStreamLocked(this);
+  }
+
+  abort(reason) {
+    if (IsWritableStream(this) === false) {
+      return Promise.reject(CreateWritableStreamBrandCheckException('abort'));
+    }
+
+    if (IsWritableStreamLocked(this) === true) {
+      return Promise.reject(new TypeError('Cannot abort a stream that already has a reader'));
+    }
+
+    WritableStreamAbort(this, reason);
+  }
+
+  getWriter() {
+    if (IsWritableStream(this) === false) {
+      throw CreateWritableStreamBrandCheckException('getWriter');
+    }
+
+    return AcquireWritableStreamDefaultWriter(this);
+  }
+}
+
+exports.WritableStream = WritableStream;
+
+// Helper functions for the WritableStream.
+
+function CreateWritableStreamBrandCheckException(name) {
+  return new TypeError('WritableStream.prototype.' + name + ' can only be used on a WritableStream')
+}
+
+// Abstract operations for the WritableStream.
+
+function AcquireWritableStreamDefaultWriter(stream) {
+  return new WritableStreamDefaultWriter(stream);
+}
+
+function IsWritableStream(x) {
+  if (!typeIsObject(x)) {
+    return false;
+  }
+
+  if (!Object.prototype.hasOwnProperty.call(x, '_writableStreamController')) {
+    return false;
+  }
+
+  return true;
+}
+
+function IsWritableStreamLocked(stream) {
+  assert(IsWritableStream(stream) === true, 'IsWritableStreamLocked should only be used on known writable streams');
+
+  if (stream._writer === undefined) {
+    return false;
+  }
+
+  return true;
+}
+
+function WritableStreamAbort(stream, reason) {
+  const state = stream._state;
+  if (state === 'closed') {
+    return Promise.resolve(undefined);
+  }
+  if (state === 'errored') {
+    return Promise.reject(stream._storedError);
+  }
+
+  assert(state === 'writable' || state === 'waiting' || state === 'closing');
+
+  const writer = stream._writer;
+
+  const error = new TypeError('Aborted');
+
+  for (const writeRequest of stream._writeRequests) {
+    writeRequest._reject(error);
+  }
+
+  if (writer !== undefined) {
+    WritableStreamDefaultWriterClosedPromiseReject(writer, error);
+
+    if (state === 'waiting') {
+      WritableStreamDefaultWriterReadyPromiseResolve(writer, undefined);
+    }
+  }
+
+  stream._state = 'errored';
+  stream._storedError = error;
+
+  return WritableStreamDefaultControllerAbort(stream._writableStreamController, reason);
+}
+
+// WritableStream API exposed for controllers.
+
+function WritableStreamAddWriteRequest(stream) {
+  const writer = stream._writer;
+  assert(IsWritableStreamDefaultWriter(writer) === true);
+
+  const state = stream._state;
+  assert(state === 'writable' || state === 'waiting');
+
+  const promise = new Promise((resolve, reject) => {
+    const writeRequest = {
+      _resolve: resolve,
+      _reject: reject
+    };
+
+    stream._writeRequests.push(writeRequest);
+  });
+
+  return promise;
+}
+
+function WritableStreamError(stream, e) {
+  const state = stream._state;
+  assert(state === 'writable' || state === 'waiting' || state === 'closing');
+
+  const writeRequests = stream._writeRequests;
+  while (writeRequests.length > 0) {
+    const writeRequest = writeRequests.shift();
+    writeRequest._reject(e);
+  }
+
+  const writer = stream._writer;
+
+  if (writer !== undefined) {
+    WritableStreamDefaultWriterClosedPromiseReject(writer, e);
+
+    if (state === 'waiting') {
+      WritableStreamDefaultWriterReadyPromiseResolve(writer, undefined);
+    }
+  }
+
+  stream._state = 'errored';
+  stream._storedError = e;
+}
+
+function WritableStreamFinishClose(stream) {
+  assert(stream._state === 'closing');
+
+  stream._state = 'closed';
+
+  const writer = stream._writer;
+
+  // writer cannot be released while close() is ongoing. So, we can assert that
+  // there's an active writer.
+  assert(writer !== undefined);
+
+  WritableStreamDefaultWriterClosedPromiseResolve(writer);
+}
+
+function WritableStreamFulfillWriteRequest(stream) {
+  assert(stream._writeRequests.length > 0);
+
+  const writeRequest = stream._writeRequests.shift();
+  writeRequest._resolve(undefined);
+}
+
+function WritableStreamUpdateBackpressure(stream, backpressure) {
+  const state = stream._state;
+  const writer = stream._writer;
+
+  if (state === 'writable') {
+    if (backpressure === false) {
+      return;
+    }
+
+    stream._state = 'waiting';
+
+    if (writer !== undefined) {
+      WritableStreamDefaultWriterReadyPromiseReset(writer);
+    }
+
+    return;
+  }
+
+  assert(state === 'waiting');
+
+  if (backpressure === true) {
+    return;
+  }
+
+  stream._state = 'writable';
+
+  if (writer !== undefined) {
+    WritableStreamDefaultWriterReadyPromiseResolve(writer, undefined);
+  }
+}
+
+class WritableStreamDefaultWriter {
+  constructor(stream) {
+    if (IsWritableStream(stream) === false) {
+      throw new TypeError('WritableStreamDefaultWriter can only be constructed with a WritableStream instance');
+    }
+    if (IsWritableStreamLocked(stream) === true) {
+      throw new TypeError('This stream has already been locked for exclusive writing by another writer');
+    }
+
+    this._ownerWritableStream = stream;
+    stream._writer = this;
+
+    const state = stream._state;
+
+    if (state === 'writable' || state === 'waiting' || state === 'closing') {
+      WritableStreamDefaultWriterClosedPromiseInitialize(this);
+    } else {
+      if (state === 'closed') {
+        WritableStreamDefaultWriterClosedPromiseInitializeAsResolved(this, undefined);
+      } else {
+        assert(state === 'errored', 'state must be errored');
+
+        WritableStreamDefaultWriterClosedPromiseInitializeAsRejected(this, stream._storedError);
+      }
+    }
+
+    if (state === 'waiting') {
+      WritableStreamDefaultWriterReadyPromiseInitialize(this);
+    } else {
+      WritableStreamDefaultWriterReadyPromiseInitializeAsResolved(this, undefined);
+    }
+  }
+
+  get closed() {
+    if (IsWritableStreamDefaultWriter(this) === false) {
+      return Promise.reject(CreateWritableStreamDefaultWriterBrandCheckException('closed'));
+    }
+
+    return this._closedPromise;
+  }
+
+  get desiredSize() {
+    if (IsWritableStreamDefaultWriter(this) === false) {
+      throw CreateWritableStreamDefaultWriterBrandCheckException('desiredSize');
+    }
+
+    if (this._ownerWritableStream === undefined) {
+      throw CreateWritableStreamDefaultWriterLockException('desiredSize');
+    }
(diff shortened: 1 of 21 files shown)


tyoshino:
Squashed and rebased version of https://github.com/whatwg/streams/pull/462.

tyoshino:
I inserted a lot of logs this time. I'd like to leave some or all of them in the reference impl for debugging in the future. Do you have any opinion about that, Domenic?

tyoshino:
In rewriting pipeTo() algorithm, I felt that it's good if we apply some convention to variable naming to distinguish variables local to the inner functions and ones of pipeTo() like how we're distinguishing internal slots and local variables. I'm tentatively prefixing them with _.

domenic:
> I'd like to leave some or all of them in the reference impl for debugging in the future. Do you have any opinion about that, Domenic?

I think we should at least comment them out when committing to the repo. In the future we could investigate a logging framework that allows us to toggle them on and off with environment variables, like https://www.npmjs.com/package/debug maybe.

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
 
-    this._closedPromise = new Promise((resolve, reject) => {
-      this._closedPromise_resolve = resolve;
-      this._closedPromise_reject = reject;
-    });
+    this._writer = undefined;
 
-    this._readyPromise = Promise.resolve(undefined);
-    this._readyPromise_resolve = null;
+    this._writeRequests = [];
```
domenic: Maybe add a comment that these are kept here instead of on the writer because we want to allow multiple queued up writes.

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
   }
 }
 
-exports.WritableStream = WritableStream;
+// Helper functions for the WritableStreamDefaultWriter.
 
-function closure_WritableStreamErrorFunction() {
-  const f = e => ErrorWritableStream(f._stream, e);
-  return f;
+function CreateWritableStreamDefaultWriterBrandCheckException(name) {
```
domenic: Name this lowercase since it isn't an abstract op. Also since it's local to this file we could just call it `writerBrandCheckException`.

(Similar comments for the other helpers. Maybe also consider moving the helpers to one section instead of per class?)

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
-  get state() {
-    if (!IsWritableStream(this)) {
-      throw new TypeError('WritableStream.prototype.state can only be used on a WritableStream');
+  abort(reason) {
+    if (IsWritableStream(this) === false) {
+      return Promise.reject(CreateWritableStreamBrandCheckException('abort'));
+    }
+
+    if (IsWritableStreamLocked(this) === true) {
+      return Promise.reject(new TypeError('Cannot abort a stream that already has a reader'));
```
domenic: s/reader/writer

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
     }
 
-    return this._state;
+    WritableStreamAbort(this, reason);
```
domenic: Missing return? Presumably missing tests for this return too.

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
+    return AcquireWritableStreamDefaultWriter(this);
+  }
+}
+
+exports.WritableStream = WritableStream;
+
+// Helper functions for the WritableStream.
+
+function CreateWritableStreamBrandCheckException(name) {
+  return new TypeError('WritableStream.prototype.' + name + ' can only be used on a WritableStream')
```
domenic: Missing semicolon (I should work on adding a lint step...)

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
+function WritableStreamAbort(stream, reason) {
+  const state = stream._state;
+  if (state === 'closed') {
+    return Promise.resolve(undefined);
+  }
+  if (state === 'errored') {
+    return Promise.reject(stream._storedError);
+  }
+
+  assert(state === 'writable' || state === 'waiting' || state === 'closing');
```
domenic: Interesting, we still distinguish 'waiting' and 'writable'? Is that correct, or should we be checking the queue size instead?

domenic: Basically my concern is with keeping multiple pieces of information coordinated. For example maybe what we want here is to check if the ready promise exists (or to make WritableStreamDefaultWriterReadyPromiseResolve do nothing if it doesn't exist).

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
+    if (backpressure === false) {
+      return;
     }
-    if (this._state === 'closed') {
-      return Promise.reject(new TypeError('cannot close an already-closed stream'));
+
+    stream._state = 'waiting';
+
+    if (writer !== undefined) {
+      WritableStreamDefaultWriterReadyPromiseReset(writer);
```
domenic: I think this is a helper function, but in the spec language we would not write it as an abstract op. Because in the spec language we don't need to track the resolve and reject callbacks, so it can be a very simple "Set writer.[[readyPromise]] to a new promise." The same might be true of many of the other promise-manipulation methods.

In general naming them with lowerCamelCase and maybe moving them to another block would help differentiate when it comes time to do the spec transition.

reference-implementation/lib/writable-stream.js
```diff
@@ -4,225 +4,661 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, PromiseInvokeOrFallbackOrNoop, Valida
...
+    }
+
+    if (this._ownerWritableStream === undefined) {
+      throw CreateWritableStreamDefaultWriterLockException('desiredSize');
+    }
+
+    return WritableStreamDefaultWriterGetDesiredSize(this)
+  }
+
   get ready() {
```
domenic: Are you still planning to eventually make this `waitForDesiredSize(n)` or similar? Per https://github.com/whatwg/streams/issues/318 ?

reference-implementation/lib/writable-stream.js
```diff
@@ -234,65 +672,109 @@ function IsWritableStream(x) {
...
+
+  const sinkClosePromise = PromiseInvokeOrNoop(controller._underlyingSink, 'close');
+  sinkClosePromise.then(
+    () => {
+      if (stream._state !== 'closing') {
+        return;
+      }
 
-  return undefined;
+      WritableStreamFulfillWriteRequest(stream);
```
domenic: Why not fulfill the write request normally? Why is the last write request special?

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/e57e25e9385d2ce243e7da0de4b95d9f35d9cbcb

I'm not sure if it's what you meant. Please take a look again.

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/e57e25e9385d2ce243e7da0de4b95d9f35d9cbcb

> (Similar comments for the other helpers. Maybe also consider moving the helpers to one section instead of per class?)

Will do later.

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/e57e25e9385d2ce243e7da0de4b95d9f35d9cbcb

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/e57e25e9385d2ce243e7da0de4b95d9f35d9cbcb

Added a test, too.

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/e57e25e9385d2ce243e7da0de4b95d9f35d9cbcb

> (I should work on adding a lint step...)

Ah, yeah. Style check.

reference-implementation/lib/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/78da22f4ee481cce7a57cd438c0d01770fd6bdf2

tyoshino: Removed prefix as you suggested above.

reference-implementation/test/transform-stream.js
```diff
@@ -207,67 +187,59 @@ test('TransformStream: by default, closing the writable waits for transforms to
...
   })
   .catch(e => t.error(e));
 });
 
 test('TransformStream: by default, closing the writable closes the readable after async enqueues and async done', t => {
-  t.plan(3);
-
   const ts = new TransformStream({
-    transform(chunk, enqueue, done) {
+    transform(chunk, done, enqueue) {
```
domenic: I don't mind, since I think we will work on transform streams more later, but why change the order of these arguments?

reference-implementation/test/writable-stream-abort.js
```diff
@@ -1,43 +1,84 @@
...
+  const writer = ws.getWriter();
+  writer.releaseLock();
+
+  const abortPromise = writer.abort();
+  abortPromise.then(() => {
+    t.fail('abortPromise fulfilled unexpectedly');
+    t.end();
+  },
+  r => {
+    t.end();
```
domenic: Should probably test that it's a TypeError

reference-implementation/lib/writable-stream.js
tyoshino: Good catch. The differentiation is now working just as a cache for reducing backpressure signal checking. I'll try to remove it.

reference-implementation/test/writable-stream.js
```diff
@@ -1,25 +1,131 @@
...
+  array.forEach(chunk => writableStreamWriter.write(chunk));
+  return writableStreamWriter.close();
+}
+
+test('desiredSize on a released writer', t => {
+  const ws = new WritableStream({});
+  const writer = ws.getWriter();
+  writer.releaseLock();
+
+  try {
```
domenic: You could use `t.throws(...)`? https://www.npmjs.com/package/tape#tthrowsfn-expected-msg

domenic:
Review finished :). Seems like you've got most of my comments already. Big ones are `waitForDesiredSize` (https://github.com/whatwg/streams/pull/488#discussion_r72938794) and the different states (https://github.com/whatwg/streams/pull/488#discussion_r72937800).

reference-implementation/test/transform-stream.js
tyoshino: Yeah, this is what I should clearly explain. I thought that like streams' controller, it might be good to pass the TransformStream controlling functions at the time of creation so that we allow the implementation to start writing some header that is independent from what will be written to the writable side. Then, we'll introduce start(). start() would take function to enqueue(), close() and error(). They don't need to be given to each of transform() and flush() but passed just as a shortcut. But, chunk and done() are specific to each transform() call. So, I thought it would be good to regroup them like this.

reference-implementation/lib/writable-stream.js
domenic: Yeah that is exactly what I meant, sorry, I was pretty unclear.

reference-implementation/test/transform-stream.js
domenic: Oh, very interesting.

I guess I am still hopeful we can get rid of done by using a promise return. But I don't remember where our discussion landed there. We'll figure it out.

reference-implementation/test/writable-stream-abort.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/d08b7c86de53a64755e25e32ddfde37847b3f701

reference-implementation/test/writable-stream.js
tyoshino: Fixed by https://github.com/whatwg/streams/pull/488/commits/d08b7c86de53a64755e25e32ddfde37847b3f701

I forgot that tape's throws has been fixed to be able to handle that.

reference-implementation/lib/writable-stream.js
tyoshino: _writeRequests now holds both pending write and pending close. That's why we're calling this here.

reference-implementation/lib/writable-stream.js
tyoshino: First, converted `"waiting"` state into _backpressure internal slot by https://github.com/whatwg/streams/pull/488/commits/9cda324e1aa73dcbf52b658908a79afa5feba075

And tried eliminating the _backpressure variable by https://github.com/whatwg/streams/pull/488/commits/f5ebbeb01a4abf0204588d2718931e4a81a5f915

What do you think? The code after removing _backpressure is a bit slow. I'll check how much tomorrow.

domenic:
This is looking really good to me.

We may want to create ReadableStreamDefaultControllerErrorIfNeeded so that readable and writable are more parallel. I noticed the spec always does a state check for readable streams in the same way.

Let me know how I can help.

tyoshino:
> We may want to create ReadableStreamDefaultControllerErrorIfNeeded so that readable and writable are more parallel. I noticed the spec always does a state check for readable streams in the same way.
> 
> Let me know how I can help.

Please go ahead to take it :)

reference-implementation/lib/writable-stream.js
tyoshino: Thanks for reminding me. I think it's ok to keep desiredSize getter. waitForDesiredSize(n) or waitForDesiredSizeChange() could be added to give the same power as the pull() signal for underlyingSource to WritableStream users would be good I still think basically. But it doesn't need to be included in this PR, I guess. WDYT?

domenic: What if we changed ready() from a getter to a function, waitForDesiredSize() with no arguments? Then in the future we could upgrade it to allow arguments. But by default it waits for desiredSize >= highWaterMark.

reference-implementation/lib/writable-stream.js
tyoshino: After removing `"waiting"`
fd41dd1ce4cbdc1afb8d764200b315b1a5f723b4
  duration:  22.5s

Before removing `"waiting"`
d08b7c86de53a64755e25e32ddfde37847b3f701
  duration:  11.4s

Twice slower

reference-implementation/lib/writable-stream.js
domenic: It would technically be a normative change as we would stop returning the same promise every time and start returning a new one for each function invocation. But I think that would be OK.

reference-implementation/lib/writable-stream.js
domenic: Interesting. I bet most of the loss comes from eliminating _backpressure?

I think we should add a non-normative note advising implementations to cache the result of WritableStreamDefaultControllerGetBackpressure (and invalidate it under... which conditions?).

reference-implementation/lib/writable-stream.js
tyoshino: So, `waitForDesiredSize()` is equivalent to .. `waitForDesiredSize(1)`, and it would fulfill when the desired size becomes equal to or more than 1?

waitForDesiredSizeChange() can be implemented by
```
const current = writer.desiredSize;
writer.waitForDesiredSize(current + 1).then(...);
```
assuming there's no legitimate use of non integer values together with this waiting API.

> desiredSize >= highWaterMark.

nitpicking: you meant `HWM >= queueSize`?

reference-implementation/lib/writable-stream.js
tyoshino: Oh, I was thinking that GetTotalQueueSize() is optimized by letting the EnqueueValueWithSize() and DequeueValue() calculate and save the total size, but it's for loop. That should be the cause.

tyoshino: fd41dd1ce4cbdc1afb8d764200b315b1a5f723b4 with totalSize caching
  duration:  228ms

It takes only 4.2s to run all tests.

reference-implementation/lib/writable-stream.js
domenic: I guess we are somewhat re-debating (1) vs. (2) from #318. I guess I was thinking (2), but (1) is probably best. If we do (1) then I think you are right.

If we do (2) then waitForDesiredSize() is equivalent to waitForDesiredSize(HWM), i.e. queueSize <= 0. It waits for complete draining instead of just dipping below the HWM.

But yeah, let's do (1).

reference-implementation/lib/writable-stream.js
domenic: This is part of why I don't like it when people use the reference implementation as a polyfill :P

Maybe we should add caching to those operations just so that we can run our own tests faster.

tyoshino: Added caching internally in queue-with-sizes.js
72157748bc5543c0cedec3fc69ed94ab3bffdfe2

Tell me if you think we should address it differently.

reference-implementation/lib/writable-stream.js
tyoshino: Oops! I see. I wrote the comment without remembering that the point is not yet finalized in #318.

domenic:
I think this is ready! What about you?

domenic:
Draft commit message:

Add locking and precise flow control to WritableStream

This fixes #319 (by adding locking and the concept of a writable stream writer) and fixes #318 (by adding precise flow control via the writer.desiredSize property). The structure parallels that introduced for ReadableStream, including the introduction of a controller class.

The infrastructure is set up to allow the future introduction of different types of writable streams (with corresponding new controllers and writers).

With this in place, the public API for writable streams is almost stable; the remaining issue is whether writers should have a ready promise, or a waitForDesiredSize() method which could in the future be customized by passing an argument (see discussions in #318).

The reference implementation's pipeTo and transform stream code has been updated, and all the tests pass, but they haven't been put in the spec yet, as their design is not yet finalized (but is much closer to now that they have a more stable WritableStream foundation).

Merge at will!

tyoshino:
Merged as 5a2753464bd6e34e3bc7a57fcaac24ea5153fbb6

reference-implementation/lib/writable-stream.js
tyoshino: > Will do later.

Done by ce435714c451cd4d8b5b6cca3d6de251fda8e827

tyoshino:
> We may want to create ReadableStreamDefaultControllerErrorIfNeeded so that readable and writable are more parallel. I noticed the spec always does a state check for readable streams in the same way.

Done by 28c20b120a107b067aa287316d3f623cddfd3936

PR #490, opened 2016-08-03
Performance improvements

gwicke:
- Avoid use of Object.defineProperty for enumerable / writable / configurable
  properties.
- Merge two adjacent try/catch statements into one.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -26,7 +26,7 @@ exports.toInteger = v => {
 
 exports.createDataProperty = (o, p, v) => {
   assert(exports.typeIsObject(o));
-  Object.defineProperty(o, p, { value: v, writable: true, enumerable: true, configurable: true });
+  o[p] = v;
 };
 
 exports.createArrayFromList = elements => {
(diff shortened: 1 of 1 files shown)


domenic:
Sorry, this is a reference implementation, and is meant to emulate the semantics exactly. It is not meant to be used in production by performance-sensitive applications. Someone should work on a polyfill that actually applies optimizations.

gwicke:
@domenic, while I agree that this is a reference implementation & exact semantics should trump performance, I am not sure that there *is* actually a change in semantics here.

- My understanding of those `Object.defineProperty ` calls is that since they don't actually deviate from regular property settings (enumerable / writable / configurable), they are semantically identical to a regular property. Is there a difference I am missing here?
- The try/catch change does not seem to change semantics either, unless there is a way this code could throw an exception:
```javascript
if (method === undefined) {
  return Promise.resolve(undefined);
}
```

domenic:
Right. The point is we want it to stay as close to the spec text as possible so it is easy to maintain and work with. Since the spec explicitly lists the descriptor, so do we.

There is also a difference between `O[p] = v` and defineProperty; the former triggers setters.

gwicke:
> There is also a difference between O[p] = v and defineProperty; the former triggers setters.

Right. However, these helpers are exclusively used on objects that are freshly constructed in the library, and thus do not have any setters.

I understand and sympathize with your concern about maintaining a more direct mapping from spec to code. Would a comment pointing out the equivalent `Object.defineProperty` call serve this purpose as well?

domenic:
No. Sorry, I'm just not interested in tuning the reference implementation for performance.

Also, the effects will be different if any code adds setters to Object.prototype.

PR #497, opened 2016-08-04
Rename headings ID atrributes

tyoshino:


tyoshino:
Hmm, so we can rename only ReadableStreamDefaultControllerShouldCallPull of them?

domenic:
I think we can rename them all. It seems like https://fetch.spec.whatwg.org/#streams is kind of broken anyway (not updated for controller/byte streams refactoring). So we should just break all the links and fix them as part of #372.

PR #502, opened 2016-08-04
Remove unchecked expectation from readable byte stream test

tyoshino:


domenic:
I was wondering about those. Thanks.

PR #503, opened 2016-08-04
Make writer.desiredSize return 0 for closed streams

tyoshino:


tyoshino:
Added the issue number to the commit message and force pushed

PR #504, opened 2016-08-04
Fix makeReadableBackpressureByteSocketStream to drain all data

### LINKED ISSUE #489: Fix makeReadableBackpressureByteSocketStream() to drain all the available data
https://github.com/ricea/streams/pull/1/files/1941c5b9244184941dbb27a23bde42d7de77e627#r72926215

The example should drain all the available data and enqueue them to the RS in onreadable if byobRequest.view couldn't receive all the bytes.

tyoshino:
Fixes #489

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3727,7 +3727,7 @@ socket.
 The following function returns <a>readable byte streams</a> that wraps a hypothetical UDP socket API, including a
 promise-returning <code>select2()</code> method that is meant to be evocative of the POSIX select(2) system call.
 
-Since UDP sockets do not have any built-in backpressure support, the backpressure signal given by
+Since the UDP protocol does not have any built-in backpressure support, the backpressure signal given by
 {{ReadableByteStreamController/desiredSize}} is ignored, and the stream ensures that when data is available from the
 socket but not yet requested by the developer, it is enqueued in the stream's <a>internal queue</a>, to avoid overflow
 of the kernel-space queue and a consequent loss of data.
--- a/index.bs
+++ b/index.bs
@@ -3724,103 +3724,77 @@ socket.
 
 <h3 id="example-rbs-push">A readable byte stream with an underlying push source and backpressure support</h3>
 
-The following function returns <a>readable byte streams</a> that allow efficient direct reading of TCP sockets, based on
-a hypothetical JavaScript socket API. The socket object is an <code><a idl class="non-normative">EventTarget</a></code>
-which emits a <code>readable</code> event (similar to the POSIX select(2) call). [[DOM]]
+The following function returns <a>readable byte streams</a> that represents an adaptive UDP-based video stream. It uses
+a hypothetical UDP socket API, including a promise-returning <code>select2()</code> method that is meant to be evocative
+of the POSIX select(2) system call. It uses the <a>backpressure</a> signaling provided by
+{{ReadableByteStreamController/desiredSize}} to send a message down the socket asking for an increased bit rate whenever
+too much buffered data is being accumulated, as this represents a case where the user is watching video slower than the
+data is arriving.
 
-This setup allows zero-copy reading directly into developer-supplied buffers. Additionally, it ensures that when data is
-available from the socket but not yet requested by the developer, it is enqueued in the stream's <a>internal queue</a>,
-to avoid overflowing the kernel-space queue. In this case, an additional copy will potentially be necessary when using a
-<a>BYOB reader</a>, to move the data from the stream's internal queue to the developer-supplied buffer. If this occurs,
-<a>backpressure</a> will immediately be applied downstream on the socket, by adjusting the TCP window size.
+The byte stream setup shown here allows zero-copy reading directly into developer-supplied buffers. Additionally, it
+ensures that when data is available from the socket but not yet requested by the developer, it is enqueued in the
+stream's <a>internal queue</a>, to avoid overflow of the kernel-space queue and a consequent loss of video data. In this
+case, an additional copy will potentially be necessary when using a <a>BYOB reader</a>, to move the data from the
+stream's internal queue to the developer-supplied buffer.
 
 <pre><code class="lang-javascript">
-  const DEFAULT_WINDOW_SIZE = 65536;
+  const STARTING_BIT_RATE = 131072;       // 1 Mbps
+  const DEFAULT_CHUNK_SIZE = 65536;       // 64 KiB
+  const PREFERRED_QUEUED_SIZE = 10485760; // 10 MiB
 
-  function makeReadableBackpressureByteSocketStream(host, port) {
-    const socket = createHypotheticalTCPSocket(host, port);
+  function makeReadableVideoSocketStream(host, port) {
+    const socket = createUDPSocket(host, port);
 
     return new ReadableStream({
       type: "bytes",
 
       start(controller) {
-        socket.setTCPWindowSize(Math.max(0, controller.desiredSize));
+        socket.send(createSetBitrateMessage(STARTING_BIT_RATE));
 
-        let buffer;
-        let byteOffset;
-        let byteLength;
-        let bytesWritten = 0;
+        readRepeatedly().catch(e => controller.error(e));
 
-        socket.onreadable = () => {
-          while (true) {
-            // Since onreadable can happen even when there's no pending BYOB
-            // requests, we need to handle both cases.
-            //
-            // In case the BYOB requests cannot drain all available data, we
-            // enqueue them into the stream.
-
-            if (buffer === undefined) {
-              bytesWritten = 0;
-
-              if (controller.byobRequest) {
-                const v = controller.byobRequest.view;
-
-                buffer = v.buffer;
-                byteOffset = v.byteOffset;
-                byteLength = v.byteLength;
-              } else {
-                buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
-                byteOffset = 0;
-                byteLength = DEFAULT_WINDOW_SIZE;
-              }
-            }
-
-            const result = socket.readInto(buffer, byteOffset + bytesWritten, byteLength - bytesWritten);
-
-            if (result.type === socket.WOULDBLOCK || result.type === socket.EOF) {
-              break;
-            } else if (result.type === socket.OK) {
-              bytesWritten += result.bytesRead;
-              if (bytesWritten === byteLength) {
-                break;
-              }
-            } else {
-              // Handle errors.
-            }
-          }
-
-          if (bytesWritten > 0) {
+        function readRepeatedly() {
+          return socket.select2().then(() => {
+            // Since the socket can become readable even when there’s
+            // no pending BYOB requests, we need to handle both cases.
+            let bytesRead;
             if (controller.byobRequest) {
-              controller.byobRequest.respond(bytesWritten);
+              const v = controller.byobRequest.view;
+              bytesRead = socket.readInto(v.buffer, v.byteOffset, v.byteLength);
+              controller.byobRequest.respond(bytesRead);
             } else {
-              controller.enqueue(new Uint8Array(buffer, 0, bytesWritten));
+              const buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
+              bytesRead = socket.readInto(buffer, 0, DEFAULT_WINDOW_SIZE);
+              controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
             }
 
-            buffer = undefined;
-          }
+            if (bytesRead === 0) {
+              controller.close();
+              return;
+            }
 
-          // The internal queue size has changed, so propagate
-          // the backpressure signal to the underlying source.
-          socket.setTCPWindowSize(Math.max(0, controller.desiredSize));
-        };
+            // Adjust the bitrate according to how much data we have queued.
+            // (If the queue is full, then we can up the bitrate for higher quality.)
+            const newBitrate = determineBestBitrate(controller.desiredSize);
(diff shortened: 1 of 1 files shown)


tyoshino:
It's too complicated. Do you have any comment on how to simplify this? If we decide to land something complicated like this, I'd like to add something in the test in the reference implementation to check correctness, too.

domenic:
Hmm. I guess I have to admit I didn't understand the problem we were solving; I was hoping you would just work out a simple solution that would make sense when I read it :). Let me try to understand better now...

In https://github.com/ricea/streams/pull/1/files/1941c5b9244184941dbb27a23bde42d7de77e627#r72926215 @ricea says

> Careful inspection of the code indicates that the event is level-triggered. In other words, returning from onreadable without reading all available data results in another "readable" event being fired immediately. Should I mention this explicitly?

Since onreadable is a hypothetical API, I am wondering what code you are inspecting, @ricea?

Taking this behavior as a given though, I don't necessarily see a problem with just letting `onreadable` fire immediately. If it does, the second time we fire we will just repeat the logic, fulfilling any further BYOB read requests (if they exist) or enqueuing the data (if no such requests exist). That seems to work OK, right? Why is the version in this PR better?

ricea:
I was inspecting the code of the example itself. It made no effort to read more data if there was more than v.byteLength / DEFAULT_WINDOW_SIZE pending. If the example is correct, this means that onreadable must keep being fired as long as there is still data available.

It seems to me that Javascript events don't normally work that way. For one thing, if you're not ready to handle the data yet you'd have to actually unregister the event handler in order to stop the event loop from spinning. That seems unusual.

select() is also level-triggered, but it's not as problematic because
 a) It's a pull API, and
 b) It's idiomatic to remove descriptors from the set when you're not ready to read/write from them yet

Having said that, given the complexity of the fix, it may be simpler just to mention that onreadable is level-triggered and leave it at that.

domenic:
Ah, I understand better now. OK, hmm. I wonder if there's a way to fix this while also making our socket API more like the low-level version. (I am already a little unhappy that setTCPWindowSize is fictional.) I guess that is in a large part what the PR is trying to do.

Maybe one way of getting a bit less complexity is to pretend we had a `socket.select2()` that returned a promise, instead of an EventTarget. The translation between low-level polling in C and a higher-level event loop language is tricky to capture, but I think that might be a better mapping. (Is `select2()` a good name, or better to stick with something schematic like `waitForReadable()`?)

Concretely, here is what I am thinking:

```js
    start(controller) {
      socket.setTCPWindowSize(Math.max(0, controller.desiredSize));

      socket.onend = () => controller.close();

      readRepeatedly().catch(e => {
        if (e.code === "EOF") {
          controller.close();
        } else {
          controller.error(e)
        }
      });
      
      function readRepeatedly() {
        return socket.select2().then(() => {
          // Since the socket can become readable even when there’s
          // no pending BYOB requests, we need to handle both cases.
          if (controller.byobRequest) {
            const v = controller.byobRequest.view;
            const bytesRead = socket.readInto(v.buffer, v.byteOffset, v.byteLength);
            controller.byobRequest.respond(bytesRead);
          } else {
            const buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
            const bytesRead = socket.readInto(buffer, 0, DEFAULT_WINDOW_SIZE);
            controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
          }

          // The internal queue size has changed, so propagate
          // the backpressure signal to the underlying source.
          socket.setTCPWindowSize(Math.max(0, controller.desiredSize));
          
          return readRepeatedly();
        });
      }
    },
```

ricea:
This is more along the lines of moving the deckchairs around than actually moving closer to the C API, but fwiw here's an alternative:

    start(controller) {
      socket.setTCPWindowSize(Math.max(0, controller.desiredSize));

      readRepeatedly().catch(e => controller.error(e));

      function readOnce(buffer, offset, size) {
        return socket.read(buffer, offset, size)
            .then((bytesRead) => {
              if (bytesRead == 0) {
                controller.close();
              } else {
                // The internal queue size has changed, so propagate
                // the backpressure signal to the underlying source.
                socket.setTCPWindowSize(Math.max(0, controller.desiredSize));
                readRepeatedly();
              }
              return bytesRead;
            });
      }

      function readRepeatedly() {
        // Since the socket can become readable even when there's
        // no pending BYOB requests, we need to handle both cases.
        if (controller.byobRequest) {
          const v = controller.byobRequest.view;
          readOnce(v.buffer, v.byteOffset, b.byteLength)
              .then(bytesRead => controller.byobRequest.respond(bytesRead));
        } else {
          const buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
          readOnce(buffer, 0, DEFAULT_WINDOW_SIZE)
              .then(bytesRead => controller.enqueue(
                  new Uint8Array(buffer, 0, bytesRead)));
        }
      }
    }

I've renamed readInto() to read() and made it return a Promise. It also returns 0 for EOF rather than throwing an exception. I feel like this is closer in spirit to the C API.

If we were going to be closer to the C API we could call read() in a loop until it throws an EWOULDWAIT exception, and then call select() as in your version. But I suspect that would be really hard to read.

ricea:
If you stop trying to make it be a push source, it gets much simpler and closer to the C API:

```javascript
start(controller) {},

pull(controller) {
  // This is called when the internal queue has been drained, and the
  // stream’s consumer can accept more data.
  return socket.select()
      .then(() => {
        const v = controller.byobRequest.view;
        try {
          const bytesRead = socket.read(v.buffer, v.byteOffset, v.byteLength);
          if (bytesRead === 0) {
            controller.close();
          } else {
            controller.byobRequest.respond(bytesRead);
          }
        } catch (e) {
          controller.error(e);
        }
      });
}
```

This probably defeats the purpose of the example, however.

domenic:
The problem with that, I believe---please correct me if I'm wrong---is that it allows the kernel buffer to potentially overflow and cause data loss. The example as-is is trying to constantly pull shuffle out of kernel space and into user space to avoid this.

ricea:
There's no risk of data loss. The kernel controls the window size automatically so that its buffer does not overflow. If there is backpressure from Javascript, read() won't be called and the window size will shrink.

domenic:
OK. So I guess a TCP socket is not a great example here because in the end it will just duplicate our file example at https://streams.spec.whatwg.org/#example-rbs-pull.

Can anyone think of a good example of something that lets us illustrate the dual enqueue/byobRequest interface? Maybe a UDP socket (and just get rid of the backpressure stuff)?

ricea:
A non-zero number of stock exchanges provide ticker information as a stream of UDP packets with no backpressure, so that could be a motivating example.

Alternatively, we could imagine a video streaming UDP protocol where you send messages back to the server to vary the bitrate.

domenic:
Those are both great ideas. So if desiredSize is low (e.g. negative) then we're accumulating too much buffered data and we can afford a higher bitrate, basically. Something like this, perhaps?

```js
  const STARTING_BIT_RATE = 131072;       // 1 Mbps
  const DEFAULT_CHUNK_SIZE = 65536;       // 64 KiB
  const PREFERRED_QUEUED_SIZE = 10485760; // 10 MiB

  function makeReadableVideoSocketStream(host, port) {
    const socket = createUDPSocket(host, port);

    return new ReadableStream({
      type: "bytes",

      start(controller) {
        socket.send(createSetBitrateMessage(STARTING_BIT_RATE));

        readRepeatedly().catch(e => controller.error(e));

        function readRepeatedly() {
          return socket.select2().then(() => {
            // Since the socket can become readable even when there’s
            // no pending BYOB requests, we need to handle both cases.
            let bytesRead;
            if (controller.byobRequest) {
              const v = controller.byobRequest.view;
              bytesRead = socket.readInto(v.buffer, v.byteOffset, v.byteLength);
              controller.byobRequest.respond(bytesRead);
            } else {
              const buffer = new ArrayBuffer(DEFAULT_WINDOW_SIZE);
              bytesRead = socket.readInto(buffer, 0, DEFAULT_WINDOW_SIZE);
              controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
            }
            
            if (bytesRead === 0) {
              controller.close();
              return;
            }

            // Adjust the bitrate according to how much data we have queued
            // (If the queue is full, then we can up the bitrate for higher quality)
            const newBitrate = determineBestBitrate(controller.desiredSize);
            socket.send(createSetBitrateMessage(newBitrate));

            return readRepeatedly();
          });
        }
      }
      pull(controller) {
        // This is called when the internal queue has been drained, and the
        // stream's consumer can accept more data. Recompute the best bitrate
        // using the new desiredSize.
        const newBitrate = determineBestBitrate(controller.desiredSize);
        socket.send(createSetBitrateMessage(newBitrate));
      },

      cancel() {
        socket.close();
      }
    }, {
      highWaterMark: PREFERRED_QUEUED_SIZE
    });
  }
```

I am picturing people who actually work on adaptive video streaming cringing at this simplistic code, but it does kind of get the point across, I think...

ricea:
One problem with the adaptive streaming example that I didn't anticipate was that it is not clear what to do when the queue is full. I was picturing reducing the bitrate, but your answer of increasing it also makes sense.

I think we can assume that packets are always read completely or not at all, and so simplify the interface back to the "onreadable" event that was there originally. If v.byteLength - v.byteOffset is less than the size of the incoming packet then that's bad, but maybe it's better to gloss over that issue.

I think this example represents the state-of-the-art of adaptive streaming 20 years ago, and people who actually work in this area will realise that and be forgiving.

domenic:
I think the `v.byteLength - v.byteOffset` too-small case is the one worth worrying about. With a promise-returning `.select2()`, it should work fine; the next call to `.select2()` will immediately fulfill and we'll start going down the no-byobRequest path. So unless the kernel buffer fills up in one event loop turn we'll be OK. Whereas with the readable event we're in trouble.

I'll push a spec revision to this PR branch for people to review :)

domenic:
OK, I pushed it; you can view at https://streams.spec.whatwg.org/branch-snapshots/updateexample/#example-rbs-push

However I'm no longer sure about this example. It seems to be predicated on the idea of wanting there to be a full queue, which is kind of the opposite of what you usually want with byte streams, since a full queue means BYOB readers are screwed. Maybe a HWM of 0 is better, on the assumption that the video-playing infrastructure will maintain its own queue, and should just be grabbing bytes from the UDP socket as directly as possible?

ricea:
Yes. If the video decoder can't keep up, then it enters a spiral of failure where we keep trying to feed it more and more data, and it gets more and more behind.

Unfortunately, I don't know how to fix it. We could drop the whole adaptive rate idea and just set HWM to 0 to illustrate a byte stream which tries to drop as little data as possible from a source without guaranteed delivery.

tyoshino:
Thanks Adam for reviewing the meaning of the example seriously. Yeah, the adaptive stream idea doesn't make much sense for ordinal use cases basically.

As Adam said https://github.com/whatwg/streams/pull/504#issuecomment-237803327 and I said in https://github.com/whatwg/streams/pull/430#issuecomment-193877949 regarding Linux for example, it's safe to ask the kernel to buffer data.

To benefit from BYOB and also drain data from the source, the consumer should just prepare a large buffer to absorb all the fluctuation in download speed. The adaptive stream can add an additional option to it to have two-level strategy which is:
- the consumer would have a large buffer of X byte where X is kept chosen so that it's not too large (to keep average memory foot print small)
- the adaptive stream makes sure to drain data from the source when the data arrival speed it too fast accepting the cost of buffer allocation.

I tried to justify the need to drain data from the source, but don't have any good idea. Given that the POSIX socket API doesn't provide any method to directly control receive window size, buffer-in-JS may be useful to make sure to buffer large data to prepare for sudden slow-down of download. But this means that we try to always have some amount of data buffered, that is, we cannot benefit from BYOB. Buffering over X byte should just be temporary thing.

So, ... consumers for which the adaptive stream would be useful should have both the characteristics of WebSocket (push source) and BYOB style reading like POSIX read(2)...

domenic:
> So, ... consumers for which the adaptive stream would be useful should have both the characteristics of WebSocket (push source) and BYOB style reading like POSIX read(2)...

I think we have an example of that: UDP streams. However, they don't have backpressure at the socket level. So either we don't illustrate backpressure in the example (which might be the simplest thing to do) or we say that there's some sort of out-of-band backpressure signaling.

Let me try a version without backpressure and we'll see what we think... again it will be at https://streams.spec.whatwg.org/branch-snapshots/updateexample/#example-rbs-push when I finish (in a few minutes).

index.bs
```diff
@@ -3722,35 +3722,36 @@ sockets. This time, however, when we pipe to a destination that cannot accept da
...
-data is arriving.
+The following function returns <a>readable byte streams</a> that wraps a hypothetical UDP socket API, including a
+promise-returning <code>select2()</code> method that is meant to be evocative of the POSIX select(2) system call.
 
-The byte stream setup shown here allows zero-copy reading directly into developer-supplied buffers. Additionally, it
-ensures that when data is available from the socket but not yet requested by the developer, it is enqueued in the
-stream's <a>internal queue</a>, to avoid overflow of the kernel-space queue and a consequent loss of video data. In this
-case, an additional copy will potentially be necessary when using a <a>BYOB reader</a>, to move the data from the
-stream's internal queue to the developer-supplied buffer.
+Since UDP sockets do not have any built-in backpressure support, the backpressure signal given by
```
tyoshino: s/sockets/sockets and protocol/ as the UDP protocol also doesn't have it.

tyoshino:
566a24fe06af3254ee3f8fcb9736d327ea347c9d lgtm. Nice rewriting including mention on possible out-of-band flow control.

PR #508, opened 2016-08-22
Requirements.md: Change example of push stream from TCP to UDP

ricea:
As discussed in #504, the kernel will not drop buffered TCP data. Instead, refer to UDP data, which the kernel will drop when the buffer is full.


tyoshino:
lgtm

Issue #492, opened 2016-08-04
Convert writable stream tests to web platform tests format


domenic:
After #488 lands we should do this

---

Update 2016-10-25: here is a remaining checklist of files, after #540 lands:
- [x] bad-strategies.js
- [x] bad-underlying-sinks.js (can go to existing to-upstream bad-underlying-sinks.js)
- [x] brand-checks.js
- [x] byte-length-queuing-strategy.js
- [x] count-queuing-strategy.js

Updated 2016-11-07: #640 is taking care of the above.

tyoshino:
I (and @ricea also maybe) are going to work on this.

What would the best work flow? Maybe, local review somewhere, and then PR to w3c/.

Creating PR for fork would be overkilling, right?

Just commit to personal forks and then get them reviewed before upstream PR?
https://github.com/tyoshino/web-platform-tests/commit/4cc6bf218ef5bdae180cb425ee8d391787b89e38

domenic:
Oh, very exciting! I think doing something similar to what we did for readable stream makes sense: use this repo as a staging ground and convert the tests here, then when we feel they're "done", move them to WPT in one PR. The benefit there is that we can continue to validate the tests against the reference implementation as we go.

tyoshino:
OK. As we've removed the directory in https://github.com/whatwg/streams/commit/85260dc2ad3cb32ef2216e6af4653435dada3e0b and is now used for checking out wpt from w3c repo, let me create a new directory dedicated for staging/reviewing.

tyoshino:
Hmm, before https://github.com/whatwg/streams/commit/85260dc2ad3cb32ef2216e6af4653435dada3e0b, we had to manually copy the streams test to a full checkout of w3c wpt so that the tests can use e.g. stuff in service-workers/resources? Sorry, I forgot that.

Now, to keep the converted wpt tests for writable streams in this repository, basically we need to have a separate directory. To allow the tests in it to use the files (service-workers/resources), we need to create a symlink (not good for Windows checkout?) or (partially) clone the repo. It's not something we want to do, I think.

tyoshino:
Hmm, it's awkward, but negation entry in .gitignore would help. See #510.

domenic:
Please leave the piping tests to me and I'll work on converting them as part of #512.

tyoshino:
> Please leave the piping tests to me and I'll work on converting them as part of #512.

Got it, domenic.

[...]

PR #510, opened 2016-08-29
Convert some general tests for ws into wpt

tyoshino:
Partial work for #492


domenic:
OK, I think I fixed wpt-runner and moved the files around enough to make the build succeed. Let me know what you think. The actual conversion LGTM.

domenic:
We should probably merge this ASAP so that we don't write any more tests in the old format, like I did in #513.

tyoshino:
lgtm. Thanks for adding the feature to the wpt-runner. Merging.

tyoshino:
Merged as eb9bb0cbada0165c3fdad61825856e480f9f60f8

Issue #461, opened 2016-05-31
light weight transformations


dominictarr:
I hope it's not too late to make a suggestion,

The current design where a transform stream is a pair of writeable/readable streams is overkill for quite a few applications of transform streams.

Often, people use streams as (one hopes) very light weight transformations, for example, to parse an incoming file into lines, or maybe just to filter out items (there is a very strong correspondence to the array functions `map`, `filter`, etc)

These transformations simple transformations do not need their own back pressure - they simply need to proxy the back pressure further along the stream.

One simple way to achive this would be to think of a readable stream as a function that returns a promise when ever you call it, and those promises will be resolved in the order created.

A transform stream could instead be a function that takes a readable stream and returns another readable stream. Except when a transformed promise resolves it's value is transformed. This is great because you save a lot of queuing/dequeuing. The transform doesn't even need to be async, it can just return an the promise from upstream wrapped to return it's own result instead, but without creating a new promise.

You could flip this around the other way, and apply functions to a writeable stream - but that would mean a transform stream is a function that promises are passed to, and I feel a function that takes a readable stream is more natural.

For duplex streams, like a tcp client, a pair of streams is appropiate, because each side may have independent back pressure. but for simple cases, it just means now every little filter and map needs two queues one incoming one outcoming.

domenic:
Hey, great to have you here!

In general the design of transform streams, and even writable streams, is still in flux and open to suggestions. (Readable streams, we're pretty happy with.) So feedback like this is definitely good to have.

The big struggle we've had so far is figuring out the balance between two positions:

- Readable streams only, with transforms being functions mapping readable streams to readable streams, and writables being functions that accept readable streams
- The Node-style ecosystem with three first class concepts and pipe chains between them

For example, service worker APIs have leaned us toward the former for various technical reasons, which IMO as a fan of the Node style is disappointing.

My way of interpreting your feedback is that we'd want to preserve the three-concept ecosystem, but provide a performant and lightweight way to make synchronous transformations work. Here would be my strawman:

```js
const transform = TransformStream.map(syncMappingFunction);

readable.pipeThrough(transform).pipeTo(writable);
```

Here:

- transform.writable and transform.readable exist, and can be used directly if desired
- But pipeThrough performs a more optimal fast-path that avoids ever reifying those two streams

[...]

Issue #359, opened 2015-05-29
More concrete examples of "optimization" of pipeTo


tyoshino:
Potential targets of optimization

- TextEncodingTransformByteStream
- IdentityTransformByteStream
- BufferingTransformByteStream
- ByteCountingIdentityTransformByteStream
    - The wrapper is not a pure identify transform but with observation functionality.
    - In https://github.com/yutakahirano/fetch-with-streams/issues/37#issuecomment-105512434, use of a wrapper is proposed for counting the number of bytes consumed from a Request/Response.

## Skipping JS processing between a ReadableByteStream and a WritableByteStream

Skip JS code invocation (write(), read(), etc.) and does the following internally.

### Sink exposes a memory region for write

1. destMemory = sink.getMemoryToWrite()
1. source.generate(destMemory)
1. sink.notifyWritten()

### Source exposes a memory region where data for consuming is stored

1. sourceMemory = source.getMemoryToRead()
1. sink.consume(sourceMemory)
1. source.notifyConsumed()

### Both takes memory

1. tempMemory = allocate()
1. source.generate(tempMemory)
1. sink.consume(tempMemory)

### sendfile(2) style

1. transfer(lowerLayerSourceDescriptor, lowerLayerSinkDescriptor)

## Skipping an "identity" transform byte streams completely

```
a.pipeTo(id, options0);
id.pipeTo(b. options1);
```

Issues:

- Thread-safety: `a.pipeTo(id)` and `id.pipeTo(b)` may happen in different threads.
- How to observe bytes for ByteCountingIdentityTransformByteStream?

tyoshino:
We've decided to introduce isDisturbed, not byte counting system. So, we should reinvestigate how we set isDisturbed.

tyoshino:
Regarding how we realize postMessage()-ing streams https://github.com/whatwg/streams/issues/244, we should consider how to deal with invocation of an optimized pipeTo() logic for cross thread case.

tyoshino:
Please take a look at my strawman API for allowing pipeTo optimization at #511.

Issue #325, opened 2015-04-09
pipe redirection


tyoshino:
I started feeling that we've been too optimistic about use of specialized pipeTo algorithm given an identify transform between the pair of rs/ws.

In the discussion about the `Request` object, I've been saying that
- we should be able to `.write()` even before the `Request` gets attached to the final sink (https://github.com/yutakahirano/fetch-with-streams/issues/30#issuecomment-90882279).
- we can run specialized `.pipeTo()` internally even with existence of such a buffer.

The following are reduced example of what I said:

`id` is writable even before `id.readable.pipeTo()` happens.

```
id.writable.write("hello");
id.writable.write("world");
id.readable.pipeTo(ws);
```

In the following two examples, `rs.pipeTo(ws)` should happen eventually and internally after draining chunks queued in `id`.

```
rs.pipeTo(id.writable);
id.readable.pipeTo(ws);  // Eventually rs.pipeTo(ws)
```

```
rs.pipeTo(id.writable);
id.readable.pipeTo(ws);  // Eventually rs.pipeTo(ws)
```

Note that not whole of the `rs.pipeTo(ws)` algorithm should always happen. Closure of `rs` should result in making only `id` closed if `id.readable.pipeTo(ws)` is `id.readable.pipeTo(ws, { preventClose: true})`.

I'd call the main data transfer part of `.pipeTo()` as `.specialDataTransfer()` which excludes close/error signal.

So, let's think of what we should do for the following example:

```
id.writable.write("hello");
id.writable.write("world");
rs.pipeTo(id.writable);
id.readable.pipeTo(ws);
```

`id` already has two chunks queued in it. If `id.readable.pipeTo(ws)` algorithm is the one of `ReadableStream.pipeTo()` as-is, we cannot drain all the data from `id` until `ws` stop exerting back-pressure.

Without an ability to stop `rs.pipeTo(id.writable)`, we never be able to switch to rs.specialDataTransfer(ws).

I guess, the solution is:
- Identity transform flushes all the data queued in it when `id.readable.pipeTo(ws)` is called regardless of back-pressure of `ws`.
- Provide an (possibly private) API, say `.redirectTo`, on the `WritableStream` for `.pipeTo()` algorithm to redirect `.specialDataTransfer()` to `ws`. `ReadableStream.pipeTo()` is required to check `.redirectTo` before every `dest.write()` and if it's set, it must switch to use `rs.specialDataTransfer(dest.redirectTo)`.

[...]

Issue #321, opened 2015-04-08
explicity operate on underlying source and sink instead of public methods


wanderview:
Certain optimizations like piping off-main-thread are much more difficult in a world where the public methods of a ReadableStream and WritableStream can be monkey patched with js.  To enable UAs to implement these optimizations the spec should be written such that it operates directly on the underlying sources and sinks.

For example, ReadableStream.pipeTo(ws) should not call ws.write(), but should write data to ws's underlying sink.

This makes the API a bit more "DOMy" and a bit less "JSy", but makes it much easier to create efficient implementations in multi-threaded browsers.

wanderview:
In IRC @domenic also suggested that we could add a WritableStream cast function if duck-typed WritableStreams are found to be needed.

yutakahirano:
We planned to introduce a generic pipeTo function that interacts with public reader / writer APIs and specific ones implemented natively. What's wrong with the idea?

yutakahirano:
> We planned to introduce a generic pipeTo function that interacts with public reader / writer APIs and specific ones implemented natively. What's wrong with the idea?

I mean, the existing (planned) pipeTo can handle arbitrary streams correctly and can handle specific ones (that possibly includes ReadableStream and WritableStream) efficiently. Your proposal seems to drop the former property and I don't see what we get instead.

domenic:
So, the current reference implementation is generic on both `this` (using `this.getReader()`) and on its argument (using `dest.write()`, or in the future probably `dest.getWriter()`). And it's also generic on the objects returned from those.

I think the genericness on `this` is not too important. It's main benefit is that it would allow duck-type readable streams that don't inherit from `ReadableStream` to do `MyReadableStream.prototype.pipeTo = ReadableStream.prototype.pipeTo` and get a pipe algorithm automatically. (It would also allow the edge case of subclasses which use "return override" but that is pretty crazy... can explain more if interested.) So let's put that aside for now and say that we abandon it, just like when formalizing .tee() I abandoned it.

[...]

PR #511, opened 2016-08-30
Add a proof of concept of optimized pipe

tyoshino:
Related to #359, #325, #97, #321, #461

### DIFF
--- a/reference-implementation/lib/fast-pipe.js
+++ b/reference-implementation/lib/fast-pipe.js
@@ -54,6 +54,10 @@ class Pipe {
     doSpecialPipe(readable, writable);
   }
 
+  get coveredRequests() {
+    return this._coveredRequests;
+  }
+
   onProgress(numBytesDone) {
     globalPipeManager.onPipeProgress(this, numBytesDone);
   }
@@ -68,23 +72,29 @@ class Pipe {
   }
 }
 
+// Corresponds to and describes each pipeTo() invocation.
 class PipeRequest {
-  constructor(readable, writable, numBytes) {
+  constructor(reader, writer, numBytes) {
     this._remainingBytes = numBytes;
 
-    this._readable = readable;
-    this._writable = writable;
+    // The reader of the source readable stream from which we pipe data.
+    this._reader = reader;
+    // The writer of the destination writable stream to which we pipe data.
+    this._writer = writer;
 
+    // The active pipe instance which is covering this request.
     this._pipe = undefined;
 
     this._done = false;
+
+    globalPipeManager.register(this);
 
     function processPipeCandidates() {
       if (this._done === true) {
         return;
       }
 
-      const candidates = this._writable.pipeCandidates;
+      const candidates = this._writer.pipeCandidates;
 
       this._pipeCandidates = [];
       for (let candidate of candidates) {
@@ -105,19 +115,28 @@ class PipeRequest {
       // pipeCandidates might be updated asynchronously.
       // GlobalPipeManager.onRequestsUpdate() may need to adopt some
       // intelligent algorithm to avoid churn.
-      const candidatesToForward = [this];
+      //
+      // flushRequired is a list of writable streams indicating that the
+      // candidates are not yet available but becomes available if writes to
+      // the writable streams are throttled to allow their parent transform
+      // stream to flush data. E.g. an IdentityTransformStream with pending
+      // queue inside it could append its writable stream to flushRequired to
+      // ask the global pipe manager to suspend piping to the writable stream
+      // for a while to enable the pipe candidates.
+      const candidatesToForward = {
+        flushRequired: [],
+        pipeRequest: this
+      };
       for (let candidate of candidates) {
         candidatesToForward.push(candidate);
       }
-      this._readable.notifyPipeCandidates(candidatesToForward);
+      this._reader.notifyPipeCandidates(candidatesToForward);
 
-      this._writable.waitWritablesChange.then(() => {
+      this._writer.waitPipeCandidatesChange.then(() => {
         processPipeCandidates();
       })
     }
     processPipeCandidates();
-
-    globalPipeManager.register(this);
   }
 
   // To be scanned by the global pipe manager and the best one will be chosen
--- a/reference-implementation/lib/fast-pipe.js
+++ b/reference-implementation/lib/fast-pipe.js
@@ -1,0 +1,160 @@
+'use strict';
+
+// Note: Readable streams and writable streams that are capable of fast piping
+// doesn't use strategy in JS.
+
+class GlobalPipeManager {
+  constructor() {
+    // Describes all the ongoing pipes including non-skipping pipeTo().
+    this._pipeRequests = {};
+
+    this._pipes = [];
+  }
+
+  onPipeProgress(pipe) {
+    for (let req of pipe.coveredRequests) {
+      req.onProgress(numBytesDone);
+    }
+
+    reorganizeIfNeeded();
+  }
+
+  registerRequest(req) {
+    this._pipeRequests[req] = true;
+
+    reorganizeIfNeeded();
+  }
+
+  unregisterRequest(req) {
+    delete this._pipeRequests[req];
+  }
+
+  reorganizeIfNeeded() {
+    // Calculate the best combinations, so that
+    // - All requests in _pipeRequests are covered.
+    // - There's no overlap of covered requests between pipes in _pipes.
+
+    // Stop pipes if needed.
+    // Create pipes if needed.
+  }
+}
+
+GlobalPipeManager.PROGRESS = 0;
+GlobalPipeManager.REQUESTS_UPDATE = 1;
+
+const globalPipeManager = new GlobalPipeManager();
+
+class Pipe {
+  constructor(readable, writable, coveredRequests) {
+    this._coveredRequests = coveredRequests;
+    for (let req of coveredRequests) {
+      req.coveredBy(this);
+    }
+
+    doSpecialPipe(readable, writable);
+  }
+
+  onProgress(numBytesDone) {
+    globalPipeManager.onPipeProgress(this, numBytesDone);
+  }
+
+  stop() {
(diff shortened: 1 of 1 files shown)


tyoshino:
This is a proof-of-concept code snippet for my answer to the optimized piping vision. Basically brain dump + a little validation of solidness.

domenic:
Very, very interesting. Could you explain the vision a bit more? I think I see the outlines, but getting a high-level view would be nice.

I also think it would be ideal if we allowed JS strategies somehow. I guess they make things fairly observable though, as size() gets called all the time... that's tricky.

I'm going to work today on a requirements-based spec for pipeTo. Let's make sure it synchronizes with your vision.

tyoshino:
OK.

I forgot to note but this is intended to be used just for demonstrating that our API enables this kind of optimization. Not necessarily intended to be a part of the specification.

### PipeRequest

First, an instance of the PipeRequest class corresponds to each pipeTo() invocation. It holds the instances and parameters given to the method:
- destination writable stream instance
- source readable stream instance
- parameters such as preventClose given to pipeTo()

It also has
- the `pipe` slot which points the Pipe instance which represents ongoing piping which is going to satisfy the PipeRequest (wholly or partially) if any.
- the `done` flag which is initially false.

Every time a new pipeTo() is invoked, a PipeRequest instance describing the pipeTo() invocation is created and:
- the PipeRequest is registered with the global pipe manager.
- the PipeRequest retrieves current **(skipped) piping candidates** from the destination writable stream from `.pipeCandidates` and passes it to the source readable stream by calling `.notifyPipeCandidates()`. This is called **pipe candidates propagation**.
- the PipeRequest subscribes to the destination writable stream to get notified of updates on the piping candidates of the destination writable stream by calling `.waitPipeCandidatesChange()`. Every time new candidates are received, it's propagated to the source readable stream.

The global pipe manager holds:
- a list of all readable stream instances being piped.
- lists representing piping candidates for each of the piped readable streams.
  - It's a list of references to PipeRequest instances. It's sorted in the order where ones went through the least number of "pipe candidates propagation" comes first. I.e. destinations closest to the readable stream comes first.

The global pipe manager chooses the best subset of triplets of:
- a readable stream
- pipe candidate (i.e. the destination writable stream)
- the transfer method to use. E.g. between a TCP socket and a file descriptor, either of the default pipeTo() algorithm operating over the public ReadableStream.*Reader/WritableStream.*Writer interfaces, or offloading by sendfile() could be used. The global pipe manager lists up available transfer methods by checking the underlying sources/sinks of the readable/writable stream pair.

The subset must cover all the active PipeRequests. E.g. skipped piping to `b` in a list representing skipped piping candidates `[a, b, c]` can cover PipeRequests `a` and `b`. `c` must be covered by any other pipe.

### Reorganizing pipes

Ongoing pipes can be stopped to be reorganized into longer or faster pipe. The duration (number of bytes to transfer) of a long piping is limited to the minimum of the requested number of bytes to transfer of all PipeRequests covered by the long piping.

### Candidate propagation

The propagation of pipe candidates may happen asynchronously. E.g. IdentityTransformStream does that. This is useful when an IdentityTransformStream already have some chunks enqueued in it. It may exert backpressure to the source readable stream of an existing pipe whose destination is the writable side of the IdentityTransformStream so that it temporarily stops write()-ing new chunks. Once all the chunks are drained, it can announce the piping candidates received at the readable side to the writable side to choose the faster one. I think this backpressure is not normal backpressure signal but should be made by announcing an empty pipeCandidates to completely stop the piping (readable to the writable of the IdentityTransformStream) and wait for the ongoing piping to stop. We need some mechanism to realize this. Without strict backpressure like this, it's possible the ongoing pipe write()s new chunks before seeing the updates pipe candidates including long piping and switch to it before the queued chunks are processed.

tyoshino:
Extended the API a bit to address the flushing issue I discussed in the last comment.

domenic:
Wow, perfect, thank you for the explanation! I understand this is not necessarily something for the spec, but I think it will be valuable to have written up somewhere that we can point people to as an example of how to use the spec's flexibility around piping to accomplish great things. It might be worth prototyping as well, if we have the time and energy; I'm not sure if that was your plan in this PR or if you were just using the .js code as a way to organize the thoughts.

So let me try to restate the actors in the system:

- **pipe requests** represent a pipeTo call, i.e. (readable, writable, options) tuples
- **pipes** represent actual data transfers between readable and writable streams
  - they include information on the transfer method, so they are roughly (readable, writable, transfer method) tuples.
- the **pipe manager** is responsible for finding a series of pipes that satisfy all ongoing pipe requests
  - it probably also is responsible for ensuring that no matter what pipes are in play, the options for each pipe _request_ are honored

With this in mind, the notifyPipeCandidates/waitPipeCandidatesChange mechanisms are all about coordinating to potentially update the set of pipes in action when the set of pipe requests in action change. This is what the reorganizing pipes section is about, I think?

I'm not sure I fully understand the candidate propagation section, or why it implies a new backpressure mechanism. My vision of identity transform streams (which is not fully thought out, so might be wrong) is that in a system like this, they get completely excluded from the list of pipes, i.e. `rs.pipeThrough(id).pipeTo(ws)` will create two pipe requests but only one pipe (from `rs` to `ws`). Then normal backpressure propagation would take place between `rs` and `ws`.

I guess you are concerned about cases like `rs.pipeThrough(id); setTimeout(() => id.readable.pipeTo(ws), 10000)`, when you talk about asynchronocity? Or is it something else?

tyoshino:
> I'm not sure if that was your plan in this PR or if you were just using the .js code as a way to organize the thoughts.

Yeah, initially I attempted to just implement the API into the ReadableStream, but it looked it's going to complicate the reference impl so much, so I wrote it in a separate file. I want to evolve the code snippet to something working for verifying our ideas. Yes.

> This is what the reorganizing pipes section is about, I think?

Right

> I guess you are concerned about cases like rs.pipeThrough(id); setTimeout(() => id.readable.pipeTo(ws), 10000), when you talk about asynchronocity? Or is it something else?

Ah, yeah. Asynchronous invocation of pipeTo()s and pending chunks.

IdentityTransformStream's readable side and writable side themselves work as ReadableStream and WritableStream, and they can be separately passed around. Even pipeThrough() may be given some writable / readable pair which are build without considering the transform streams concept at all. I'd like to confirm that this understanding of mine is correct, first. Or, do you have some idea in your mind that we brand check that a given pair of writable / readable to pipeThrough is "transform stream"?

So, the writable side of an identity transform may get write()-en manually even when it's exerting backpressure. Then, some chunks are queued inside the transform stream.

Once pipeTo() is invoked on both the readable side and the writable side of the transform stream with pending chunks, we can consider optimizing this pair of pipeTo()-s. But we need to have the transform stream flush the chunks before establishing such skipped transfer.

Hmm, we can also resolve this by having a requirement enforced on all pipeTo() implementations to stop writing when the normal backpressure signal is exerted as well as the current pipeTo() reference implementation is doing. Yeah, it may work.

domenic:
> IdentityTransformStream's readable side and writable side themselves work as ReadableStream and WritableStream, and they can be separately passed around. Even pipeThrough() may be given some writable / readable pair which are build without considering the transform streams concept at all. I'd like to confirm that this understanding of mine is correct, first. Or, do you have some idea in your mind that we brand check that a given pair of writable / readable to pipeThrough is "transform stream"?

I haven't thought this through fully. But my idea was that yes, we would add an unobservable brand check and fast path if we know it's a transform stream created with the TransformStream constructor (or with `TransformStream.identity()` or whatever).

> Once pipeTo() is invoked on both the readable side and the writable side of the transform stream with pending chunks, we can consider optimizing this pair of pipeTo()-s. But we need to have the transform stream flush the chunks before establishing such skipped transfer.

Got it. I understand the problem now.

> Hmm, we can also resolve this by having a requirement enforced on all pipeTo() implementations to stop writing when the normal backpressure signal is exerted as well as the current pipeTo() reference implementation is doing. Yeah, it may work.

The draft in https://github.com/whatwg/streams/pull/512 says "While WritableStreamDefaultWriterGetDesiredSize(writer) is ≤ 0 or is null, the user agent must not read from reader." I think this is what we want, although I guess it is talking about reading, not writing. Maybe it should say both.

domenic:
> IdentityTransformStream's readable side and writable side themselves work as ReadableStream and WritableStream, and they can be separately passed around. Even pipeThrough() may be given some writable / readable pair which are build without considering the transform streams concept at all. I'd like to confirm that this understanding of mine is correct, first. Or, do you have some idea in your mind that we brand check that a given pair of writable / readable to pipeThrough is "transform stream"?

I haven't thought this through fully. But my idea was that yes, we would add an unobservable brand check and fast path if we know it's a transform stream created with the TransformStream constructor (or with `TransformStream.identity()` or whatever).

> Once pipeTo() is invoked on both the readable side and the writable side of the transform stream with pending chunks, we can consider optimizing this pair of pipeTo()-s. But we need to have the transform stream flush the chunks before establishing such skipped transfer.

Got it. I understand the problem now.

> Hmm, we can also resolve this by having a requirement enforced on all pipeTo() implementations to stop writing when the normal backpressure signal is exerted as well as the current pipeTo() reference implementation is doing. Yeah, it may work.

The draft in https://github.com/whatwg/streams/pull/512 says "While WritableStreamDefaultWriterGetDesiredSize(writer) is ≤ 0 or is null, the user agent must not read from reader." I think this is what we want, although I guess it is talking about reading, not writing. Maybe it should say both.

tyoshino:
> we would add an unobservable brand check and fast path

I see. Anyway we need to perform some special interaction with the streams to realize the optimization. Even the propagation idea is requiring the streams to expose special signal / method.

One good thing of leaving the decision to the streams and have the global manager to fetch some predefined signal from them as done in the propagation idea above is that each transform stream can control the timing to announce longer piping. E.g. one could consume some initial header to do something special, and then pass through data to the next streams.

> Maybe it should say both.

Yeah! I'm not yet sure about necessity of different backpressure signals. Let's just try.

afriadishariff-lab (commented):
reference-implementation/lib/fast-pipe.js

Issue #476, opened 2016-07-06
Should write() always error the stream when it rejects the promise it returned?


tyoshino:
writer.closed rejects when some error occurs in the writable stream. And therefore, those who're not interested in timing of successful completion of each write() but only interested in any error event may write() and forget but just watch writer.closed.

BYOB read() method has a logic that validates the argument and returns a rejected promise without erroring the stream when the validation fails. If we introduce similar logic to write(), the write()-and-forget() described above won't work.

Is this implying that such no-erroring rejection is bad design?

domenic:
Hmm, interesting. I think the BYOB design is good... it tries to distinguish between two types of errors:

- This particular chunk is bad and I won't allow it.
- I tried to write, but doing so failed, and you probably should not try again; treat this as broken.

I guess right now we do not support this distinction very well in the underlying sink/controller API. For example, let's say you wanted to implement a writable stream that only accepted strings, or bytes, or something. I think it would be good to reject a particular write, without necessarily killing the entire writable stream. But right now all you can do is return a rejected promise, which will cause the stream to become errored.

I see a few options:

- Say that if you fail validation, we should just error the whole stream. Probably for BYOB read() too. You screwed up and you don't get to recover. This is nice and simple, at least.
- Allow separate validation and erroring capabilities.  I guess the most natural way to do this is to say that returning a rejected promise from write() does not error the stream, and you have to do that with a separate controller.error() call?

There is also a gray area. If a write fails, not because of the producer's fault, but because of some transient issue, then it is simpler to error the stream---we probably should not let them continue writing, as they might skip a chunk that's important.

Hmm, that kind of logic is moving me more toward the "error the whole stream" option. E.g. consider if you had a program error that somehow ended up with chunks [string1, string2, blob3, string4, string5]. Probably it would be better to write string1 + string2 then get an errored writable stream, than to write string1 + string2 + string4 + string5 and only get the error notification on the one write.

Also, part of this comes down to, what do the underlying sink APIs look like? Looking through http://man7.org/linux/man-pages/man2/write.2.html#ERRORS it seems to me like most of these are not recoverable.

So for writable streams error-forever may be better. For BYOB it seems a bit less clear since I don't think there's the possibility of that kind of skipping and of missing error notifications. So maybe the asymmetry is OK?

tyoshino:
[...]

PR #512, opened 2016-08-30
Rigorously specify and test pipeTo

domenic:
This is a first draft at the "algorithm" for pipeTo, which contains a few concrete steps and a lot of more general requirements. I'd love feedback on whether this style of speccing is clear enough, or if I missed any important requirements, or if the way in which I grouped the requirements is confusing, or anything else. I'm kind of nervous about this part :)

Preview at https://streams.spec.whatwg.org/branch-snapshots/spec-pipe-to-omg/#rs-pipe-to

Things to do before this is merge ready:

- [x] <del>Factor this out into a StreamPipe() abstract op that other specs could use. (ReadableStreamPipe??)</del> <ins>Let's wait until we have consumers first</ins>
- [x] Factor our releasing a writable stream lock into an abstract op
- [x] Ensure we have tests that cover all of the requirements here (and realign any existing ones to match better with the requirements' wording if they are close)
- [x] Ensure we are not missing any requirements that are expressed in the tests
- [x] Ensure in particular that the various error conditions are tested correctly and align with the spec, e.g. what if both the source and dest error.
- [x] Check to make sure no asserts are violated or incorrect errors are thrown/promises are rejected, e.g. by doing something that's not supposed to be done while in a given state.
- [x] Align the reference implementation more closely with this version of the spec, by using more internal methods and seeing if we can get better correspondence between the requirement lines here and the JS lines in the reference implementation. It's possible they can't be that close, which is fine, as long as the tests pass.


ricea:
I like this style. I find it easy to read and easy to write tests for.

index.bs
```diff
@@ -603,16 +603,66 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+       * If _dest_.[[state]] is `"closing"` or `"closed"` and _preventCancel_ is *false*,
+         1. Let _destClosed_ be a new *TypeError*.
+         1. Perform ! ReadableStreamCancel(*this*, _destClosed_).
+         1. If the resulting promise fulfills, perform error cleanup with _destClosed_. Otherwise, perform error cleanup
+            with the resulting promise rejection.
+       * If _dest_.[[state]] is `"closing"` or `"closed"` and _preventCancel_ is *true*, perform error cleanup with a
+         new *TypeError*.
+     * <i>Cleanup</i>: if any of the above requirements ask to perform cleanup, then:
+       * Wait until any outstanding reads or writes finish (i.e. the corresponding promises settle). If that causes one
+         of the above error conditions to be true, jump back to those requirements instead of proceeding with cleanup.
```
tyoshino: Depending on resolution of #476, we might need to mention the error situation where write() promise rejects but the destination writable stream doesn't get errored.

reference-implementation/test/readable-stream-templated.js
```diff
@@ -20,20 +19,6 @@ templatedRSClosed('ReadableStream (closed via cancel)',
 
 const theError = new Error('boo!');
 
-templatedRSErrored('ReadableStream (errored via call in start)',
```
tyoshino: you've chose not to port the instance for this line? I think it's ok if the basic functionality of controller.error() is tested somewhere else, but just to make sure you're aware of the removal.

tyoshino:
- Regarding that all the operations are performed by directly invoking the abstract operations:
  - I think this is great.
- The way the algorithm is specified:
  - Yeah, it's nice. The algorithm in the reference implementation is also written in the form reacting to each situation. We just have `state` to resolve race. The spec text is representing what we have well :)
- reviewed the code changes. left some comments but lgtm.

domenic:
Update: I made a lot of progress on this today. In the process I discovered some contradictions in the spec text as written, which was kind of expected. E.g. races between the various error conditions, or races between performing cleanup (now renamed shutdown) and finishing a write. The spec has been amended, and also the reference implementation has been rewritten to follow the spec.

There are still a number of failing (unconverted) tests, so I guess the algorithm is probably not quite right yet.

I was still not able to convert many tests because we have a lot of good tests that manage to put a stream in a given state asynchronously, possibly after some chunks have already passed through it. I am not sure yet whether I want to solve this in some kind of generative way (maybe like the templated tests) or just by porting them individually. I'll try to work on it tomorrow.

domenic:
This is allllmost done. Some TODOs listed in the commit message, all minor. Last time I said "tomorrow" in this thread that didn't work out, but I'm hopeful...

reference-implementation/to-upstream-wpts/resources/test-utils.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+self.getterRejects = (t, obj, getterName, target) => {
```
ricea: I would prefer assertGetterRejects(), to make it clear that this is an assertion.

I thought about assert_getter_rejects(), but then people might expect to find the definition in testharness.js.

I realise that testharness's "promise_rejects" doesn't include "assert", but I find that a bit counter-intuitive to be honest.

domenic: This whole file already exists upstream; I just am copying it here because of limitations of the test runner.

domenic:
This is done and ready for review! It's on top of #526 so the first two commits can be ignored.

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
+      function performShutdown(isError, error) {
+        shuttingDown = true;
 
-      Promise.all([_lastRead, _writer.ready]).then(
-        ([{ value, done }]) => {
-          if (_state !== 'piping') {
-            return;
-          }
+        waitForCurrentWrite().then(() => {
+          WritableStreamDefaultWriterRelease(writer);
```
tyoshino: Factor out L201-208 and call it from L191 and L192?

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
+          action().then(
+            () => performShutdown(originalIsError, originalError),
+            newError => performShutdown(true, newError)
+          );
+        });
+      }
 
-      _lastRead = _reader.read();
+      function performShutdown(isError, error) {
+        shuttingDown = true;
```
tyoshino: what's the reason why we can omit `if (shuttingDown == true`) here?

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
-    }
-
-    function doPipe() {
-      // console.log('pipeTo(): doPipe()');
+        shuttingDown = true;
+
+        waitForCurrentWrite().then(() => {
+          action().then(
+            () => performShutdown(originalIsError, originalError),
+            newError => performShutdown(true, newError)
```
tyoshino: shouldn't we take the original error if any?

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
-      // We don't use writer.closed. pipeTo() is responsible only for what it has written.
-      _lastWrite.then(
-        () => {
-          releaseWriter();
-          finishWithFulfillment();
-        },
-        reason => {
-          releaseWriter();
-          finishWithRejection(reason);
+      function performAsyncShutdown(action, originalIsError, originalError) {
```
tyoshino: Can we describe the difference between this and performShutdown in the name?

performShutdownWithAction?

reference-implementation/lib/writable-stream.js
```diff
@@ -359,6 +370,24 @@ function WritableStreamDefaultWriterClose(writer) {
...
 
+
+function WritableStreamDefaultWriterCloseWithErrorPropagation(writer) {
+  const stream = writer._ownerWritableStream;
+
+  assert(stream !== undefined);
+
+  const state = stream._state;
+  if (state === 'closed' || state === 'closing') {
+    return Promise.resolve();
```
tyoshino: Shouldn't this case handled as an error as it's already closed when we tried to propagate the close signal?

reference-implementation/lib/readable-stream.js
domenic: Sure, will rename.

reference-implementation/lib/readable-stream.js
domenic: Hmm. An interesting question, but I don't think so. This way makes sure both errors are seen by someone. The original error is seen by the action (abort or cancel) and then the error thrown (or propagated via a rejected promise) from the action is seen by the caller of pipeTo. I think that is best.

reference-implementation/lib/readable-stream.js
domenic: _shuttingDown_ might be true here, if we are coming from async shutdown. So that would be expected, not surprising.

_shuttingDown_ is basically meant to prevent a second async shutdown from happening if one is already in progress.

reference-implementation/lib/readable-stream.js
domenic: Done

reference-implementation/lib/writable-stream.js
domenic: Great catch. This is never hit, by design, so I changed it to an assert

reference-implementation/lib/writable-stream.js
```diff
@@ -60,6 +60,17 @@ class WritableStream {
 
 exports.WritableStream = WritableStream;
 
+Object.assign(exports, {
+  AcquireWritableStreamDefaultWriter,
+  IsWritableStream,
+  WritableStream,
+  WritableStreamAbort,
+  WritableStreamDefaultWriterClose,
```
isonmad: Why is WritableStreamDefaultWriterClose exported? readable-stream.js doesn't import it.

reference-implementation/lib/writable-stream.js
```diff
@@ -60,6 +60,17 @@ class WritableStream {
 
 exports.WritableStream = WritableStream;
```
isonmad: Is this redundant with the below assignment of WritableStream now?

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
-        ).then(
-          () => {
-            releaseWriter();
-            finishWithFulfillment();
-          }
-        );
-        _state = 'closingDest';
+      // Closing must be propagated backward
+      writer._closedPromise.then(() => {
+        const destClosed = new TypeError('the destination writable stream closed before all data could be piped to it');
```
isonmad: Since this can only happen if running pipeTo on a destination that was already closed beforehand, maybe this error should be more specific and say that's what happened? This can't possibly happen in the middle of piping, can it?

reference-implementation/lib/readable-stream.js
```diff
@@ -86,273 +89,126 @@ class ReadableStream {
...
+        new TypeError('ReadableStream.prototype.pipeTo\'s first argument must be a WritableStream'));
+    }
 
     preventClose = Boolean(preventClose);
     preventAbort = Boolean(preventAbort);
     preventCancel = Boolean(preventCancel);
 
-    const source = this;
+    const reader = AcquireReadableStreamDefaultReader(this);
+    const writer = AcquireWritableStreamDefaultWriter(dest);
```
isonmad: Why do the brand checks above return a rejected promise (instead of throwing synchronously), when `AcquireReadableStreamDefaultReader` and `AcquireWritableStreamDefaultWriter` might throw synchronously anyway if either stream is already locked? Shouldn't either all of these throw, or all of these be promises?

isonmad:
Couldn't you actually entirely replace the 'Closing must be propagated backward' code with a new abstract op `WritableStreamGetState(dest)` and a check whether it's `'closing'` or `'closed'` at the start of `pipeTo`, right after the brand checks? Unlike the other 3 propagation conditions, it can't ever suddenly happen after pipeTo starts.

reference-implementation/to-upstream-wpts/resources/rs-utils.js
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+  }
+
+  function randomChunk(size) {
+    let chunk = '';
+
+    for (let i = 0; i < size; ++i) {
+      // Add a random character from the basic printable ASCII set.
+      chunk += String.fromCharCode(Math.round(Math.random() * 84) + 32);
```
domenic: Agreed, but again this is a preexisting file upstream.

ricea: Sorry. I should have checked.

reference-implementation/lib/readable-stream.js
tyoshino: Oh, right. My comment was not clear about what's the problem I saw. I meant that it looks when shutdown() is invoked directly, not via shutdownWithAction(), we cannot prevent a second shutdown from happening if one is already in progress.

reference-implementation/lib/readable-stream.js
tyoshino: It's nice to propagate all the errors to someone, but it means that when preventCancel (or preventAbort) is not in use, the original cause of error is lost. It sounds bad.

domenic:
Updated for @isonmad's feedback. Thanks!

isonmad:
Why start the `pipeLoop` at all if the destination is a closed stream? What use case is there where someone wants to pipeTo to a closed writablestream?

domenic:
The use case is when you don't know ahead of time whether your destination is closed or not.

domenic:
@ricea @tyoshino any ETA on when you'll be able to review this?

tyoshino:
@domenic Oh, sorry. I was thinking that you've been thinking about what to do with my comments here https://github.com/whatwg/streams/pull/512#pullrequestreview-4242170

isonmad:
The only goal being to cancel the readable stream if it's piped to an already closed writable (if `preventCancel` is false)?

Can't that be done without involving the whole `pipeLoop`? It doesn't really need all the `shutdown` infrastructure does it.

tyoshino:
> @domenic Oh, sorry. I was thinking that you've been thinking about what to do with my comments here #512 (comment)

Ah, regarding https://github.com/whatwg/streams/pull/512/files/119449f4f6815d54f2c3d42dc439248f660da534#discussion_r83581544, I was confused. I just realized that when prevent.* is true, there's only one error we need to propagate. So, my comment should be fixed to "It sounds weird that the source (action() or the streams) of the error propagated to the pipePromise differs depending on the prevent.* flag.". As we can see the source clearly based on which value we passed to the prevent.* flag, I don't strongly object to this now, but some note needed to explain this behavior.

index.bs
```diff
@@ -598,16 +598,75 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+  1. Set _preventClose_ to ! ToBoolean(_preventClose_), set _preventAbort_ to ! ToBoolean(_preventAbort_), and set
+     _preventCancel_ to ! ToBoolean(_preventCancel_).
+  1. If ! IsReadableStreamLocked(*this*) is *true*, return a promise rejected with a *TypeError* exception.
+  1. If ! IsWritableStreamLocked(_dest_) is *true*, return a promise rejected with a *TypeError* exception.
+  1. If ! IsReadableByteStreamController(*this*.[[readableStreamController]]) is true, let _reader_ be either !
+     AcquireReadableStreamBYOBReader(*this*) or ! AcquireReadableStreamDefaultReader(*this*), at the user agent's
+     discretion.
+  1. Otherwise, let _reader_ be ! AcquireReadableStreamDefaultReader(*this*).
+  1. Let _writer_ be ! AcquireWritableStreamDefaultWriter(_dest_).
+  1. Let _promise_ be a new promise.
```
tyoshino: align the order with the reference implementation?

index.bs
```diff
@@ -598,16 +598,75 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+     to the locking provided by the reader and writer, the exact manner in which this happens is not observable to
+     author code, and so there is flexibility in how this is done. The following constraints apply regardless of the
+     exact algorithm used:
+     * <strong>Public API must not be used:</strong> while reading or writing, or performing any of the operations
+       below, the JavaScript-modifiable reader, writer, and stream APIs (i.e. methods on the appropriate prototypes)
+       must not be used. Instead, the streams must be manipulated directly.
+     * <strong>Backpressure must be enforced:</strong>
+       * While WritableStreamDefaultWriterGetDesiredSize(_writer_) is ≤ *0* or is *null*, the user agent must not read
+         from _reader_.
+       * If _reader_ is a <a>BYOB reader</a>, WritableStreamDefaultWriterGetDesiredSize(_writer_) should be used to
```
tyoshino: can we put some note here why we have the if-otherwise but recommending the same method?

index.bs
```diff
@@ -598,16 +598,75 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+     * <strong>Shutdown must stop all activity:</strong> if _shuttingDown_ becomes *true*, the user agent must not
+       initiate further reads from _reader_ or writes to _writer_. (Ongoing reads and writes may finish.)
+     * <strong>Errors must be propagated forward:</strong> if *this*.[[state]] is or becomes `"errored"`, then
+       1. If _preventAbort_ is *false*, <a href="#rs-pipeTo-shutdown-with-action">shutdown with an action</a> of !
+          WritableStreamAbort(_dest_, *this*.[[storedError]]) and with *this*.[[storedError]].
+       1. Otherwise, <a href="#rs-pipeTo-shutdown">shutdown</a> with *this*.[[storedError]].
+     * <strong>Errors must be propagated backward:</strong> if _dest_.[[state]] is or becomes `"errored"`, then
+       1. If _preventCancel_ is *false*, <a href="#rs-pipeTo-shutdown-with-action">shutdown with an action</a> of !
+          ReadableStreamCancel(*this*, _dest_.[[storedError]]) and with _dest_.[[storedError]].
+       1.  Otherwise, <a href="#rs-pipeTo-shutdown">shutdown</a> with _dest_.[[storedError]].
```
tyoshino: double space between `1.` and `Otherwise`

index.bs
```diff
@@ -598,16 +598,75 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+       1. If _preventCancel_ is *false*, <a href="#rs-pipeTo-shutdown-with-action">shutdown with an action</a> of !
+          ReadableStreamCancel(*this*, _destClosed_) and with _destClosed_.
+       1. Otherwise, <a href="#rs-pipeTo-shutdown">shutdown</a> with _destClosed_.
+     * <i id="rs-pipeTo-shutdown-with-action">Shutdown with an action</i>: if any of the above requirements ask to
+       shutdown with an action _action_, optionally with an error _originalError_, then:
+       1. If _shuttingDown_ is *true*, abort these substeps.
+       1. Set _shuttingDown_ to *true*.
+       1. Wait until any ongoing write finishes (i.e. the corresponding promises settle).
+       1. Let _p_ be the result of performing _action_.
+       1. If _p_ fulfills, <a href="#rs-pipeTo-finalize">finalize</a>, passing along _originalError_ if it was given.
```
tyoshino: use the wording `Upon fulfillment` for consistency with other part of the spec?

index.bs
```diff
@@ -3001,6 +3060,22 @@ nothrow>WritableStreamDefaultWriterClose ( <var>writer</var> )</h4>
...
+<p class="note">This abstract operation helps implement the error propagation semantics of
+{{ReadableStream/pipeTo()}}.</p>
+
+<emu-alg>
+  1. Let _stream_ be _writer_.[[ownerWritableStream]].
+  1. Assert: _stream_ is not *undefined*.
+  1. Let _state_ be _stream_.[[state]].
+  1. Assert: _state_ is not `"closing"` or `"closed"`.
+  1. If _state_ is `"errored"`, return a new promise rejected with _stream_.[[storedError]].
+  1. Assert: _state_ is `"writable"`.
```
tyoshino: add this assert to the reference implementation?

tyoshino:
Regarding https://github.com/whatwg/streams/pull/512/files/119449f4f6815d54f2c3d42dc439248f660da534#discussion_r83581051, please add some tests where two or more of prevent.* flags are set to true.

I've finished reviewing the newly added wpt test cases, spec text and reference impl. I'll check that all the existing tests that are worth keeping are converted into the wpt.

index.bs
domenic: done

index.bs
domenic: Sorry, I don't quite understand. What were you thinking of adding?

index.bs
domenic: fixed

index.bs
domenic: fixed

index.bs
domenic: done

reference-implementation/to-upstream-wpts/piping/close-propagation-forward.js
```diff
@@ -0,0 +1,378 @@
...
+    assert_array_equals(rs.eventsWithoutPulls, []);
+    assert_array_equals(ws.events, ['close']);
+
+    return Promise.all([
+      rs.getReader().closed,
+      ws.getWriter().closed
+    ]);
+  });
+
+}, 'Closing must be propagated forward: becomes closed while empty; preventClose omitted; fulfilled close promise');
```
domenic: Ah yeah, good point, changed

reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
```diff
@@ -0,0 +1,577 @@
...
+      writer.releaseLock();
+
+      return promise_rejects(t, error1, rs.pipeTo(ws), 'pipeTo must reject with the write error')
+        .then(() => {
+          assert_array_equals(rs.eventsWithoutPulls, ['cancel', error1]);
+          assert_array_equals(ws.events, ['write', 'Hello']);
+        });
+    });
+
+}, 'Errors must be propagated backward: becomes errored before piping due to write; preventCancel omitted; ' +
```
domenic: I think it is pretty redundant, but probably it already existed in some form in the existing tests, so I thought it would be good to port it and keep the coverage.

reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
```diff
@@ -0,0 +1,577 @@
...
+    write(chunk) {
+      if (chunk === 'c') {
+        return Promise.reject(error1);
+      }
+      return undefined;
+    }
+  });
+
+  return promise_rejects(t, error1, rs.pipeTo(ws, { preventCancel: true }),
+                                      'pipeTo must reject with the same error')
```
domenic: fixed

reference-implementation/to-upstream-wpts/piping/general.js
```diff
@@ -0,0 +1,156 @@
...
+promise_test(t => {
+
+  const rs = new ReadableStream();
+  const ws = new WritableStream();
+
+  ws.getWriter();
+
+  assert_false(rs.locked, 'sanity check: the ReadableStream does not start locked');
+  assert_true(ws.locked, 'sanity check: the WritableStream starts locked');
+
```
domenic: fixed

domenic:
> Can't that be done without involving the whole pipeLoop? It doesn't really need all the shutdown infrastructure does it.

It could. I find the spec and reference implementation cleaner this way.

> As we can see the source clearly based on which value we passed to the prevent.* flag, I don't strongly object to this now, but some note needed to explain this behavior.

Added some bullet points

> please add some tests where two or more of prevent.* flags are set to true.

Hmm. Given that only one flag can be used in any given pipeTo, I'm not sure how valuable this would be. Could you give an example of what you'd be hoping to test?

Although, I guess I just realized we do not have any tests for multiple close/error conditions being true at once :(. So e.g. piping a closed readable stream to an errored writable stream and so on. The spec isn't even super clear about what takes precedence there. I guess that's another reasonably large chunk of work to do. I'll try to do it Monday.

tyoshino:
> Hmm. Given that only one flag can be used in any given pipeTo, I'm not sure how valuable this would be. Could you give an example of what you'd be hoping to test?

E.g. please try simply adding `preventCancel: true` to the `Errors must be propagated forward: starts errored; preventAbort = ${String(truthy)} (truthy)` test in `error-propagation-forward.js` and see how many times `shutdown()` is invoked. It's invoked twice and `finalize()` is also invoked twice. The second `finalize()` fails harmlessly on the redundant call of WritableStreamDefaultWriterRelease() which throws. It's not affecting the publicly visible behavior, but I think we should suppress the redundant invocation of `finalize()`.

That's the background of my comment at https://github.com/whatwg/streams/pull/512#discussion_r83581051.

In general, I think it's worth testing some cases where two or three of prevent.* flags are set to a truthy value (though no need to have full coverage) as it's possible and should be common to do so, in addition to what you've done which is, of course, needed and important.

> Although, I guess I just realized we do not have any tests for multiple close/error conditions being true at once :(. So e.g. piping a closed readable stream to an errored writable stream and so on. The spec isn't even super clear about what takes precedence there. I guess that's another reasonably large chunk of work to do. I'll try to do it Monday.

Agreed.

reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
tyoshino: Got it. Thanks!

index.bs
tyoshino: Ohh, never mind. Sorry.

domenic:
> It's not affecting the publicly visible behavior, but I think we should suppress the redundant invocation of finalize().

Why? The entire setup is designed specifically so that it's harmless and doesn't affect public behavior; we shouldn't need to do any extra bookkeeping for suppression.

tyoshino:
> Why? The entire setup is designed specifically so that it's harmless and doesn't affect public behavior; we shouldn't need to do any extra bookkeeping for suppression.

Okay. I still think we should have the if-clause and some tests with multiple prevent.* set to true, but it shouldn't block this from landing. lgtm. I'll open an issue for that.

tyoshino:
> I'll open an issue for that.

Created. Let's discuss this there.
https://github.com/whatwg/streams/issues/557

ricea:
lgtm based on my own superficial review and tyoshino's in-depth lgtm.

domenic:
I have to leave early today unfortunately. I want to get some good tests for multiple error/closure scenarios before merging this. So I've pushed up some WIP ones that are mostly failing. The tests might be bad or the spec or the logic; not sure yet. No need to pay them much attention unless you have extra free time. Just wanted to give an update.

I think we can also resolve https://github.com/whatwg/streams/issues/557 before merging this too.

PR #513, opened 2016-08-30
Factor out WritableStreamDefaultWriterRelease

domenic:
This will help with specifying pipeTo, as seen in #512. This also aligns the spec, tests, and reference implementation on sharing a single error object across both closed and ready promises; previously this was untested and the reference implementation did so but the spec did not.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2953,15 +2953,7 @@ lt="WritableStreamDefaultWriter(stream)">new WritableStreamDefaultWriter(<var>st
   1. Let _stream_ be *this*.[[ownerWritableStream]].
   1. If _stream_ is *undefined*, return *undefined*.
   1. Assert: _stream_.[[writer]] is not *undefined*.
-  1. Let _state_ be _stream_.[[state]].
-  1. If _state_ is `"writable"` or `"closing"`, reject _writer_.[[closedPromise]] with a *TypeError* exception.
-  1. Otherwise, set _writer_.[[closedPromise]] to a new promise rejected with a *TypeError* exception.
-  1. If _state_ is `"writable"` and !
-     WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, reject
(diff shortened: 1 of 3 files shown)


domenic:
Rebased and moved the test to WPT format; would appreciate a quick review :)

reference-implementation/to-upstream-wpts/writable-streams/general.js
```diff
@@ -80,3 +80,22 @@ test(() => {
...
   );
 }, 'ws.getWriter() on an errored WritableStream');
+
+promise_test(() => {
+  const ws = new WritableStream({});
+
+  const writer = ws.getWriter();
+  writer.releaseLock();
+
+  return writer.closed.then(
```
ricea: Maybe use the weird promise_rejects() function from testharness.js? Something like

return promise_rejects(t, 'TypeError', writer.closed, 'closed promise rejects')
    .then(promise_rejects(t, 'TypeError', writer.ready, 'ready promise rejects');

domenic: Yeah, but then I would have to write about this much code to test that they're equal (line 96), so I'm not sure it gains much :-/.

index.bs
```diff
@@ -3032,6 +3024,26 @@ nothrow>WritableStreamDefaultWriterGetDesiredSize ( <var>writer</var> )</h4>
   1. Return ! WritableStreamDefaultControllerGetDesiredSize(_stream_.[[writableStreamController]]).
 </emu-alg>
 
+<h4 id="writable-stream-default-writer-release" aoid="WritableStreamDefaultWriterRelease"
+nothrow>WritableStreamDefaultWriterRelease ( <var>writer</var> )</h4>
+
+
```
tyoshino: single vertical space?

tyoshino:
lgtm. Sorry for delay!!!

PR #514, opened 2016-08-31
lint glob should exclude Emacs backup files

tyoshino:


tyoshino:
Sorry, I should have chosen to just limit the files to grab to .js files. Thanks for reviewing this and upstreaming the question, anyway, Domenic.

Issue #518, opened 2016-09-16
Pass controller to transformer methods in TransformStream


ricea:
Currently methods of the transformer object are passed individual parameters for each action they can take. For example, the transform method is called like this:

```javascript
transform(chunk, done, enqueue, closeReadable, error)
```

where `done`, `enqueue`, `closeReadable` and `error` are all functions that `transform()` can call in processing `chunk`.

For consistency with ReadableStream and WritableStream, it would be better if the methods were passed a controller object. The definition would be something like

```javascript
class TransformStreamDefaultController {
  constructor(stream)

  done()
  enqueue(chunk)
  closeReadable()
  error()
}
```

In a Promise-based implementation `done()` and `error()` would go away.

Then the signature of `transform()` would change to

```javascript
transform(chunk, controller)
```

isonmad:
Since WritableStream only passes the controller to `start()`, and since the transformer can similarly store the controller for later use, does `transform()` need a second parameter at all?

ricea:
My view is that someone has to store the controller. If the TransformStream does it, then that code only exists in one place. If the transformer does it, then every transformer implementation has to include code to store the controller, including implementing a start() method even if they didn't need one.

Issue #498, opened 2016-08-04
Argument order of transform() function (or attempt to remove done)


tyoshino:
https://github.com/whatwg/streams/pull/488#discussion_r72941214

Issue #331, opened 2015-04-15
TransformStream: There's no means available to fail transform() asynchronously


tyoshino:
An exception thrown by `transform()` is caught to error both `readable` and `writable`. But if `transform()` wants to do something asynchronous that may fail, we cannot tell the TransformStream the failure. `transformDone` just tells it to retrieve the next chunk.

tyoshino:
Related: #185

domenic:
Yeah I think fixing #185 is the right fix for this.

PR #519, opened 2016-09-16
Update TransformStream API & misc. fixups

isonmad:
First pass at issue #518 to align with the other stream APIs. Ended up also addressing #498, #331, and #185.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -44,11 +44,10 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
   try {
     controller.enqueue(chunk);
   } catch (e) {
-    if (transformStream._errored === false) {
-      // This happens when the given strategy is bad.
-      const reason = new TypeError('Failed to enqueue to readable side');
-      TransformStreamErrorInternal(transformStream, reason);
-    }
+    // This happens when the given strategy is bad.
+    const reason = new TypeError('Failed to enqueue to readable side');
+    TransformStreamErrorIfNeeded(transformStream, reason);
+
     throw transformStream._storedError;
   }
 
@@ -56,10 +55,9 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
   try {
     backpressure = controller.desiredSize <= 0;
   } catch (e) {
-    if (transformStream._errored === false) {
-      const reason = new TypeError('Failed to calculate backpressure of readable side');
-      TransformStreamError(transformStream, reason);
-    }
+    const reason = new TypeError('Failed to calculate backpressure of readable side');
+    TransformStreamErrorIfNeeded(transformStream, reason);
+
     throw transformStream._storedError;
   }
 
@@ -101,8 +99,16 @@ function TransformStreamChunkDone(transformStream) {
 
 // Abstract operations.
 
+function TransformStreamErrorIfNeeded(transformStream, e) {
+  if (transformStream._errored === false) {
+    TransformStreamErrorInternal(transformStream, e);
+  }
+}
+
 function TransformStreamErrorInternal(transformStream, e) {
   // console.log('TransformStreamErrorInternal()');
+
+  assert(transformStream._errored === false);
 
   transformStream._errored = true;
   transformStream._storedError = e;
@@ -154,9 +160,7 @@ function TransformStreamTransformIfNeeded(transformStream) {
           transformStream._errorFunction);
     }
   } catch (e) {
-    if (transformStream._errored === false) {
-      TransformStreamErrorInternal(transformStream, e);
-    }
+    TransformStreamErrorIfNeeded(transformStream, e);
   }
 }
 
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -319,6 +319,9 @@ module.exports = class TransformStream {
 
     this.readable = new ReadableStream(source, transformer.readableStrategy);
 
+    assert(this._writableController !== undefined);
+    assert(this._readableController !== undefined);
+
     const transformStream = this;
     const startResult = InvokeOrNoop(transformer, 'start', [transformStream._controller]);
     startPromise_resolve(startResult);
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -155,9 +155,7 @@ function TransformStreamTransformIfNeeded(transformStream) {
       transformStream._transformer.transform(
           chunk,
           TransformStreamChunkDone.bind(undefined, transformStream),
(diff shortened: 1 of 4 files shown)


ricea:
If I understand correctly, this doesn't enforce any ordering guarantees between the start/transform/flush methods. I would like to have the guarantee that transform() is not called until start()'s Promise resolves (and not at all if it rejects), and flush() is not called until transform()'s Promise resolves.

The reason is that if transform() is called before the Promise returned by start() resolves, then a call to enqueue() by transform() could end up happening before before a call to enqueue() by start(). Similarly with transform() and flush().

It's possible to enforce the ordering within the transformer, but that seems likely to cause a steep learning curve as every developer rediscovers the issue.

I can file a separate issue for this if you like.

reference-implementation/lib/transform-stream.js
```diff
@@ -229,25 +229,15 @@ class TransformStreamSink {
...
-      } catch (e) {
-        if (transformStream._errored === false) {
-          TransformStreamErrorInternal(transformStream, e);
-          throw e;
-        }
-      }
-    }
+    const flushPromise = PromiseInvokeOrNoop(transformStream._transformer, 'flush', [transformStream._controller]);
+    // Return a promise that is fulfilled with undefined on success.
+    return flushPromise.then(() => TransformStreamCloseReadableInternal(transformStream),
```
tyoshino: What if controller.error() is invoked inside flush() but flush() resolved? I think this path should perform TransformStreamCloseReadableInternal() when the transform stream is neither errored nor closed.

reference-implementation/lib/transform-stream.js
```diff
@@ -16,6 +17,17 @@ function TransformStreamCloseReadable(transformStream) {
     throw new TypeError('Readable side is already closed');
   }
 
+  if (transformStream._writableDone === true) {
+    throw new TypeError('TransformStream is already closing');
+  }
```
tyoshino: Did you try to limit the way to close the readable side to only by resolving flush() once flush() is invoked?  What's the motivation for that?

isonmad:
> I would like to have the guarantee that transform() is not called until start()'s Promise resolves (and not at all if it rejects), and flush() is not called until transform()'s Promise resolves.

`start()` doesn't return a promise yet, but adding one with that invariant makes sense. But flush() is only called from sink.close(), and that's already guaranteed not to be called until the last sink.write() promise resolves, which is only resolved when a transform() promise resolves, if I understand it all correctly.

reference-implementation/lib/transform-stream.js
isonmad: Having two redundant ways to perform the same action seemed dubious (especially if, as in your other comment, the result is potentially contradictory, instead of redundant). It's already an error to call `close()` twice, this seemed roughly analogous to that.

reference-implementation/lib/transform-stream.js
isonmad: You're right, that should be fixed, though I'm tempted to make `controller.error()` throw after `flush()` is called, like with `controller.close()`.

Otherwise, what should happen if controller.error() is called with one error, and flush throws or rejects the promise with a different error? What should happen if controller.close() is called, but then the flushPromise is rejected? It seemed simpler to give flush only a single way of signaling success or failure.

tyoshino:
Sorry for delay. I'll review this soon.

reference-implementation/test/transform-stream.js
```diff
@@ -372,3 +407,118 @@ test('Transform stream should call transformer methods as methods', t => {
...
+    }
+  });
+
+  t.ok(startCalled, 'start is called synchronously');
+
+  const writer = ts.writable.getWriter();
+  writer.write('a');
+  writer.close().then(() => {
+    t.ok(flushDone, 'flushPromise resolved');
+    t.end();
```
tyoshino: unnecessary as you have t.plan()

reference-implementation/lib/transform-stream.js
tyoshino: I understand your point. But we could also say that close() becoming disallowed in flush() would make the API a little complicated. Regarding the priority concern, I think we should make the transform stream accept the signal which arrived to it earliest. That said, this kind of rule also complicates the API.

Another question I'm not yet fully sure about is whether the resolution of flushPromise which you've changed to be propagated to writer.close() and readable's final state should always match or not. If close() / error() are allowed inside flush(), we can e.g. close() readable side while making writer.close() rejected. My gut feeling is basically they should match each other, but it seems worth being discussed a bit.

tyoshino:
lgtm. thanks!

tyoshino:
I'd like to add your name to the acknowledgement section of the spec. Do you want to use `isonmad` or some other name?

I'm going to squash the commits and merge it. If you want to write the commit summary for the squashed commit, please let me know. I can also write it for you if you want.

domenic:
Just wanted to say thank you @isonmad for all this work! It's really excellent to get this kind of contribution.

I guess we can merge this now with "isonmad" in the acknowledgments, but do feel free to let us know what would be better!

isonmad:
That works fine. If it has to all be one monolithic commit, I guess the commit message would be

    move TransformStream to promises based api (#519)
    
    Fix a bunch of bugs along the way and add tests for them.
    
    Closes #518, #498, #331, and #185.

tyoshino:
Thanks so much for your work, @isonmad. I've added your name to the acknowledgement section in https://github.com/whatwg/streams/commit/294d720458f1f560fc9b748146e9108a2c28a014

I'm going to make some follow up commits.

reference-implementation/lib/transform-stream.js
```diff
@@ -181,12 +183,14 @@ class TransformStreamSink {
...
 
     assert(transformStream._resolveWrite === undefined);
+    assert(transformStream._rejectWrite === undefined);
 
     transformStream._chunkPending = true;
     transformStream._chunk = chunk;
 
     const promise = new Promise(resolve => {
       transformStream._resolveWrite = resolve;
+      transformStream._rejectWrite = resolve;
```
tyoshino: this should be set to the rejection callback of the promise

reference-implementation/lib/transform-stream.js
```diff
@@ -121,8 +123,8 @@ function TransformStreamErrorInternal(transformStream, e) {
 
   transformStream._chunk = undefined;
 
-  if (transformStream._resolveWriter !== undefined) {
-    transformStream._resolveWriter(undefined);
+  if (transformStream._rejectWriter !== undefined) {
+    transformStream._rejectWriter(e);
```
tyoshino: s/rejectWriter/rejectWrite/

anyway, I just noticed that this is unnecessary since we call _writableController.error() at L118. I'm going to just remove this in #533

PR #521, opened 2016-09-22
fix ReadableStream spec inconsistencies

isonmad:
Erred on the side of adding asserts rather than removing them.


domenic:
Thank you so much! Merged as 3fd124677939ffa65a8ff801249c9cd4771bb6be, 4be6fd1ff79ec26934b816e31c08a97cafaff38b, and 7ba2589db8bfdceaeba9585b9e66815ffcfd3379. What name can we list you under in the Acknowledgments? Is "isonmad" preferred?

PR #523, opened 2016-10-05
Change async test to a promise_test

ricea:
The test 'ws.getWriter() on an errored WritableStream' was using the
test() function. However, since it is an asynchronous test it needs to
use the promise_test() function. Fix it.


tyoshino:
lgtm. thanks!

PR #524, opened 2016-10-05
Port constructor tests to web-platform-tests

ricea:
Move tests relating to the WritableStream constructor and
newly-constructed objects to the testharness.js framework.

Only syntactic and style changes have been made in this CL. There are no
semantic changes, except that some property existence checks have been
changed to use assert_idl_attribute(), which is stricter. Also where the
original test verified that a promise resolved, the ported version just
returns the promise and lets the test harness check it.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/constructor.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/constructor.js
@@ -1,0 +1,73 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+}
+
+promise_test(() => {
+  let controller;
+  const ws = new WritableStream({
+    start(c) {
+      controller = c;
+    }
+  });
+
+  // Now error the stream after its construction.
+  const passedError = new Error('horrible things');
+  controller.error(passedError);
+
+  const writer = ws.getWriter();
+
+  assert_equals(writer.desiredSize, null, 'desiredSize should be null');
+  return writer.closed.catch(r => {
+    assert_equals(r, passedError, 'ws should be errored by passedError');
+  });
+}, 'controller argument should be passed to start method');
+
+promise_test(() => {
+  const ws = new WritableStream({}, {
+    highWaterMark: 1000,
+    size() { return 1; }
+  });
+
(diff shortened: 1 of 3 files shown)


reference-implementation/to-upstream-wpts/writable-streams/constructor.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const writer = ws.getWriter();
+
+  assert_equals(typeof writer.write, 'function', 'writer should have a write method');
+  assert_equals(typeof writer.abort, 'function', 'writer should have an abort method');
+  assert_equals(typeof writer.close, 'function', 'writer should have a close method');
+
+  assert_equals(writer.desiredSize, 1, 'desiredSize should start at 1');
+
+  assert_idl_attribute(writer, 'ready', 'writer should have a ready attribute');
```
domenic: Since these technically aren't IDL attributes, it's probably misleading to use assert_idl_attribute. I'm a bit surprised the test passes.

reference-implementation/to-upstream-wpts/writable-streams/constructor.js
```diff
@@ -0,0 +1,73 @@
...
+    size() { return 1; }
+  });
+
+  const writer = ws.getWriter();
+
+  assert_equals(writer.desiredSize, 1000, 'desiredSize should be 1000');
+  return writer.ready.then(v => {
+    assert_equals(v, undefined, 'ready promise should fulfill with undefined');
+  });
+}, 'highWaterMark should be copied to desiredSize');
```
tyoshino: how about s/copied/reflected/?

reference-implementation/to-upstream-wpts/writable-streams/constructor.js
ricea: I decided that they were morally equivalent. assert_idl_attribute() is basically a check that the property exists on the prototype.

For the time being I have weaken the check to just require that the property is defined. The original test checked that the property was truthy, but that is unnatural in testharness.js. assert_true requires an actual boolean, not just truthiness, so you'd have to write assert_true(Boolean(writer.ready)) which feels awkward.

PR #526, opened 2016-10-06
Reject writer.ready promise on errors, instead of fulfilling

domenic:
In the process of testing this, ported several tests to web-platform-tests format, and introduced the recording stream helpers that I am using extensively in #512.

---

Apologies if we discussed this before and came to a good reason for this behavior; I couldn't find anything by searching. Anyway, this seems much more intuitive to me... the previous behavior caused some of my pipe code to fail since I wrote the code expecting that ready would only be fulfilled when it was safe to write.

PR #527, opened 2016-10-07
Port write() tests to web-platform-tests

ricea:
Move tests relating to the WritableStream sink's stream() method to the
testharness.js framework.

To reduce the risk of functional changes, changes to the tests are
intentionally minimal. The only kinds of changes are:

* Use testharness.js syntax
* Match web-platform-tests style
* Use promise_rejects() instead of explicitly checking .then()
* Use assert_equals() instead of if(!equal) fail()


domenic:
Can you convert the async_tests to promise_tests? async_test is unreliable in general (it runs in parallel instead of sequentially) and ugly to write and read. You can use the delay() helper landing in #526/already in WPTs instead of setTimeout.

Otherwise LGTM I guess; given the intentional minimalism there's not too much to review.

ricea:
Sorry, I should have said explicitly that the use of async_test was temporary. It's really hard to review the changes when a test is moved between files like this.

PR #528, opened 2016-10-10
Update ReadableStreamTee to allow cloning for only the second branch

domenic:
This is part of fixing https://github.com/yutakahirano/fetch-with-streams/issues/67.

/cc @youennf @yutakahirano. Because of how I changed this (just modifying the meaning of the second arg being true) this should require no subsequent changes to Fetch.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -706,21 +706,26 @@ readable stream is <a>locked to a reader</a>.
 </emu-alg>
 
 <h4 id="readable-stream-tee" aoid="ReadableStreamTee" throws>ReadableStreamTee ( <var>stream</var>,
-<var>shouldClone</var> )</h4>
+<var>cloneForBranch2</var> )</h4>
 
-This abstract operation is meant to be called from other specifications that may wish to
-<a lt="tee a readable stream">tee</a> a given readable stream. Its second argument governs whether or not the data from
-the original stream will be <a abstract-op lt="StructuredClone">structured cloned</a> before becoming visible in the
-returned branches. [[!HTML]]
+This abstract operation is meant to be called from other specifications that may wish to <a lt="tee a readable
+stream">tee</a> a given readable stream.
+
+The second argument, <var>cloneForBranch2</var>, governs whether or not the data from the original stream will be <a
+abstract-op lt="StructuredClone">structured cloned</a> before appearing in the second of the returned branches. This is
+useful for scenarios where both branches are to be consumed in such a way that they might otherwise interfere with each
+other, such as by <a abstract-op lt="Transfer">transfering</a> their <a>chunks</a>. However, it does introduce a
+noticable asymmetry between the two branches. [[!HTML]]
 
 <emu-alg>
   1. Assert: ! IsReadableStream(_stream_) is *true*.
-  1. Assert: Type(_shouldClone_) is Boolean.
+  1. Assert: Type(_cloneForBranch2_) is Boolean.
   1. Let _reader_ be ? AcquireReadableStreamDefaultReader(_stream_).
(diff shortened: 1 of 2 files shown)


youennf:
That seems ok as a minimal change.
I would like to understand though why a direct clone method is not possible/appropriate.
The other naive question I have is whether we can do so something better in the case the source is mandated to be a byte stream source.

reference-implementation/lib/readable-stream.js
```diff
@@ -488,23 +488,20 @@ function create_ReadableStreamTeePullFunction() {
...
 
+      const value1 = value;
+      const value2 = value;
+
       // There is no way to access the cloning code right now in the reference implementation.
       // If we add one then we'll need an implementation for StructuredClone.
-
+      // if (teeState.canceled2 === false && cloneForBranch2 === true) {
+      //   value2 = StructuredClone(value2);
+      // }
```
tyoshino: Even when the branch 1 has been cancelled, we clone values. I think it's ok and is more consistent, but leaving this comment to make sure you're aware of.

index.bs
```diff
@@ -782,14 +769,11 @@ ReadableStreamTee pull function <var>F</var> is called, it performs the followin
...
-      1. Perform ? ReadableStreamDefaultControllerEnqueue(_branch1_, _value1_).
-    1. If _teeState_.[[canceled2]] is *false*,
-      1. Let _value2_ be _value_.
-      1. If _shouldClone_ is *true*, set _value2_ to ? <a abstract-op>StructuredClone</a>(_value_).
-      1. Perform ? ReadableStreamDefaultControllerEnqueue(_branch2_, _value2_).
+    1. Let _value1_ and _value2_ be _value_.
+    1. If _teeState_.[[canceled2]] is *false* and _cloneForBranch2_ is *true*, set _value2_ to ? <a
+       abstract-op>StructuredClone</a>(_value2_).
+    1. If _teeState_.[[canceled1]] is *false*, perform ? ReadableStreamDefaultControllerEnqueue(_branch1_, _value1_).
+    1. If _teeState_.[[canceled2]] is *false*, perform ? ReadableStreamDefaultControllerEnqueue(_branch1_, _value2_).
```
tyoshino: _branch2_

index.bs
```diff
@@ -28,11 +28,6 @@ Ignored Vars: e
 </pre>
 
 <pre class="anchors">
-urlPrefix: https://html.spec.whatwg.org/multipage/; spec: HTML
-    text: StructuredClone; urlPrefix: infrastructure.html; type: abstract-op
-    text: Transfer; url: infrastructure.html#transfer-abstract-op; type: abstract-op
-    text: WebSocket; url: comms.html#websocket; type: interface
-    text: transferable objects; url: infrastructure.html; type: dfn
```
tyoshino: what made these unnecessary?

domenic:
> I would like to understand though why a direct clone method is not possible/appropriate.

To read the data from a readable stream, you must necessarily remove it from the stream's internal queue. So there is no real way to create a clone; the very act of reading is destructive.

reference-implementation/lib/readable-stream.js
domenic: Ah, that is interesting; thanks for pointing it out. It is probably better this way, yeah.

index.bs
domenic: Recent changes to the HTML spec to export these terms automatically instead of requiring manual import.

domenic: I pulled this and some other changes that are unrelated to teeing out into https://github.com/whatwg/streams/commit/0dc890351d298e288161c991582d998a9d5e5f51 (committed to master)

index.bs
domenic: Thanks, fixed

tyoshino:
3a95925 lgtm :)

yutakahirano:
lgtm

PR #531, opened 2016-10-11
Port some TransformStream tests to wpt

marvinhagemeister:
This includes constructor, flush and general tests. I'm new to the web platform test suite, and was wondering if we have some kind of `deepEqual` assertion? I didn't find one and as fallback compared each object property with its own assertion.


sideshowbarker:
>  was wondering if we have some kind of `deepEqual` assertion? I didn't find one 

The harness has an undocumented `assert_object_equals` but it’s deprecated so shouldn't be used.

> and as fallback compared each object property with its own assertion

That’s the preferred way as far as project policy goes. So I think you already did the right thing.

(All said not knowing exactly what `deepEqual` is but assuming it’s for testing object equivalence.)

domenic:
@marvinhagemeister, this is really helpful! However, I'm afraid this might be in conflict with #519, which is going to be merged very soon :(. A more stable target for test porting would be #492, where you could coordinate with @ricea on which _writable_ stream tests still need to be converted.

marvinhagemeister:
@domenic Ah shoot! I missed that. Are some tests which are safe to port?

marvinhagemeister:
@domenic I think the CI setup needs some love. There is a failing testcase but it doesn't fail the build

domenic:
> Are some tests which are safe to port?

Mostly the writable stream tests. There the potential conflict is that @ricea is working on porting them too, but as long as you both use #492 to coordinate which ones you intend to work on, it should be safe.

(Most piping-related tests are ported in #512.)

> I think the CI setup needs some love. There is a failing testcase but it doesn't fail the build

Eek, that's not good either. I'll try to figure that out too.

reference-implementation/to-upstream-wpts/transform-stream/constructor.https.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
domenic: Let's merge the constructor tests into the general tests

reference-implementation/to-upstream-wpts/transform-stream/constructor.js
```diff
@@ -0,0 +1,15 @@
...
+  self.importScripts('/resources/testharness.js');
+}
+
+test(() => {
+  const ts = new TransformStream({ transform() { } });
+}, 'TransformStream can be constructed with a transform function');
+
+test(() => {
+  const error = new TypeError('something bad');
+  assert_throws(error, () => new TransformStream(), 'TransformStream cannot be constructed with no arguments');
```
domenic: Nit: Use `assert_throws(new TypeError(), ...)` here instead of constructing an error with an arbitrary string.

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+// TODO
+test(() => {
+  let flushCalled = false;
+  const ts = new TransformStream({
+    transform() {
+      return new Promise(resolve => setTimeout(resolve, 10));
+    },
+    flush() {
+      flushCalled = true;
+      return new Promise(); // never resolves
```
domenic: This is not valid promise usage; it should be `new Promise(() => {})`. A preexisting problem I guess!

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+  assert_false(flushCalled, 'closing the writable does not immediately call flush if writes are not finished');
+
+  let rsClosed = false;
+  ts.readable.getReader().closed.then(() => {
+    rsClosed = true;
+  });
+
+  setTimeout(() => {
+    assert_true(flushCalled, 'flush is eventually called');
+    assert_equals(rsClosed, false, 'if flushPromise does not resolve, the readable does not become closed');
```
domenic: You can use `assert_false` here.

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+      c.enqueue('y');
+    }
+  });
+
+  const reader = ts.readable.getReader();
+
+  const writer = ts.writable.getWriter();
+  writer.write('a');
+  writer.close();
+  reader.read().then(result1 => {
```
domenic: return the promise here, and remove the `.catch(e => t.error(e))` (`t` is not defined)

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+  writer.write('a');
+  writer.close();
+  assert_false(flushCalled, 'closing the writable does not immediately call flush if writes are not finished');
+
+  let rsClosed = false;
+  ts.readable.getReader().closed.then(() => {
+    rsClosed = true;
+  });
+
+  setTimeout(() => {
```
domenic: Instead of using setTimeout, do `return delay(50).then(...)`. This will require importing test-utils for these tests, in both the HTML and JS files. See other tests for examples.

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+  });
+
+  const reader = ts.readable.getReader();
+
+  const writer = ts.writable.getWriter();
+  writer.write('a');
+  writer.close();
+
+  reader.read().then(result1 => {
```
domenic: Do `return Promise.all([...])` here on the two promises, with no `.catch()`s.

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+  const ts = new TransformStream({
+    start(controller) {
+      c = controller;
+    },
+    transform() {
+    },
+    flush() {
+      c.enqueue('x');
+      c.enqueue('y');
+      return new Promise(resolve => setTimeout(resolve, 10));
```
domenic: Once you've imported `delay()`, you can do `return delay(10)` here.

reference-implementation/to-upstream-wpts/transform-stream/flush.js
```diff
@@ -0,0 +1,117 @@
...
+      assert_equals(result2.done, false, 'the second chunk read is the second one enqueued in flush');
+    });
+  })
+  .catch(e => t.error(e));
+
+  reader.closed.then(() => {
+    assert_true(true, 'readable reader becomes closed');
+  })
+  .catch(e => { throw e; });
+}, 'TransformStream flush gets a chance to enqueue more into the readable, and can then async close');
```
domenic: Newline at EOF

reference-implementation/to-upstream-wpts/transform-stream/general.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+}
+
+test(() => {
+  const ts = new TransformStream({ transform() { } });
+
+  assert_true(Object.getOwnPropertyDescriptor(Object.getPrototypeOf(ts), 'writeable'), 'it has a writable property');
```
domenic: This existing test of the property descriptor is kind of bad. Maybe replace it with something like https://github.com/w3c/web-platform-tests/blob/a838bae2db982f96ddb6760ca88e699d4514884e/streams/readable-streams/general.js#L56-L61 instead.

reference-implementation/to-upstream-wpts/transform-stream/general.js
```diff
@@ -0,0 +1,22 @@
...
+  assert_true(Object.getOwnPropertyDescriptor(Object.getPrototypeOf(ts), 'readable'), 'it has a readable property');
+  assert_true(ts.readable instanceof ReadableStream, 'readable is an instance of ReadableStream');
+}, 'TransformStream instances must have writable and readable properties of the correct types');
+
+test(() => {
+  const ts = new TransformStream({ transform() { } });
+
+  const writer = ts.writable.getWriter();
+  assert_equals(writer.desiredSize, 1, 'writer.desiredSize should be 1');
+}, 'TransformStream writable starts in the writable state');
```
domenic: Newline at EOF

domenic:
Oh, also, please add yourself the acknowledgments!

Test runner fixed in https://github.com/whatwg/streams/commit/5f5b019493549b083b5ce73f30d36fb8e1bbe506

marvinhagemeister:
Awesome, thanks a lot for your quick review! I'll update the PR tomorrow (already midnight here).

reference-implementation/to-upstream-wpts/transform-stream/constructor.https.html
marvinhagemeister: done!

reference-implementation/to-upstream-wpts/transform-stream/flush.js
marvinhagemeister: Thanks for the tip! Didn't know that as well 👍

reference-implementation/to-upstream-wpts/transform-stream/flush.js
marvinhagemeister: Oh, that somehow slipped through the rebase

reference-implementation/to-upstream-wpts/transform-stream/general.js
marvinhagemeister: Fixed

marvinhagemeister:
@domenic The PR is updated with the requested changes. Thanks again for your review, really appreciate it 👍

PR #533, opened 2016-10-12
No need to reject writePromise in TransformStreamErrorInternal

tyoshino:
Follow up for #519

As we error the writable stream in TransformStreamErrorInternal, we don't need to take care of the promise returned from the underlying sink write().


ricea:
lgtm

PR #535, opened 2016-10-13
Port test that sink methods are called as methods

ricea:
Port the test 'WritableStream should call underlying sink methods as
methods' to web-platform-tests. The tape test relied on assertions
working after the test function had completed. The new test creates a
promise for each method, then fulfills it with the value of 'this'. The
main logic then verifies that each promise is fulfilled with the correct
value of 'this'.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/general.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/general.js
@@ -99,3 +99,50 @@ promise_test(() => {
     }
   );
 }, 'closed and ready on a released writer');
+
+promise_test(t => {
+  let promises = {};
+  let resolvers = {};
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/writable-streams/general.js
```diff
@@ -99,3 +99,50 @@ promise_test(() => {
...
   );
 }, 'closed and ready on a released writer');
+
+promise_test(t => {
+  let promises = {};
+  let resolvers = {};
+  ['start', 'write', 'close', 'abort'].forEach(methodName =>
+       promises[methodName] = new Promise(resolve => resolvers[methodName] = resolve));
+
+  class Sink {
```
tyoshino: write a comment that we ignore second and later calls.

PR #537, opened 2016-10-14
Change 3 tests from async_test to promise_test

ricea:
Change 3 WritableStream web-platform-tests from async_test to
promise_test. Also make them more idiomatic, and use
recordingWritableStream to avoid having assert statements inside sink
methods.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/close.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/close.js
@@ -3,6 +3,7 @@
 if (self.importScripts) {
   self.importScripts('/resources/testharness.js');
   self.importScripts('../resources/test-utils.js');
+  self.importScripts('../resources/recording-streams.js');
 }
 
 promise_test(t => {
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/writable-streams/close.js
```diff
@@ -61,108 +62,72 @@ promise_test(t => {
...
 
-    writer.ready.then(t.step_func(v => {
-      readyFulfilledAlready = true;
-      assert_equals(v, undefined, 'ready promise was fulfilled with undefined');
-      t.done();
-    }));
-  }), 0);
-}, 'If close is called on a WritableStream in waiting state, ready will be fulfilled immediately even if close ' +
+    return writer.ready.then(v => {
+      assert_false(asyncCloseFinished, 'ready promise should be called before async close completes');
```
tyoshino: called -> fulfilled

reference-implementation/to-upstream-wpts/writable-streams/close.js
```diff
@@ -61,108 +62,72 @@ promise_test(t => {
...
+  let asyncCloseFinished = false;
+  const ws = recordingWritableStream({
     close() {
-      return new Promise(resolve => {
-        setTimeout(t.step_func(() => {
-          t.ok(readyFulfilledAlready, 'ready should have fulfilled already');
-          resolve();
-        }), 50);
-      });
+      return delay(50).then(() => asyncCloseFinished = true);
```
tyoshino: this test was not waiting for t.ok(readyFulfilledAlready). right? good catch!

ricea: I reversed the logic so that I wouldn't need an assert inside the sink method. Simplifying the test was just a happy side-effect.

reference-implementation/to-upstream-wpts/writable-streams/close.js
ricea: done

PR #538, opened 2016-10-14
Add controller arguments to sink write and close

ricea:
Previously the controller was only passed to the start() method of a
WritableStream sink. It can be convenient to omit the start() method if
no processing is needed. Pass the controller argument to the write() and
close() methods so that it can be used without implementing start().

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/constructor.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/constructor.js
@@ -25,6 +25,38 @@ promise_test(() => {
 }, 'controller argument should be passed to start method');
 
 promise_test(() => {
+  const passedError = new Error('tea and scones');
+  const ws = new WritableStream({
+    write(chunk, controller) {
+      controller.error(passedError);
+    }
+  });
+
+  const writer = ws.getWriter();
+  writer.write('a');
+
+  return writer.closed.catch(r => {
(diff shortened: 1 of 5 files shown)


reference-implementation/to-upstream-wpts/writable-streams/constructor.js
```diff
@@ -25,6 +25,38 @@ promise_test(() => {
...
+    }
+  });
+
+  const writer = ws.getWriter();
+  writer.write('a');
+
+  return writer.closed.catch(r => {
+    assert_equals(r, passedError, 'ws should be errored by passedError');
+  });
+}, 'controller argument should be passed to write method');
```
domenic: Nit: maybe "and should be able to error the stream"?

tyoshino:
lgtm once domenic's comment is addressed

isonmad:
Isn't the sole method on the writable controller object `.error`? Is it really useful when both `write()` and `close()` can just reject their promise for the same effect?

ricea:
@isonmad: For `write()` the function is the same as for `start()` -- you can error the stream even when there is no write pending. The benefit is that you don't have to artificially implement `start()` just to do this.

The arguments for `close()` are much weaker:

- Consistency.
- `controller` may gain more methods in future.

I could be persuaded either way.

PR #539, opened 2016-10-14
Lint fixes to new web-platform-tests

ricea:
Not all issues reported by eslint have been fixed:
* eslint cannot load scripts from importScripts, so sees all those
  symbols as undefined.
* eslint does not know about 'self'.
* 'Expected to return a value at the end of this method' for sink
  methods in recording-streams.js was harmless so not fixed.
* write(chunk) gives an error "'chunk' is defined but never used". Aids
  readability so not fixed.
* "'t' is defined but never used" is reported for many tests. t
  parameter has been retained for consistency.
* "Arrow function should not return assignment" ignored.

### DIFF
--- a/reference-implementation/to-upstream-wpts/resources/recording-streams.js
+++ b/reference-implementation/to-upstream-wpts/resources/recording-streams.js
@@ -1,4 +1,4 @@
-"use strict";
+'use strict';
 
 self.recordingReadableStream = (extras = {}, strategy) => {
   let controllerToCopyOver;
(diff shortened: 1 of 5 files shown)


domenic:
Hmm. I'd like to fix all issues. Why was eslint not failing the build?

ricea:
The command-line does not include to-upstream-wpts. I added it locally to see what would happen.

domenic:
Oh oops. OK, I'll work on fixing these issues as well and adding it to the build so we don't get in this situation going forward.

tyoshino:
Great, Domenic. and lgtm on this.

PR #540, opened 2016-10-14
Port writable stream abort tests to wpt

marvinhagemeister:
Just ported another test file to wpt. There is one thing where I'm unsure if this is the correct way for assertion. In some tests the constructor of the error object is compared. I couldn't get it to work without casting it to a string first.

For example:
```js
const ws = new WritableStream();
writer.write().then(
   () => { throw new Error('something failed'); },
   r => assert_equals(r.constructor.toString(), TypeError.toString(), 'err message')
);
```


domenic:
The pattern we've been using for that is to put

```
const error1 = new Error('error1');
error1.name = 'error1';
```

at the top of the file (plus maybe error2 if you have multiple errors and want to distinguish), then later doing

```js
return promise_rejects(t, error1, writer.write(), 'err message');
```

marvinhagemeister:
Ah I see. I'll adapt my code to use `promise_rejects` 👍

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +41,256 @@ promise_test(t => {
     assert_not_equals(writer.ready, readyPromise, 'the ready promise property should change');
     return promise_rejects(t, new TypeError(), writer.ready, 'the ready promise should reject with a TypeError');
   });
-
 }, 'Aborting a WritableStream should cause the writer\'s fulfilled ready promise to reset to a rejected one');
+
+function promise_fulfills(expectedValue, promise, msg) {
```
domenic: This is not necessary. Inside your promise_test, just do

```js
return promise.then(value => {
  assert_equals(value, expectedValue, msg);
});
```

the test will fail if it returns a rejected promise.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +41,256 @@ promise_test(t => {
...
+  },
+  r => {
+    assert_equals(r.constructor.toString(), TypeError.toString(), 'abort() should reject with a TypeError');
+  });
+}, 'abort() on a released writer rejects');
+
+promise_test(() => {
+  const ws = new WritableStream({
+    write() {
+      throw new Error('Unexpected write() call');
```
domenic: It's a decent bit of work, but I think these sorts of tests would be improved by using the recordingWritableStream, and then checking the recorded events in a final `.then` block. An example: https://github.com/whatwg/streams/blob/ff1d3c9af8a01f4b03a508281844b72a92b533c7/reference-implementation/to-upstream-wpts/writable-streams/close.js#L65-L81

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +41,256 @@ promise_test(t => {
...
+    writer.write(1);
+    writer.write(2);
+  })
+  .then(() => delay(100));
+}, 'Aborting a WritableStream immediately prevents future writes');
+
+promise_test(() => {
+  let writeCount = 0;
+
+  const ws = new WritableStream({
```
domenic: Another good example of where recordingWritableStream would help you. You wouldn't need writeCount or any asserts, just check the events in a final `.then` block.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +41,256 @@ promise_test(t => {
...
+        'with the given reason')
+    ),
+    writer.closed.then(
+      () => { throw new Error('closed fulfilled unexpectedly'); },
+      r => assert_equals(r.constructor.toString(), TypeError.toString(), 'closed should reject with a TypeError')
+    )
+  ]);
+}, 'Aborting a WritableStream puts it in an errored state, with stored error equal to the abort reason');
+
+test(() => {
```
domenic: This will need to be a promise_test when it moves to use assert_promise_rejects

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +41,256 @@ promise_test(t => {
...
+  );
+}, 'Closing a WritableStream and aborting it while it closes causes the stream to error');
+
+test(() => {
+  const ws = new WritableStream();
+  const writer = ws.getWriter();
+
+  writer.close();
+
+  setTimeout(() => {
```
domenic: Use promise_test and delay()

marvinhagemeister:
Sorry for the late response, I got ill for a few days. Thanks, for the review. I'll make these changes over the day 👍

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
marvinhagemeister: I'm unsure where to import `assert_promise_rejects` from. This function seems to be duplicated inside the `web-platform-tests` repo. I tried including one of the existing functions, but get an URL error by testharness

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
marvinhagemeister: Thanks for the tip. `recordingWritableStream` is much cleaner 👍

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
marvinhagemeister: agree

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
domenic: I meant promise_rejects, sorry.

domenic:
It looks like you just accidentally created a duplicate of `promise_rejects` named `assert_promise_rejects`. I'm sorry for misleading you by saying the wrong name! But yeah, that can be deleted and we can just use `promise_rejects`.

marvinhagemeister:
Thanks for clearing that up. Yeah, I got confused by the existing tests in inside the `web-platform-tests` folder. Adjusted the PR.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -1,14 +1,16 @@
 'use strict';
+/* global delay, recordingWritableStream */
```
domenic: If you rebase this will not be necessary

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -1,14 +1,16 @@
...
 
 if (self.importScripts) {
+  self.importScripts('../resources/test-utils.js');
```
domenic: Move this after testharness.js

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+  return delay(0)
+    .then(() => {
+      const writer = ws.getWriter();
+
+      writer.abort();
+      writer.write(1);
+      writer.write(2);
+    })
+    .then(() => {
+      assert_equals(ws.events.length, 2, 'the stream should have received 2 events');
```
domenic: You should use assert_array_equals here like the other recording stream tests. Same below.

The events[1] entry is the argument passed to abort, by the way, and is not related to writes.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+  const errorInSinkAbort = new Error('Sorry, it just wasn\'t meant to be.');
+  const ws = new WritableStream({
+    abort() {
+      throw errorInSinkAbort;
+    }
+  });
+
+  const writer = ws.getWriter();
+
+  return promise_rejects(t, errorInSinkAbort, writer.abort(undefined),
```
domenic: Can you use the `error1` pattern used elsewhere? With a single `error1` at the top of the file? Here and below.

Also everywhere that creates a separate `passedReason` can use that `error1`.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+      recordedReason = reason;
+    }
+  });
+
+  const writer = ws.getWriter();
+
+  const passedReason = new Error('Sorry, it just wasn\'t meant to be.');
+  writer.abort(passedReason);
+
+  assert_equals(recordedReason, passedReason);
```
domenic: This test can use recordingWritableStream instead, with assert_array_equals on the contents.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+  const passedReason = new Error('Sorry, it just wasn\'t meant to be.');
+  writer.abort(passedReason);
+
+  return Promise.all([
+    promise_rejects(t, new TypeError(), writer.write(), 'writing should reject with the given reason'),
+    promise_rejects(t, new TypeError(), writer.close(), 'closing should reject with the given reason'),
+    promise_rejects(t, new TypeError(), writer.abort(), 'aborting should reject with the given reason'),
+    promise_rejects(t, new TypeError(), writer.closed, 'closed should reject with the given reason'),
+  ]);
+}, 'Aborting a WritableStream puts it in an errored state, with stored error equal to the abort reason');
```
domenic: This test description is outdated; change it to "with a TypeError as the stored error"

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+
+// TODO return promise
```
domenic: Just store the promise with `const writePromise = promise_rejects(...)` and then return it as the last line

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+promise_test(t => {
+  const ws = new WritableStream();
+  const writer = ws.getWriter();
+
+  promise_rejects(t, new TypeError(), writer.write('a'), 'writing should reject with a TypeError');
+
+  const passedReason = new Error('Sorry, it just wasn\'t meant to be.');
+  writer.abort(passedReason);
+}, 'Aborting a WritableStream causes any outstanding write() promises to be rejected with the abort reason');
```
domenic: Test description is again outdated; "with a TypeError"

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+promise_test(t => {
+  const ws = new WritableStream();
+  const writer = ws.getWriter();
+
+  writer.close();
+
+  const passedReason = new Error('Sorry, it just wasn\'t meant to be.');
+  writer.abort(passedReason);
+
+  return promise_rejects(t, new TypeError('Aborted'), writer.closed, 'the stream should be errored with a TypeError');
```
domenic: Don't include `'Aborted'` here; the test framework doesn't look at messages. (Yeah, it's kind of wonky.)

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+  const ws = new WritableStream();
+  const writer = ws.getWriter();
+
+  writer.close();
+
+  return delay(0)
+    .then(() => writer.abort())
+    .then(
+      v => assert_equals(v, undefined, 'abort promise should fulfill with undefined'),
+      () => { throw new Error(); }
```
domenic: This `() => { throw new Error(); }` is not necessary. In fact the second `.then()` can be removed entirely since we already test the undefinedness earlier.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -43,5 +43,193 @@ promise_test(t => {
...
+test(() => {
+  const ws = new WritableStream({
+    close(...args) {
+      assert_equals(args.length, 0, 'close() was called (with no arguments)');
+    }
+  });
+
+  const writer = ws.getWriter();
+
+  writer.abort();
```
domenic: Again best to use recordingWritableStream. As-is, this test will pass even if close is never called.

It might be good to add something to recordingWritableStream's close() that asserts that `arguments.length === 0`.

domenic:
@marvinhagemeister I'll take care of the rest of the issues I noted. Thanks so much for your help getting this file moved over :).

PR #542, opened 2016-10-16
brand-check transformstream methods

isonmad:
Based on the ReadableStream brand-checking.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -279,6 +279,16 @@ class TransformStreamSource {
 
 class TransformStreamDefaultController {
   constructor(transformStream) {
+    if (IsTransformStream(transformStream) === false) {
+      throw new TypeError('TransformStreamDefaultController can only be ' +
+                          'constructed with a TransformStream instance');
+    }
+
+    if (transformStream._transformStreamController !== undefined) {
+      throw new TypeError('TransformStreamDefaultController instances can ' +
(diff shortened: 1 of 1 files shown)


tyoshino:
lgtm but please wait for Domenic

reference-implementation/lib/transform-stream.js
```diff
@@ -253,18 +279,40 @@ class TransformStreamSource {
 
 class TransformStreamDefaultController {
   constructor(transformStream) {
+    if (IsTransformStream(transformStream) === false) {
+      throw new TypeError('TransformStreamDefaultController can only be ' +
+                          'constructed with a TransformStream instance');
+    }
+
+    if (transformStream._transformStreamController !== undefined) {
```
ricea: I just filed #543. Depending on the outcome, it may mean changing this.

reference-implementation/lib/transform-stream.js
```diff
@@ -208,7 +233,8 @@ class TransformStreamSink {
 
     transformStream._writableDone = true;
 
-    const flushPromise = PromiseInvokeOrNoop(transformStream._transformer, 'flush', [transformStream._controller]);
+    const flushPromise = PromiseInvokeOrNoop(transformStream._transformer,
+                         'flush', [transformStream._transformStreamController]);
```
ricea: This is inconsistent with transformPromise above, where a temporary variable called "controller" is created first. Also startResult, below.

PR #544, opened 2016-10-19
Add tests for WritableStreamDefaultWriter and Controller constructors

ricea:
The WritableStreamDefaultWriter and WritableStreamDefaultController
constructors are not exported. However they can be accessed fron the
"constructor" property of the relevant objects.

The WritableStreamDefaultController constructor should always throw an
exception if used explicitly. Verify that it does.

WritableStreamDefaultWriter can be constructed with an unlocked stream
and should then work just like a writer returned from
getWriter(). Verify that it does.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/write.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/write.js
@@ -197,3 +197,15 @@ promise_test(() => {
           assert_equals(writeCount, numberOfWrites, `should have called sink's write ${numberOfWrites} times`));
   });
 }, 'a large queue of writes should be processed completely');
+
+promise_test(() => {
+  const stream = recordingWritableStream();
+  const w = stream.getWriter();
+  const WritableStreamDefaultWriter = w.constructor;
+  w.releaseLock();
+  const writer = new WritableStreamDefaultWriter(stream);
+  return writer.ready.then(() => {
+    writer.write('a');
+    assert_array_equals(stream.events, ['write', 'a'], 'write() should be passed to sink');
+  });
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/writable-streams/write.js
```diff
@@ -197,3 +197,15 @@ promise_test(() => {
           assert_equals(writeCount, numberOfWrites, `should have called sink's write ${numberOfWrites} times`));
   });
 }, 'a large queue of writes should be processed completely');
+
+promise_test(() => {
+  const stream = recordingWritableStream();
+  const w = stream.getWriter();
+  const WritableStreamDefaultWriter = w.constructor;
+  w.releaseLock();
+  const writer = new WritableStreamDefaultWriter(stream);
```
tyoshino: Hmm, this looks something we didn't intend to enable. We could disallow this by letting the constructor to look at some internal slot of the stream which can be set only by getWriter(), e.g. WDYT, @domenic ?

domenic: Nah, this is intended, just a little silly. IMO constructors should only be prohibited from working when doing so would break an invariant (like 1 controller <-> 1 stream). That's not the case here so it's good to have a working constructor, even if it's hidden by default.

tyoshino: Okay. So, we should keep in mind that invocation of `.getWriter()` shouldn't be used for detecting obtaining of a writer. We've already switched to not use public method invocation for controlling its behavior for pipeTo() implementation. So, yes, I understand it's ok.

PR #545, opened 2016-10-19
Editorial: change underlyingSource to underlyingSink

ricea:
The algorithm for the WritableStream constructor made reference to
"underlyingSource". Change it to "underlyingSink".


tyoshino:
lgtm

PR #550, opened 2016-10-21
make TransformStream respect backpressure

### LINKED ISSUE #330: TransformStream: transform() is never invoked via pull()

isonmad:
This should basically fix #330 / make it technically irrelevant, as now transform is only called from one place, technically. But it can now be triggered by pull() indirectly.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -343,7 +343,7 @@ module.exports = class TransformStream {
     this._readableClosed = false;
 
     this._resolveWrite = undefined;
-
+    this._readableBackpressure = false;
     this._chunkPending = false;
     this._chunk = undefined;
 
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -81,15 +81,7 @@ function TransformStreamCloseReadableInternal(transformStream) {
   transformStream._readableClosed = true;
 }
 
-function TransformStreamResolveWrite(transformStream) {
-  if (transformStream._errored === true) {
-    return;
-  }
-
-  assert(transformStream._transforming === true);
-
-  assert(transformStream._resolveWrite !== undefined);
-
+function TransformStreamReadyPromise(transformStream) {
   assert(transformStream._readyPromise_resolve === undefined);
 
   let readyPromise;
@@ -100,6 +92,20 @@ function TransformStreamResolveWrite(transformStream) {
       transformStream._readyPromise_resolve = resolve;
     });
   }
+
+  return readyPromise;
+}
+
+function TransformStreamResolveWrite(transformStream) {
+  if (transformStream._errored === true) {
+    return;
+  }
+
+  assert(transformStream._transforming === true);
+
+  assert(transformStream._resolveWrite !== undefined);
+
+  const readyPromise = TransformStreamReadyPromise(transformStream);
 
   readyPromise.then(() => {
     transformStream._transforming = false;
@@ -201,7 +207,10 @@ class TransformStreamSink {
 
     transformStream._writableController = c;
 
-    return this._startPromise;
+    // delay all sink.write() calls until there is no longer backpressure.
+    return this._startPromise.then(() => {
+      return TransformStreamReadyPromise(transformStream);
+    });
   }
 
   write(chunk) {
@@ -393,13 +402,13 @@ module.exports = class TransformStream {
       startPromise_resolve = resolve;
     });
 
-    const sink = new TransformStreamSink(this, startPromise);
-
-    this._writable = new WritableStream(sink, transformer.writableStrategy);
-
     const source = new TransformStreamSource(this, startPromise);
 
     this._readable = new ReadableStream(source, transformer.readableStrategy);
+
+    const sink = new TransformStreamSink(this, startPromise);
+
+    this._writable = new WritableStream(sink, transformer.writableStrategy);
 
     assert(this._writableController !== undefined);
     assert(this._readableController !== undefined);
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -266,6 +266,7 @@ class TransformStreamSource {
   }
 
   pull() {
+    // console.log('TransformStreamSource.pull()');
     this._transformStream._readableBackpressure = false;
     TransformStreamTransformIfNeeded(this._transformStream);
   }
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -96,25 +96,6 @@ function TransformStreamReadyPromise(transformStream) {
   return readyPromise;
 }
 
-function TransformStreamResolveWrite(transformStream) {
-  if (transformStream._errored === true) {
-    return;
-  }
-
-  assert(transformStream._transforming === true);
-
-  assert(transformStream._resolveWrite !== undefined);
-
-  const readyPromise = TransformStreamReadyPromise(transformStream);
-
-  readyPromise.then(() => {
-    transformStream._transforming = false;
-
-    transformStream._resolveWrite(undefined);
-    transformStream._resolveWrite = undefined;
-  });
-}
-
 function TransformStreamErrorIfNeeded(transformStream, e) {
   if (transformStream._errored === false) {
     TransformStreamErrorInternal(transformStream, e);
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -342,10 +338,11 @@ module.exports = class TransformStream {
     this._writableDone = false;
     this._readableClosed = false;
 
-    this._resolveWrite = undefined;
-
-    this._chunkPending = false;
-    this._chunk = undefined;
+    // readableBackpressure begins in an unknown state, to be determined at
```
ricea: I find this confusing. Perhaps use an explicit enum like "undefined", "no", "yes" or similar?

reference-implementation/test/transform-stream.js
```diff
@@ -200,8 +200,10 @@ test('TransformStream: by default, closing the writable closes the readable afte
   writer.write('a');
```
ricea: We're going to need a test verifying that backpressure actually works.

reference-implementation/lib/transform-stream.js
```diff
@@ -79,21 +81,19 @@ function TransformStreamCloseReadableInternal(transformStream) {
...
-
-  assert(transformStream._resolveWrite !== undefined);
-
-  transformStream._transforming = false;
+function TransformStreamReadyPromise(transformStream) {
+  assert(transformStream._readyPromise_resolve === undefined);
 
-  transformStream._resolveWrite(undefined);
-  transformStream._resolveWrite = undefined;
+  let readyPromise;
```
ricea: I think this would be easier to understand without the `readyPromise` variable. How about

```javascript
if (transformStream._readableBackpressure === false) {
    return Promise.resolve();
}
return new Promise(resolve => {
  transformStream._readyPromise_resolve = resolve;
});
```

reference-implementation/lib/transform-stream.js
```diff
@@ -116,39 +116,25 @@ function TransformStreamErrorInternal(transformStream, e) {
...
   const controller = transformStream._transformStreamController;
   const transformPromise = PromiseInvokeOrNoop(transformStream._transformer,
                              'transform', [chunk, controller]);
 
-  transformPromise.then(() => TransformStreamResolveWrite(transformStream),
-                     e => TransformStreamErrorIfNeeded(transformStream, e));
+  return transformPromise.then(() => {
+    transformStream._transforming = false;
+  }).then(() => TransformStreamReadyPromise(transformStream),
+          e => TransformStreamErrorIfNeeded(transformStream, e));
```
ricea: I don't think this will be easy to read when translated into an algorithm in the spec.

How about

```javascript
return transformPromise.then(
    () => {
      transformStream._transforming = false;
      TransformStreamReadyPromise(transformStream);
    },
    e => TransformStreamErrorIfNeeded(transformStream, e));
```

reference-implementation/lib/transform-stream.js
```diff
@@ -186,31 +172,18 @@ class TransformStreamSink {
 
     transformStream._writableController = c;
 
-    return this._startPromise;
+    // delay all sink.write() calls until there is no longer backpressure.
+    return this._startPromise.then(() => {
+      return TransformStreamReadyPromise(transformStream);
```
ricea: Don't need a block here. Can just write

```javascript
return this._startPromise.then(() => TransformStreamReadyPromise(transformStream));
```

ricea:
lgtm

We can add some backpressure tests later.

@isonmad, let me know if you're ready for me to land this or you want to add more commits.

reference-implementation/lib/transform-stream.js
```diff
@@ -262,12 +238,32 @@ class TransformStreamSource {
...
+    // invariant: pull is never called while readableBackpressure is false.
+    // Enforced by the promises returned by source.start and source.pull.
+    assert(transformStream._readableBackpressure === true);
+    assert(transformStream._backpressureChangePromise !== undefined);
+
+    TransformStreamSetBackpressure(transformStream, false);
+
+    transformStream._backpressureChangePromise.then(backpressure => assert(backpressure === true));
+
+    return transformStream._backpressureChangePromise;
```
tyoshino: This is looping back the backpressure signal on desiredSize to pull() but mux-ed with backpressure-changing TransformStreamEnqueueToReadable() invocation. I think we shouldn't have to have such a loop back here. Let me think more.

tyoshino:
Could you please take a look at this fork?
https://github.com/whatwg/streams/commit/dedffc87e05e71e6eb279fbeb3296bc5115d7167

enqueue() passes controls the pull() suppression signal, and pull() and enqueue() control the backpressure signal.

reference-implementation/lib/transform-stream.js
```diff
@@ -262,12 +238,32 @@ class TransformStreamSource {
 
     transformStream._readableController = c;
 
-    return this._startPromise;
+    // delay the first source.pull call until there is backpressure
+    return this._startPromise.then(() => {
```
isonmad: Like the comment above this line says, the purpose of this `.then()` is to delay `pull()` being called. It doesn't actually *do* anything except delay the fulfillment of the promise returned by `source.start`, so putting it in the constructor would do nothing.

reference-implementation/lib/transform-stream.js
```diff
@@ -354,17 +359,20 @@ module.exports = class TransformStream {
...
 
+    const sink = new TransformStreamSink(this, startPromise);
+
+    this._writable = new WritableStream(sink, transformer.writableStrategy);
+
     assert(this._writableController !== undefined);
     assert(this._readableController !== undefined);
 
+    const desiredSize = this._readableController.desiredSize;
+    TransformStreamSetBackpressure(this, desiredSize <= 0);
```
isonmad: Are there ever going to be any active reads when we're still inside the TransformStream constructor? I thought it was physically impossible for someone to have called `ts.readable.getReader().read()` at this point.

And it already depends on `pull()` invocation and sets backpressure to false whenever `pull()` is called, so I'm not sure what you're suggesting here.

isonmad:
@ricea I think this is good to go, unless tyoshino objects.

reference-implementation/lib/transform-stream.js
tyoshino: Once the TransformStream constructor completes, user can start read()-ing from the readable side. If the strategy is set to HWM=0, i.e. configured to only pull() only when there's one or more read()s, pull() invocation is the only signal to remove backpressure. So, if you start with calling TransformStreamSetBackpressure(this, true), and TransformStreamSource() starts with being blocked on _backpressureChangePromise's fulfillment, I thought that there's no one who can fulfill it as pull() is suppressed.

reference-implementation/lib/transform-stream.js
isonmad: > here you just return the passed promise stored in _startPromise

That's untrue. The promise returned by `.then()` is returned.

reference-implementation/lib/transform-stream.js
isonmad: `TransformStreamSetBackpressure(this, true)` does the exact opposite of suppressing `pull()`. It sets backpressure to true, meaning pull is NOT suppressed.

tyoshino:
> the queue is empty and highWaterMark === 0, in which case there is backpressure and pull() calls are not being suppressed

This depends on how you interpret the signals. The note at `get desiredSize` might be seen as implying that assumption. But we're not explicitly saying that `desiredSize <= 0` is equivalent to whether or not there's backpressure. What people would see is that there're two signals, desiredSize and pull() invocation. Until the desiredSize interface was added for more precise flow-control, there was only pull().

We should show the same signals to the transform stream's controller. You've added desiredSize. I suggest that we translate pull() to transform invocation.

reference-implementation/lib/transform-stream.js
tyoshino: Sorry. I meant `"then" you just return ...`.

isonmad:
> This depends on how you interpret the signals.

Let me clarify what I meant when I said that.

- the queue is empty and `highWaterMark === 0`, in which case, there *was* backpressure right up until the moment of the new `read()` call. In which case, because there was backpressure until right this very moment, `pull()` calls are not currently being suppressed.

reference-implementation/lib/transform-stream.js
tyoshino: I was unclear. I wanted to say the following.

You can pass a promise already then-ed with this callback in TransformStream's constructor to the constructor of TransformStreamSource, and then you just return the passed promise stored in _startPromise (which is already then-ed).

tyoshino:
Oh, ok. I was misunderstanding the setup in the constructor. I'm sorry about that.

I'll reconstruct my suggestion.

isonmad:
If it's just refactoring suggestions can this be merged first? I have some unrelated changes I want to submit and want to avoid merge conflicts.

tyoshino:
OK. Go ahead, please. I'm just almost convinced that this logically works. But regarding readability, I have some suggestions.

tyoshino:
I've finished understanding the code. Sorry for the confusion in review. This does lgtm.

I'd like to do some refactoring to record the result of my analysis. Please take a look at https://github.com/whatwg/streams/pull/571.

PR #554, opened 2016-10-23
Reference promise spec definitions

isonmad:
There might still be room for simplication in `PromiseInvokeOrNoop` by using the term [promise-calling](https://w3ctag.github.io/promises-guide/#promise-calling), but it's not clear how to use that term in practice when you only have an array of arguments of unknown length. That and PromiseInvokeOrNoop invokes the function as a method, and promise-calling's definition seems to just be for regular functions.

Does any other spec currently use `<a>promise-calling</a>` at all?

Also, upon-fulfillment's definition of "upon fulfillment of *p* with value *v*" assumes you always care about the value of *v*, should it have a more explicit syntax for when *v* is unused?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4073,8 +4073,8 @@ itself will evolve in these ways.
     procedure was performed before the algorithm starts.
   <li> We similarly use the default argument notation <code>= {}</code> in a couple of cases.
   <li> We use "<emu-val>this</emu-val>" instead of "<emu-val>this</emu-val> value".
-  <li> We use <a href="https://w3ctag.github.io/promises-guide/#shorthand-phrases">the shorthand phrases from the W3C
-    TAG promises guide</a> to operate on promises at a higher level than the ECMAScript spec does.
+  <li> We use the shorthand phrases from the [[!PROMISES-GUIDE]] to operate on promises at a higher level than the
+    ECMAScript spec does.
 </ul>
 
 <h2 id="acks" class="no-num">Acknowledgments</h2>
--- a/index.bs
+++ b/index.bs
@@ -722,7 +722,7 @@ noticable asymmetry between the two branches. [[!HTML]]
   1. Assert: Type(_cloneForBranch2_) is Boolean.
   1. Let _reader_ be ? AcquireReadableStreamDefaultReader(_stream_).
   1. Let _teeState_ be Record {[[closedOrErrored]]: *false*, [[canceled1]]: *false*, [[canceled2]]: *false*,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2752,10 +2752,10 @@ nothrow>WritableStreamUpdateBackpressure ( <var>stream</var>, <var>backpressure<
   1. Let _writer_ be _stream_.[[writer]].
   1. If _writer_ is *undefined*, return.
   1. If _backpressure_ is *true*,
-    1. Set _writer_.[[readyPromise]] to a new promise.
+    1. Set _writer_.[[readyPromise]] to <a>a new promise<a>.
```
ricea: Second \<a> should be \</a>

index.bs
```diff
@@ -857,8 +857,8 @@ nothrow>ReadableStreamAddReadIntoRequest ( <var>stream</var> )</h4>
 
 <emu-alg>
   1. Set _stream_.[[disturbed]] to *true*.
-  1. If _stream_.[[state]] is `"closed"`, return a new promise resolved with *undefined*.
-  1. If _stream_.[[state]] is `"errored"`, return a new promise rejected with _stream_.[[storedError]].
+  1. If _stream_.[[state]] is `"closed"`, return <a>a promise resolved with</a> *undefined*.
+  1. If _stream_.[[state]] is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
```
ricea: "a promise rejected with" seems to be inconsistently linked. I counted 34 places where it is not linked. Am I missing the reason for this?

domenic:
Agreed promise-calling is not usable in this context.

isonmad:
@domenic But what about upon-fulfillment's syntax?

Out of curiosity, has literally any spec found promise-calling usable in their contexts? It's defined as a gerund instead of an imperative verb, making it awkward to use, and it seems extremely limited in utility.

ricea:
lgtm, thanks.

I will wait for @domenic's response before merging.

domenic:
> But what about upon-fulfillment's syntax?

IMO it's fine as-is; people should be able to understand what's meant if you omit the "with v".

> Out of curiosity, has literally any spec found promise-calling usable in their contexts? It's defined as a gerund instead of an imperative verb, making it awkward to use, and it seems extremely limited in utility.

All specs that use Web IDL promise callbacks use it, which is quite a few. But they do not use the definition directly; instead they use Web IDL's callback invocation mechanisms which perform similar steps.

PR #555, opened 2016-10-24
Fix use of `this` in AcquireWritableStreamDefaultWriter

ricea:
AcquireWritableStreamDefaultWriter references `this` despite being an
abstract operation called with a `stream` parameter.

Replace `this` with `stream`.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2648,7 +2648,7 @@ specifications, instead of just being part of the implementation of this spec's 
 throws>AcquireWritableStreamDefaultWriter ( <var>stream</var> )</h4>
 
 <emu-alg>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2648,7 +2648,7 @@ specifications, instead of just being part of the implementation of this spec's
 throws>AcquireWritableStreamDefaultWriter ( <var>stream</var> )</h4>
 
 <emu-alg>
-  1. Return ? Construct(`<a idl>WritableStreamDefaultWriter</a>`, « *this* »).
+  1. Return ? Construct(`<a idl>WritableStreamDefaultWriter</a>`, « <var>source</var> »).
```
tyoshino: s/source/stream/
enclose with _ than <var>

tyoshino:
lgtm!

PR #563, opened 2016-10-25
Update Examples.md for api changes

isonmad:
Unless Examples.md should just be deleted entirely if there are strictly better examples elsewhere?

Also delete TeeStream.md since teeing was actually implemented in a very different way in 2015.


domenic:
I guess we can delete "Buffering the Entire Stream Into Memory" since it duplicates https://streams.spec.whatwg.org/#example-37a7e90d

domenic:
Actually it seems like they're all pretty redundant. Let me know if you agree. writeArrayToStreamWithReporting is redundant with https://streams.spec.whatwg.org/#example-b9d5039d + https://streams.spec.whatwg.org/#example-222a6060 and promptAndWrite is redundant with https://streams.spec.whatwg.org/#example-e3388a0d, IMO.

PR #568, opened 2016-10-27
check if sink/source methods are callable in constructor

isonmad:
Before, it waited until there was a reason to call them, then errored the stream when it couldn't. But this can easily be caught much earlier in the constructor instead the same way strategies are already validated.

It looks like it would've never even been noticeable if `abort` or `cancel` were set to a non-function.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3592,6 +3592,16 @@ throws>ValidateAndNormalizeQueuingStrategy ( <var>size</var>, <var>highWaterMark
   1. Return Record {[[size]]: _size_, [[highWaterMark]]: _highWaterMark_}.
 </emu-alg>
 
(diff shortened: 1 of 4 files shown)


domenic:
We generally don't want to do this kind of up-front checking. You can supply an arbitrary object whose shape changes over time, and we call its methods at runtime.

isonmad:
So it's TransformStream's constructor that should be changed not to check this then?

domenic:
> So it's TransformStream's constructor that should be changed not to check this then?

Yeah

> Is there a reason the strategy object isn't similarly allowed to be changed arbitrarily, and gets decomposed instead?

Strategies are not treated as an object with methods; it is treated as  "record" that contains a number and a function. This helps avoid strategies being stateful, whereas an underlying source being stateful is normal.

isonmad:
TransformStream changes in #573.

PR #570, opened 2016-10-27
Tests for Sink methods returning thenable

ricea:
Sink methods that can return a Promise can also return an arbitrary
object that implements a then() method. The then() method should be
called as if it was a Promise object.

Add web platform tests to verify this behaviour.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/aborting.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/aborting.js
@@ -203,3 +203,19 @@ test(() => {
     assert_array_equals(ws.events, ['close']);
   });
 }, 'WritableStream should call underlying sink\'s close if no abort is supplied');
+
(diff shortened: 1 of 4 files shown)


reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -203,3 +203,19 @@ test(() => {
     assert_array_equals(ws.events, ['close']);
   });
 }, 'WritableStream should call underlying sink\'s close if no abort is supplied');
+
+promise_test(() => {
+  let thenCalled = false;
+  const ws = new WritableStream({
+    abort() {
+      return {
+        then(resolve) {
```
domenic: Nit: the proper argument names are `(onFulfilled, onRejected)` not `(resolve, reject)`

tyoshino:
lgtm

PR #571, opened 2016-10-27
Refactor TransformStream's backpressure handling

tyoshino:
Add notes (in the form of assert description where possible) about the new
backpressure controlling logic introduced by isonmad to ease understanding.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -123,7 +123,7 @@ function TransformStreamReadableReadyPromise(transformStream) {
 function TransformStreamSetBackpressure(transformStream, backpressure) {
   // console.log(`TransformStreamSetBackpressure(${backpressure})`);
 
-  // Passes also when called for initialization.
+  // Passes also when called during construction.
   assert(transformStream._backpressure !== backpressure,
          'TransformStreamSetBackpressure() should be called only when backpressure is changed');
 
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -21,6 +21,8 @@ function TransformStreamCloseReadable(transformStream) {
 }
 
 function TransformStreamEnqueueToReadable(transformStream, chunk) {
+  // console.log('TransformStreamEnqueueToReadable()');
+
   if (transformStream._errored === true) {
     throw new TypeError('TransformStream is already errored');
   }
@@ -45,10 +47,16 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
     throw transformStream._storedError;
   }
 
-  const backpressure = controller.desiredSize <= 0;
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() is made immediately (but
+    // asynchronously) after this because of pending read()s and fix _backpressure back to false.
+    //
+    // We don't have to worry about that such a pull() may happen synchronously to the enqueue(), and therefore is not
+    // queued at this point. It's guaranteed that there is pending start() or the last pull() kept pending which
+    // prevent such a synchronous pull().
+    TransformStreamSetBackpressure(transformStream, true);
   }
 }
 
@@ -97,14 +105,17 @@ function TransformStreamErrorInternal(transformStream, e) {
   }
 }
 
-function TransformStreamReadyPromise(transformStream) {
-  assert(transformStream._backpressureChangePromise !== undefined);
+// Used for preventing the next write() call on TransformStreamSink until there
+// is no longer backpressure.
+function TransformStreamBackpressureGonePromise(transformStream) {
+  assert(transformStream._backpressureChangePromise !== undefined,
+         '_backpressureChangePromise should have been initialized');
 
-  if (transformStream._readableBackpressure === false) {
+  if (transformStream._backpressure === false) {
     return Promise.resolve();
   }
 
-  transformStream._backpressureChangePromise.then(backpressure => assert(backpressure === false));
+  assert(transformStream._backpressure === true, '_backpressure should have been initialized');
 
   return transformStream._backpressureChangePromise;
 }
@@ -112,9 +123,12 @@ function TransformStreamReadyPromise(transformStream) {
 function TransformStreamSetBackpressure(transformStream, backpressure) {
   // console.log(`TransformStreamSetBackpressure(${backpressure})`);
 
-  assert(transformStream._readableBackpressure !== backpressure);
+  // Passes also when called for initialization.
+  assert(transformStream._backpressure !== backpressure,
+         'TransformStreamSetBackpressure() should be called only when backpressure is changed');
 
   if (transformStream._backpressureChangePromise !== undefined) {
+    // The fulfillment value is just for the assertion.
     transformStream._backpressureChangePromise_resolve(backpressure);
   }
 
@@ -122,7 +136,12 @@ function TransformStreamSetBackpressure(transformStream, backpressure) {
     transformStream._backpressureChangePromise_resolve = resolve;
   });
 
-  transformStream._readableBackpressure = backpressure;
+  transformStream._backpressureChangePromise.then(resolution => {
+    assert(resolution !== backpressure,
+           '_backpressureChangePromise should be fulfilled only when backpressure is changed');
+  });
+
+  transformStream._backpressure = backpressure;
 }
 
 function TransformStreamTransform(transformStream, chunk) {
@@ -130,7 +149,7 @@ function TransformStreamTransform(transformStream, chunk) {
 
   assert(transformStream._errored === false);
   assert(transformStream._transforming === false);
-  assert(transformStream._readableBackpressure === false);
+  assert(transformStream._backpressure === false);
 
   transformStream._transforming = true;
 
@@ -141,8 +160,8 @@ function TransformStreamTransform(transformStream, chunk) {
   return transformPromise.then(
     () => {
       transformStream._transforming = false;
-      // Do not fulfill until there's no backpressure.
-      return TransformStreamReadyPromise(transformStream);
+
+      return TransformStreamBackpressureGonePromise(transformStream);
     },
     e => TransformStreamErrorIfNeeded(transformStream, e));
 }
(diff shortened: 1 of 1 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -97,40 +105,51 @@ function TransformStreamErrorInternal(transformStream, e) {
   }
 }
 
-function TransformStreamReadyPromise(transformStream) {
-  assert(transformStream._backpressureChangePromise !== undefined);
+// Used for preventing the next write() call on TransformStreamSink until there
+// is no longer backpressure.
+function TransformStreamBackpressureGonePromise(transformStream) {
```
isonmad: I don't see how this name change makes it clearer. ReadyPromise is analogous to the [[readyPromise]] that writablestream writers have.

reference-implementation/lib/transform-stream.js
```diff
@@ -363,14 +386,18 @@ module.exports = class TransformStream {
...
+    }));
 
     this._writable = new WritableStream(sink, transformer.writableStrategy);
 
     assert(this._writableController !== undefined);
     assert(this._readableController !== undefined);
 
     const desiredSize = this._readableController.desiredSize;
+    // Tentatively set _backpressure based on desiredSize. When desiredSize is 0, it's possible that a pull() is made
+    // immediately after this because of pending read()s and fix _backpressure back to false.
```
isonmad: There can't be pending reads because this is the constructor, there's nothing 'tentative' about the fact that there is or isn't backpressure when the stream is first constructed.

reference-implementation/lib/transform-stream.js
```diff
@@ -363,14 +386,18 @@ module.exports = class TransformStream {
 
     this._readable = new ReadableStream(source, transformer.readableStrategy);
 
-    const sink = new TransformStreamSink(this, startPromise);
+    const sink = new TransformStreamSink(this, startPromise.then(() => {
+      return TransformStreamBackpressureGonePromise(this);
+    }));
```
isonmad: I don't see the point of doing this in the constructor instead of inside sink.start. It's conceptually something that .start does, and it's symmetrical with what source.start does. It's not really any clearer this way.

reference-implementation/lib/transform-stream.js
```diff
@@ -238,13 +256,17 @@ class TransformStreamSource {
...
     transformStream._readableController = c;
 
-    // delay the first source.pull call until there is backpressure
     return this._startPromise.then(() => {
-      if (transformStream._readableBackpressure === true) {
-        return true;
+      // Prevent the first pull() call until there is backpressure.
+
+      assert(transformStream._backpressureChangePromise !== undefined,
+             '_backpressureChangePromise should have been initialized');
```
isonmad: This added assert is good.

reference-implementation/lib/transform-stream.js
tyoshino: OK. How about ReadableReadyPromise?

reference-implementation/lib/transform-stream.js
tyoshino: OK. Reverted

reference-implementation/lib/transform-stream.js
tyoshino: Oops, yeah, this is not right for here. Fixed.

tyoshino:
> I still don't see how this improves code comprehensibility, honestly.

I've addressed your points. Is there anything else doesn't make sense to you?

reference-implementation/lib/transform-stream.js
```diff
@@ -45,10 +47,16 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
...
-  const backpressure = controller.desiredSize <= 0;
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() is made immediately (but
+    // asynchronously) after this because of pending read()s and fix _backpressure back to false.
+    //
+    // We don't have to worry about that such a pull() may happen synchronously to the enqueue(), and therefore is not
```
ricea: Sorry, I don't understand what "synchronously to" means here. Maybe you could explain the dangerous case as a series of steps, and then explain why it never happens?

reference-implementation/lib/transform-stream.js
```diff
@@ -97,40 +105,51 @@ function TransformStreamErrorInternal(transformStream, e) {
...
+  assert(transformStream._backpressure === true, '_backpressure should have been initialized');
 
   return transformStream._backpressureChangePromise;
 }
 
 function TransformStreamSetBackpressure(transformStream, backpressure) {
   // console.log(`TransformStreamSetBackpressure(${backpressure})`);
 
-  assert(transformStream._readableBackpressure !== backpressure);
+  // Passes also when called for initialization.
```
ricea: How about "during construction" instead of "for initialization"?

reference-implementation/lib/transform-stream.js
```diff
@@ -97,40 +105,51 @@ function TransformStreamErrorInternal(transformStream, e) {
...
+  assert(transformStream._backpressure !== backpressure,
+         'TransformStreamSetBackpressure() should be called only when backpressure is changed');
 
   if (transformStream._backpressureChangePromise !== undefined) {
+    // The fulfillment value is just for the assertion.
```
ricea: "the assertion" is ambiguous because it's not clear which assertion is being referred to. How about "just for a sanity check"?

tyoshino: Done

reference-implementation/lib/transform-stream.js
```diff
@@ -45,10 +47,16 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
...
   }
 
-  const backpressure = controller.desiredSize <= 0;
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() is made immediately (but
+    // asynchronously) after this because of pending read()s and fix _backpressure back to false.
```
tyoshino: Done

reference-implementation/lib/transform-stream.js
```diff
@@ -45,10 +47,16 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
     throw transformStream._storedError;
   }
 
-  const backpressure = controller.desiredSize <= 0;
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() is made immediately (but
```
tyoshino: Done

reference-implementation/lib/transform-stream.js
```diff
@@ -45,10 +47,16 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
...
-  const backpressure = controller.desiredSize <= 0;
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() will happen immediately (but
+    // asynchronously) after this because of pending read()s and set _backpressure back to false.
+    //
+    // We don't have to worry about that such a pull() may happen inside the enqueue() call, and therefore is not
```
ricea: How about:

> If pull() could be called from inside enqueue(), then this logic would be wrong. This cannot happen because there is always a promise pending from start() or pull() when _backpressure is false.

tyoshino: Thanks for the nice rewriting. Done

ricea:
lgtm

reference-implementation/lib/transform-stream.js
tyoshino: I felt that it's more readable to me to place logic exchanging signal between sink and source or doing something based on stuff stored inside TransformStream. TransformStream is already controlling the completion of sink.start by passing startPromise. I was viewing this as an additional delaying logic from TransformStream. And, this signal is based on observation on the source. That's my thought behind this suggestion.

But your points also makes sense. So, I'll go with your original structure.

tyoshino:
Thank you isonmad, ricea for review.

reference-implementation/lib/transform-stream.js
isonmad: The change isn't fully reverted; the original explanatory comment is gone, and the code didn't need to be in a block per [ricea's review at the original PR](https://github.com/whatwg/streams/pull/550#discussion_r84678242).

reference-implementation/lib/transform-stream.js
```diff
@@ -371,6 +395,8 @@ module.exports = class TransformStream {
     assert(this._readableController !== undefined);
 
     const desiredSize = this._readableController.desiredSize;
+    // Set _backpressure based on desiredSize. As there is no read() at this point, so we can just interpret
```
isonmad: "As there is no read()at this point, we can"

reference-implementation/lib/transform-stream.js
```diff
@@ -45,10 +47,15 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
...
+  const maybeBackpressure = controller.desiredSize <= 0;
 
-  if (transformStream._readableBackpressure !== backpressure) {
-    TransformStreamSetBackpressure(transformStream, backpressure);
+  if (maybeBackpressure === true && transformStream._backpressure === false) {
+    // This allows pull() again. When desiredSize is 0, it's possible that a pull() will happen immediately (but
+    // asynchronously) after this because of pending read()s and set _backpressure back to false.
+    //
+    // If pull() could be called from inside enqueue(), then this logic would be wrong. This cannot happen
+    // because there is always a promise pending from start() or pull() when _backpressure is false.
```
isonmad: Okay, with these two lines now I finally understand what you were worried about this entire time. I really had no idea what you were talking about.

reference-implementation/lib/transform-stream.js
tyoshino: Going to address in #577

reference-implementation/lib/transform-stream.js
tyoshino: Oh, sorry! Addressing in #577

reference-implementation/lib/transform-stream.js
tyoshino: Sorry for not being clear enough. Thanks for review.

PR #572, opened 2016-10-27
Replace some example IDs with more descriptive ones

tyoshino:


tyoshino:
Sorry for belated review, but these two of IDs introduced in #564 should be improved to be more descriptive and specific. Please review.

PR #573, opened 2016-10-27
Don't mandate transform method in TransformStream

### LINKED ISSUE #565: Default transform() method should be the identity transform
Currently `transformer.transform()` is compulsory. The TransformStream constructor will throw a TypeError if you do not set it.

I think instead if `transformer.transform` is undefined it should be equivalent to

```javascript
   controller.enqueue(chunk);
}
```

This enables two useful things:
* A transformer which just adds a fixed header and footer can be implemented with just `start()` and `flush()`.
* A transformer which just buffers can by implemented by specifying an empty _transformer_ and a strategy.

Both these things can be done by specifying the `transform()` method explicitly, but skipping it

1. Is shorter,
2. Enables optimisations by making the transformation unobservable from Javascript.

isonmad:
This was bugging me for ages but I assumed the consensus had already settled on that design.

Closes #565.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -77,7 +77,7 @@ exports.PromiseInvokeOrNoop = (O, P, args) => {
   }
 };
 
-exports.PromiseInvokeOrFallbackOrNoop = (O, P1, args1, P2, args2) => {
+exports.PromiseInvokeOrFallback = (O, P1, args1, F, args2) => {
   let method;
   try {
     method = O[P1];
@@ -86,7 +86,7 @@ exports.PromiseInvokeOrFallbackOrNoop = (O, P1, args1, P2, args2) => {
   }
 
   if (method === undefined) {
-    return exports.PromiseInvokeOrNoop(O, P2, args2);
+    return F(args2);
   }
 
   try {
@@ -94,6 +94,10 @@ exports.PromiseInvokeOrFallbackOrNoop = (O, P1, args1, P2, args2) => {
   } catch (e) {
     return Promise.reject(e);
   }
+};
+
+exports.PromiseInvokeOrFallbackOrNoop = (O, P1, args1, P2, args2) => {
+  return exports.PromiseInvokeOrFallback(O, P1, args1, exports.PromiseInvokeOrNoop, [O, P2, args2]);
 };
 
 // Not implemented correctly
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -77,7 +77,7 @@ exports.PromiseInvokeOrNoop = (O, P, args) => {
   }
 };
 
-exports.PromiseInvokeOrFallback = (O, P, args, F, argsF) => {
+exports.PromiseInvokeOrPerformFallback = (O, P, args, F, argsF) => {
   let method;
   try {
     method = O[P];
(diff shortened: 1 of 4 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -134,9 +139,15 @@ function TransformStreamTransform(transformStream, chunk) {
 
   transformStream._transforming = true;
 
+  const transformer = transformStream._transformer;
   const controller = transformStream._transformStreamController;
-  const transformPromise = PromiseInvokeOrNoop(transformStream._transformer,
-                             'transform', [chunk, controller]);
+  let transformPromise;
+  if (transformer.transform === undefined) {
```
domenic: Oh, this is not great. We should only get the `transformer.transform` once. This gets it twice, once here and once inside PromiseInvokeOrNoop. I am not sure what a good design is.

isonmad: Oh right. InvokeOrNoop already solved this: you get it, then call method.apply later if it's not undefined. But we need to either open code it or design an entirely new helper function that will only be used in exactly one caller, like `PromiseInvokeOrFallbackOrNoop` is.

domenic:
LGTM; that helper turned out pretty reasonable. I guess we should wait for @ricea and @tyoshino though, unless this is blocking any further work you wanted to do today.

isonmad:
Whoops. Reviewing it myself I saw a mistake. Tests were passing though, so I think `WritableStreamDefaultControllerAbort` must not have any tests right now, it should've broken that.

isonmad:
Historical note. Previously these kind of checks were added in cb45feb2f9295aada63a5ec96ae23bfcbe5e2626 and were removed in 1b208967327caf1396ef3c9459901ffab55adc14 as mentioned in https://github.com/whatwg/streams/issues/256#issuecomment-67711994.

reference-implementation/lib/helpers.js
```diff
@@ -77,7 +77,7 @@ exports.PromiseInvokeOrNoop = (O, P, args) => {
   }
 };
 
-exports.PromiseInvokeOrFallbackOrNoop = (O, P1, args1, P2, args2) => {
+exports.PromiseInvokeOrFallback = (O, P1, args1, F, args2) => {
```
ricea: This operation needs a different name to reflect its unique semantics.

How about `PromiseInvokeOrCallFallback` ?

reference-implementation/lib/transform-stream.js
```diff
@@ -134,9 +139,11 @@ function TransformStreamTransform(transformStream, chunk) {
 
   transformStream._transforming = true;
 
+  const transformer = transformStream._transformer;
   const controller = transformStream._transformStreamController;
-  const transformPromise = PromiseInvokeOrNoop(transformStream._transformer,
-                             'transform', [chunk, controller]);
+
+  const transformPromise = PromiseInvokeOrFallback(transformer, 'transform', [chunk, controller],
+                             TransformStreamDefaultTransform, [chunk, controller]);
```
ricea: Indentation here is odd. Indentation in this file doesn't seem particularly consistent, but I would prefer the second line to be indented after the `(` on the first line.

reference-implementation/test/transform-stream.js
```diff
@@ -4,18 +4,10 @@ const test = require('tape-catch');
 const readableStreamToArray = require('./utils/readable-stream-to-array.js');
 
 
-test('Pass-through sync TransformStream: can read from readable what is put into writable', t => {
+test('identity TransformStream: can read from readable what is put into writable', t => {
```
tyoshino: identity -> Identity
for consistency with other test descriptions

reference-implementation/lib/helpers.js
tyoshino: Agreed

The arguments should be also renamed. Maybe
- P1 -> P
- args1 -> args
- args2 -> argsF

domenic: Here F is an abstract op, right? Then we should say "PromiseInvokeOrPerformFallback" instead of "PromiseInvokeOrCallFallback"

isonmad:
> Could you file an issue for the testing gap you noticed?

Looking into it, this would have been caught is if all the `helper.js` ops were made into function-functions instead of arrow functions, and had an appropriate `assert(arguments.length === 5);` at the top. Or asserting that all arguments weren't undefined individually, because as-is it 'works' if all are undefined.

PR #577, opened 2016-10-28
Follow up for #571

tyoshino:
Address issues pointed out by isonmad after merging.


isonmad:
You mean a followup for #571?

tyoshino:
> You mean a followup for #571?

Right. Just fixed. Thanks

PR #579, opened 2016-10-30
eslint fixes and standardize exports style

isonmad:
Put exports in one place per file and use the same style in each one.

Along the way as I tried editing run-web-platform-tests.js, I found out it had lint errors.

### DIFF
--- a/reference-implementation/.eslintrc.json
+++ b/reference-implementation/.eslintrc.json
@@ -19,7 +19,7 @@
     // Possible errors
     "comma-dangle": ["error", "never"],
     "no-cond-assign": ["error", "except-parens"],
-    "no-console": "off",
+    "no-console": "error",
     "no-constant-condition": "error",
     "no-control-regex": "error",
     "no-debugger": "error",
--- a/reference-implementation/.eslintrc.json
+++ b/reference-implementation/.eslintrc.json
@@ -19,7 +19,7 @@
     // Possible errors
     "comma-dangle": ["error", "never"],
     "no-cond-assign": ["error", "except-parens"],
-    "no-console": "error",
+    "no-console": "off",
     "no-constant-condition": "error",
     "no-control-regex": "error",
     "no-debugger": "error",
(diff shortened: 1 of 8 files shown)


ricea:
lgtm, thanks.

domenic:
Can you update the .eslintignore patterns and such to ensure eslint is actually applied to the files you're fixing, and thus fails the build?

reference-implementation/.eslintrc.json
```diff
@@ -19,7 +19,7 @@
     // Possible errors
     "comma-dangle": ["error", "never"],
     "no-cond-assign": ["error", "except-parens"],
-    "no-console": "error",
+    "no-console": "off",
```
domenic: Don't turn this off globally. Turn it off inside the file that uses console, using `/*eslint-disable no-console */`

reference-implementation/run-web-platform-tests.js
```diff
@@ -21,11 +21,11 @@ wptRunner(toUpstreamTestsPath, { rootURL: 'streams/', setup })
   })
   .then(failures => {
     totalFailures += failures;
-    process.exit(totalFailures);
+    process.exitCode = totalFailures;
```
domenic: Don't change this just to get around the restriction. Use `/*eslint-disable*/`

reference-implementation/.eslintrc.json
isonmad: Ah, I wasn't familiar enough with eslint to know I could do that. Good idea.

reference-implementation/run-web-platform-tests.js
isonmad: Wait, why not? Why would the process need to exit immediately, when it's about to exit on its own naturally regardless as this is the last thing that gets run and nothing else is keeping process alive? This seems cleaner.

domenic: I guess that's true; thanks for pushing back.

isonmad:
> Can you update the .eslintignore patterns and such to ensure eslint is actually applied to the files you're fixing, and thus fails the build?

I was trying to figure out what this meant, since there are no .eslintignore files anywhere in the repo. But now I see the problem, it's the wildcards in lib/package.json.

isonmad:
Okay, I moved eslint to a blacklist of things that shouldn't be linted, instead of a whitelist of things that should; is there anything else blocking this being merged now?

PR #580, opened 2016-10-30
always check if methods are callable first

isonmad:
InvokeOrNoop invokes [Call](https://tc39.github.io/ecma262/#sec-call), which only takes actual functions, not anything that has `.apply` on it.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -66,6 +66,11 @@ exports.InvokeOrNoop = (O, P, args) => {
   if (method === undefined) {
     return undefined;
   }
+
+  if (typeof method !== 'function') {
+    throw new TypeError(`${P} is not a function`);
+  }
+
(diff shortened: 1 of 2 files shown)


ricea:
lgtm

tyoshino:
Change index.bs also?

tyoshino:
Ah, it's been not sync with the reference impl. Let's update it in batch :)

ricea:
I think index.bs was always correct. It only needs to be done explicitly for the reference implementation because it uses the .apply method to call the function.

Similarly, the Chrome implementation doesn't have this problem because it uses Function.prototype.call().

tyoshino:
Ah, OK. The problem was the `IsCallable` step in `Call (F, V [ , argumentsList ])` was not performed correctly. Got it.

Regarding not in sync, we need to apply the rename of PromiseInvokeOrFallbackOrNoop to index.bs.

PR #581, opened 2016-10-31
Editorial: export terms needed by Fetch

annevk:


annevk:
Note that I suspect that rather than taking this patch you might want to actually update your build script to insert this metadata automatically. That was a little bit beyond what I was comfortable taking on.

annevk:
Maybe these should be annotated with `type="abstract-op"` too?

PR #583, opened 2016-10-31
Add done() to the end of w-p-t Javascript files

ricea:
When running a web-platform-test in a Worker, the Javascript file must
end with a done() statement so that testharness.js knows that parsing is
complete. The extra done() statement is a no-op when running directly in
the page or via the wpt-runner.

Add done() statements to all relevant Javascript files.


domenic:
Wow, nice catch, thank you.

PR #584, opened 2016-10-31
assert invariants in InvokeOrNoop and friends

isonmad:
If this is a good idea I'll add a commit to change index.bs next.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3620,7 +3620,9 @@ A few abstract operations are used in this specification for utility purposes. W
 </div>
 
 <emu-alg>
+  1. Assert: _O_ is not *undefined*.
   1. Assert: ! IsPropertyKey(_P_) is *true*.
+  1. Assert: _args_ is a List.
   1. Let _method_ be ? GetV(_O_, _P_).
   1. If _method_ is *undefined*, return *undefined*.
   1. Return ? Call(_method_, _O_, _args_).
@@ -3646,8 +3648,11 @@ A few abstract operations are used in this specification for utility purposes. W
 </div>
 
 <emu-alg>
+  1. Assert: _O_ is not *undefined*.
   1. Assert: ! IsPropertyKey(_P1_) is *true*.
+  1. Assert: _args1_ is a List.
   1. Assert: ! IsPropertyKey(_P2_) is *true*.
+  1. Assert: _args2_ is a List.
   1. Let _method_ be GetV(_O_, _P1_).
   1. If _method_ is an abrupt completion, return <a>a promise rejected with</a> _method_.[[Value]].
   1. Let _method_ be _method_.[[Value]].
@@ -3666,7 +3671,9 @@ A few abstract operations are used in this specification for utility purposes. W
 </div>
 
 <emu-alg>
+  1. Assert: _O_ is not *undefined*.
   1. Assert: ! IsPropertyKey(_P_) is *true*.
+  1. Assert: _args_ is a List.
   1. Let _returnValue_ be InvokeOrNoop(_O_, _P_, _args_).
   1. If _returnValue_ is an abrupt completion, return <a>a promise rejected with</a> _result_.[[Value]].
   1. Otherwise, return <a>a promise resolved with</a> _returnValue_.[[Value]].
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/helpers.js
```diff
@@ -62,6 +62,10 @@ exports.IsFiniteNonNegativeNumber = v => {
 };
 
 exports.InvokeOrNoop = (O, P, args) => {
+  assert(exports.typeIsObject(O) === true);
```
domenic: This is not accurate. We use GetV, not Get. So in theory `O` could be anything.

reference-implementation/lib/helpers.js
```diff
@@ -62,6 +62,10 @@ exports.IsFiniteNonNegativeNumber = v => {
...
+  assert(typeof P === 'string');
```
domenic: This is already in the spec, sort of, as "Assert: ! IsPropertyKey(P) is true." Probably there should be a (non-exported?) helper for that in this file, which checks if typeof is string or symbol.

reference-implementation/lib/helpers.js
```diff
@@ -62,6 +62,10 @@ exports.IsFiniteNonNegativeNumber = v => {
...
+  assert(Array.isArray(args));
```
domenic: This is not accurate; the spec says "If args was not passed, let args be a new empty List."

reference-implementation/lib/helpers.js
isonmad: Ah, right, I didn't notice that distinction.

So we have to allow `underlyingSource` et al. to be a boolean, string, or symbol, and treat it like they passed `{}` i.e. no methods?

reference-implementation/lib/helpers.js
isonmad: I just noticed that, but... why does it say that? Every single caller passes a List, and this op isn't exported.

reference-implementation/lib/helpers.js
domenic: It's actually a bit weirder. For example this code will work:

```js
Number.prototype.start = controller => { ... };
new ReadableStream(5);
```

This is purely for consistency with existing platform APIs/JavaScript destructuring.

reference-implementation/lib/helpers.js
isonmad: Yeah, should probably do that, but then again, all callers are guaranteed to pass a constant string, so do we need to allow Symbols if they're never used?

domenic: I think it's just good to be consistent with Invoke().

reference-implementation/lib/helpers.js
isonmad: Given that the impetus for writing this PR was an accidental bug where `args` was not passed, when it was intended to be passed, I think it hurts, yeah.

domenic: OK, let's remove it uniformly from all these helpers then.

isonmad: The helpers were already a bit inconsistent so it's just InvokeOrNoop we'd remove it from; `PromiseInvokeOrFallbackOrNoop` never created empty Lists if `args1` or `args2` were undefined.

domenic: Also PromiseInvokeOrNoop. The others always end up calling into someone who takes care of it for them.

isonmad: Wait, right, never mind about inconsistency. Cuz you mean all of them indirectly call InvokeOrNoop, right? Since PromiseInvokeOrNoop calls that now.

domenic: Yeah, although in the spec it still has a "If args was not passed, let args be a new empty List" line which is I guess now unnecessary. We should be sure to remove that.

index.bs
```diff
@@ -3620,8 +3620,9 @@ A few abstract operations are used in this specification for utility purposes. W
 </div>
 
 <emu-alg>
+  1. Assert: _O_ is not *undefined*.
   1. Assert: ! IsPropertyKey(_P_) is *true*.
-  1. If _args_ was not passed, let _args_ be a new empty List.
+  1. Assert: _args_ is a List.
```
ricea: If we did this following the "If _args_ was not passed, let _args_ be a new empty List." step then we could satisfy the expectation that args is optional.

isonmad: As I said in https://github.com/whatwg/streams/pull/584#discussion_r85836343 that would defeat the initial point of this PR.

PR #590, opened 2016-11-02
Add code coverage reporting the reference implementation/tests

domenic:
See https://dl.dropboxusercontent.com/u/20140634/streams-spec-coverage/index.html for a preview. The gaps are interesting, although mostly minor, which is great.

### DIFF
--- a/reference-implementation/.eslintignore
+++ b/reference-implementation/.eslintignore
@@ -1,1 +1,2 @@
 web-platform-tests/
+coverage/
--- a/reference-implementation/.gitignore
+++ b/reference-implementation/.gitignore
@@ -1,7 +1,7 @@
 node_modules/
 npm-debug.log
 
-bench/results/
 web-platform-tests/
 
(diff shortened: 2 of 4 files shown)


isonmad:
I already filed https://github.com/domenic/wpt-runner/pull/3 yesterday, since it looked like a bug in wpt-runner. Did you see that?

domenic:
Ah, no, I got bit by the bug where sometimes on GitHub you aren't watching your own repos. Thanks for finding that.

isonmad:
There's a bug like that? Damn. Was hoping the mention at https://github.com/whatwg/streams/pull/579#ref-issue-186588912 would get noticed if nothing else.

domenic:
Hmm now updating wpt-runner causes a few new failures due to testharness.js updates...

domenic:
Test failures straightened out; this should be usable now.

ricea:
Awesome! lgtm

PR #591, opened 2016-11-02
test non-function sink.abort

isonmad:
This codepath was missing code coverage as revealed in #590.


domenic:
Gotta fix some broken tests first, let me finish that up...

isonmad:
For posterity: those two failures were fixed in 9312f2c7c3954ff11e412747e85b408d65b103cd and https://github.com/w3c/web-platform-tests/pull/4162.

domenic:
Ideally we should not be adding more tests in the tap format for writable streams. Would it be reasonable to ask that you put these in the to-upstream-wpts version of bad-underlying-sinks.js?

PR #592, opened 2016-11-03
sync spec and implementation of autoAllocateChunkSize

isonmad:
autoAllocateChunkSize was added in #430 as commit e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393.

The spec was out of sync with the reference implementation: it tried to round
non-integers into integers instead of rejecting them.

The reference implementation was out of sync with the spec: it was allowing
an autoAllocateChunkSize of 0 through instead of throwing a RangeError.


domenic:
Great catch, thank you.

PR #594, opened 2016-11-03
test non-function strategy.size

isonmad:
This was missing code coverage according to #590.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
+++ b/reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
@@ -3,9 +3,7 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
-<script src="/streams/resources/test-initializer.js"></script>
-
-<style src="/resources/testharness.css"></style>
+<script src="../resources/test-initializer.js"></script>
 
 <script src="bad-strategies.js"></script>
 <script>
--- a/reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
+++ b/reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
@@ -4,7 +4,8 @@
 <script src="/resources/testharnessreport.js"></script>
 <script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
 <script src="/streams/resources/test-initializer.js"></script>
-<style src="/resources/testharness.css"/>
+
+<style src="/resources/testharness.css"></style>
 
 <script src="bad-strategies.js"></script>
 <script>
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+}
+
+test(() => assert_throws(new TypeError(), () => new WritableStream({}, { size: 'a string' })),
```
ricea: While it's impressive that you could fit this test on one line, I found it difficult to understand.

I'm sorry to be boring, but if you could split the test body into two statements I would appreciate it.

reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
```diff
@@ -0,0 +1,14 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
+<script src="/streams/resources/test-initializer.js"></script>
+<style src="/resources/testharness.css"/>
+
+<script src="bad-strategies.js"></script>
+<script>
+'use strict';
+worker_test('bad-strategies.js');
+service_worker_test('bad-strategies.js');
```
ricea: `worker_test()` already runs `service_worker_test()`, so this is unnecessary.

isonmad: It does? Where is `worker_test()` defined anyway?

ricea: The "real" version is at https://github.com/w3c/web-platform-tests/blob/master/streams/resources/test-initializer.js

reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
+<script src="/streams/resources/test-initializer.js"></script>
```
ricea: Can you make this `"../resources/test-initializer.js"` for consistency with the other tests?

reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
```diff
@@ -0,0 +1,14 @@
...
+
+<style src="/resources/testharness.css"></style>
```
ricea: Is this needed? The other test files don't have it.

reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.https.html
isonmad: Sure. What do you think of a follow-up commit to change all of them to use the absolute URL though? It's not actually a file that's local in this repo like `recording-streams.js` is.

domenic: I think it would be better to use relative, as that makes it easier to move the streams directory around, in theory.

PR #595, opened 2016-11-03
update travis config

isonmad:
'stable' isn't supposed to be used anymore as a version alias, in the post-1.0 world. And optionally, should node 6.x keep being tested against as a baseline rather than exclusively testing the latest version?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
-node_js: node
+node_js:
+ - node
+ - 6
 sudo: false
 
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-node_js: stable
+node_js: node
 sudo: false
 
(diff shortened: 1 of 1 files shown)


domenic:
Stable is fine. It means latest stable.

I see no reason to test against node v6.

isonmad:
That's not exactly true, [according to the nvm devs.](https://github.com/creationix/nvm/blob/v0.32.1/README.markdown#usage)

> stable: this alias is deprecated, and only truly applies to node v0.12 and earlier. Currently, this is an alias for node.

In fact, as far as the official Travis documentation is concerned, [there is no 'stable', there is only node.](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Available-Versions)

domenic:
Yeah, I've had this argument with them a few times. When they actually remove it, we can do something, but I don't think they will. It's a fake deprecation.

PR #597, opened 2016-11-04
WPT-ify some tests

isonmad:
Also fix some inaccurate test summaries; they said 'abort should reject with the abort reason' when that wasn't what the assert asserted.


reference-implementation/to-upstream-wpts/transform-stream/general.js
```diff
@@ -41,4 +43,245 @@ test(() => {
...
+  const ts = new TransformStream();
+
+  const writer = ts.writable.getWriter();
+  writer.write('a');
+  assert_equals(writer.desiredSize, 0, 'writer.desiredSize should be 0 after write()');
+
+  return ts.readable.getReader().read().then(result => {
+    assert_equals(result.value, 'a',
+      'result from reading the readable is the same as was written to writable');
+    assert_false(result.done, 'stream is not done');
```
ricea: This is a good example of why phrasing descriptions using the word 'should' is helpful. Writing 'stream should not be done' makes it a lot clearer what the expected result is.

reference-implementation/to-upstream-wpts/writable-streams/bad-underlying-sinks.js
```diff
@@ -9,6 +9,36 @@ if (self.importScripts) {
...
 error1.name = 'error1';
 
+test(() => {
+  assert_throws(error1, () => {
+    new WritableStream({
+      get start() {
+        throw error1;
+      }
+    });
+  }, 'constructor throws same error from throwing start getter');
```
isonmad: Thanks for the pointers and the link! Is this better?

reference-implementation/to-upstream-wpts/writable-streams/bad-underlying-sinks.js
```diff
@@ -47,6 +77,48 @@ promise_test(t => {
...
+    get write() {
+      throw error1;
+    }
+  });
+
+  const writer = ws.getWriter();
+
+  return promise_rejects(t, error1, writer.write('a'), 'write should reject with the thrown error')
+  .then(() => {
+    return promise_rejects(t, error1, writer.closed, 'closed should reject with the thrown error');
```
isonmad: Wait, what do you mean?

ricea: That's weird. I intended to write {} but it came out as ().

What I was trying to say is you could just do `() => promise_rejects(t, error1, writer.closed, 'closed should reject with the thrown error')`.

ricea:
lgtm. Thanks for the port!

ricea:
lgtm again

PR #598, opened 2016-11-04
use abstract ops in TransformStream implementation

isonmad:


reference-implementation/lib/readable-stream.js
```diff
@@ -253,7 +253,14 @@ class ReadableStream {
   }
 }
 
-module.exports = { ReadableStream, IsReadableStreamDisturbed };
+module.exports = {
+  ReadableStream,
+  IsReadableStreamDisturbed,
```
ricea: You appear to have exported this but not imported it?

isonmad: But I didn't export it, someone else did.

ricea: Ooops. Sorry.

In that case, lgtm.

Issue #174, opened 2014-08-21
Need new names for { input, output }


domenic:
Originally raised in https://github.com/sysapps/tcp-udp-sockets/issues/69.

My thinking has been that you write things in to the input (a writable stream), and you read them out of the output (a readable stream).

Apparently this is at odds with most other programming libraries, for which you write to output and read from input. Unix stdin/stdout is an example, but also some Java stream APIs.

We could just switch them, and that's my fallback position. I am curious if there is a better name pair.

PR #601, opened 2016-11-05
add "bytes" readableType to TransformStream transformer

isonmad:
Leaves room for writableType as well, which the WritableStream constructor will reject if it's anything other than undefined, at the moment.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -3,7 +3,10 @@ const assert = require('assert');
 const { InvokeOrNoop, PromiseInvokeOrPerformFallback, PromiseInvokeOrNoop, typeIsObject } = require('./helpers.js');
 const { ReadableStream, ReadableStreamDefaultControllerClose,
         ReadableStreamDefaultControllerEnqueue, ReadableStreamDefaultControllerError,
-        ReadableStreamDefaultControllerGetDesiredSize } = require('./readable-stream.js');
+        ReadableStreamDefaultControllerGetDesiredSize,
+        ReadableByteStreamControllerClose, ReadableByteStreamControllerEnqueue,
+        ReadableByteStreamControllerError, ReadableByteStreamControllerGetBYOBRequest,
+        ReadableByteStreamControllerGetDesiredSize } = require('./readable-stream.js');
 const { WritableStream, WritableStreamDefaultControllerError } = require('./writable-stream.js');
 
 // Methods on the transform stream controller object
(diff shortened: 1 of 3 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -379,23 +435,26 @@ class TransformStream {
 
     this._transformStreamController = new TransformStreamDefaultController(this);
 
+    this._readableType = transformer.readableType;
```
ricea: Is there a reason not to validate the value of `readableType` and `writableType` here? It seems weird to me.

domenic:
I tend to agree we should be conservative.

reference-implementation/lib/transform-stream.js
isonmad: I was relying on the other two stream constructors already validating it, but I suppose not.

isonmad:
Is it not useful to allow optimized 'objects-to-bytes' serializing transforms, even before the other 2 combinations of bytes-to-bytes and bytes-to-objects?

tyoshino:
This attempt clarified some issues regarding extension of TransformStream to more variants corresponding to the (current and future) RS/WS variants. So, it's really useful. Thanks. E.g. I think we don't want to make the TransformStreamController be an all-in-one class with disabled methods (e.g. the byobRequest getter). OTOH, we also don't want to have a lot of TransformStream variants defined for each combination.

The TransformStream class is basically a helper for implementing stuff following the transform streams concept and explanation of one reasonable backpressure handling. No one is disallowed to directly use the ReadableXXXStream and WritableXXXStream to build a TransformStream. I guess we shouldn't bother ourselves for maintaining a lot of wrappers.

I agree that objects-to-bytes use cases are not uncommon. But I think we should be careful not to inflate the spec. Can we componentize the TransformStream class to avoid the all-in-one controller and also avoid combinatorial explosion? Such attempt could also result in some additional complexity, but my gut feeling is we should explore that.

domenic:
An obvious objects-to-bytes use case is TextEncoder.

ricea:
I created #616 to discuss TransformStream byte streams independently of this particular pull request. Hopefully we can decide the direction there and then come back here.

ricea:
Random observation: I find writableType/readableType confusing and would prefer to have inputType/outputType.

domenic:
Originally transform streams were { input, output } but people found this confusing :(. https://github.com/whatwg/streams/issues/174

ricea:
I didn't realise that this had already been bikeshedded. 

I would like to be pedantic and say that writableType/readableType are from the point of view of the author of the transform, whereas {writable, readable} are from the point of view of the consumer of the transform. But even I don't find this particularly convincing, so I withdraw my objection.

tyoshino:
> I would like to be pedantic and say that writableType/readableType are from the point of view of the author of the transform, whereas {writable, readable} are from the point of view of the consumer of the transform. 

This is an interesting point of view. But yes, let's keep the readable/writable naming.

tyoshino:
> An obvious objects-to-bytes use case is TextEncoder.

Ah, yeah! I didn't realize that well. So, given what we're trying to ship for the first time, TextDecoder and TextEncoder are asymmetric, I'm a little more convinced that we could include this in the spec tentatively.

PR #603, opened 2016-11-05
port piping tests to WPT

isonmad:
This includes the last user of several helper functions
which had already been ported over to WPT in
commit fd44398d410beb6879fb27290e690603bed04c4d.

Although, not sure what to do with that last piping test that was disabled in
commit 032c4a8ef065a8353ed5987ed2554f3cec63566d in 2014. It couldn't work
as-is, as despite several updates to it to track API changes it actually
still relies on `controller.enqueue()` returning a boolean.

Rewrite or delete entirely and write different backpressure tests?


reference-implementation/to-upstream-wpts/piping/general.js
```diff
@@ -154,4 +155,55 @@ promise_test(() => {
...
+  const rs = new ReadableStream({
+    start(c) {
+      c.enqueue('a');
+      c.enqueue('b');
+      c.enqueue('c');
+      c.close();
+    }
+  });
+
+  const ts = new TransformStream();
```
ricea: I don't think we need the `ts` variable. I think it would be clearer just to create the TransformStream object inline.

ricea:
lgtm

PR #604, opened 2016-11-06
WPT-ify most of the remaining tape tests

isonmad:


reference-implementation/to-upstream-wpts/writable-streams/brand-checks.js
```diff
@@ -0,0 +1,96 @@
...
+const WritableStreamDefaultWriter = writer.constructor;
+const WriterProto = WritableStreamDefaultWriter.prototype;
+
+test(() => {
+  getterThrows(WriterProto, 'desiredSize', fakeWritableStreamDefaultWriter());
+  getterThrows(WriterProto, 'desiredSize', realReadableStreamDefaultWriter());
+}, 'WritableStreamDefaultWriter.prototype.desiredSize enforces a brand check');
+
+promise_test(t => {
+  const results = [];
```
domenic: I think all your Promise.all tests would be easier to understand if they were just inline, e.g. `return Promise.all([\n...\n...\n]);`

isonmad: Sure. That works for this one, but some tests used a for loop so that doesn't work for them without a bigger rework.

reference-implementation/to-upstream-wpts/writable-streams/bad-strategies.js
```diff
@@ -4,6 +4,89 @@ if (self.importScripts) {
...
+      size() {
+        return size;
+      },
+      highWaterMark: 5
+    });
+
+    const writer = ws.getWriter();
+
+    return Promise.all([
+      promise_rejects(t, new RangeError(), writer.write('a').catch(r => {
```
ricea: Use promise_rejects is just making this more complicated and harder to understand, so I suggest doing something closer to the original. For example (untested):

```javascript
return writer.write('a').then(
    () => assert_unreached('write should not fulfill');
    r => {
      assert_equals(r.name, 'RangeError', `write should reject with a RangeError for ${size}`);
      let theError = r;
      return writer.closed.then(
          () => assert_unreached('closed should not fulfill'),
          e => assert_equals(e, theError, `closed should reject with the same error for ${size}`));
      ));
```

This removes the implicit test that the write() promise fulfills before the closed promise.  If you are feeling energetic you could add an explicit test of that fact.

reference-implementation/to-upstream-wpts/transform-stream/errors.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const ts = new TransformStream({
+    transform() {
+      throw thrownError;
+    }
+  });
+
+  const reader = ts.readable.getReader();
+
+  results.push(promise_rejects(t, thrownError, reader.read(),
```
ricea: I agree with @domenic that the |results| array doesn't really help with readability.

reference-implementation/to-upstream-wpts/transform-stream/errors.js
```diff
@@ -0,0 +1,65 @@
...
+
+  results.push(promise_rejects(t, thrownError, reader.closed,
+                               'readable\'s closed should be rejected with the thrown error'));
+
+  const writer = ts.writable.getWriter();
+
+  results.push(promise_rejects(t, thrownError, writer.closed,
+                               'writable\'s closed should be rejected with the thrown error'));
+
+  writer.write('a');
```
ricea: It's odd that this call to write() is just sitting here with no obvious purpose. I know you didn't add it, and I'm not asking you to remove it. I'm just making a random observation.

ricea: The test rewrite made it clear that write() actually is necessary for the test. Thank you!

ricea:
lgtm, this is great, thank you.

Issue #548, opened 2016-10-20
Tests contain non-zero timers


ricea:
Short timers make tests flaky. From the [Testharness.js API Documentation](http://testthewebforward.org/docs/testharness-library.html):

> Note that timeouts generally need to be a few seconds long in order to produce stable results in all test environments.

Long timeouts slow down all developers by making the the tests take longer to run. They also waste resources on automatic testing infrastructure.

Mostly timers are being used to ensure that the implementation never takes some action that it shouldn't. For example (from [writable-streams/start.js](https://github.com/whatwg/streams/blob/master/reference-implementation/to-upstream-wpts/writable-streams/start.js)):

```javascript
promise_test(() => {
  const ws = recordingWritableStream({
    start() {
      return Promise.reject();
    }
  });

  // Wait and verify that write or close aren't called.
  return delay(100)
      .then(() => assert_array_equals(ws.events, [], 'write and close should not be called'));
}, 'underlying sink\'s write or close should not be invoked if the promise returned by start is rejected');
```

This test never[1] flakes on a correct implementation. Suppose hypothetically that there existed an implementation which called `close()` here after a asynchronously performing an action which takes an arbitrary amount of time. Then this test would be flaky on that implementation: it would fail when the action took <100ms, and pass otherwise.

I propose that we make the simplifying assumption that no reasonable implementation will do the wrong thing after a delay. We can then disregard the above hypothetical implementation as being unreasonable, and so consider its false positive on the above test to be acceptable.

I furthermore propose that we consider two times around the event loop to be long enough to see whether  an implementation is going to do the wrong thing or not. My reasoning is that the implementation may run a callback after the user code, which then posts another async callback which finally does the wrong thing.

I think we should have a function `flushAsyncEvents()` which is equivalent to `delay(0).then(delay(0))`. The above test would then look like

```javascript
promise_test(() => {
  const ws = recordingWritableStream({
    start() {
      return Promise.reject();
    }
  });

  // Verify that write or close aren't called.
  return flushAsyncEvents()
      .then(() => assert_array_equals(ws.events, [], 'write and close should not be called'));
}, 'underlying sink\'s write or close should not be invoked if the promise returned by start is rejected');
```

[1] Test frameworks usually have a hard timeout, which means that even passing tests will fail occasionally. The longer a test takes, the more likely it is to hit the hard timeout. This is seen in practice on the Blink layout tests.

domenic:
[...]

PR #605, opened 2016-11-09
Remove trivial non-zero delays in wpts

ricea:
Add a util function flushAsyncEvents() which spins the event loop to
ensure that the implementation doesn't do something it shouldn't
asynchronously.

Use this function to replace trivial uses of delay(100) and similar  to
improve test determinism.

This change relates to #548.

Tests which use multiple delays to express ordering requirements have
not been changed. Those tests require individual analysis.

### DIFF
--- a/reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
+++ b/reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
@@ -286,7 +286,7 @@ promise_test(t => {
   const ws = recordingWritableStream({
     write() {
       if (ws.events.length > 2) {
-        return delay(10).then(() => {
+        return flushAsyncEvents().then(() => {
           throw error1;
         });
       }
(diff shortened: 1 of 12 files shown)


reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
```diff
@@ -286,7 +286,7 @@ promise_test(t => {
   const ws = recordingWritableStream({
     write() {
       if (ws.events.length > 2) {
-        return delay(10).then(() => {
+        return flushAsyncEvents().then(() => {
```
ricea: Okay, where we are not waiting for something to not happen, but instead just testing async behaviour, I have changed the flushAsyncEvents() to delay(0). delay(0) is currently the usual way to say "do this asynchronously" in these tests.

The net result of this is that a lot of the changes just end up changing delay(n) to delay(0), but that still satisfies the goal of reducing use of delay(0).

tyoshino:
lgtm. thanks

PR #607, opened 2016-11-10
Mark promise rejections for "state promises" as handled

### LINKED ISSUE #547: Marking promise rejections as handled
See https://bugs.chromium.org/p/chromium/issues/detail?id=654701 for background /cc @yutakahirano @jakearchibald. Also https://github.com/whatwg/streams/issues/468 is kind of related, but that issue was about pipeTo which was confusing since pipeTo was under flux at the time.

A simple repro ([run in Chrome](http://jsbin.com/sopogoyoyi/edit?html,output)) is

``` js
window.onunhandledrejection = e => {
  console.log(e.reason);
};

  start(controller) {
    controller.close();
  }
});
const reader = rs.getReader();
reader.releaseLock();
```

This will log an unhandled rejection (and also show up in the console in Chrome, but that's a browser-specific devtools thing).

I think the correct solution is that all "state promises" like `.closed` should be marked as having their rejection handled (i.e. their [[PromiseIsHandled]] slot should be set to true).

domenic:
This fixes #547, by marking the state promises (reader.closed, writer.closed, and writer.ready) as handled whenever they are rejected. It also overhauls the tests to handle all rejections, as apparently this will be enforced by testharness.js soon: see https://github.com/w3c/testharness.js/commit/7716e2581a86dfd9405a9c00547a7504f0c7fe94.

In the process, this adds additional test coverage for many rejections: since we have to handle them anyway, we might as well test them with promise_rejects. It also led to a fix to the completely-broken test "WritableStream should call underlying sink's close if no abort is supplied" in aborting.js. Finally, it uncovered an issue where ongoing write() calls would reject if abort() is called subsequently, even if the ongoing write succeeds according to the underlying sink.

---

This is not yet ready to merge as a test is still failing, for the write() + abort() case. It is the test "Aborting a WritableStream prevents further writes after any that are in progress", which I have updated to expect that the write() fulfills, but currently it fails with the aborted TypeError.

The problem is essentially that abort() synchronously rejects all pending write requests, whereas WritableStreamDefaultControllerProcessWrite does not unshift the pending write request until the underlying sink's write() call succeeds, which happens asynchronously.

We should instead unshift the pending write request the moment the underlying sink write() call is made, I think. This matches better the idea that once the underlying sink write() has started, it is entirely up to the underlying sink whether the writer.write() call should fulfill or reject. But a quick attempt to do that leads to lots of test timeouts and other failures. I would appreciate help investigating ways of making the test pass.

Alternately we could mark the test as skipped for now (I guess in WPT land that would mean commenting it out and filing an issue) just to get the promise rejection stuff underway.

### DIFF
--- a/reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
+++ b/reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
@@ -41,9 +41,11 @@ promise_test(t => {
   });
 
   const writer = ws.getWriter();
-  writer.write('Hello');
 
-  return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error')
+  return promise_rejects(t, error1, writer.write('Hello'), 'writer.write() must reject with the write error')
+    .then(() => {
+      return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error');
+    })
     .then(() => {
       writer.releaseLock();
 
@@ -72,9 +74,11 @@ promise_test(t => {
   });
 
   const writer = ws.getWriter();
-  writer.write('Hello');
 
-  return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error')
+  return promise_rejects(t, error1, writer.write('Hello'), 'writer.write() must reject with the write error')
+    .then(() => {
+      return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error');
+    })
     .then(() => {
       writer.releaseLock();
 
@@ -100,9 +104,11 @@ for (const falsy of [undefined, null, false, +0, -0, NaN, '']) {
     });
 
     const writer = ws.getWriter();
-    writer.write('Hello');
 
-    return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error')
+    return promise_rejects(t, error1, writer.write('Hello'), 'writer.write() must reject with the write error')
+      .then(() => {
+        return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error');
+      })
       .then(() => {
         writer.releaseLock();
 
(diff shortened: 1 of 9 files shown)


reference-implementation/to-upstream-wpts/piping/error-propagation-backward.js
```diff
@@ -41,9 +41,11 @@ promise_test(t => {
   });
 
   const writer = ws.getWriter();
-  writer.write('Hello');
 
-  return promise_rejects(t, error1, writer.closed, 'writer.closed must reject with the write error')
+  return promise_rejects(t, error1, writer.write('Hello'), 'writer.write() must reject with the write error')
+    .then(() => {
```
ricea: How about doing without the {} here?

domenic:
OK, this should be ready to go. Here is a proposed commit message:

```
Mark promise rejections for "state promises" as handled

This fixes #547, by marking the state promises (reader.closed, writer.closed, and writer.ready) as handled whenever they are rejected. It also overhauls the tests to handle all rejections, as apparently this will be enforced by testharness.js soon: see https://github.com/w3c/testharness.js/commit/7716e2581a86dfd9405a9c00547a7504f0c7fe94.

In the process, this adds additional test coverage for many rejections: since we have to handle them anyway, we might as well test them with promise_rejects. It also led to a fix to the completely-broken test "WritableStream should call underlying sink's close if no abort is supplied" in aborting.js. Finally, it uncovered an issue where ongoing write() calls would reject if abort() is called subsequently, even if the ongoing write succeeds according to the underlying sink; this led to temporarily commenting out the test "Aborting a WritableStream prevents further writes after any that are in progress", and opening #611 to track fixing that.

In terms of spec changes, this also discovered that WritableStreamDefaultWriter was returning rejected promises that its caller had no use for. This has been fixed as well.
```

ricea:
lgtm

tyoshino:
I found that the reference implementation will be readable and easier to maintain if we move the `.catch(() => {})` to the promise fulfilling/rejecting/resetting helper functions ( https://github.com/whatwg/streams/commit/41c46c263eca3c1ce25fceff97dc7f4a24469569 ).

I attempted to do this refactoring, but noticed that considering that the purpose of the reference implementation is for checking the correctness of the specification, it might be better to keep `.catch(() => {})` placed right next to each helper function invocation to keep them in sync with the spec text.

Thoughts? If you agree, I just do nothing.

ricea:
@tyoshino I already had this conversation inside my head. Maybe I should have mentioned it to someone else. Anyway, I came to the same conclusion that you did: that it's important to do the steps in the same order and granularity as the standard, even if it sometimes makes the reference implementation a bit unnatural.

tyoshino:
@ricea Thanks. Then ok!

PR #608, opened 2016-11-10
skip WPT dir if not present

isonmad:
As is it treats you not having a fully checked out web-platform-tests directory as a failure instead of skipping it.

### DIFF
--- a/reference-implementation/run-web-platform-tests.js
+++ b/reference-implementation/run-web-platform-tests.js
@@ -1,6 +1,7 @@
 // This runs the web platform tests against the reference implementation, in Node.js using jsdom, for easier rapid
 // development of the reference implementation and the web platform tests.
 'use strict';
+const fs = require('fs');
 const path = require('path');
 const wptRunner = require('wpt-runner');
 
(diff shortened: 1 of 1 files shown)


domenic:
Why? If you don't pass the readable stream tests, you should fail.

isonmad:
The WPT repo is out of tree and kind of enormous; if it's completely mandatory to run the out of tree tests every time you want to run the in tree tests, should it be a git submodule instead of something manually checked out if you run a script?

domenic:
I find submodules harder to use than Bash scripts, so I went with this approach. Note that after you've done an initial checkout future runs of the script are basically instant.

ricea:
It might be good to use submodules so that we could keep the WPT repo on a known-good revision. At the moment if an upstream change broke our tests we'd have to either hack run-web-platform-tests.js or wait for the problem to be fixed in the WPT repo.

Disclaimer: I've never used git submodules so I don't know how painful this would be in practise.

domenic:
Let's close this. If someone wants to investigate using git submodules and documenting how to integrate them then that might be worthwhile.

PR #609, opened 2016-11-10
rework recording-streams helper

isonmad:
This is the first building block towards being able to record the events of multiple streams in a single array for future backpressure tests.

The old, flattened `events` array is easy to use with `assert_array_equals`, so it's preserved with a few getter properties on top of the more detailed logging.

The next step is creating a 'factory' object that creates recording writable, readable, and transform streams, and records both when the sink/source/transformer methods are invoked, and when their returned promises fulfill, all in one array so you can see the relative ordering of them all to each other without relying on timeouts. `recorderFactory.createReadable()` et al. would take a new `name` argument to label which stream an event came from, etc.

### DIFF
--- a/reference-implementation/to-upstream-wpts/piping/flow-control.js
+++ b/reference-implementation/to-upstream-wpts/piping/flow-control.js
@@ -84,12 +84,8 @@ promise_test(() => {
 
   const rs = recordingReadableStream();
 
-  const startPromise = Promise.resolve();
   let resolveWritePromise;
   const ws = recordingWritableStream({
-    start() {
-      return startPromise;
-    },
     write() {
       if (!resolveWritePromise) {
         // first write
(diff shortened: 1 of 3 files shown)


domenic:
This seems a lot more complicated :(. I'd appreciate keeping recording-streams.js as-is, and creating this new helper as a separate file, if it's necessary?

ricea:
I strongly agree with the principle of recording and testing the relative ordering of events, rather than relying on timeouts.

However, the complexity of these test utilities makes me very uncomfortable.

One of the principles I apply when writing test fixtures or utilities is that it's better to have several small specific ones than one large general one. There's a couple of reasons for this:

1. Generally you only need to understand one test case at a time (the one that's broken). Therefore minimising the size of the dependencies for each test is a higher priority than minimising the total size of the test suite.
2. Test fixtures and utils are generally not themselves tested. So the need for them to be "obviously correct" is even more extreme than for other kinds of code.

The existing recordingReadableStream is already at the upper limit of complexity I am comfortable with in test utilities.

I agree with @domenic about keeping recording-streams.js as-is. Then I would recommend thinking about the simplest possible test utility that would let you write the tests you need right now. I'd rather have five 10-line test utilities than one 30-line one.

domenic:
Let's close this in its current form. I hope you're still up for writing backpressure tests though, @isonmad!

PR #612, opened 2016-11-14
Editorial: consistently use minimal returns

### LINKED ISSUE #587: Editorial: "return undefined." vs. "return"
https://streams.spec.whatwg.org/#default-writer-release-lock

> return undefined

There seems to be inconsistent usage between "return undefined." and "return.". I think if a method or operation has no explicit return in its main path, then other paths should use "return." rather than "return undefined.".

There also appear to be some operations that have an unnecessary explicit "return undefined." at the end, but maybe I am just missing the reason for it.

domenic:
That is, just say "return" instead of "return undefined", and don't return explicitly at the end of algorithms. Fixes #587.

---

Unlike fbe8b561a0021d0ef3e7955f8c7380baef47c1f0 I'm not as confident I caught everything here, so I'm putting this up as a PR instead of commiting directly to master, in case someone else wants to take a second look.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3624,7 +3624,7 @@ A few abstract operations are used in this specification for utility purposes. W
   1. Assert: ! IsPropertyKey(_P_) is *true*.
   1. Assert: _args_ is a List.
   1. Let _method_ be ? GetV(_O_, _P_).
-  1. If _method_ is *undefined*, return.
+  1. If _method_ is *undefined*, return *undefined*.
   1. Return ? Call(_method_, _O_, _args_).
 </emu-alg>
 
--- a/index.bs
+++ b/index.bs
@@ -826,10 +826,10 @@ noticable asymmetry between the two branches. [[!HTML]]
   1. Set _pull_.[[branch1]] to _branch1Stream_.[[readableStreamController]].
   1. Set _pull_.[[branch2]] to _branch2Stream_.[[readableStreamController]].
   1. <a>Upon rejection</a> of _reader_.[[closedPromise]] with reason _r_,
-    1. If _teeState_.[[closedOrErrored]] is *true*, return.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -826,7 +826,7 @@ noticable asymmetry between the two branches. [[!HTML]]
   1. Set _pull_.[[branch1]] to _branch1Stream_.[[readableStreamController]].
   1. Set _pull_.[[branch2]] to _branch2Stream_.[[readableStreamController]].
   1. <a>Upon rejection</a> of _reader_.[[closedPromise]] with reason _r_,
-    1. If _teeState_.[[closedOrErrored]] is *true*, return *undefined*.
+    1. If _teeState_.[[closedOrErrored]] is *true*, return.
```
ricea: I feel that this one should be return *undefined*, since ReadableStreamTee does sometimes return a defined value.

domenic: This is actually inside the "closure" created by "upon rejection". But probably it's better to not use "return" in such steps at all, as it's somewhat ambiguous. Let me restructure.

index.bs
```diff
@@ -3628,7 +3624,7 @@ A few abstract operations are used in this specification for utility purposes. W
   1. Assert: ! IsPropertyKey(_P_) is *true*.
   1. Assert: _args_ is a List.
   1. Let _method_ be ? GetV(_O_, _P_).
-  1. If _method_ is *undefined*, return *undefined*.
+  1. If _method_ is *undefined*, return.
```
ricea: I think this would be a good place to explicitly return *undefined*.

This is the only one I've found.

index.bs
ricea: Thanks. Sorry for not reading this carefully enough.

ricea:
I have checked every return statement and I don't think you missed any.

lgtm.

PR #613, opened 2016-11-16
Invoking methods should not call .apply()

ricea:
The specification documents InvokeOrNoop() and friends as using the
Ecmascript Call() abstract operation. The reference implementation was
calling .apply() on the method instead. Fix it to use
Function.prototype.apply.call() instead.

Also add tests to verify that .apply() and .call() on the underlyingSink
and transformer methods are not called.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -45,8 +45,8 @@ exports.IsFiniteNonNegativeNumber = v => {
 };
 
 function Call(F, V, args) {
-  if (typeof method !== 'function') {
-    throw new TypeError(`${P} is not a function`);
+  if (typeof F !== 'function') {
+    throw new TypeError('Argument is not a function');
   }
 
   return Function.prototype.apply.call(F, V, args);
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -45,6 +45,10 @@ exports.IsFiniteNonNegativeNumber = v => {
 };
 
 function Call(F, V, args) {
+  if (typeof method !== 'function') {
+    throw new TypeError(`${P} is not a function`);
+  }
+
   return Function.prototype.apply.call(F, V, args);
 }
 
@@ -56,10 +60,6 @@ exports.InvokeOrNoop = (O, P, args) => {
   const method = O[P];
   if (method === undefined) {
     return undefined;
-  }
-
-  if (typeof method !== 'function') {
-    throw new TypeError(`${P} is not a function`);
   }
 
   return Call(method, O, args);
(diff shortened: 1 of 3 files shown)


reference-implementation/lib/helpers.js
```diff
@@ -58,7 +62,7 @@ exports.InvokeOrNoop = (O, P, args) => {
     throw new TypeError(`${P} is not a function`);
   }
 
-  return method.apply(O, args);
+  return Call(method, O, args);
```
tyoshino: It seems including the L61-63 in the Call method you've just introduced would give them better correspondence because the lines are performing the step `If IsCallable(F) is false, throw a TypeError exception.` of the Call operation ( https://tc39.github.io/ecma262/#sec-call ).

domenic:
I'm not opposed to this if these tests are helpful for implementers, such as yourself.

However, I've been resistant in the past because where do we stop? For example, this is robust against passing in an `{ apply() { } }` object. But it's not robust against overriding `Function.prototype.apply`. Should we be guarding against that? Similarly, we operate on promises a lot by using the built-in `Promise.resolve` and friends, which is not robust against overrides.

In general this kind of comes down to what the role of tests and the reference implementation are. I'm of the opinion the reference implementation is not meant to be perfect in every way; otherwise it would become much less easy to read and develop. Instead it's meant to provide something useful we can write tests against.

Similarly, the goals of tests is not to necessarily hit every possible way something could be implemented wrong; they kind of assume a base level of things being done correctly, in many cases, and focus on testing the important logic.

But, again, any philosophical considerations here are entirely overridden by what works best for implementers. If these sorts of tests are useful, then we should include them. And that means updating the reference implementation to pass them. So if this is useful to you, I'm happy to see this merged in.

ricea:
I noticed that the Chrome implementations had this bug. I considered adding Chrome-specific tests for it, but I concluded that if other browsers implemented streams in Javascript they might have the same issue.[1]

Having added a web platform test I then needed to make the reference implementation pass it. I can modify the commit message to emphasize the added tests and just mention that the reference implementation passes.

I think for polyfills it's expected that you can tweak their behaviour by replacing the implementation of Promise, etc., and the same applies to the reference implementation.

The place where I personally draw the line is that the reference implementation should be robust against bad input but not against poisoning the global scope.

[1] Although I just checked Firefox's in-progress implementation and they don't have this bug. Actually, they appear to have a check in their parser to prevent this whole class of bugs: https://dxr.mozilla.org/mozilla-central/source/js/src/frontend/Parser.cpp#8475

tyoshino:
Oops. Moving the IsCallable requires passing the method name for throwing an exception... I think we can just omit the function name from the description string.

ricea:
Agreed. I have done so. I will have to do something fancier for the Chrome implementation.

PR #621, opened 2016-11-25
Do not assert() when writer is detached during close()

ricea:
When writer.releaseLock() was called during us.close(), an assert would
fire in WritableStreamFinishClose() as it expected the writer to still
be attached.

Since detaching the writer during close() is permitted, remove the
assert.

Also add tests that this works correctly.


domenic:
It's a bit surprising to me that close() and .closed have different behaviors, but I guess it is an edge case. LGTM.

PR #623, opened 2016-12-04
Specify ReadableStream.[[Transfer]]

isonmad:
See also #276.

As @domenic mentioned in https://github.com/whatwg/streams/issues/244#issuecomment-257041201, this effort seems more or less blocked on https://github.com/whatwg/html/issues/935, but this at least gives a general idea of the parts it would need to touch.

In particular, there's an apparent problem with transferring errored streams (namely `[[storedError]]`), as well as the open question of what even ends up in [[storedError]] if the underlyingSource throws or calls `controller.error()` it from the original realm with something uncloneable.

StructuredClone doesn't seem to be actually possible to truly polyfill as it iterates over objects in a different order than for-in does, and there doesn't seem to be an obvious way to replicate it; is there any reason StructuredClone isn't/ shouldn't be exposed on its own by the JS engine, for admittedly really niche use cases like this?

***
<a href="https://whatpr.org/streams/623.html" title="Last updated on Jan 15, 2021, 7:26 AM UTC (ebe73f8)">Preview</a> | <a href="https://whatpr.org/streams/623/872188a...ebe73f8.html" title="Last updated on Jan 15, 2021, 7:26 AM UTC (ebe73f8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -376,7 +376,7 @@ Instances of {{ReadableStream}} are created with the internal slots described in
     </tr>
   </thead>
   <tr>
-    <td>\[[Detached]]
+    <td>{{[[Detached]]}}</a>
     <td>A boolean flag set to <emu-val>true</emu-val> when the stream has been transferred away to another realm
   </tr>
   <tr>
@@ -452,7 +452,7 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
 <emu-alg>
   1. Set *this*.[[state]] to `"readable"`.
   1. Set *this*.[[reader]] and *this*.[[storedError]] to *undefined*.
-  1. Set *this*.[[Detached]] to *false*.
+  1. Set *this*.<a idl>[[Detached]]</a> to *false*.
   1. Set *this*.[[disturbed]] to *false*.
   1. Set *this*.[[readableStreamController]] to *undefined*.
   1. Let _type_ be ? GetV(_underlyingSource_, `"type"`).
@@ -742,7 +742,7 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
 
 The following internal method is implemented by each {{ReadableStream}} instance.
 
-<h5 id="rs-internal-method-transfer">\[[Transfer]](<var>targetRealm</var>)</h5>
+<h5 id="rs-internal-method-transfer"><a idl lt="[[Transfer]]()">\[[Transfer]](<var>targetRealm</var>)</a></h5>
 
 <emu-alg>
   1. If ! IsReadableStreamLocked(*this*) is *true*, throw a *TypeError* exception.
@@ -758,7 +758,7 @@ The following internal method is implemented by each {{ReadableStream}} instance
   1. Repeat for each Record {[[value]], [[size]]} _pair_ that is an element of _queue_,
     1. Set _pair_.[[value]] to ! <a abstract-op>StructuredClone</a>(_pair_.[[value]], _targetRealm_).
   1. Set _controller_.[[targetRealm]] to _targetRealm_.
-  1. Set _this_.[[Detached]] to *true*.
+  1. Set _this_.<a idl>[[Detached]]</a> to *true*.
   1. Return _that_.
 </emu-alg>
 
@@ -814,7 +814,7 @@ readable stream is <a>locked to a reader</a> or has been transferred away to ano
 
 <emu-alg>
   1. Assert: ! IsReadableStream(_stream_) is *true*.
-  1. If _stream_.[[Detached]] is *true*, return *true*.
+  1. If _stream_.<a idl>[[Detached]]</a> is *true*, return *true*.
   1. If _stream_.[[reader]] is *undefined*, return *false*.
   1. Return *true*.
 </emu-alg>
@@ -970,7 +970,7 @@ nothrow>ReadableStreamAddReadIntoRequest ( <var>stream</var> )</h4>
 <var>reason</var> )</h4>
 
 <emu-alg>
-  1. Assert: _stream_.[[Detached]] is *false*.
+  1. Assert: _stream_.<a idl>[[Detached]]</a> is *false*.
   1. Set _stream_.[[disturbed]] to *true*.
   1. If _stream_.[[state]] is `"closed"`, return <a>a promise resolved with</a> *undefined*.
   1. If _stream_.[[state]] is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
--- a/index.bs
+++ b/index.bs
@@ -376,7 +376,7 @@ Instances of {{ReadableStream}} are created with the internal slots described in
     </tr>
   </thead>
   <tr>
-    <td>{{[[Detached]]}}</a>
+    <td>{{[[Detached]]}}
     <td>A boolean flag set to <emu-val>true</emu-val> when the stream has been transferred away to another realm
   </tr>
   <tr>
--- a/index.bs
+++ b/index.bs
@@ -461,10 +461,10 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
     1. If _highWaterMark_ is *undefined*, let _highWaterMark_ be *0*.
     1. Set *this*.[[readableStreamController]] to ? Construct(`<a idl>ReadableByteStreamController</a>`, « *this*,
        _underlyingSource_, _highWaterMark_ »).
-  1. Otherwise, if _type_ is *undefined*,
+  1. Otherwise, if _type_ is *undefined* or _type_ is `"cloning"`,
     1. If _highWaterMark_ is *undefined*, let _highWaterMark_ be *1*.
     1. Set *this*.[[readableStreamController]] to ? Construct(`<a idl>ReadableStreamDefaultController</a>`, « *this*,
-       _underlyingSource_, _size_, _highWaterMark_ »).
+       _underlyingSource_, _size_, _highWaterMark_, _type_ »).
   1. Otherwise, throw a *RangeError* exception.
 </emu-alg>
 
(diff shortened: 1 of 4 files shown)


ricea:
I want to apologise for not reviewing this yet. I'm very interested in it, but I've just been busy finishing up stuff for the end of the year. Hopefully someone else will get to it soon.

index.bs
```diff
@@ -733,6 +738,30 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+  1. Let _controller_ be *this*.[[readableStreamController]].
+  1. If _controller_.[[targetRealm]] is *undefined*, throw a *TypeError* exception.
+  1. Let _that_ be a new instance of <a idl>ReadableStream</a> in _targetRealm_.
+  1. Set _that_.[[state]] to  *this*.[[state]].
+  1. Set _that_.[[disturbed]] to *this*.[[disturbed]].
+  1. Set _controller_.[[controlledReadableStream]] to _that_.
+  1. Set _that_.[[readableStreamController]] to _controller_.
+  1. Let _queue_ be _controller_.[[queue]].
+  1. Repeat for each Record {[[value]], [[size]]} _pair_ that is an element of _queue_,
+    1. Set _pair_.[[value]] to ! <a abstract-op>StructuredClone</a>(_pair_.[[value]], _targetRealm_).
```
domenic: This should use ?, not !, as StructuredClone could throw

index.bs
```diff
@@ -3589,11 +3634,13 @@ throughout the rest of this standard.
 </emu-alg>
 
 <h4 id="enqueue-value-with-size" aoid="EnqueueValueWithSize" throws>EnqueueValueWithSize ( <var>queue</var>,
-<var>value</var>, <var>size</var> )</h4>
+<var>value</var>, <var>size</var>, <var>targetRealm</var> )</h4>
```
domenic: It seems nicer to me if EnqueueValueWithSize stays as a naive implementation of the queue-with-sizes data structure, and the structured cloning happens elsewhere. Is doing that much uglier?

reference-implementation/lib/readable-stream.js
```diff
@@ -12,6 +14,7 @@ const { AcquireWritableStreamDefaultWriter, IsWritableStream, IsWritableStreamLo
 
 const InternalCancel = Symbol('[[Cancel]]');
 const InternalPull = Symbol('[[Pull]]');
+const InternalTranfer = Symbol('[[Transfer]]');
```
domenic: typo Tran[s]fer

reference-implementation/lib/readable-stream.js
```diff
@@ -251,6 +257,33 @@ class ReadableStream {
     const branches = ReadableStreamTee(this, false);
     return createArrayFromList(branches);
   }
+
+  [InternalTranfer](/* targetRealm */) {
```
domenic: I think in code I would reify the realms as their globals.

reference-implementation/lib/readable-stream.js
```diff
@@ -251,6 +257,33 @@ class ReadableStream {
...
+    }
+    if (this._state === 'errored') {
+      throw new TypeError('Cannot transfer an errored stream');
+    }
+    const controller = this._readableStreamController;
+    if (controller._targetRealm === undefined) {
+      throw new TypeError('Only cloning streams are transferable');
+    }
+    /* can't exactly polyfill realm-transfer */
+    const that = new ReadableStream();
```
domenic: So here this would be `new targetRealm.ReadableStream()`

index.bs
```diff
@@ -1447,6 +1478,12 @@ Instances of {{ReadableStreamDefaultController}} are created with the internal s
     </tr>
   </thead>
   <tr>
+    <td>\[[targetRealm]]
```
domenic: This approach, of using the ReadableStreamDefaultController in one realm instead of just creating a new stream + controller pair, is very interesting, but not what I expected. Why did you choose that?

To be specific, I would have expected the algorithm to be something like

```js
const reader = this.getReader();

const that = new ReadableStream({
  pull(c) {
    return reader.read().then(
      ({ value, done }) => {
        if (done) {
          c.close();
          return;
        }

        c.enqueue(StructuredClone(value));
      },
      e => c.error(e)
    );
  },
  // probably more
});
```

reference-implementation/lib/readable-stream.js
domenic: This would allow you to write web platform tests using iframes (although they might not do so great in our test runner... but maybe they would!)

isonmad:
> If you ignore the question of "so how does this work for Indexed DB's use of structured clone"

Can you elaborate? I'm not familiar with IndexedDB. This PR doesn't make streams cloneable anyway, trying to clone them is still an error as long as ReadableStream lacks a `[[Clone]]` internal method. They're still only transferable.

> This seems similar to the question of what happens if the underlyingSource calls controller.enqueue(somethingUncloneable). The answer is to error the destination stream, right?

Not necessarily.  Why should a `cloning` stream be errored, when `bytes` streams don't error the stream when you provide an unacceptable chunk either? You can just have `enqueue` throw if cloning/serializing to an Agent Record fails. If the underlyingSource doesn't catch the thrown DataCloneError, *then* the stream obviously gets errored.

And 'erroring the stream' still leaves the original question of what even to actually put in that storedError slot at all--

> With a TypeError of some kind I guess.

Right, that. Or a generic DataCloneError? We could attempt a StructuredClone of the error object actually, since it may or may not be uncloneable, and just place a DataCloneError in the slot if the clone attempt throws.

> Reflect.ownKeys

That I wasn't aware of, neat.

index.bs
isonmad: It actually can't at this point, because every `[[value]]` here is already the result of a previous StructuredClone call. The only call to StructuredClone that could throw is at enqueue time.

domenic:
> Can you elaborate? I'm not familiar with IndexedDB.

IndexedDB uses structured clone, kind of. But then it writes to disk, instead of recreating the object in another realm. So the current structure of StructuredClone is kind of broken for it.

> This PR doesn't make streams cloneable anyway, trying to clone them is still an error as long as ReadableStream lacks a [[Clone]] internal method. They're still only transferable.

Good point. Maybe it is even more fine then.

> Not necessarily. Why should a cloning stream be errored, when bytes streams don't error the stream when you provide an unacceptable chunk either? You can just have enqueue throw if cloning/serializing to an Agent Record fails. If the underlyingSource doesn't catch the thrown DataCloneError, then the stream obviously gets errored.

I said error the *destination* stream, not the original being-cloned stream.

> We could attempt a StructuredClone of the error object actually, since it may or may not be uncloneable, and just place a DataCloneError in the slot if the clone attempt throws.

Yes, we'd definitely try to clone it at first. I guess "DataCloneError" DOMException is better if that fails.

index.bs
isonmad: This approach just seemed intuitive to me.

- For one, I wanted to make throw-on-enqueue possible, so the clone call has to occur when a chunk first enters the controller's queue, not when it leaves it, as would happen in your algorithm.
- The underlyingSource still lives in the original realm, and has a reference to the controller object. (Conceptually, 'controller' is sort of a term for two different things that are conflated, a public API object provided to the underlyingSource and the hidden implementation details of the stream. The API object is stuck in the original realm of the underlyingSource, but the 'implementation details' of the stream are in charge of moving things between realms, and I was conceiving that as synonymous with a single controller, I guess.)
- I thought it would also be nice to have a stream that clones its chunks *without* having to transfer it to a worker and back again, and there's no need for a second readablestream at all in that case.
- It just seems like less code.

isonmad:
> I said error the destination stream, not the original being-cloned stream.

Okay I'm semantically confused. I'm considering it conceptually the same stream, 'transferred' to a new realm. Same underlyingSource and everything.

index.bs
isonmad: Oh, right, I also had a preference for throwing the `RangeError` for invalid strategy return values before potentially throwing the `DataCloneError`.

domenic:
> Okay I'm semantically confused. I'm considering it conceptually the same stream, 'transferred' to a new realm. Same underlyingSource and everything. The original ReadableStream object is 'detached' and an empty shell of its former self.

No, they're definitely not the same stream. The underlying data is what gets transfered, but there are two different ReadableStream objects, and thus two different streams.

domenic:
Although I still am unsure on what model we want to pursue here, and I should spend more time thinking about it, I was talking with a coworker today who was very interested in this use case. To demonstrate it to him exactly how it would work I produced two gists:

- https://gist.github.com/domenic/e1e2238f1447e6f6783f1fba3ee601a6 for only ReadableStreams being transferable
- https://gist.github.com/domenic/ea5ebedffcee27f552e103963cf8585c for using an identity transform stream to make things a bit more intuitive

ricea:
I've been thinking about it a bit. The transferred stream ends up with a different strategy. This is weird. But I think it's unavoidable.

For byte streams it's not too bad. The HWM may be different from the original stream. It's attractive to me to let the browser set that to some "optimal" value which depends on the transfer overhead.

For other streams I guess we just end up with the equivalent of CountQueuingStrategy(1). Not great.

We could recognise the original value CountQueuingStrategy.prototype.size and ByteLengthQueuingStrategy.prototype.size and treat them specially. That seems like such a terrible idea that I don't want to pursue it.

domenic:
I think that might not be so bad... For example, consider a readable stream with HWM = 5 in a worker that gets transferred to the main thread, where it ends up with HWM = 1. The HWM isn't observable from the main thread, since all you have is a reader. Meanwhile, the stream creator over in the worker thread is still getting appropriate backpressure signals, because they still see HWM = 5.

A writable stream is a bit more troubling. Assuming the same setup, the producer in the main thread will always see desiredSize = 1. I wonder if we should consider some kind of asynchronous proxying of desiredSize over to the main thread... that makes things two-directional, which seems weird. This reminds me of when we were discussing whether the caller of getWriter() should determine the HWM, instead of the writable stream creator.

The fact that readables and writables are so different here implies to me I might be missing something. Thoughts welcome.

isonmad:
> The fact that readables and writables are so different here implies to me I might be missing something.

The underlyingSource of the transferred stream could only see backpressure if its queue isn't being dequeued from, and it fills up to the HWM=5. Likewise, after writing one chunk, why wouldn't the main thread continue to see `writer.desiredSize` as 0 for as long as its internal queue wasn't dequeued from, either?

The producer realm has to somehow know when not to dequeue something and send it off to the other side. it can't just send chunks off without some kind of ready signal. Isn't it the same in both cases?

ricea:
We've implemented transferable streams in a different way, so I'm closing this.

PR #619, opened 2016-11-22
Aborting a stream should wait for pending writes

ricea:
This is a work-in-progress for #611.

I haven't updated the standard language to match yet. I would like to get an initial round of feedback on the changes before I do that.

PTAL.

PR #634, opened 2016-12-19
Do not allow abort to happen when a close is about to happen

### LINKED ISSUE #632: Do not allow abort() to interrupt a closing writable stream?
I thought this would have been taken care of by https://github.com/whatwg/streams/commit/591a6eda9ea0b84881a370519d6aca53fab9e256, but it seems like there is a subtle difference between "closing" (a close request has been queued) and [[inClose]] === true (underlying sink close() is executing).

I am wondering if we can remove the wart currently noted as

> Due to the way writable streams asynchronously close, it is possible for both close and abort to be called, in cases where the producer aborts the stream while it is in the "closing" state. Notably, since a stream always spends at least one turn in the "closing" state, code like ws.close(); ws.abort(...); will cause both to be called, even if the close method itself has no asynchronous behavior. A well-designed underlying sink object should be able to deal with this. 

I started some work on this in https://github.com/whatwg/streams/tree/close-then-abort but my method of doing so broke the requirement that close() not happen before start() finishes. I think there might be another better method, but it's not clear to me what it is quite yet...

domenic:
As of #619, writer.abort() does not disturb the stream while an underlying-sink write or close operation is ongoing. However, it was still possible to cause both underlying sink abort and close to happen, with code like `writer.close(); writer.abort()`. This was because of the asynchronous way in which the stream's state transitions from "closing" to actually having [[inClose]] set to true.

This addresses that problem by peeking at the queue when abort is called. If the state is [[closing]], and the queue contains no writes, we will not perform the underlying abort, and instead just let the underlying close happen. This counts as a success, so we immediately return a promise fulfilled with undefined.

Fixes #632.

---

Careful review appreciated. A couple points worth questioning:

- This mechanism of peeking at the queue seems a bit suspect. I wonder if [[inClose]] is redundant with this, and we could consolidate somehow?
- I'm not sure immediately returning a promise fulfilled with undefined is correct. What if closing fails? It would be better to return the appropriate underlying sink close promise, but that doesn't exist yet... hmm. Maybe we should just call WritableStreamDefaultControllerClose?
- This behaves differently if the abort() happened during a write, as shown by the test. That is: write(); close(); abort(); will call the underlying abort, whereas just close(); abort(); will cause it to become closed. I think this seems correct, but I'm not sure.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/aborting.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/aborting.js
@@ -175,22 +175,25 @@ promise_test(t => {
   return writePromise;
 }, 'Aborting a WritableStream causes any outstanding write() promises to be rejected with a TypeError');
 
-promise_test(t => {
-  const ws = new WritableStream();
+promise_test(() => {
+  const ws = recordingWritableStream();
   const writer = ws.getWriter();
 
   const closePromise = writer.close();
-  writer.abort(error1);
+  const abortPromise = writer.abort(error1);
 
   return Promise.all([
-    promise_rejects(t, new TypeError(), writer.closed, 'closed should reject with a TypeError'),
-    promise_rejects(t, new TypeError(), closePromise, 'close() should reject with a TypeError')
-  ]);
-}, 'Closing but then immediately aborting a WritableStream causes the stream to error');
+    writer.closed,
+    closePromise,
+    abortPromise
+  ]).then(() => {
+    assert_array_equals(ws.events, ['close']);
+  });
+}, 'Closing but then immediately aborting a WritableStream results in a successful close and a fulfilled abort()');
 
-promise_test(t => {
+promise_test(() => {
   let resolveClose;
-  const ws = new WritableStream({
+  const ws = recordingWritableStream({
     close() {
       return new Promise(resolve => {
         resolveClose = resolve;
@@ -202,14 +205,20 @@ promise_test(t => {
   const closePromise = writer.close();
 
   return delay(0).then(() => {
-    writer.abort(error1);
+    const abortPromise = writer.abort(error1);
     resolveClose();
+
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.closed, 'closed should reject with a TypeError'),
-      closePromise
-    ]);
+      writer.closed,
+      closePromise,
+      abortPromise
+    ])
+    .then(() => {
+      assert_array_equals(ws.events, ['close']);
+    });
   });
-}, 'Closing a WritableStream and aborting it while it closes causes the stream to error');
+}, 'Closing a WritableStream and aborting it while it is in the process of closing results in a successful close and ' +
+   'fulfilled abort()');
 
 promise_test(() => {
   const ws = new WritableStream();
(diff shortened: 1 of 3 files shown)


domenic:
I think there might be an alternate way of implementing this, and of implementing the non-interruptibility of close/write, which is to leverage the [[queue]] more. In that formulation, abort() would clear the queue and then put an "abort" item in the queue, instead of its current behavior of immediately transitioning to the error state---which causes a number of strange things to happen.

ricea:
This looks okay to my brief review, but I'd like to think a bit more about whether the behavioural changes make sense.

`write(); close(); abort()` behaving differently from `close(); abort();` is WIA. My view of the contraints is:
1. Underlying-sink write() and close() should be treated as atomic
2. abort() should happen ASAP but without violating 1.
3. Underlying-sink abort() should never be called after underlying-sink close(). ← New with this change

I am optimistic about the potential of leveraging the [[queue]] to get the right behaviour. That might be the simpler implementation I was grasping for but couldn't quite reach.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -402,56 +405,59 @@ promise_test(() => {
     writer.close();
     const abortPromise = writer.abort();
     return flushAsyncEvents().then(() => {
-      assert_false(abort_called, 'abort should not be called while close is pending');
+      assert_array_equals(ws.events, ['close'], 'abort should not be called while close is pending');
       resolveClose();
-      return abortPromise.then(() => assert_false(abort_called, 'abort should not be called after close completes'));
+      return abortPromise.then(() => {
+        assert_array_equals(ws.events, ['close'], 'abort should not be called while close is pending');
```
ricea: close is no longer pending here. Perhaps go back to 'after close completes' wording?

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -402,56 +405,59 @@ promise_test(() => {
...
+    return Promise.all([
+      closePromise.then(() => { ordering.push('close'); }),
+      abortPromise.then(() => { ordering.push('abort'); })
+    ])
+    .then(() => {
+      assert_array_equals(ordering, ['abort', 'close']);
+    });
   });
-}, 'writer close() promise should resolve before abort() promise');
+}, 'writer abort() promise should fulfill before close() promise (since it bails early)');
```
ricea: I'm not keen on this behaviour. Usually we can rely on abort() to tell us whether the stream was cleaned up or not, but in this case we can't.

domenic:
I don't care much about the ordering, so happy to change that. I'm still pretty unhappy with how this changes the spec though and would prefer to mess with [[queue]]. Maybe it is better to just get the behavior/test changes landed though and fix the internals to be more elegant later. I'll see what I can do.

ricea:
Yeah, doing the behaviour changes here and then messing with [[queue]] in a separate CL seems like a good way to unblock this.

domenic:
I'm going to put this on hold until @tyoshino's PR #655 is merged since that is a bigger refactoring and might provide a better basis for this work.

tyoshino:
Sorry for belated reply. #655 has removed `WritableStreamError()` invocation from `WritableStreamAbort()`. Now, when there's ongoing `sink.write()` or `sink.close()`, `WritableStreamAbort()` does only:
- rejecting `writer.ready`
- schedule aborting by setting `[[pendingAbortRequest]]`

So, all of ricea's list in https://github.com/whatwg/streams/pull/634#issuecomment-270561679 are satisfied.

tyoshino:
> - I'm not sure immediately returning a promise fulfilled with undefined is correct. What if closing fails? It would be better to return the appropriate underlying sink close promise, but that doesn't exist yet... hmm. Maybe we should just call WritableStreamDefaultControllerClose?

I think it should be delayed **if** we want to keep the behavior that `WritableStreamAbort()` returns a promsie resolved with **undefined**. Even before #655, we specified that `sink.close()` rejection results in erroring the stream. So, it's unknown whether we'll end up with `"errored"` or `"closed"`. I think it's not great to return a fulfilled promise immediately.

Or, we could change `WritableStreamAbort()` to return a promise rejected with a **TypeError** indicating that: the stream has already been requested to `close()` and it's about to or already processing or finished processing the `close()`, and therefore it doesn't make sense to run `WritableStreamAbort()` now.

ricea:
My preference is for a fulfilled abort() to mean "the stream has been cleaned up and no more data will be written". Which implies that abort() has to wait for sink.close() to complete.

If instead a fulfilled abort() means "sink.abort() was called and didn't reject" then this has to apply in all cases, ie. `close().then(() => abort())` has to reject.

domenic:
Closing as this was fixed by https://github.com/whatwg/streams/commit/a25f410cbccfbaf56479ca3224fe57bab8b2b6fa. We're tracking porting the tests from this PR in #694.

PR #635, opened 2016-12-21
don't invoke [[strategySize]] as method

isonmad:
The spec text says to Call it with *this* as undefined. That wasn't what the reference implementation was doing.


domenic:
Great find. Would you be able to submit a test to WPT for readable streams as well?

PR #638, opened 2017-01-05
Throw a TypeError in ReadableStreamBYOBReader constructor if stream isn't a byte stream

tschneidereit:


ricea:
This looks correct to me. I'm going to wait for someone else's opinion on whether we should remove the check from getReader() since it is now redundant. @domenic ?

We will also need the same change made in the reference implementation and a test adding to the web-platform-tests but I can do that if you don't want to.

domenic:
Yeah I was planning to push some commits to this PR to remove the getReader check, add till to the acks, and update at least the reference implementation. Till says he's working on converting the tests to WPT and will add a test to those when he's done but it wouldn't hurt to update the existing suite too.

Feel free to do so yourself if you have time before I next wake up (at least 12 hours from now) :)

ricea:
@domenic PTAL

tschneidereit:
@ricea thank you for the additional fixes and tests!

Issue #639, opened 2017-01-06
Redundant step: "Assert: stream.[[state]] is "closing" or "error..."


tyoshino:
https://streams.spec.whatwg.org/commit-snapshots/22b3f03bf803868bf65d9b57f862cc7e3cc80188#writable-stream-finish-close

> Assert:​ stream\.\[\[state\]\] is "closing" or "errored"\.

Looks the assert in the following else clause is enough.

PR #640, opened 2017-01-06
Factor out rejection handling logic from WritableStreamDefaultControllerProcessClose()

tyoshino:
This improves readability of WritableStreamDefaultControllerProcessClose() slightly by making it easier to see that all the promises are correctly processed in both of the fulfillment callback and rejection callback.

The assertion at the top of WritableStreamFinishClose() has been removed because the assertion in the else clause of the following if-block is enough.

~~This improves readability of WritableStreamDefaultControllerProcessClose() slightly by making it easier to see that all the promises are correctly processed in both of the fulfillment callback and rejection callback.~~

~~The assertion that was placed at the top of WritableStreamFinishClose() has been removed because the assertion in the else clause of the following if-block is enough.~~

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2826,17 +2826,32 @@ visible through the {{WritableStream}}'s public API.
 <var>stream</var> )</h4>
 
 <emu-alg>
-  1. Assert: _stream_.[[state]] is `"closing"` or `"errored"`.
+  1. Assert: _stream_.[[pendingCloseRequest]] is not *undefined*.
+  1. <a>Resolve</a> _stream_.[[pendingCloseRequest]] with *undefined*.
+  1. Set _stream_.[[pendingCloseRequest]] to *undefined*.
+  1. If _stream_.[[pendingAbortRequest]] is not *undefined*,
+    1. <a>Resolve</a> _stream_.[[pendingAbortRequest]] with *undefined*.
+    1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
   1. Let _writer_ be _stream_.[[writer]].
   1. If _stream_.[[state]] is `"closing"`,
     1. If _writer_ is not *undefined*, <a>resolve</a> _writer_.[[closedPromise]] with *undefined*.
     1. Set _stream_.[[state]] to `"closed"`.
-  1. Otherwise if _writer_ is not *undefined*,
-    1. Assert: _stream_.[[state]] is `"errored"`.
-    1. <a>Reject</a> _writer_.[[closedPromise]] with _stream_.[[storedError]].
-    1. Set _writer_.[[closedPromise]].[[PromiseIsHandled]] to *true*.
+    1. Return.
+  1. Assert: _stream_.[[state]] is `"errored"`.
+  1. If _writer_ is *undefined*, return.
+  1. <a>Reject</a> _writer_.[[closedPromise]] with _stream_.[[storedError]].
+  1. Set _writer_.[[closedPromise]].[[PromiseIsHandled]] to *true*.
+</emu-alg>
+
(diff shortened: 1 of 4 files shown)


tyoshino:
Oops. This is violating the design principle. WritableStreamFinishClose() is encapsulating the algorithm for modifying the state of the stream in response to signal from the controller. So, rather than merging WritableStreamFinishClose() into WritableStreamDefaultControllerProcessClose(), the steps in the rejection callback should be factored out and placed next to WritableStreamFinishClose(). I'll rewrite the patch.

tyoshino:
Changed the patch to factor out the logic as an operation of the WritableStream exposed to controllers.

reference-implementation/lib/writable-stream.js
```diff
@@ -188,7 +188,14 @@ function WritableStreamError(stream, e) {
 }
 
 function WritableStreamFinishClose(stream) {
-  assert(stream._state === 'closing' || stream._state === 'errored');
+  assert(stream._pendingCloseRequest !== undefined);
+  stream._pendingCloseRequest._resolve(undefined);
+  stream._pendingCloseRequest = undefined;
+
+  if (stream._pendingAbortRequest !== undefined) {
```
ricea: Where did this come from? Does it change visible behaviour?

tyoshino: Oh, it's from L202 of the previous commit. I changed the order to make this and `WritableStreamFinishCloseWithError()` look similar. This reordering shouldn't cause any behavior change.

ricea: Sorry, the diff confused me.

It changes the order that the abort() promise is resolved wrt. the .closed promise. This is observable in principle. When I tried to write tests for promise resolve ordering, the results were not what I expected, so I never added any. Probably no-one cares.

ricea:
lgtm

index.bs
```diff
@@ -2828,16 +2828,30 @@ visible through the {{WritableStream}}'s public API.
...
   1. If _stream_.[[state]] is `"closing"`,
     1. <a>Resolve</a> _stream_.[[writer]].[[closedPromise]] with *undefined*.
     1. Set _stream_.[[state]] to `"closed"`.
   1. Otherwise,
     1. Assert: _stream_.[[state]] is `"errored"`.
     1. <a>Reject</a> _stream_.[[writer]].[[closedPromise]] with _stream_.[[storedError]].
     1. Set _stream_.[[writer]].[[closedPromise]].[[PromiseIsHandled]] to *true*.
+</emu-alg>
+
+<h4 id="writable-stream-finish-close-with-error" aoid="WritableStreamFinishCloseWithError" nothrow>WritableStreamFinishCloseWithError (
```
domenic: Line too long

reference-implementation/lib/writable-stream.js
tyoshino: > This is observable in principle

Ah, right. Thanks for pointing it out.

index.bs

tyoshino:
> Looks good but I would like a test for the promise resolution order. Even though the order isn't important, we should add a test to ensure everyone is interoperable on it. (We can add a comment to the test saying that we should feel free to change the results in the future, as long as we file bugs on all implementers to update to the latest tests.)

I'm adding this and found some bugs. Fixing them now.

> Also be sure when committing to use a < 70 character first line on the commit message, and include "closes #639" in the commit message body.

Will fix. Thanks

tyoshino:
70% done (only reference implementation). The diff became huge, but I think it's worth doing as I think now it's not so easy to check correctness of the algorithm around aborting. I'll try to finish this early next week.

tyoshino:
The change has been drastically rewritten. It's no longer a small refactoring. I'm creating a new PR with an appropriate description.

tyoshino:
Please see the successor PR #655

PR #641, opened 2017-01-06
When underlyingSink.close() resolves, assert state

ricea:
Previously, WritableStreamDefaultControllerProcessClose checked the
state was "closing" or "errored" when underlyingSink.close()
resolved. In fact it can't be anything else, so make it an assert
instead.

Thanks to @tyoshino for spotting this.


tyoshino:
lgtm. thanks!

PR #642, opened 2017-01-10
Include server-worker include in pipe-through test

ricea:
pipe-through.https.html uses worker_test() and therefore needs the /service-workers/service-worker/resources/test-helpers.sub.js script included. Add it.


tyoshino:
lgtm

PR #643, opened 2017-01-10
Fix a few bugs in spec algorithms

tschneidereit:
These are all issues where the reference implementation is either correct or, in the `ReadableByteStreamControllerPullInto` case, doesn't make a testable difference. (For the reference implementation we could test that it invokes the `view`'s `buffer` and `byteOffset` property getters only once, but that doesn't make sense for actual implementations because those must directly access the underlying slots.)


domenic:
Thank you so much! I reworded the commit headlines to fit in one line but didn't change the contents at all.

tyoshino:
lgtm. thanks!

PR #645, opened 2017-01-11
Correctly handle releaseLock() during close()

ricea:
Code which calls `writer.releaseLock()` while `writer.close()` is in
progress can cause an exception in WritableStreamFinishClose because
`stream.[[writer]]` is undefined.

Existing tests did not find the issue because it is quite hard to
detect.

Add undefined checks to fix this. Also add tests to verify the correct
behaviour.


tyoshino:
lgtm

Issue #644, opened 2017-01-11
pipeTo() algorithm shuts down too early on readable close()?


ricea:
I think there's an issue with the pipeTo() algorithm that prevents writing all the data. When the underlying source calls `close()` the state of the ReadableStream changes immediately to "CLOSED". If there is a `read()` on the readable ongoing at that point then pipeTo shutdown starts immediately, prohibiting the data that was read from being written to the WritableStream.

I'm not 100% certain that this is a spec issue yet but I am making a note of it to look at it again later. I will close this bug if I conclude there is no actual issue.

ricea:
A related problem is that if the source queues several items and calls `close()`, then the sink errors while writing one of those items, the standard implies that since the readable became CLOSED first, pipeTo should fulfill without error. I think it should reject with the error from the sink. I need to check whether there is a test for this case and add one if there isn't.

ricea:
I've concluded that there is an issue here.

ricea:
> It sounds like the problem is in "Shutdown must stop all activity". We want any reads to always be followed by writes... as long as the dest is "writable", I guess. Shutdown says "Wait until any ongoing write finishes" already, but if the write isn't ongoing yet that doesn't work.

I will try to come up with some wording for this.

For the Chromium implementation, I'm thinking of removing the 'Wait until any ongoing write finishes (i.e. the corresponding promises settle).' step from 'shutdown with action' because now that we treat underlying sink write() as uninterruptible it doesn't make any difference any more, and that step is troublesome to implement. Maybe we should also remove it from the standard?

> I think this works with the current spec:
> 
> close() ends up in "Closing must be propagated forward"
> We go to "shutdown with an action" of WritableStreamDefaultWriterCloseWithErrorPropagation(writer).
> We wait for any ongoing writes to finish. One errors, causing dest.[[storedError]] to be set.
> We let p be the result of performing WritableStreamDefaultWriterCloseWithErrorPropagation(writer).
> p is rejected with dest.[[storedError]].
> We finalize with dest.[[storedError]].

I see. I will write a test for this, after checking that it isn't already covered.

domenic:
> For the Chromium implementation, I'm thinking of removing the 'Wait until any ongoing write finishes (i.e. the corresponding promises settle).' step from 'shutdown with action' because now that we treat underlying sink write() as uninterruptible it doesn't make any difference any more, and that step is troublesome to implement. Maybe we should also remove it from the standard?

[...]

PR #647, opened 2017-01-12
What pipeTo does when both streams are closed

ricea:
Previously it was ambiguous what implementations should do if both
source and destination are closed or errored at the start.

Clarify which states take priority in determining the behaviour. The new
text is based on the multiple-propagation.js tests and can be summarised
as "errors before closed, source before destination".

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -653,25 +653,26 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
        initiate further reads from _reader_, and must only perform writes of already-read <a>chunks</a>, as described
        below. In particular, the user agent must check the below conditions before performing any reads or writes,
        since they might lead to immediate shutdown.
-     * <strong>Errors must be propagated forward:</strong> if *this*.[[state]] is or becomes `"errored"`, then
-       1. If _preventAbort_ is *false*, <a href="#rs-pipeTo-shutdown-with-action">shutdown with an action</a> of !
-          WritableStreamAbort(_dest_, *this*.[[storedError]]) and with *this*.[[storedError]].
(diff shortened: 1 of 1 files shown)


ricea:
This is quite repetitive. There may be simpler ways to express the same thing.

index.bs
```diff
@@ -648,23 +665,18 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
        initiate further reads from _reader_ or writes to _writer_. (Ongoing reads and writes may finish.) In particular,
        the user agent must check the below conditions on *this*.[[state]] and _dest_.[[state]] before performing any
        reads or writes, since they might lead to immediate shutdown.
```
domenic: Maybe the PR could simply add "in order" after "before performing any reads or writes"?

ricea: I think because of #644 I want to clarify what statuses are checked for at what time. Something like

* readable "ERRORED" and "CLOSED" trigger shutdown before a read
* readable "ERRORED" triggers shutdown before a write
* writable "CLOSING" and "CLOSED" are only checked at the beginning
* writable "ERRORED" triggers shutdown immediately at any time

Maybe the answer is to split out the "when to trigger" part from the "what to do" part.

ricea: That wasn't quite right. We also need to detect readable "ERRORED" and "CLOSED" while we are waiting for backpressure to go away.

ricea: Based on our discussion in #644, I think we can probably combine the wording changes for that with the "in order" approach to achieve this without significant restructuring.

ricea:
@domenic Here's a new attempt. Although it's not obvious from the github diff view, only one new line is added. The rest of the diff just adds indentation and changes `*` to `1.`.

domenic:
Perfect, LGTM. I'll let you merge as the commit message probably needs some tweaking.

PR #648, opened 2017-01-12
Convert readable byte stream tests to WPT

tschneidereit:
These tests should all be entirely equivalent to the original tape tests. They weren't converted mechanically though, so at least a cursory review would be appreciated.


domenic:
@tyoshino would you be able to help with this since you wrote most of these? I will also review myself but you are the best person.

reference-implementation/to-upstream-wpts/readable-byte-stream/readable-byte-stream.js
```diff
@@ -0,0 +1,1960 @@
...
+test(() => {
+  // Constructing ReadableStream with an empty underlying byte source object as parameter shouldn't throw.
+  new ReadableStream({ type: 'bytes' });
+}, 'ReadableStream with byte source can be constructed with no errors');
+
+async_test(t => {
+  let startCalled = false;
+  let controller;
+
+  new ReadableStream({
```
domenic: Because of how the async_test framework works, I believe you need to wrap any functions called async that could possibly throw, such as `pull`, in `t.step_func`

domenic: In general you may be able to replace some of these with uses of the recording stream helpers, modified to allow `type: 'bytes'`. But maybe not.

reference-implementation/to-upstream-wpts/readable-byte-stream/readable-byte-stream.js
```diff
@@ -0,0 +1,1960 @@
...
+    pull() {
+      assert_unreached('pull must not be called');
+      t.done();
+    },
+    type: 'bytes'
+  }, {
+    highWaterMark: 0
+  });
+
+  Promise.resolve().then(() => {
```
domenic: Looks like this could be a promise_test

reference-implementation/to-upstream-wpts/readable-byte-stream/readable-byte-stream.js
```diff
@@ -0,0 +1,1960 @@
...
+
+async_test(t => {
+  const stream = new ReadableStream({
+    type: 'bytes'
+  });
+
+  const reader = stream.getReader();
+  reader.releaseLock();
+
+  reader.closed.then(() => {
```
domenic: Should be a promise_test with promise_rejects. Same for most of the async_tests that follow and end by checking reader.closed.

reference-implementation/to-upstream-wpts/readable-byte-stream/readable-byte-stream.js
```diff
@@ -0,0 +1,1960 @@
...
+  });
+
+  const reader = stream.getReader({ mode: 'byob' });
+
+  const promise = promise_rejects(t, testError, reader.read(new Uint8Array(1)), 'read(view) must fail');
+  return promise_rejects(t, testError, promise.then(() => reader.closed));
+}, 'ReadableStream with byte source: Throwing in pull in response to read(view) function must error the stream');
+
+promise_test(t => {
+  const passedError = new TypeError('foo');
```
domenic: I've found it a nicer pattern, at the top of the file, create an `error1` and maybe `error2`, and then you can use assert_throws/promise_rejects more easily. See e.g. https://github.com/whatwg/streams/blob/b412a96bbb5905b1abcc25c8032439a6c081e96c/reference-implementation/to-upstream-wpts/writable-streams/aborting.js (Ctrl+F "error1")

I guess this is a preexisting issue so no big deal if you'd rather not do this cleanup.

tschneidereit:
@domenic, I addressed all your review comments. Turning all remaining `async_test`s into `promise_test`s turned out to be pretty straight-forward, as was the other suggested cleanup.

tschneidereit:
... now without eslint failures.

domenic:
Awesome. LGTM, but let's give at least another day for @tyoshino to chime in if he has time. Otherwise we can merge and make any changes he suggests at a later date.

tyoshino:
Sorry, I couldn't review this today. Feel free to proceed.

Issue #646, opened 2017-01-11
Constructing a readable byte stream can reach code that assumes a default reader


domenic:
Found by @tschneidereit:

- [new ReadableStreamDefaultController](https://streams.spec.whatwg.org/#rs-default-controller-constructor) step 11.a.iv. ->
- [ReadableStreamDefaultControllerCallPullIfNeeded](https://streams.spec.whatwg.org/#readable-stream-default-controller-call-pull-if-needed) step 1 ->
- [ReadableStreamDefaultControllerShouldCallPull](https://streams.spec.whatwg.org/#readable-stream-default-controller-should-call-pull) step 5 ->
- [ReadableStreamGetNumReadRequests](https://streams.spec.whatwg.org/#readable-stream-get-num-read-requests) which assumes _stream_.[[reader]].[[readRequests]] exists, which is not true for byte readers.

Untested, but I think the repro code is then something like:

```js
const rs = new ReadableStream({ type: "bytes" });
const reader = rs.getReader();
```

@tschneidereit mentions that his Gecko implementation is failing the test "ReadableStream with byte source: autoAllocateChunkSize", whereas the reference implementation for some reason is not.

---

I think the correct fix here is to bail in ReadableStreamDefaultControllerCallPullIfNeeded immediately if the reader is not a default reader. @tyoshino, does that sound right? We should also track down the missing test coverage though.

ricea:
I found when trying to reproduce #645 that exceptions were silently getting dropped when happening in async code, even though I'd expect an "unhandled promise rejection". I have no idea why.

It's good to have coverage for these things anyway, although writing the tests can be mind-bendingly difficult. If an exception happens at the end of an async function it can be unobservable.

domenic:
Maybe it is a problem with the tape framework in particular; I think the web platform tests are generally good at catching these sorts of things.

isonmad:
> new ReadableStreamDefaultController

Wait, how would that happen in the first place if you're constructing a byte stream?

domenic:
Argh, thanks @isonmad, I guess I did not understand the problem correctly... @tschneidereit thoughts?

tschneidereit:
I'm afraid I don't understand what was going on either, anymore. I can't reproduce it and didn't save what I had that triggered this. #649 is good regardless, so I vote for closing this as invalid.

domenic:
Sounds good; let us know if you can reproduce!

tyoshino:
[...]

PR #649, opened 2017-01-13
add assert to byte stream controller [[Pull]]()

isonmad:
Looking at #646, this was the closest thing that fit the issue description. There isn't a bug, but an assert helps clarify that the method depends on its only caller being a default reader.


ricea:
I don't know whether or not  #646 is a real problem, but this change is good. lgtm.

tyoshino:
lgtm

PR #650, opened 2017-01-13
Meta: enable HTML checker

annevk:


annevk:
It fails because it's somehow still in npm mode by the time it reaches the Java invocation.

sideshowbarker:
> It fails because it's somehow still in npm mode by the time it reaches the Java invocation.

It seems to be failing because it never evaluates the `install` section, so never actually runs the ` curl -O https://sideshowbarker.net/nightlies/jar/vnu.jar` call, then because of that we end up getting `Error: Unable to access jarfile vnu.jar`.

That all seems to fall apart because of some other earlier problem—which I’m not sure what it is, but the first indication I see `sh: 1: eslint: not found`.

domenic:
@sideshowbarker https://checker.html5.org/?doc=https%3A%2F%2Fstreams.spec.whatwg.org%2Fbranch-snapshots%2Fannevk%2Fchecker%2F shows no errors but the Travis output shows a bunch of custom element-related errors. Is https://sideshowbarker.net/nightlies/jar/vnu.jar out of sync perhaps?

sideshowbarker:
> the Travis output shows a bunch of custom element-related errors. Is https://sideshowbarker.net/nightlies/jar/vnu.jar out of sync perhaps?

Yes it is. I only now realize I neglected to make some changes to the code the jar uses that would enable it to handle custom elements in the same way the Web-based checker does. So I’ll get those changes made and push out a new nightly of the jar today.

.travis.yml
```diff
@@ -17,3 +18,4 @@ script:
   - npm test
   - cd ..
   - bash ./deploy.sh
+  - curl -O https://sideshowbarker.net/nightlies/jar/vnu.jar && /usr/lib/jvm/java-8-oracle/jre/bin/java -jar vnu.jar --skip-non-html /home/travis/build/whatwg/streams/streams.spec.whatwg.org
```
annevk: Does this mean that even with install we'd install it each time? Or this a more wasteful setup?

sideshowbarker: If we use the command-line checker (jar) I don’t know of any way in Travis to avoid the need for Travis to download the jar each time.

If we wanted to avoid the download, we could switch (back) to using the HTTP API to https://checker.html5.org/, in which case the documents are POSTed to there for checking.

The reason I switched away from doing is that way is, we lose the path information for the files, and only get the file names reported. In some cases that means we can’t determine which file actually had errors (without actually examining the files).

annevk: I'm fine with the current setup. Was just wondering what the difference was between the various ways of making things work.

.travis.yml
```diff
@@ -6,14 +6,18 @@ addons:
...
+    - oracle-java8-set-default
 
 env:
   global:
     - secure: "VayIJwnO14YDmIureSnEE8cNw2gEYJ1b4V8F3YDw+aRVgFB8CGf247JmnOYfEEfadxvjkItAmjAmRzwU+u9U9hosSKuBPSelZ9tmgo3g8LpJnaB578ODBRt/CbLOSSyQyAICxIYJjs/pgV/DJXwdWqW4becWNCw0wgGHqiwDNIQ="
     - secure: "Ga/n9txTxjBDFxsyhbthRZyqVTZA1cOmhH5WaQEWoMcG9yU+RvAdEjqdFmUFHE7SArHIKovyzoupQ9FELESeu8/lkbDtin3MZn3Aj6QXHioBSMihGbduHsGK8yJ1WyjDYXC3fN8XdwNQRZrVFq0ocauPkml/w8uauEsiMo3PX8A="
 before_install:
   - cd reference-implementation
+install:
+  - curl -O https://sideshowbarker.net/nightlies/jar/vnu.jar
```
isonmad: This .jar wasn't found because before_install, above, changed the current directory to reference-implementation/, so the jar is there instead. So this should probably be done in `before_install` instead of moving it to `script`?

isonmad:
> It seems to be failing because it never evaluates the install section,

No, it runs, but it runs in `reference-implementation/`, the place that `before_install` `cd`'d into.

.travis.yml
domenic: There's not much difference between the install step and the script step. They both run every time. The install step just has a default depending on your language (which was overriden in the original).

domenic:
Per https://github.com/whatwg/meta/issues/6 this may evolve more in the future but for now let's merge it.

PR #651, opened 2017-01-14
fix autoAllocateChunkSize

isonmad:
It was special-casing the first read request, and adding a second one would assert ever since it was first merged in https://github.com/whatwg/streams/pull/430#issuecomment-195370054.

```js
var rs = new ReadableStream({type:'bytes', autoAllocateChunkSize:1});
var r = rs.getReader();
var p1 = r.read();
var p2 = r.read(); // assertion failure
```

Also combined two redundant nested if statements, since `totalQueuedBytes > 0` implies `ReadableStreamGetNumReadRequests(stream) === 0`, AFAICT.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1970,8 +1970,8 @@ dispatch from the readable stream implementation to either these or their counte
 <emu-alg>
   1. Let _stream_ be *this*.[[controlledReadableStream]].
   1. Assert: ! ReadableStreamHasDefaultReader(_stream_) is *true*.
-  1. If ! ReadableStreamGetNumReadRequests(_stream_) is *0*,
-    1. If *this*.[[totalQueuedBytes]] > *0*,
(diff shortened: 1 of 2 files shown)


domenic:
Can you add a test showing the change in behavior?

isonmad:
In what file would I put the test though?

domenic:
https://github.com/whatwg/streams/blob/master/reference-implementation/test/readable-byte-stream.js I guess? Although to avoid collisions with https://github.com/whatwg/streams/pull/648 perhaps the best thing to do is just copy the structure from there (i.e. add the .html/.js files) and then I'll make sure to rebase #648 on top.

tschneidereit:
LGTM, except for the missing tests.

tyoshino:
Code lgtm.

I don't remember why I put the assert on _autoAllocateChunkSize in the else block... should be just a mistake.

PR #655, opened 2017-01-16
WritableStream abort logic clean up

tyoshino:
This patch factors out the code to handle fulfillment/rejection of sink write and sink close operation into separate methods to make it easier to follow the code and check the correctness of the logic.

This patch also clarifies the interface boundary, i.e. factoring out operations that are exposed by the WritableStream to controllers.

This PR is recreation of #640.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4,7 +4,7 @@ Group: WHATWG
 H1: Streams
 Shortname: streams
 Repository: whatwg/streams
-Inline Github Issues: true
+Inline Github Issues: title
 Status: LS
 Boilerplate: omit conformance, omit feedback-header
 No Editor: true
@@ -2768,17 +2768,34 @@ writable stream is <a>locked to a writer</a>.
   1. Let _state_ be _stream_.[[state]].
   1. If _state_ is `"closed"`, return <a>a promise resolved with</a> *undefined*.
   1. If _state_ is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
-  1. Assert: _state_ is `"writable"` or `"closing"`.
   1. Let _error_ be a new *TypeError* indicating that the stream has been aborted.
-  1. Perform ! WritableStreamError(_stream_, _error_).
+  1. If _stream_.[[pendingAbortRequest]] is not *undefined*, return <a>a promise rejected with</a> _error_.
+  1. Assert: _state_ is `"writable"` or `"closing"`.
   1. Let _controller_ be _stream_.[[writableStreamController]].
   1. Assert: _controller_ is not *undefined*.
-  1. If _controller_.[[writing]] is *true* or _controller_.[[inClose]] is *true*,
-    1. Set _stream_.[[pendingAbortRequest]] to <a>a new promise</a>.
-    1. If _controller_.[[writing]] is *true*, return the result of <a>transforming</a> _stream_.[[pendingAbortRequest]]
-       with a fulfillment handler that returns ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
-    1. Otherwise, return _stream_.[[pendingAbortRequest]].
-  1. Return ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
+  1. Let _readyPromiseIsPending_ be *false*.
+  1. If _state_ is `"writable"` and !
+     WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, set
+     _readyPromiseIsPending_ to *true*.
+  1. If _controller_.[[writing]] is *false* and _controller_.[[inClose]] is *false*,
+    1. If _stream_.[[writer]] is not *undefined*, perform !
+       WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _error_, _readyPromiseIsPending_).
+    1. Perform ! WritableStreamFinishAbort(_stream_).
+    1. Return ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
+  1. Let _promise_ be <a>a new promise</a>.
+  1. Set _stream_.[[pendingAbortRequest]] to Record {[[promise]]: _promise_, [[reason]]: _reason_}.
+  1. If _stream_.[[writer]] is not *undefined*, perform !
+     WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _error_, _readyPromiseIsPending_).
+  1. Return _promise_.
+</emu-alg>
+
+<h4 id="writable-stream-finish-abort" aoid="WritableStreamFinishAbort" nothrow>WritableStreamFinishAbort (
+<var>stream</var> )</h4>
+
+<emu-alg>
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Set _stream_.[[storedError]] to a new *TypeError* indicating that the stream has been aborted.
+  1. Perform ! WritableStreamRejectPromisesInReactionToError(_stream_).
 </emu-alg>
 
 <h3 id="ws-abstract-ops-used-by-controllers">Writable Stream Abstract Operations Used by Controllers</h3>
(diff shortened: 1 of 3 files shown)


tyoshino:
This patch doesn't include any change on the spec text. I'll change it once we agree on this change.

tyoshino:
I'd like to summarize the current (or new for some places where I made change) state transition rule of the WritableStream. There're some Qs I'd like to discuss.

- Until a WS starts, no write request becomes "a pending write"
- Once a WS starts, the head write request may picked to be a pending write if allowed by the backpressure, etc.
  - Q: Is this "one pending write at a time" restriction a part of WritableStream's semantics or WritableStreamDefaultController's semantics?
- `writer.closed` change is delayed if there's any pending write or pending close until the pending write/close finishes
- `writer.ready` gets updated immediately in response to `controller.error()` and `writer.abort()`
- `controller.error()` call made during there's pending abort wins the abort.
  - `writer.abort()` gets rejected with the error given to the `controller.error()`
  - Q: Do you like this new behavior?
- `writer.write()` gets rejected with different reasons before and after completion of sink write rejection
  - That is, for example, if we call `writer.abort()` and then `writer.writer()` when there's one pending `writer.write()` and two `writer.write()`s in `_writeRequests` and then the pending `writer.write()` finishes by rejection of sink write, the two `writer.write()`s in `_writeRequests` get rejected with the result of sink write while the `writer.write()` after `writer.abort()` gets rejected with `new TypeError('Aborted')`

Implementation notes
- For logic simplicity, state processing e.g. calculation of `readyPromiseIsPending` is done at some early timing before the method starts modifying the state variables even if it increases the code size.
- Steps in the WSFinish(?:Pending)?(Close|Write) methods are ordered so that it's easy to compare them to make it easier to find missing operation.

tyoshino:
Fixes #656

tyoshino:
To add to https://github.com/whatwg/streams/pull/655#issuecomment-272812907,

Error priority (higher first) for .closed:

1. `controller.error()`
1. sink write rejection
1. `writer.abort()`

Error priority (higher first) for .ready: First invocation of these three above.

You can see this difference by looking at the recently added unit tests.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -530,6 +530,76 @@ promise_test(t => {
...
+    const writePromise2 = writer.write('a');
+
+    writer.closed.catch(() => {
+      closedRejected = true;
+    });
+
+    return Promise.all([
+      promise_rejects(t, new TypeError(), writePromise2, 'writePromise2 must be rejected'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be rejected after abort()'),
+      flushAsyncEvents()
```
ricea: Missing return

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -530,6 +530,76 @@ promise_test(t => {
...
+      promise_rejects(t, new TypeError(), writePromise3, 'writePromise3 must be rejected'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be rejected after write completion'),
+      promise_rejects(t, error2, writer.closed, 'writer.closed must be rejected after write completion'),
+      promise_rejects(t, error2, abortPromise, 'abortPromise must be rejected after write completion')
+    ]);
+  }).then(() => {
+    writer.releaseLock();
+
+    return Promise.all([
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be reset to be rejected with a TypeError'),
```
ricea: Nitpick:  the assertion description implies that it verifies more than it actually does. If we want to actually check that it's been reset, we have to save the previous value and verify it is not equal.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
isonmad: How? The return statement is 3 lines above this.

ricea: Sorry, I misread it. Never mind.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
tyoshino: Ah, good point. Resetting is internal thing. What we should verify is that the promises are rejected with an error which indicates "already released". Fixed.

reference-implementation/to-upstream-wpts/writable-streams/close.js
```diff
@@ -213,4 +213,97 @@ promise_test(() => {
...
+  // Wait until the WritableStream starts so that the close() call gets processed.
+  return delay(0).then(() => {
+    const writer = ws.getWriter();
+
+    const closePromise = writer.close();
+    const abortPromise = writer.abort();
+
+    const events = [];
+    return Promise.all([
+      closePromise.then(() => {
```
domenic: I would redo these as

```js
promise_rejects(t, new TypeError(), error1,
  'closePromise must reject with the error returned from the sink\'s close method')
.then(() => events.push('closePromise'))
```

Same for all tests. In general using assert_unreached for promises seems like a code smell.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,242 @@ promise_test(t => {
...
+  const ws = new WritableStream({
+    write() {
+      return new Promise((resolve, reject) => {
+        rejectWrite = reject;
+      });
+    }
+  });
+
+  let writePromise;
+  let writePromiseRejected = false;
```
domenic: In general I prefer arrays of events and assert_array_equals to boolean flags, but it's not a big deal.

reference-implementation/to-upstream-wpts/writable-streams/close.js
tyoshino: OK. Did you mean that I should redo rejection expecting ones e.g. at L259 as your example? Just to make sure.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
tyoshino: Changed to use an array

tyoshino:
Addressed the comments and rewrote the description of assertions.

I'll update the PR description to list all the visible behavior change.

reference-implementation/to-upstream-wpts/writable-streams/close.js
```diff
@@ -213,4 +219,96 @@ promise_test(() => {
   });
 }, 'the promise returned by async abort during close should resolve');
 
+// Though the order of the promises is not important, we're checking it for interoperability. We can change the order
```
ricea: I suggest "arbitrary" rather than "not important". Also, maybe "order in which the promises are fulfilled or rejected"?

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,262 @@ promise_test(t => {
...
+      promise_rejects(t, error2, writePromise,
+                      'writePromise must reject with the error returned from the sink\'s write method'),
+      promise_rejects(t, error2, abortPromise,
+                      'abortPromise must reject with the error returned from the sink\'s write method'),
+      promise_rejects(t, error2, writer.closed,
+                      'writer.closed must reject with the error returned from the sink\'s write method'),
+      flushAsyncEvents()
+    ]);
+  }).then(() => {
+    const writePromise3 = writer.write('a');
```
tyoshino: Yeah, added.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,262 @@ promise_test(t => {
...
+    return Promise.all([
+      writePromise,
+      promise_rejects(t, error2, abortPromise,
+                      'abortPromise must reject with the error passed to the controller\'s error method'),
+      promise_rejects(t, error2, writer.closed,
+                      'writer.closed must reject with the error passed to the controller\'s error method'),
+      flushAsyncEvents()
+    ]);
+  }).then(() => {
+    const writePromise4 = writer.write('a');
```
tyoshino: Done

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,262 @@ promise_test(t => {
...
+
+    const writePromise2 = writer.write('a');
+
+    return Promise.all([
+      promise_rejects(t, new TypeError(), writePromise2, 'writePromise2 must reject with an error indicating abort'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must reject with an error indicating abort'),
+      flushAsyncEvents()
+    ]);
+  }).then(() => {
+    assert_array_equals(events, [], 'writePromise, abortPromise and writer.closed must be not fulfilled/rejected yet');
```
tyoshino: Done

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,262 @@ promise_test(t => {
...
+      promise_rejects(t, new TypeError(), writePromise2,
+                      'writePromise2 must reject with an error indicating the stream has already been errored'),
+      promise_rejects(t, error2, writer.ready,
+                      'writer.ready must reject with the error passed to the controller\'s error method'),
+      flushAsyncEvents()
+    ]);
+  }).then(() => {
+    assert_array_equals(events, [], 'writePromise and writer.closed must be rejected yet');
+
+    abortPromise = writer.abort(error1);
```
tyoshino: Thanks for finding this. Yes, it's important. Actually, in this test case, abort() fails immediately since it checks the state.

reference-implementation/to-upstream-wpts/writable-streams/aborting.js
```diff
@@ -529,6 +530,262 @@ promise_test(t => {
...
+
+    return Promise.all([
+      promise_rejects(t, new TypeError(), writePromise2,
+                      'writePromise2 must reject with an error indicating the stream has already been errored'),
+      promise_rejects(t, error2, writer.ready,
+                      'writer.ready must reject with the error passed to the controller\'s error method'),
+      flushAsyncEvents()
+    ]);
+  }).then(() => {
+    assert_array_equals(events, [], 'writePromise and writer.closed must be rejected yet');
```
tyoshino: Fixed. Also changed the phrase to "fulfilled/rejected". writePromise will fulfill in this test case.

reference-implementation/to-upstream-wpts/writable-streams/close.js
tyoshino: Thanks. Updated.

reference-implementation/lib/writable-stream.js
```diff
@@ -109,6 +109,25 @@ function IsWritableStreamLocked(stream) {
   return true;
 }
 
+function WritableStreamEnsureReadyPromiseRejectedWith(stream, error, isPending) {
```
tyoshino: Good catch. Fixed.

tyoshino:
Oops, sorry. Some review comments are not yet addressed correctly. Fixing them.

tyoshino:
Rebased

wpt PR: https://github.com/w3c/web-platform-tests/pull/4587

Here're links for checking the diff before rebasing:
- ricea's comment: 26ba2c5de34c785778bc5ad3ca7d14eed7c6c4ef
- domenic's comment: ca44eaad6a4ca5571f48aa79271405053483e5aa
- ricea's comment: 1156eef44323ee725a94037b80d767fecf954763

domenic:
The changes look good to me, mainly based on the increase test coverage, but I'm still looking forward to a description of the visible behavior changes :)

ricea:
I would prefer the changes to the standard text to be in the same commit because I'm nervous about them getting out-of-sync.

tyoshino:
Yes. I'm going to make the spec text sync. I didn't mean that I'm going to have two separate commits in the comment.

I created a draft of the commit message to describe all the details in response to Domenic's request in weekend, but haven't done yet. I'm going to get it done today.

Adam, are you fine with the change (reference impl)?

ricea:
Yes. writable-stream.js lgtm.

tyoshino:
[STILL DRAFT]

Bug fix:

- Fixed `sink.close()` rejection handling logic to reject closed even if the stream has been aborted or errored by `controller.error()`
  - Before this fix, `WritableStreamDefaultControllerErrorIfNeeded()` was used which is no-op when the state is set to `"errored"`

Changes:

- Now `writer.abort()` doesn't set the `[[state]]` of the stream to `"errored"` when there's pending `sink.write()` or `sink.close()` but prevents further operations on writer by presence of the `[[pendingAbortRequest]]`.
  - `writer.ready` rejects immediately
  - Now `controller.error()` succeeds even after writer.abort() is called. So, `writer.abort()` and `writer.closed` is rejected with the error given to the `controller.error()` while `writer.ready` is kept rejected with an error indicating abort.
  - Now pending `writer.abort()` rejects with `[[storedError]]` if it's been errored with `controller.error()`
- `sink.close()` rejection doesn't change the rejection reason of `writer.ready` if it's already rejected due to `writer.abort()` or `controller.error()`
  - Q: Is this ok? When a writer is reobtained, the ready promise is set to be rejected with the latest error.
- Changed `sink.write()` rejection to set `[[storedError]]` to the rejection reason of `sink.write()` only when there wasn't `controller.error()` call

Changes (fulfillment/rejection order):

- Changed the order of promise rejection on `sink.write()` completion. Now `writer.abort()` then `writer.closed`.
- Rejection order on `sink.write()` completion:
  - Before: write(), closed, abort()
  - After: write(), abort(), closed
- Rejection order on `sink.write()` rejection
  - Before: write(), closed, abort(), ready
  - After: write(), abort(), ready, closed
  - Q: Worse?
- `controller.error()` rejects:
  - Before: 
  - After: ready then closed

Refactoring:

- Refactored abort() finalization steps to clarify where a certain process is scheduled to run
- Factored out `[[pendingWriteRequest]]` setting logic as a WritableStream abstract operation exposed to controllers
- Removed redundant `[[queue]]` clearing from sink write rejection handling logic
- Factored out:
  - `WritableStreamDefaultControllerUpdateBackpressureIfNeeded()`
  - `WritableStreamRejectClosedPromiseIfAny()`
  - `WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith()`
- Call what's needed directly in `sink.write()` rejection than using `WritableStreamDefaultControllerErrorIfNeeded()`
- Finish calculating conditions at the top of methods to avoid thinking about where a certain state has been updated or not yet (e.g. `[[error]]`)
- Have redundant call to promise rejection helper method if needed to keep rejection order consistent

ricea:
> - sink.close() rejection doesn't change the rejection reason of writer.ready if it's already rejected due to writer.abort() or controller.error()
>   - Q: Is this ok? When a writer is reobtained, the ready promise is set to be rejected with the latest error.

I think it would be more consistent if the rejection reason changed, but unlikely to make any practical difference to anyone. So not worth making the logic more complicated for.

ricea:
- Rejection order on sink.write() rejection
  - Before: write(), closed, abort(), ready
  - After: write(), abort(), ready, closed
  - Q: Worse?

I like it better actually. Ready normally rejects ASAP to indicate that writing now is useless, whereas closed rejects later so that it can provide a guarantee that everything has reached a stable state. Here both things happen together, but it's good to be consistent. Ideally ready would reject first or second, but it seems unimportant.

tyoshino:
> I don't understand the reason for this change. The promises from Writer methods can remain pending after the the Writer is detached, but until now the .closed and .ready promises have always been rejected immediately.

`.closed` is not. `.ready` was rejected immediately and still.

tyoshino:
Almost finished. I've updated https://github.com/whatwg/streams/pull/655#issuecomment-274404928 by adding notes about the fixes made after the post was made with __NEW__ mark so that ricea can comment to the new ones easily.

tyoshino:
> I think it would be more consistent if the rejection reason changed, but unlikely to make any practical difference to anyone. So not worth making the logic more complicated for.

Thanks. ricea. I'll keep this unaddressed.

tyoshino:
> I like it better actually. Ready normally rejects ASAP to indicate that writing now is useless, whereas closed rejects later so that it can provide a guarantee that everything has reached a stable state. Here both things happen together, but it's good to be consistent. Ideally ready would reject first or second, but it seems unimportant.

Thanks, ricea. Sorry I didn't state what my concern was in the post at that time.I think ready -> closed order is fine, but wondered that it might make sense that abort() reject after write() and close() in queue.

tyoshino:
Finished. Though the discussion with ricea hasn't finished, I start updating the spec text by changing the part not affected by the discussion.

ricea:
> .closed is not. .ready was rejected immediately and still.

Sorry, I misunderstood the change. There is no problem after all.

lgtm 6128a39.

domenic:
Sounds good to me as well.

Do you see a path for fixing https://github.com/whatwg/streams/issues/632 on top of this work?

tyoshino:
Thank you ricea and domenic for checking.

Done adding spec text side change. Self-reviewed and checked rendering result. Not rebased yet.

I'll take a look at #632.

tyoshino:
Squashed and rebased

The last commit before squash and rebase is 8ed23ee2c88befe63b8ee3b0886cefcd3527db6f.

tyoshino:
Sorry. Force-push-ed again to remove __NEW__, __Q__ from the commit message.

tyoshino:
Again, force-pushed to sync the submodule. Only submodule change.

index.bs
```diff
@@ -2768,17 +2768,30 @@ writable stream is <a>locked to a writer</a>.
...
+  1. Assert: _state_ is `"writable"` or `"closing"`.
   1. Let _controller_ be _stream_.[[writableStreamController]].
   1. Assert: _controller_ is not *undefined*.
-  1. If _controller_.[[writing]] is *true* or _controller_.[[inClose]] is *true*,
-    1. Set _stream_.[[pendingAbortRequest]] to <a>a new promise</a>.
-    1. If _controller_.[[writing]] is *true*, return the result of <a>transforming</a> _stream_.[[pendingAbortRequest]]
-       with a fulfillment handler that returns ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
-    1. Otherwise, return _stream_.[[pendingAbortRequest]].
-  1. Return ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
+  1. If _state_ is `"writable"` and ! WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, let _readyPromiseIsPending_ be *true*.
```
ricea: This construction bothers me a bit because my brain interprets it as the declaration of the variable happening inside the if statement. However, I don't have a better idea.

Feel free to disregard this comment.

index.bs
```diff
@@ -2804,50 +2817,134 @@ visible through the {{WritableStream}}'s public API.
...
-    WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, <a>reject</a>
-    _writer_.[[readyPromise]] with _e_.
-    1. Otherwise, set _writer_.[[readyPromise]] to <a>a promise rejected with</a> _e_.
-    1. Set _writer_.[[readyPromise]].[[PromiseIsHandled]] to *true*.
+  1. If _wasAborted_ is *false*, return.
+  1. Perform ! WritableStreamFinishAbort(_stream_, _state_).
+  1. Let _abortRequest_ be _stream_.[[pendingAbortRequest]].
+  1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. Let _promise_ be ! WritableStreamDefaultControllerAbort(_controller_, _abortRequest_.[[reason]]).
+  1. Transform _promise_ with a fulfillment handler which takes the argument _result_ and <a>resolves</a> _abortRequest_.[[promise]] with _result_ and a rejection handler which takes the argument _reason_ and <a>rejects</a> _abortRequest_.[[promise]] with _reason_.
```
ricea: Linkify Transform? #665

index.bs
```diff
@@ -2804,50 +2817,134 @@ visible through the {{WritableStream}}'s public API.
...
+  1. If _wasAborted_ is *true*,
+    1. <a>Reject</a>  _stream_.[[pendingAbortRequest]] with _reason_.
+    1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. If _state_ is `"errored"`,
+    1. Perform ! WritableStreamRejectPromisesInReactionToError(_stream_).
+    1. Return.
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Set _stream_.[[storedError]] to _reason_.
+  1. If _wasAborted_ is *false* and _stream_.[[writer]] is not *undefined*,
+    1. WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _reason_, _readyPromiseIsPending_).
```
ricea: I think this can be on the same line as the if condition.

Also there should be a '!' before WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith.

index.bs
```diff
@@ -2804,50 +2817,134 @@ visible through the {{WritableStream}}'s public API.
...
+    1. <a>Reject</a>  _stream_.[[pendingAbortRequest]] with _reason_.
+    1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. If _state_ is `"errored"`,
+    1. Perform ! WritableStreamRejectClosedPromiseIfAny(_stream_).
+    1. Return.
+  1. Assert: _state_ is `"closing"`.
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Set _stream_.[[storedError]] to _reason_.
+  1. If _wasAborted_ is *false* and _stream_.[[writer]] is not *undefined*,
+    1. WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _reason_, _readyPromiseIsPending_).
```
ricea: Again can be on the same line and missing "Perform !"

index.bs
tyoshino: Yeah, as it's uncommon idiom, I wondered how it looks to you reviewers.

Given that it made you think a bit, I rewrote it by separating it into let and updating statement. I think it's ok. Thanks

index.bs
tyoshino: Good catch! Fixed

index.bs
tyoshino: Yeah. Moved and added "perform !"

index.bs
tyoshino: done

index.bs
```diff
@@ -3156,15 +3261,14 @@ nothrow>WritableStreamDefaultWriterRelease ( <var>writer</var> )</h4>
   1. Assert: _stream_.[[writer]] is _writer_.
   1. Let _releasedError_ be a new *TypeError*.
   1. Let _state_ be _stream_.[[state]].
-  1. If _state_ is `"writable"` or `"closing"`, or _stream_.[[pendingAbortRequest]] is not *undefined*, <a>reject</a>
+  1. Let _controller_ be _stream_.[[writableStreamController]].
+  1. If _state_ is `"writable"` or `"closing"` or _controller_.[[inClose]] is *true* or _controller_.[[writing]] is *true*, or _stream_.[[pendingAbortRequest]] is not *undefined*, <a>reject</a>
```
ricea: The _stream_.[[pendingAbortRequest]] condition doesn't exist in the reference implementation (unless I'm looking at the wrong version).

index.bs
```diff
@@ -2774,8 +2774,8 @@ writable stream is <a>locked to a writer</a>.
   1. Assert: _state_ is `"writable"` or `"closing"`.
   1. Let _controller_ be _stream_.[[writableStreamController]].
   1. Assert: _controller_ is not *undefined*.
-  1. If _state_ is `"writable"` and ! WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, let _readyPromiseIsPending_ be *true*.
-  1. Otherwise, let _readyPromiseIsPending_ be *false*.
+  1. Let _readyPromiseIsPending_ be *false*.
+  1. If _state_ is `"writable"` and ! WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, set _readyPromiseIsPending_ to *true*.
```
ricea: Much better! Thank you!

ricea:
Reviewed everything up to d560c9b.

index.bs
tyoshino: Great catch. Right, I've removed the condition but forgot to do so on the spec text.

Done.

tyoshino:
Addressed the review comments and wrapped long lines where possible.

ricea:
lgtm. I will leave decision on the wrapping to @domenic.

domenic:
I pushed some style tweaks (mostly line wrapping) for the spec text. LGTM now. Remember we want to land tests in WPT first then update this branch. This could be a bit tricky given the changes in https://github.com/w3c/web-platform-tests/pull/4589 so if you want me to handle it let's just make sure to sign off on https://github.com/w3c/web-platform-tests/pull/4587 then I can do both merges.

tyoshino:
Thank you domenic for the commit with great catches in bac156e2a85180749782a7d620ccd892d0fe4ca1. LGTM.

> Remember we want to land tests in WPT first then update this branch. This could be a bit tricky given the changes in w3c/web-platform-tests#4589

I've updated run-web-platform-tests.js in 59820d55150eb545744e5f2d84ba722caa469b5d as you suggested in the OP of the PR, and updated submodule in b6a2cbf5d09df8d676c8edd3e83dd6684f799ee4.

> so if you want me to handle it let's just make sure to sign off on w3c/web-platform-tests#4587 then I can do both merges.

Thanks for merging.

tyoshino:
push is failing on travis at deploy.sh

tyoshino:
Looks bikeshed API is down. Investigating.

tyoshino:
Looks this is due to https://github.com/tabatkins/bikeshed/pull/843

tyoshino:
https://tabatkins.github.io/bikeshed/#metadata-inline-github-issues is not yet updated.

According to the error message, we can choose from 'title', 'full' or false/no/n/off.

Tentatively changed it to `title` from `true` in 83dbf131a5c801b1d1e6a7fad1079fe37b362ac9.

tyoshino:
Squashed and updated the commit message to explain the additional works required for submission. The last commit before this was 83dbf131a5c801b1d1e6a7fad1079fe37b362ac9.

tyoshino:
Done. Finally. Thank you for your review.

Issue #582, opened 2016-10-31
"Set totalSize to totalSize + pair.[[size]]."


ricea:
https://streams.spec.whatwg.org/#get-total-queue-size

> Set totalSize to totalSize \+ pair\.\[\[size\]\]\.

Maybe there should be a note here that this algorithm can't be used as-is because it results in O(N^2) behaviour? I just tripped over this myself.

domenic:
That's what's intended by https://streams.spec.whatwg.org/#queue-with-sizes:

> although in implementations it would of course be backed by a more efficient data structure

But maybe we could make it more prominent with a `<div class="warning">` in the problematic algorithm in particular. Or even just overhaul the whole thing to actually be efficient.

isonmad:
One thing I don't get is the inconsistency between byte queues and non byte queues. Bytestream controllers have [[totalQueuedBytes]], which is constantly updated, but the default controller uses GetTotalQueueSize, and the reference implementation doesn't even follow that as-written and keeps a running total like bytestream controllers do, because it's more efficient.

Shouldn't both be either one or the other?

ricea:
Thanks @isonmad, I hadn't noticed that byte queues already do the right thing. It doesn't seem like we're gaining much in terms of readability by doing the wrong thing for non-byte queues.

The reference implementation tracks the size by adding a `.totalSize` attribute to the Array holding the queue. We can't standardise that. The Chrome implementation uses a separate [[queueSize]] internal slot on the controller, which works, but I got confused between queue _size_ and queue _length_ when I was implementing it. I think we need a different name.

The [[totalQueuedBytes]] name used by byte queues is also unsuitable, but maybe [[totalQueuedSize]] is clear enough.

Bikeshedding welcome.

tyoshino:
We just built the spec text of the byte stream by translating the reference implementation without paying attention to this point. I don't quite remember but the fact the record used for the byte stream (see ReadableByteStreamControllerEnqueueChunkToQueue()) is different from the one for the non-byte stream and that chunks of the byte stream can be partially consumed might have made me reluctant to think about the consistency.

`totalQueuedSize` sounds ok.

isonmad:
Wait, isn't there a reason you'd want to spec it as adding up the numbers every time for the case where each size can return an arbitrary Number, if not implement it?

Like if the size starts to exceed `Number.MAX_SAFE_INTEGER`, could subtracting one chunk's size from the total end up not the same result as adding up all remaining chunks again.

[...]

PR #657, opened 2017-01-17
precision issues with very large chunks and desiredSize

isonmad:
Per https://github.com/whatwg/streams/issues/582#issuecomment-273016992 there's an observable problem with the current shortcut.

Should it be changed to follow the spec directly and add up the chunk sizes every time, or use some other way to cheat?


ricea:
There's a similar issue with small chunk sizes, ie.

First chunk is size 1e-16
Second chunk is size 1

ricea:
I found a case where the totalSize ends up > 0:

First chunk is size 2e-16
Second chunk is size 1

PR #661, opened 2017-01-18
Fix queue total size tracking logic

### LINKED ISSUE #582: "Set totalSize to totalSize + pair.[[size]]."

domenic:
Previously, we believed that the spec's inefficient-but-clear method, of adding up all the chunk sizes whenever it wanted to get the total queue size, was equivalent to keeping a running count. This was discussed in #582; note also #491 where we changed the reference implementation to the running-count strategy for speed reasons. As discovered in https://github.com/whatwg/streams/issues/582#issuecomment-273016992, and further elaborated on in #657, these two methods are not in fact equivalent, due to the vagaries of floating-point arithmetic. <del>(That is, they are equivalent if you assume the addition is floating-point addition, but that is less than clear from the spec.)</del> <ins>I was confused; see below. Let's remember to remove that sentence from the commit message when merging.</ins>

As such, this commit switches the spec to the more efficient running-count method, as that's realistically the only implementable version. It also adds tests to ensure floating point arithmetic is being used, exposing such cases.

This commit also includes a few unobservable cleanups:

- It introduces the ResetQueue abstract operation, to ensure that the queue total size gets reset to zero when the queue is cleared. This should not matter because no code paths check the queue's total size after it has been cleared, but keeping the two slots in sync seems virtuous.

- It updates the internal slot name for ReadableByteStreamController instances from [[totalQueuedBytes]] to [[queueTotalSize]], to be consistent with those for the other queue-containers. We do not yet use the queue-with-sizes abstract operations (except ResetQueue) on ReadableByteStreamController instances as the queue management is significantly more complicated there. But aligning makes the spec easier to read.

Closes #582. Closes #657.

---

Things to contemplate:

- I realized after making this whole change that we probably could have gotten away with simply reminding everyone that all arithmetic is floating point arithmetic. Unless I am very confused, I think the running-count methods and the sum-as-needed methods are equivalent as long as both use the same type of arithmetic. So really the only necessary part of this pull request is 

  ```diff
   +It's also worth noting that, as in [[!ECMASCRIPT]], all numbers are represented as double-precision floating point
   +values, and all arithmetic operations performed on them must be done in the usual way for such values.
   ```

- I would really like to unify the readable byte stream queue so that it uses the queue with sizes operations, instead of its own bespoke stuff. But I'll leave that for future work. There's no obvious way to do so that isn't at least a bit awkward for someone.

---

Tests: https://github.com/w3c/web-platform-tests/pull/4568 (for the updated semantics discussed in #582).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3808,6 +3808,8 @@ total in the \[[queueTotalSize]] slot, is <em>not</em> equivalent to adding up t
   1. Remove _pair_ from _container_.[[queue]], shifting all other elements downward (so that the second becomes the
      first, and so on).
   1. Set _container_.[[queueTotalSize]] to _container_.[[queueTotalSize]] − _pair_.[[size]].
+  1. If _container_.[[queueTotalSize]] < *0*, set _container_.[[queueTotalSize]] to *0*. (This can occur due to
+     floating-point arithmetic.)
   1. Return _pair_.[[value]].
 </emu-alg>
 
--- a/index.bs
+++ b/index.bs
@@ -30,6 +30,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
     text: %Uint8Array%; url: #sec-typedarray-objects; type: constructor
     text: ArrayBuffer; url: #sec-arraybuffer-objects; type: interface
     text: DataView; url: #sec-dataview-objects; type: interface
+    text: Number; url: #sec-ecmascript-language-types-number-type; type: interface
     text: Uint8Array; url: #sec-typedarray-objects; type: interface
     text: typed array; url: #sec-typedarray-objects; type: dfn
     text: the typed array constructors table; url: #table-49; type: dfn
@@ -1470,6 +1471,10 @@ Instances of {{ReadableStreamDefaultController}} are created with the internal s
     <td>A List representing the stream's internal queue of <a>chunks</a>
   </tr>
   <tr>
+    <td>\[[queueTotalSize]]
+    <td>The total size of all the chunks stored in \[[queue]] (see [[#queue-with-sizes]])
+  </tr>
+  <tr>
     <td>\[[started]]
     <td>A boolean flag indicating whether the <a>underlying source</a> has finished starting
   </tr>
@@ -1505,7 +1510,7 @@ ReadableStreamDefaultController(<var>stream</var>, <var>underlyingSource</var>, 
   1. If _stream_.[[readableStreamController]] is not *undefined*, throw a *TypeError* exception.
   1. Set *this*.[[controlledReadableStream]] to _stream_.
   1. Set *this*.[[underlyingSource]] to _underlyingSource_.
-  1. Set *this*.[[queue]] to a new empty List.
+  1. Perform ! ResetQueue(*this*).
   1. Set *this*.[[started]], *this*.[[closeRequested]], *this*.[[pullAgain]], and *this*.[[pulling]] to *false*.
   1. Let _normalizedStrategy_ be ? ValidateAndNormalizeQueuingStrategy(_size_, _highWaterMark_).
   1. Set *this*.[[strategySize]] to _normalizedStrategy_.[[size]] and *this*.[[strategyHWM]] to
@@ -1588,7 +1593,7 @@ polymorphic dispatch from the readable stream implementation to either these or 
 <h5 id="rs-default-controller-private-cancel">\[[Cancel]](<var>reason</var>)</h5>
 
 <emu-alg>
-  1. Set *this*.[[queue]] to a new empty List.
+  1. Perform ! ResetQueue(*this*).
   1. Return ! PromiseInvokeOrNoop(*this*.[[underlyingSource]], `"cancel"`, « _reason_ »)
 </emu-alg>
 
@@ -1597,7 +1602,7 @@ polymorphic dispatch from the readable stream implementation to either these or 
 <emu-alg>
   1. Let _stream_ be *this*.[[controlledReadableStream]].
   1. If *this*.[[queue]] is not empty,
-    1. Let _chunk_ be ! DequeueValue(*this*.[[queue]]).
+    1. Let _chunk_ be ! DequeueValue(*this*).
     1. If *this*.[[closeRequested]] is *true* and *this*.[[queue]] is empty, perform ! ReadableStreamClose(_stream_).
     1. Otherwise, perform ! ReadableStreamDefaultControllerCallPullIfNeeded(*this*).
     1. Return <a>a promise resolved with</a> ! CreateIterResultObject(_chunk_, *false*).
@@ -1690,7 +1695,7 @@ asserts).
         1. Perform ! ReadableStreamDefaultControllerErrorIfNeeded(_controller_, _chunkSize_.[[Value]]).
         1. Return _chunkSize_.
       1. Let _chunkSize_ be _chunkSize_.[[Value]].
-    1. Let _enqueueResult_ be ! EnqueueValueWithSize(_controller_.[[queue]], _chunk_, _chunkSize_).
+    1. Let _enqueueResult_ be ! EnqueueValueWithSize(_controller_, _chunk_, _chunkSize_).
     1. If _enqueueResult_ is an abrupt completion,
       1. Perform ! ReadableStreamDefaultControllerErrorIfNeeded(_controller_, _enqueueResult_.[[Value]]).
       1. Return _enqueueResult_.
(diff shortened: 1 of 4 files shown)


isonmad:
>  Unless I am very confused, I think the running-count methods and the sum-as-needed methods are equivalent as long as both use the same type of arithmetic.

Except you eventually have to subtract from the running-count total when you dequeue something, and that isn't equivalent, as said in https://github.com/whatwg/streams/issues/582#issuecomment-273016992 and tested in https://github.com/whatwg/streams/pull/657

domenic:
> Except you eventually have to subtract from the running-count total when you dequeue something, and that isn't equivalent, as said in #582 (comment) and tested in #657

Ah, right, thanks! It's a + b - a vs. just b that we're comparing, not two ways of adding up a + b.

isonmad:
Looks like the ResetQueue stuff needs to be added to Chromium, since it does exactly what the reference implementation does right now.

```js
  var strategy = {
    size(x) {
      return x;
    },
    highWaterMark: 0
  };
  var controller;
  var rs = new ReadableStream({start(c){controller = c;}}, strategy);

  var reader = rs.getReader();
  console.log(controller.desiredSize); // prints 0 when empty

  controller.enqueue(2);
  controller.enqueue(Number.MAX_SAFE_INTEGER);
  console.log(reader.read());
  reader.read().then(() => console.log(controller.desiredSize)); // prints 1 when empty
```

domenic:
Oh, I didn't realize that was observable. OK, let me add some tests. Thank you for catching that.

domenic:
Note that the specific case you mention is fine. We don't call ResetQueue when the queue is empty. We call it when closing or erroring the stream.

I guess that is a bit weird that sometimes when the queue is empty during the course of operation desiredSize will be nonzero because of floating-point shenanigans, and sometimes it will be reset to 0. We could either accept this, or manually make DequeueValue reset the total size to 0 if the queue becomes empty.

domenic:
Marking "do not merge yet" until we get some clarity on which approach we want to take for ResetQueue, and add tests for it. (And then, per #662, send those tests to the WPT project.)

reference-implementation/lib/queue-with-sizes.js
```diff
@@ -2,39 +2,43 @@
 const assert = require('assert');
 const { IsFiniteNonNegativeNumber } = require('./helpers.js');
 
-exports.DequeueValue = queue => {
-  assert(queue.length > 0, 'Spec-level failure: should never dequeue from an empty queue.');
-  const pair = queue.shift();
+exports.DequeueValue = container => {
+  assert('_queue' in container && '_queueTotalSize' in container,
+    'Spec-level failure: ResetQueue should only be used on containers with [[queue]] and [[queueTotalSize]].');
```
isonmad: Copypasted assert text.

reference-implementation/lib/queue-with-sizes.js
```diff
@@ -2,39 +2,43 @@
...
+  const pair = container._queue.shift();
+  container._queueTotalSize -= pair.size;
 
   return pair.value;
 };
 
-exports.EnqueueValueWithSize = (queue, value, size) => {
+exports.EnqueueValueWithSize = (container, value, size) => {
+  assert('_queue' in container && '_queueTotalSize' in container,
+    'Spec-level failure: ResetQueue should only be used on containers with [[queue]] and [[queueTotalSize]].');
```
isonmad: Same here.

reference-implementation/lib/queue-with-sizes.js
```diff
@@ -2,39 +2,43 @@
...
-// This implementation is not per-spec. Total size is cached for speed.
-exports.GetTotalQueueSize = queue => {
-  if (queue._totalSize === undefined) {
-    queue._totalSize = 0;
-  }
-  return queue._totalSize;
-};
+exports.PeekQueueValue = container => {
+  assert('_queue' in container && '_queueTotalSize' in container,
+    'Spec-level failure: ResetQueue should only be used on containers with [[queue]] and [[queueTotalSize]].');
```
isonmad: Same here.

tschneidereit:
> I guess that is a bit weird that sometimes when the queue is empty during the course of operation desiredSize will be nonzero because of floating-point shenanigans, and sometimes it will be reset to 0. We could either accept this, or manually make DequeueValue reset the total size to 0 if the queue becomes empty.

Or we could clamp the queue size to >= 0 in DequeueValue, ensuring that we'll always get predictable results. I might be missing something that makes this undesirable.

Apart from that point, I can't see anything amiss here, looks great!

domenic:
OK, I've updated to add a clamping step to DequeueValue, since it seems both you and @isonmad are in favor of predictable results there, even when users are doing silly things.

@tyoshino @ricea want to review? Also need sign-off for https://github.com/w3c/web-platform-tests/pull/4568.

reference-implementation/lib/readable-stream.js
```diff
@@ -859,7 +859,10 @@ class ReadableStreamDefaultController {
 
     this._underlyingSource = underlyingSource;
 
-    this._queue = [];
+    // Need to set the slots so that the assert doesn't fire. In the spec the slots already exist implicitly.
+    this._queue = this._queueTotalSize = undefined;
```
ricea: Optional: I'd prefer not to use chained assignment. As far as I can tell the rest of the reference implementation sticks to one assignment per statement.

index.bs
```diff
@@ -3767,46 +3788,57 @@ CountQueuingStrategy({ <var>highWaterMark</var> })</h4>
...
+determined sizes. Various specification objects contain a queue-with-sizes, represented by the object having two paired
+internal slots, always named \[[queue]] and \[[queueTotalSize]]. \[[queue]] is a List of Records with \[[value]] and
+\[[size]] fields, and \[[queueTotalSize]] is a JavaScript {{Number}}, i.e. a double-precision floating point number.
+
+The following abstract operations are used when operating on objects that contain queues-with-sizes, in order to ensure
+that the two internal slots stay synchronized.
 
-A number of abstract operations are specified here to make working with queues-with-sizes more pleasant, and used
-throughout the rest of this standard.
+<p class="warning">Due to the vagaries of floating point arithmetic, the framework specified here, of keeping a running
```
ricea: Optional: maybe we should say "Due to the limited precision of floating point arithmetic", just to be slightly more specific.

index.bs
```diff
@@ -3767,46 +3788,57 @@ CountQueuingStrategy({ <var>highWaterMark</var> })</h4>
...
+total in the \[[queueTotalSize]] slot, is <em>not</em> equivalent to adding up the size of all <a>chunks</a> in
+\[[queue]]. This is most apparent when the total size is very small or very large.</p>
```
ricea: Maybe this is overly pessimistic. We could say "This only makes a difference when there is huge (~10<sup>15</sup>) variance in size between chunks, or trillions of chunks are queued."

index.bs
```diff
@@ -3767,46 +3788,57 @@ CountQueuingStrategy({ <var>highWaterMark</var> })</h4>
...
-  1. Let _pair_ be the first element of _queue_.
-  1. Remove _pair_ from _queue_, shifting all other elements downward (so that the second becomes the first, and so on).
+  1. Assert: _container_ has [[queue]] and [[queueTotalSize]] internal slots.
+  1. Assert: _container_.[[queue]] is not empty.
+  1. Let _pair_ be the first element of _container_.[[queue]].
+  1. Remove _pair_ from _container_.[[queue]], shifting all other elements downward (so that the second becomes the
+     first, and so on).
+  1. Set _container_.[[queueTotalSize]] to _container_.[[queueTotalSize]] − _pair_.[[size]].
+  1. If _container_.[[queueTotalSize]] < *0*, set _container_.[[queueTotalSize]] to *0*. (This can occur due to
+     floating-point arithmetic.)
```
ricea: Optional: "rounding errors" rather than "floating-point arithmetic".

index.bs
```diff
@@ -4371,6 +4403,9 @@ itself will evolve in these ways.
     ECMAScript spec does.
 </ul>
 
+It's also worth noting that, as in [[!ECMASCRIPT]], all numbers are represented as double-precision floating point
+values, and all arithmetic operations performed on them must be done in the usual way for such values.
```
ricea: How about "all arithmetic operations must be performed in the standard way..."?

PR #662, opened 2017-01-19
Move to the upstream versions of WPTs

domenic:
In https://github.com/w3c/web-platform-tests/commit/55e9a48c3b1218b64998e8a3f30ab3d274353f5f I upstreamed our current test corpus, minus transform streams, to web platform tests. This PR deletes those upstreamed tests from our tree, and moves to using a Git submodule (hat-tip @isonmad) for pulling them in.

I want to merge this sooner rather than later, as it of course conflicts with the majority of open PRs. I just want to make sure all active contributors are aware of the change in workflow and how this will affect things going forward. (/cc @isonmad, @tschneidereit, @tyoshino, @ricea)

After this change, for non-transform stream tests, we'll work to send two PRs per spec change: one to web platform tests, and one to this repo. We'll review the tests, merge them, then update the PR here to change the submodule metadata. I tried to document in this in the reference-implementation/README.md.

We can also work directly in to-upstream-wpts; that will still work. But it's less preferred, and doesn't help for cases where we need to modify or add to existing tests.

Does this sound good to people?

Note: the following in-flight PRs actively under consideration would need to move to this new strategy:

- [ ] #634 
- [ ] #651 (once it gets tests)
- [ ] #655 
- [ ] #661


reference-implementation/README.md
```diff
@@ -12,30 +12,47 @@ Test coverage is not complete, but we do aim for it to be. Adding tests would be
...
+The test runner here is a Node.js emulated-DOM environment, with the reference implementation loaded into it.
 
-The web platform tests for streams are found in the [streams directory](https://github.com/w3c/web-platform-tests/tree/master/streams) of the web platform tests repository, and maintained via pull requests to that repository.
+#### Upstream web platform tests
 
-For local development while working on the spec, you can run
+The web platform tests for streams are found in the [streams directory](https://github.com/w3c/web-platform-tests/tree/master/streams) of the web platform tests repository, and maintained via pull requests to that repository. They are then pulled into this repository via a [Git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules).
+
+This means that in order to land a test change for these tests, you'll need to make a pull request to the web platform tests repository, and then update the submodule pointer in this repository (probably in the same pull request as your spec change). That can be done via the command
```
ricea: I think it's possible to work directly in the web-platform-tests subdirectory. You'd still have to merge the new tests there before merging here, however. I haven't actually tried it though. Once we have more experience with the new workflow we can probably make these instructions more explicit.

reference-implementation/package.json
```diff
@@ -2,8 +2,9 @@
   "private": true,
   "description": "Reference implementation and tests for the WHATWG Streams Standard",
   "scripts": {
-    "pretest": "bash ./update-web-platform-tests.sh",
-    "test": "npm run lint && node --expose_gc run-tests.js | tap-spec && npm run wpt",
+    "pretest": "git submodule init && git submodule update",
```
ricea: Looks like `git submodule update --init` will work here.

reference-implementation/README.md
```diff
@@ -12,30 +12,47 @@ Test coverage is not complete, but we do aim for it to be. Adding tests would be
...
 
-The original tests are written using the [tape](https://github.com/substack/tape) framework.
+### Legacy tests
 
-- To run them, type `npm test` in this folder.
-- To run a specific test file, type e.g. `node test-runner test/readable-stream.js`.
+The original tests are written using the [tape](https://github.com/substack/tape) framework. They live in the `test` subfolder of this folder.
+
+- To run them, type `npm run legacy-test` in this folder.
+- To run a specific test file, type e.g. `node run-tests.js test/readable-stream.js`.
```
tschneidereit: I guess this should mention a test file that actually exists. Given that that's the only one that'll stick around, that should probably be `abstract-ops.js`.

domenic: IMO once we get it down to just abstract-ops.js we should eliminate the tests as maintaining the separate framework is a burden. But yeah, I'll change it for now.

reference-implementation/README.md
domenic: Yeah, I tried to allude to that a couple paragraphs down, but we'll see how well it works in practice.

domenic:
> I had to run "git checkout -- ." in the web-platform-tests directory before the tests would pass, but maybe that was just something weird about my environment.

I think this is what the pretest step is supposed to do. Let me make that part of `npm run wpt` instead of part of the overall `npm test`.

PR #663, opened 2017-01-19
Reduce number of iterations in write test

### LINKED ISSUE #660: write.https.html test is slow with full memory checking
Chromium has a bug related to this test, http://crbug.com/671165 due it timing out on MSAN bots. I am currently testing a change which reduces the number of iterations in the test 'a large queue of writes should be processed completely' (https://codereview.chromium.org/2644473002). In future Chromium will import the tests from the web-platform-tests so I would like to upstream this change. I don't think it makes the test less useful.

ricea:
Test 'a large queue of writes should be processed completely' was timing
out in some configurations.

Reduce the number of iterations from 10000 to 1000.

Fixes #660.

### DIFF
--- a/reference-implementation/to-upstream-wpts/writable-streams/write.js
+++ b/reference-implementation/to-upstream-wpts/writable-streams/write.js
@@ -165,7 +165,7 @@ promise_test(t => {
 }, 'when sink\'s write throws an error, the stream should become errored and the promise should reject');
 
 promise_test(() => {
-  const numberOfWrites = 10000;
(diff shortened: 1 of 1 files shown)


ricea:
The imported web-platform-tests are timing out for the Worker versions. Waiting for a fix upstream.

ricea:
Change of plan: I'm going to land this in the w-p-t repository.

tyoshino:
sg

domenic:
moved to https://github.com/w3c/web-platform-tests/pull/4584

PR #664, opened 2017-01-19
Mark pipeTo promise created by pipeThrough as handled

### LINKED ISSUE #652: Should pipeThrough() mark the promise from pipeTo() as handled?
It seems really annoying that if one link in the chain fails every single pipeThrough() in the chain will complain about it.

If a transform in the middle of the pipe fails, are the separate complaints useful? I can think of two scenarios where they might be:
1. Something after the failed step in the chain has a bug which causes it to fail on abort/cancel. Then that error will be propagated, masking the real problem.
2. preventAbort is set, stopping propagation.

domenic:
Closes #652.

Tests: https://github.com/w3c/web-platform-tests/pull/4572

Ideally the sequence here is:

- I get a LGTM on the patch and tests
- I merge the tests
- I push another commit here updating to the latest WPTs
- I squash merge so that we land a single commit on whatwg/streams that includes both the spec/reference implementation update, and the new test coverage.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -564,7 +564,9 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
 </div>
 
 <emu-alg>
-  1. Perform ? Invoke(*this*, `"pipeTo"`, « _writable_, _options_ »).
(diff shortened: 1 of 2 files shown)


ricea:
lgtm once you've added the web-platform-tests submodule update.

PR #671, opened 2017-01-25
Factor out readyPromiseIsPending calculation

tyoshino:
First, this patch factors out readyPromiseIsPending calculation into a function named WritableStreamIsReadyPromisePending().

I also tried to move the readyPromiseIsPending calculation to as earlier position as possible in the caller methods so that we're less likely to call the newly introduced method after changing some of state variables by mistake.

However, while doing so, I found that calling this method in WritableStreamDefaultControllerError() is problematic. Since WritableStreamDefaultControllerError() can be called while the instantiated controller is not yet set to the stream i.e. inside sink.start().

There's similar hack at the bottom of WritableStreamDefaultControllerError() though it seems no longer working (before c6c0a6e68a427be34f0bd1166c991f811379c32a, it was placed in WritableStreamError() and retrieving the controller from the stream instance, and therefore had the same issue). Rather than getting bothered by these corner case handling now and in the future, this patch lets the stream finish instantiation of a controller and then invoke sink.start().

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2774,7 +2774,7 @@ writable stream is <a>locked to a writer</a>.
   1. Assert: _state_ is `"writable"` or `"closing"`.
   1. Let _controller_ be _stream_.[[writableStreamController]].
   1. Assert: _controller_ is not *undefined*.
-  1. Let _readyPromiseIsPending_ be WritableStreamIsReadyPromisePending(_stream_, _oldState_, *false*).
+  1. Let _readyPromiseIsPending_ be ! WritableStreamIsReadyPromisePending(_stream_, _oldState_, *false*).
   1. If _controller_.[[writing]] is *false* and _controller_.[[inClose]] is *false*,
     1. If _stream_.[[writer]] is not *undefined*, perform !
        WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _error_, _readyPromiseIsPending_).
@@ -2855,7 +2855,7 @@ nothrow>WritableStreamFinishPendingWriteWithError ( <var>stream</var>, <var>reas
   1. Let _state_ be _stream_.[[state]].
   1. Let _wasAborted_ be *false*.
   1. If _stream_.[[pendingAbortRequest]] is not *undefined*, set _wasAborted_ to *true*.
-  1. Let _readyPromiseIsPending_ be WritableStreamIsReadyPromisePending(_stream_, _oldState_, _wasAborted_).
+  1. Let _readyPromiseIsPending_ be ! WritableStreamIsReadyPromisePending(_stream_, _oldState_, _wasAborted_).
   1. If _wasAborted_ is *true*,
     1. <a>Reject</a>  _stream_.[[pendingAbortRequest]] with _reason_.
     1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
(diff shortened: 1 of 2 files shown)


tyoshino:
In addition to this change, I'm planning to attempt to add a variable to the WritableStream indicating current backpressure status and remove the WritableStreamDefaultControllerGetBackpressure() call to evaluate current backpressure.

Probing the controller every time also has some benefit e.g. reducing variables, but I guess code (spec text) will be reduced more if we do so.

index.bs
```diff
@@ -2683,6 +2683,7 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
      potential types in the future, without backward-compatibility concerns.</p>
   1. Set *this*.[[writableStreamController]] to ? Construct(`<a idl>WritableStreamDefaultController</a>`, « *this*,
      _underlyingSink_, _size_, _highWaterMark_ »).
+  1. Return ? WritableStreamDefaultControllerStart(*this*.[[writableStreamController]]).
```
domenic: Why factor this out? It doesn't seem to meet the criteria in https://github.com/whatwg/streams/blob/master/Contributing.md#spec-editorial-conventions

index.bs
```diff
@@ -2932,6 +2924,15 @@ nothrow>WritableStreamFinishPendingCloseWithError ( <var>stream</var>, <var>reas
   1. Perform ! WritableStreamRejectClosedPromiseIfAny(_stream_).
 </emu-alg>
 
+<h4 id="writable-stream-is-ready-promise-pending" aoid="WritableStreamIsReadyPromisePending"
+nothrow>WritableStreamIsReadyPromisePending ( <var>stream</var>, <var>state</var>, <var>wasAborted</var> )</h4>
```
domenic: I guess this is a preexisting problem, but I am not sure this is the best name. I would expect "ReadyPromiseIsPending" to just return whether stream.[[writer]].[[readyPromise]].[[PromiseState]] is "pending". This instead does a complex calculation on the states and the backpressure. What is the actual thing we are trying to capture here? Maybe "WritableStreamIsReadyForWrites"?

index.bs
```diff
@@ -2932,6 +2924,15 @@ nothrow>WritableStreamFinishPendingCloseWithError ( <var>stream</var>, <var>reas
...
+
+<emu-alg>
```
domenic: I really don't like passing in state and wasAborted. Can we get those from the stream instead? It seems they are almost always computed the same way, so that would just be more good deduplication.

index.bs
tyoshino: Please see the following sentence in the OP. This is not for code reduction or operation exporting, but to set _writableStreamController before we call user defined methods to which we pass the created controller. IIUC, the criteria is for judging whether or not to do a change that doesn't change the meaning of the code.

> There's similar hack at the bottom of WritableStreamDefaultControllerError() though it seems no longer working (before c6c0a6e, it was placed in WritableStreamError() and retrieving the controller from the stream instance, and therefore had the same issue). Rather than getting bothered by these corner case handling now and in the future, this patch lets the stream finish instantiation of a controller and then invoke sink.start().

tyoshino: Maybe it was my mistake that I said "factoring out" here. It's not refactoring but changing the timing of _writableStreamController substitution.

domenic: But why not just inline it? Moving it here is fine but I don't understand why it needs its own abstract op.

tyoshino: Ah, I got your point. I'll try to rewrite so.

index.bs
tyoshino: Thanks. I'll use WritableStreamIsReadyForWrites. Do you think IsWritableStreamReadyForWrites better?

index.bs
tyoshino: Yeah, good point. The new function should just look at `stream`, and it should be placed at some safe place before updating the state variables. Will fix.

index.bs
tyoshino: Sorry, again. This logic touches the slots encapsulated in the controller. So, it should be a part of the WritableStreamDefaultController interfaces. That said, it was placed incorrectly. It should be in the section containing abstract operations exposed to the WritableStream. Fixing.

tyoshino:
I've merged this change into the PR #672 while incorporating domenic's comments.

Issue #669, opened 2017-01-24
Don't touch controller directly in WritableStreamDefaultWriterRelease() 


tyoshino:
Follow up for #655. Once the PR is committed, `WritableStreamDefaultWriterRelease()` would look at the controller directly. This shouldn't be a part `WritableStreamDefaultWriterRelease()` which is one of the WritableStream's abstract operation.

tyoshino:
a25f410cbccfbaf56479ca3224fe57bab8b2b6fa has added `[[backpressure]]` to WritableStream to track the backpressure signal and changed WritableStreamDefaultWriterRelease to look at it. So, this issue has been fixed.

Issue #633, opened 2016-12-16
Consolidate "closing" state and [[pendingClosePromise]]?


domenic:
I think they are redundant, that is, everywhere we check "writing", we can check if [[pendingClosePromise]] is not undefined and state is "writable".

It's nicer to not have the redundancy since right now the spec has to make sure they stay in sync manually.

ricea:
I think you are correct.

I felt that including these things in [[state]] was useful as explanation of the lifecycle of the stream, but I'm really not sure now.

tyoshino:
This is done by a25f410cbccfbaf56479ca3224fe57bab8b2b6fa!

PR #672, opened 2017-01-25
Factor out condition to check pending operation

tyoshino:


tyoshino:
I've also added a bigger change cd7ad8d9f5d0aaaba9a461e860a399e1b2b910bd thought it's WIP (spec text is not in sync with the impl). This would fix #632, #669 and #633.

tyoshino:
The WIP commit cd7ad8d9f5d0aaaba9a461e860a399e1b2b910bd is also an answer for https://github.com/whatwg/streams/pull/655#issuecomment-272812907 and domenic's reply https://github.com/whatwg/streams/pull/655#pullrequestreview-17110046. All the abort delaying logic will be a part of the WritableStream interface.

tyoshino:
If we throw away the `"closing"` state, we need to check `[[closeRequest]]`. But we can avoid inflating code by inflated conditions by factoring out conditions. #671 is one of such a possible refactoring.

domenic:
Really excited about this change; it is a very nice cleanup and fix. Please remove "do not merge yet" when it is ready for review :)

index.bs
```diff
@@ -3470,6 +3471,15 @@ nothrow>WritableStreamDefaultControllerGetDesiredSize ( <var>controller</var> )<
   1. Return _controller_.[[strategyHWM]] − _queueSize_.
 </emu-alg>
 
+<h4 id="writable-stream-default-controller-should-delay-close-update"
```
isonmad: An old draft name?

tyoshino:
Commit message draft

Bug fix:
- `WritableStreamDefaultWriter` should be constructed with its `[[readyPromise]]` rejected if the stream has a pending `abort()` or is `"errored"`
- `abort()` should be rejected with the error passed to `controller.error()` called during a pending `write()`/`close()`. Not the rejection reason of the `write()`/`close()` (fix in `WritableStreamFinishPendingCloseWithError()`)

Promise resolution order change:
- Moved `[[readyPromise]]` resolution in `WritableStreamDefaultWriterRelease()` to be consistent with the other places
- Reject `[[readyPromise]]` before aborting pending `abort()`

Refactoring:
- Renamed `[[pendingWriteRequest]]` to `[[inflightWriteRequest]]`
- Changed `[[writing]]` users to use `[[inflightWriteRequest]]`
- Added `[[inflightCloseRequest]]` to replace `[[inClose]]`
- Renamed `[[pendingCloseRequest]]` to `[[closeRequest]]`
  - More aligned with `[[writeRequests]]`
  - Less confusing as it's role is not so similar to `[[pendingAbortRequest]]`
- Removed `"closing"` state and let the users investigate whether or not `[[closeRequest]]` is *undefined* instead
- Let WritableStream remember the last backpressure signal in `[[backpressure]]`
  - `eliminates WritableStreamDefaultControllerGetBackpressure()` calls
- Factored out common logic into `WritableStreamHasOperationMarkedInflight()`
- Factored `readyPromiseIsPending` calculation logic into `WritableStreamIsReadyForWrites()` (merged from #671)
  - Also moved the `readyPromiseIsPending` calculation to as earlier position as possible in the caller methods so that we're less likely to call the newly introduced method after changing some of state variables by mistake. However, while doing so, found that calling this method in `WritableStreamDefaultControllerError()` is problematic. Since `WritableStreamDefaultControllerError()` can be called while the instantiated controller is not yet set to the stream i.e. inside `sink.start()`. There's similar hack at the bottom of `WritableStreamDefaultControllerError()` though it seems no longer working (before c6c0a6e68a427be34f0bd1166c991f811379c32a, it was placed in `WritableStreamError()` and retrieving the controller from the stream instance, and therefore had the same issue).
  - Rather than getting bothered by these corner case handling now and in the future, this patch let the stream finish instantiation of a controller and then invoke `sink.start()`. The split logic is put in a method named `WritableStreamDefaultControllerStart()`. It's a part of the controller since it touches variables encapsulated in the controller. This change doesn't break the mechanism in the `WritableStreamDefaultController()`'s constructor to prevent it from being invoked manually.
- Factored out pending `abort()` rejection logic into `WritableStreamRejectPendingAbortRequest()`

Editorial:
- Moved `WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith()` to the right section

index.bs
tyoshino: As noted in https://github.com/whatwg/streams/pull/672#issuecomment-275084660, spec is not yet sync with the impl. But thanks for review. I'll make sure that this gets fixed.

tyoshino:
Reference impl ready.

TODO:
- Add tests
- Sync spec text

domenic:
Great overview commit message. However it doesn't really mention how #632 was fixed, or am I missing something?

> WritableStreamIsReadyForWrites() 

This is better than my suggestion of putting Is... first, agreed.

Can you and ricea@ be sure to check that, since this closes #632, it also takes care of the failing case in Blink's pipeTo mentioned in https://bugs.chromium.org/p/chromium/issues/detail?id=684543?

reference-implementation/lib/writable-stream.js
```diff
@@ -122,22 +139,14 @@ function WritableStreamAbort(stream, reason) {
     return Promise.reject(error);
   }
 
-  assert(state === 'writable' || state === 'closing', 'state must be writable or closing');
+  assert(state === 'writable', 'state must be writable or closing');
```
domenic: assert message needs updating

reference-implementation/lib/writable-stream.js
```diff
@@ -182,50 +187,39 @@ function WritableStreamAddWriteRequest(stream) {
   return promise;
 }
 
-function WritableStreamFinishPendingWrite(stream) {
-  assert(stream._pendingWriteRequest !== undefined);
-  stream._pendingWriteRequest._resolve(undefined);
-  stream._pendingWriteRequest = undefined;
+function WritableStreamFinishInflightWrite(stream) {
```
domenic: I guess we should treat "in flight" as two words, so inFlight/InFlight in code.

reference-implementation/lib/writable-stream.js
```diff
@@ -533,6 +541,24 @@ function IsWritableStreamDefaultWriter(x) {
   return true;
 }
 
+function WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(stream, error, isPending) {
```
domenic: It seems nicer to have this recompute WritableStreamIsReadyForWrites each time. Is passing it in just an optimization, or is it important for correctness (e.g. computing it early in the function)?

reference-implementation/lib/writable-stream.js
```diff
@@ -586,7 +611,7 @@ function WritableStreamDefaultWriterCloseWithErrorPropagation(writer) {
   assert(stream !== undefined);
 
   const state = stream._state;
-  if (state === 'closing' || state === 'closed') {
+  if ((state === 'writable' && stream._closeRequest !== undefined) || state === 'closed') {
```
domenic: Could this be simplified to just checking stream._closeRequest?

index.bs
```diff
@@ -3470,6 +3468,14 @@ nothrow>WritableStreamDefaultControllerGetDesiredSize ( <var>controller</var> )<
   1. Return _controller_.[[strategyHWM]] − _queueSize_.
 </emu-alg>
 
+<h4 id="writable-stream-has-operation-marked-inflight" aoid="WritableStreamHasOperationMarkedInflight"
+nothrow>WritableStreamHasOperationMarkedInflight ( <var>stream</var> )</h4>
+
+<emu-alg>
+  1. If _stream_.[[inflightWriteRequest]] is not *undefined* or _controller_.[[pendingCloseRequest]] is not *undefined*, return *true*.
```
ricea: Should be [[closeRequest]]

reference-implementation/lib/writable-stream.js
tyoshino: Yes. Simplified

index.bs
tyoshino: Fixed!

tyoshino:
Addressed issues, squashed and rebased. The last commit before that is fb744f0a465e6e90abb2249821fd3f5bd0afb724

I removed all the wasAborted logic. All the promise rejection logic now doesn't depend on the passed stream's variables (storedError).

tyoshino:
> > WritableStreamIsReadyForWrites()
>
> This is better than my suggestion of putting Is... first, agreed.

I checked this again and found that this was not the right name for this function. It was returning whether or not the readyPromise is pending or not (not case includes rejected). So, "is ready" was misleading.

After some revision, I've removed this method.

tyoshino:
> However it doesn't really mention how #632 was fixed, or am I missing something?

As I explained in the issue, the problem has already gone and we just need to remove the note from the spec. I clarified this point in the description draft!

tyoshino:
TODO
- I haven't updated the description [draft](https://github.com/whatwg/streams/pull/672#issuecomment-275343563). I'll sync it with the current code
- I'll check if the spec text is correctly updated. It builds but I haven't checked if it's sync with the reference impl.

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,248 @@ function WritableStreamAddWriteRequest(stream) {
...
+      }
+      writer._readyPromise.catch(() => {});
+    }
+
+    WritableStreamRejectPromisesInReactionToError(stream, reason);
 
     return;
   }
 
   stream._state = 'errored';
```
ricea: I don't understand why this two lines are not shared with line 249 above (ie. moved above the loop condition).

reference-implementation/lib/writable-stream.js
```diff
@@ -760,6 +782,7 @@ function WritableStreamDefaultControllerGetChunkSize(controller, chunk) {
   try {
     return strategySize(chunk);
   } catch (chunkSizeE) {
+    // TODO: Should we notify the sink of this error?
```
ricea: I took this comment out because I thought we had consensus not to do anything about this until we have finally().

tyoshino: I felt like this has been removed by you sometime before, but not fully sure and so I kept this while merging and forgot to mention this in the PTAL comments. Thanks for confirming. Removed.

reference-implementation/lib/writable-stream.js
```diff
@@ -932,26 +947,28 @@ function WritableStreamDefaultControllerGetBackpressure(controller) {
...
 
   stream._state = 'errored';
   stream._storedError = e;
 
-  if (stream._pendingAbortRequest === undefined && stream._writer !== undefined) {
-    let readyPromiseIsPending = false;
-    if (oldState === 'writable' &&
-        WritableStreamDefaultControllerGetBackpressure(stream._writableStreamController) === true) {
-      readyPromiseIsPending = true;
+  // TODO(tyoshino): Given that writer.abort() fails immediately when the stream has been errored, shouldn't
```
ricea: I think we should try as much as possible to guarantee that the stream is quiescent when abort() returns. This also affects the timing of the return from pipeTo(), since that waits for WritableStreamAbort().

reference-implementation/lib/writable-stream.js
tyoshino: Trying to factor out common logic resulted in having code that is hard to compare with the others (WritableStreamFinish.* methods) to make sure what's needed in each of them are really included, given that several conditions are entangled.

So, I've flattened all the methods and sought better way. But yes, here it looks okay to dedupe these. Done.

reference-implementation/lib/writable-stream.js
tyoshino: Oh, I see. That was one of the keys of the change that you made closed, etc. delayed.

So, ... but, this rationale also applies to that `WritableStreamAbort()` call after `controller.error()` returns immediately even if the WritableStream has not yet became quiescent (the stream hasn't yet reflected completion of the underlying sink's method to itself).

reference-implementation/lib/writable-stream.js
```diff
@@ -21,23 +21,31 @@ class WritableStream {
...
+    this._inFlightWriteRequest = undefined;
 
     // The promise that was returned from writer.close(). Stored here because it may be fulfilled after the writer
     // has been detached.
-    this._pendingCloseRequest = undefined;
+    this._closeRequest = undefined;
+
+    // Close request is removed from _closeRequest when close() is called on the underlying sink. This prevents it
+    // from being erroneously rejected on error. If a close() call is in-flight, the request is stored here.
+    this._inFlightCloseRequest = undefined;
```
domenic: It seems like it'd be possible for implementations to shrink the object by re-using [[closeRequest]] and then adding an extra bit to the [[state]] + [[backpressure]] field. Do you think it would help the spec to use a boolean instead of a separate slot? My guess is that it's better as a separate slot (to parallel [[inFlightWriteRequest]]) but I wanted to check with you.

reference-implementation/lib/writable-stream.js
```diff
@@ -21,23 +21,31 @@ class WritableStream {
...
 
     // The promise that was returned from writer.abort(). This may also be fulfilled after the writer has detached.
     this._pendingAbortRequest = undefined;
 
+    // The backpressure signal set by the controller.
+    this._backpressure = undefined;
```
domenic: I think this should be initialized to false? You don't seem to making use of it being undefined instead of false (all comparisons are === true) so keeping it always a boolean seems good to me.

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,242 @@ function WritableStreamAddWriteRequest(stream) {
...
   const writer = stream._writer;
   if (writer !== undefined) {
-    defaultWriterClosedPromiseReject(writer, stream._storedError);
+    defaultWriterClosedPromiseReject(writer, reason);
     writer._closedPromise.catch(() => {});
   }
 }
 
-function WritableStreamRejectPromisesInReactionToError(stream) {
+function WritableStreamRejectPromisesInReactionToError(stream, reason) {
```
domenic: As far as I can tell reason === storedError is always true. Why did you change it to be an argument?

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,242 @@ function WritableStreamAddWriteRequest(stream) {
...
-      WritableStreamDefaultControllerGetBackpressure(stream._writableStreamController) === true) {
-    readyPromiseIsPending = true;
-  }
-
-  if (wasAborted === true) {
-    stream._pendingAbortRequest._reject(reason);
-    stream._pendingAbortRequest = undefined;
-  }
-
   if (state === 'errored') {
```
domenic: This code block appears above inside WritableStreamFinishInFlightWrite too. Maybe it should be factored out into a helper function (maybe even as part of WritableStreamRejectPromisesInReactionToError?)

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,242 @@ function WritableStreamAddWriteRequest(stream) {
...
+      writer._readyPromise.catch(() => {});
+    }
+
+    WritableStreamRejectPromisesInReactionToError(stream, reason);
+
+    return;
   }
 
-  WritableStreamRejectPromisesInReactionToError(stream);
+  stream._pendingAbortRequest._reject(reason);
```
domenic: This one is almost the same as the above two.

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,242 @@ function WritableStreamAddWriteRequest(stream) {
...
+  stream._inFlightCloseRequest = undefined;
 
   const state = stream._state;
 
-  let wasAborted = false;
-  if (stream._pendingAbortRequest !== undefined) {
-    wasAborted = true;
-  }
+  if (state === 'errored') {
+    if (stream._pendingAbortRequest !== undefined) {
```
domenic: Hmm another duplicated code block, but this time ending with WritableStreamRejectClosedPromiseIfAny instead of WritableStreamRejectPromisesInReactionToError. It feels like there is some abstraction missing...

reference-implementation/lib/writable-stream.js
tyoshino: Made some dedupe.

reference-implementation/lib/writable-stream.js
tyoshino: I've factored out the code to handle the errored case.

I don't want to call WritableStreamRejectPromisesInReactionToError() in close() handling code just because it's harmless. It's weird to have the logic looks for stuff remaining in `[[writeRequests]]` and `[[closeRequest]]` after finishing close() processing.

Sometimes it helps to have all-in-one helper to take care of various situations. But for this case, I feel like that we shouldn't. Increasing call depth together with dependency from leaf function calls would sometimes lead to smaller but harder-to-read code.

reference-implementation/lib/writable-stream.js
tyoshino: I struggled to have all the code look at the right thing (variables before/after update) when fixing the order of promise rejection/resolution. Saving old variables to `oldBlah` would also help but mess the code too. So, I've just let all the helpers take the new value directly than depending on that the caller makes sure to update the corresponding variables (e.g. `stream._storedError = reason`) before calling the helpers.

One benefit of letting this helper depend on `[[storedError]]` helps readers understand the invariants, i.e. the pending requests get rejected with the same exception as one in `[[storedError]]`. I understand that. So, there's trade-off.

tyoshino: Given that the current code updates `[[state]]` and `[[storedError]]` first, I can live with reverting this to depend on `[[storedError]]`, but still prefer having this take the new value directly.

reference-implementation/lib/writable-stream.js
tyoshino: I agree that it looks redundant to have the two separate slots. As you guessed, I chose this way for parallelism with `[[inFlightWriteRequest]]`.

We could also consolidate the inFlight indication for close and write into one variable.

tyoshino: > We could also consolidate the inFlight indication for close and write into one variable.

I guess I should try this after landing the patch.

reference-implementation/lib/writable-stream.js
tyoshino: How about landing the patch with this TODO kept?

tyoshino: I'd like to hear @domenic's opinion on this too.

ricea: I think it would be good if the behaviour was the some regardless of the order of controller.error() and abort(), but it's not sufficiently important to be worth adding additional complexity.

In general I don't feel strongly about what happens after controller.error() because I assume it means the underlying transport has failed and any ongoing writes are just going to be lost anyway.

tyoshino: Thanks, Adam. I agree that it's subtle point.

reference-implementation/lib/writable-stream.js
ricea: > I guess I should try this after landing the patch.

+1 for getting this landed before changing anything else.

I think in practice all browser implementations are going to pack the internal slots somewhat. Chrome does, and if I understand correctly Firefox goes further and defines the object layout using native code. I think that the key things for the standard are finding the representation of the state that is easiest to understand and reason about, and explicitly mentioning things like "only one of [[closeRequest]] and [[inFlightWriteRequest]] is used at a time" to help other implementations optimise.

tyoshino:
Fixed some bugs and checked
- text is sync with the reference impl
- the anchors' ID is updated appropriately for renamed methods
- updated the [description](https://github.com/whatwg/streams/pull/672#issuecomment-275343563)

The whole patch is ready for review now.

reference-implementation/lib/writable-stream.js
domenic: I think it is significantly clearer if we use [[storedError]]; I'd really prefer having a single canonical location where the error is stored and making it clear that the stored error is always the error that these rejections happen with.

reference-implementation/lib/writable-stream.js
domenic: I'm not sure I completely understand this issue; a quick code snippet or test case illustrating the two alternative sunder discussion would be good.

I'm going to assume the question is, given `const p = var writer.abort(); controller.error();`, should `p` reject immediately, or should it follow the underlying sink's abort() return value?

If my understanding is correct, then my opinion is: if an underlying sink abort is underway already, it seems reasonable to let that run to completion instead of rejecting the writer.abort() promise.

reference-implementation/lib/writable-stream.js
```diff
@@ -182,215 +190,226 @@ function WritableStreamAddWriteRequest(stream) {
...
   stream._pendingAbortRequest._resolve();
   stream._pendingAbortRequest = undefined;
 
+  const error = new TypeError('Closed after abort request');
+
   stream._state = 'errored';
-  stream._storedError = new TypeError('Abort requested but closed successfully');
+  stream._storedError = error;
+
+  // TODO(tyoshino): No need to reject readyPromise?
```
domenic: What's the plan for this TODO?

reference-implementation/lib/writable-stream.js
```diff
@@ -407,23 +426,33 @@ class WritableStreamDefaultWriter {
 
     const state = stream._state;
 
-    if (state === 'writable' || state === 'closing') {
+    if (state === 'writable') {
+      if (stream._pendingAbortRequest !== undefined) {
+        // TODO(tyoshino): Test this
```
domenic: Status on this TODO?

reference-implementation/lib/writable-stream.js
```diff
@@ -407,23 +426,33 @@ class WritableStreamDefaultWriter {
...
     } else if (state === 'closed') {
+      defaultWriterReadyPromiseInitializeAsResolved(this);
       defaultWriterClosedPromiseInitializeAsResolved(this);
     } else {
       assert(state === 'errored', 'state must be errored');
 
-      defaultWriterClosedPromiseInitializeAsRejected(this, stream._storedError);
+      const storedError = stream._storedError;
+      // TODO(tyoshino): Get consensus on this behavior i.e. readyPromise rejection reason may differ between
+      // release.
```
domenic: This seems good to me as-is, so I would delete this TODO

reference-implementation/lib/writable-stream.js
```diff
@@ -769,39 +770,40 @@ function WritableStreamDefaultControllerGetDesiredSize(controller) {
...
 
   WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
 }
 
+function WritableStreamDefaultControllerStart(controller) {
+  const startResult = InvokeOrNoop(controller._underlyingSink, 'start', [controller]);
+  Promise.resolve(startResult).then(
+    () => {
+      controller._started = true;
+      // TODO: Test this.
```
domenic: Another test TODO

index.bs
```diff
@@ -2680,25 +2693,19 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
   behavior will be the same as a {{CountQueuingStrategy}} with a <a>high water mark</a> of 1.
 </div>
```
domenic: It would be good to add the note about how the inFlightX slots are mutually exclusive with the X slots here (as a `<p class="note">`).

index.bs
```diff
@@ -2680,25 +2693,19 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
...
-  method itself has no asynchronous behavior. A well-designed <a>underlying sink</a> object should be able to deal with
-  this.
-</div>
-
 <emu-alg>
   1. Set *this*.[[state]] to `"writable"`.
-  1. Set *this*.[[storedError]], *this*.[[writer]], *this*.[[writableStreamController]], *this*.[[pendingAbortRequest]],
-     *this*.[[pendingCloseRequest]], and *this*.[[pendingWriteRequest]] to *undefined*.
+  1. Set *this*.[[storedError]], *this*.[[writer]], *this*.[[writableStreamController]],
+     *this*.[[inFlightWriteRequest]], *this*.[[closeRequest]], *this*.[[inFlightCloseRequest]],
```
domenic: missing "and"

index.bs
```diff
@@ -2784,34 +2791,33 @@ writable stream is <a>locked to a writer</a>.
...
-     WritableStreamDefaultControllerGetBackpressure(_stream_.[[writableStreamController]]) is *true*, set
-     _readyPromiseIsPending_ to *true*.
-  1. If _controller_.[[writing]] is *false* and _controller_.[[inClose]] is *false*,
-    1. If _stream_.[[writer]] is not *undefined*, perform !
-       WritableStreamDefaultWriterEnsureReadyPromiseRejectedWith(_stream_.[[writer]], _error_, _readyPromiseIsPending_).
+  1. Let _writer_ be _stream_.[[writer]].
+  1. If _writer_ is not *undefined*,
+    1. If ! WritableStreamCloseQueuedOrInFlight(_stream_) is *false* and _stream_.[[backpressure]] is *true*,
+       <a>reject</a> _writer_.[[readyPromise]] with _error_.
+    1. Otherwise, set _writer_.[[readypromise]] to <a>a promise rejected with</a> _error_.
```
domenic: uppercase "p" in readyPromise

index.bs
```diff
@@ -2837,166 +2843,199 @@ visible through the {{WritableStream}}'s public API.
...
+  1. Assert: _state_ is `"writable"`.
+  1. If _stream_.[[pendingAbortRequest]] is *undefined*,
+    1. Set _stream_.[[state]] to `"closed"`.
     1. Let _writer_ be _stream_.[[writer]].
     1. If _writer_ is not *undefined*, <a>resolve</a> _writer_.[[closedPromise]] with *undefined*.
-    1. Set _stream_.[[state]] to `"closed"`.
     1. Return.
   1. <a>Resolve</a> _stream_.[[pendingAbortRequest]] with *undefined*.
   1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. Let _error_ to a new *TypeError* indicating that the stream was closed successfully after abort request.
```
domenic: Hmm I'm not sure on the error phrasing here. Could you describe the situation in a bit more detail?

index.bs
```diff
@@ -3503,42 +3520,45 @@ nothrow>WritableStreamDefaultControllerGetDesiredSize ( <var>controller</var> )<
...
     1. Return.
-  1. Perform ! WritableStreamDefaultControllerUpdateBackpressureIfNeeded(_controller_, _oldBackpressure_).
+  1. Let _stream_ be _controller_.[[controlledWritableStream]].
+  1. If ! WritableStreamCloseQueuedOrInFlight(_stream_) is *false*,
+    1. Let _backpressure_ be ! WritableStreamDefaultControllerGetBackpressure(_controller_).
+    1. Perform ! WritableStreamUpdateBackpressure(_stream_, _backpressure_).
   1. Perform ! WritableStreamDefaultControllerAdvanceQueueIfNeeded(_controller_).
 </emu-alg>
 
+<h4 id="writable-stream-default-controller-start" aoid="WritableStreamDefaultControllerStart" nothrow>
```
domenic: This needs to be "throws" since step 1 could throw.

reference-implementation/lib/writable-stream.js
tyoshino: > I'm going to assume the question is

Correct, but the situation I'm concerned is when `writer.abort()` is pending due to any in-flight close()/write(). Once they call `controller.error()`, the stream's state transitions to `"errored"`. But until the close()/write() fulfills/rejects the promise it returned and is processed by the stream, we don't reject the `writer.abort()`.

```
const w = writer.write();
const p = writer.abort();  // the abort becomes pending
controller.error();  // the stream becomes errored but abort() is kept pending until sink.write()'s handler is invoked
```

I'm not objecting to this behavior. I just wondered if it's ok that this is inconsistent with `controller.error()` first and then `writer.abort()`, i.e. in reverse order. WritableStreamAbort() returns immediately when the stream is already in the `"errored"` state even before the ongoing close()/write() fulfills/rejects the promise it returned and is processed by the stream.

reference-implementation/lib/writable-stream.js
tyoshino: OK. I'll remove it.

index.bs
tyoshino: I rephrased this because I thought it's better to clarify that the requested abort() became pending and the abort() was aborted eventually.

I understand that this is still not much clear. I'll rephrase it again.

reference-implementation/lib/writable-stream.js
tyoshino: Ah, I'll fine an issue instead of the TODO for this point.

index.bs
tyoshino: I was struggling with this since
- it should represent the situation of the stream. not the abort operation
- it should tell what kind of bad thing has happened and how it was bad for the stream

Since the stream has finished closing successfully, the only bad things we should talk about are:
- the fact that there was an attempt to abort the stream (which didn't take place eventually)
- that the pending abort has been resolved without running sink.abort().

tyoshino: Just telling that the stream has been aborted like some of the other places would be misleading since the abort didn't happen due to successful completion of the in-flight close. So, I thought I should tell some more details.

tyoshino: I've changed all the "indicating that ..." texts to distinguish "requested to abort" and "completed aborting". With this, I could give a clearer message here.

index.bs

index.bs

index.bs
tyoshino: Fixed.

index.bs
tyoshino: Added. Incorporated the discussion about possible optimization, too.

index.bs
```diff
@@ -2693,10 +2693,16 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
   behavior will be the same as a {{CountQueuingStrategy}} with a <a>high water mark</a> of 1.
 </div>
 
+<div class="note">
+  The \[[inFlightCloseRequest]] slot and \[[closeRequest]] slot are mutually exclusive. No element will be removed from
+  \[[writeRequests]] while \[[inFlightWriteRequest]] is not <code>undefined</code>. Implementations may optimize the way
```
ricea: English nit:

"Implementations may optimize storage for these slots" would sound more natural.

tyoshino: Thanks! Will fix it.

index.bs
```diff
@@ -2681,24 +2694,24 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
...
 <div class="note">
-  Due to the way writable streams asynchronously close, it is possible for both <code>close</code> and
-  <code>abort</code> to be called, in cases where the <a>producer</a> aborts the stream while it is in the
-  <code>"closing"</code> state. Notably, since a stream always spends at least one turn in the <code>"closing"</code>
-  state, code like <code>ws.close(); ws.abort(...);</code> will cause both to be called, even if the <code>close</code>
-  method itself has no asynchronous behavior. A well-designed <a>underlying sink</a> object should be able to deal with
-  this.
+  The \[[inFlightCloseRequest]] slot and \[[closeRequest]] slot are mutually exclusive. No element will be removed from
+  \[[writeRequests]] while \[[inFlightWriteRequest]] is not <code>undefined</code>. Implementations may optimize the way
+  to hold these slots based on these assumptions.
```
ricea: These are invariants, not assumptions. Sorry I failed to spot this earlier.

tyoshino: Done! Thanks

reference-implementation/lib/writable-stream.js
tyoshino: OK. It seems I was stuck by the difficulties experienced during the refactoring which is no longer valid now. I also started feeling that the "InReactionToError" suffix is helping me enough understand that it reads storedError and sync readyPromise with it.

I've reverted them to use storedError.

reference-implementation/lib/writable-stream.js
tyoshino: Filed https://github.com/whatwg/streams/issues/692

reference-implementation/lib/writable-stream.js
tyoshino: Filed https://github.com/whatwg/streams/issues/693 for this.

reference-implementation/lib/writable-stream.js
tyoshino: Sorry. Never mind. It seems I was confused.

Abort()-ing a stream with in-flight close resets the readyPromise to a rejected promise, and it's final for the readyPromise as well as other places where we reject (or reset-to-rejected) the readyPromise. So, here we don't have to do anything for the readyPromise.

reference-implementation/lib/writable-stream.js
tyoshino: Added a test

reference-implementation/lib/writable-stream.js
tyoshino: I no longer remember what I wanted to do. It seems this logic is well tested by start.js basically. I've removed this TODO.

domenic:
OK, the spec and reference implementation changes LGTM!

Are we sure that the only normative change here worth testing is the additional test in https://github.com/w3c/web-platform-tests/pull/5101 ? It seems like the three first bullets in https://github.com/whatwg/streams/pull/672#issuecomment-275343563 should each get a test.

Also, if this is really solving #632, then shouldn't the test changes in https://github.com/whatwg/streams/commit/33773df16229cdc99147b4d5ad29a0493fe6c41a#diff-4c7e8a1ce50d6c68c3d24df6461fcfd6 also be applied? (I am not sure though; it has been a while since that commit. But please take a look and check if possible.)

tyoshino:
Thanks for review, domenic and ricea. Thanks for the tweaking.

I've squashed and rebased the commits. The last commit before that is f462f023d377fe52ecaa2ae787ac1aca2ceede38.

> It seems like the three first bullets in #672 (comment) should each get a test.

Yes. Sorry but I'm travelling tomorrow and doesn't have enough time to address this. ricea kindly said he'll follow up this.

> Also, if this is really solving #632, then shouldn't the test changes in

We don't follow the expectation in the `Closing but then immediately aborting a WritableStream results in a successful close and a fulfilled abort()` test. The situation described in the note has been still fixed, I think. `writer.close()` may get aborted by `writer.abort()` if `writer.close()` is still queued, but in that case, `sink.close()` is not yet called and only `sink.abort()` gets called. So, the sink doesn't get both.

The `Closing a WritableStream and aborting it while it is in the process of closing results in a successful close and fulfilled abort()` test is worth being merged, I think.

The improvement on `underlying abort() should not be called until underlying write() completes` test and following ones is just worth being merged.

I or @ricea should do these, yes!

tyoshino:
Filed #694 for https://github.com/whatwg/streams/pull/672#issuecomment-285693724

tyoshino:
It seems I used a wrong commit 5dfffa5 for the wpt submodule in a25f410cbccfbaf56479ca3224fe57bab8b2b6fa. Fixing by #695.

PR #676, opened 2017-01-25
factor out pipeTo into abstract op ReadableStreamPipeTo

### LINKED ISSUE #658: Is there an abstract operation for pipeTo?
Do you have a plan to have an abstract operation? I want to use it from the fetch spec.

isonmad:
Closes #658, if it weren't already closed at least.


domenic:
I really don't think we should do this until we have a consumer to use it.

PR #680, opened 2017-02-10
Execute strategySize first in write()

### LINKED ISSUE #675: strategy.size() can reenter WritableStream logic before its caller finishes all the work
Looks ReadableStream has the same issue.

The other user provided functions e.g. ones on underlyingSource/Sink which are invoked after their caller finishes all the work that touches the state variables of the streams. But in the callers of strategy.size(), there're lines that touch the state variables (both reading and writing) after strategy.size() is invoked. strategy.size() can implemented to call writer function or controller functions, and therefore reentry may happen. Should we have some guard against such a reentry?

ricea:
A strategy size() function can re-entrantly call controller, writer or writable
methods, leaving the stream in an unexpected state. To ensure predictable
behaviour, call size() before observing or modifying any state.

Previously, if size() threw an exception, writer.write() would reject with that
exception. After this change, it rejects with a TypeError instead.

Fixes #675.

### DIFF
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -505,14 +505,8 @@ class WritableStreamDefaultWriter {
       return Promise.reject(defaultWriterBrandCheckException('write'));
     }
 
-    const stream = this._ownerWritableStream;
-
-    if (stream === undefined) {
+    if (this._ownerWritableStream === undefined) {
       return Promise.reject(defaultWriterLockException('write to'));
-    }
-
-    if (stream._state === 'closing') {
-      return Promise.reject(new TypeError('Cannot write to an already-closed stream'));
     }
 
     return WritableStreamDefaultWriterWrite(this, chunk);
@@ -656,8 +650,16 @@ function WritableStreamDefaultWriterWrite(writer, chunk) {
 
   assert(stream !== undefined);
 
+  const controller = stream._writableStreamController;
+
+  const chunkSize = WritableStreamDefaultControllerGetChunkSize(controller, chunk);
+
+  if (stream !== writer._ownerWritableStream) {
+    return Promise.reject(defaultWriterLockException('write to'));
+  }
+
   const state = stream._state;
-  if (state === 'closed' || state === 'errored') {
+  if (state !== 'writable') {
     return Promise.reject(new TypeError(
       `The stream (in ${state} state) is not in the writable state and cannot be written to`));
   }
@@ -665,11 +667,9 @@ function WritableStreamDefaultWriterWrite(writer, chunk) {
     return Promise.reject(new TypeError('Aborted'));
   }
 
-  assert(state === 'writable');
-
   const promise = WritableStreamAddWriteRequest(stream);
 
-  WritableStreamDefaultControllerWrite(stream._writableStreamController, chunk);
+  WritableStreamDefaultControllerWrite(controller, chunk, chunkSize);
 
   return promise;
 }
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/writable-stream.js
```diff
@@ -654,7 +644,17 @@ function WritableStreamDefaultWriterRelease(writer) {
...
 
-  assert(stream !== undefined);
+  if (stream === undefined) {
+    return Promise.reject(defaultWriterLockException('write to'));
+  }
+
+  const controller = stream._writableStreamController;
+
+  const chunkSize = WritableStreamDefaultControllerGetChunkSize(controller, chunk);
+
```
ricea: Drive-by comment: maybe we should check that `stream === writer._ownerWritableStream` before proceeding?

strategySize might have
1. Detached the writer
2. Created a new writer and done a bunch of writes with it

It's probably better just to eliminate the possibility rather than try to work out whether it would be a problem or not.

Needs test(s).

reference-implementation/lib/writable-stream.js
```diff
@@ -505,16 +505,6 @@ class WritableStreamDefaultWriter {
       return Promise.reject(defaultWriterBrandCheckException('write'));
     }
 
-    const stream = this._ownerWritableStream;
```
domenic: I am sad about moving these checks as I like the separation of invariant-checking in the public API only. It seems like at least the first check, of stream being undefined, can stay here. What about the second one?

reference-implementation/lib/writable-stream.js
```diff
@@ -747,6 +747,21 @@ function WritableStreamDefaultControllerClose(controller) {
   WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
 }
 
+function WritableStreamDefaultControllerGetChunkSize(controller, chunk) {
```
domenic: This appears to be called from only one place, so inlining it fits better with our style.

reference-implementation/lib/writable-stream.js
ricea: I added tests to my wpt change and added the extra check here.

reference-implementation/lib/writable-stream.js
ricea: Originally I didn't inline it because it was a controller operation rather than a writer operation.

Afterwards I realised that once we have a ByteController this will become a polymorphic method call. Maybe the "you ain't gonna need it" principle applies, but I'm pretty sure that we will add a byte controller and that this separation will make sense at that time.

reference-implementation/lib/writable-stream.js
ricea: I moved the stream check back into write().

We need to check `state` after strategy.size() returns, since it may have changed. We could additionally check it before, if we wanted to guarantee that size() is only called in the writable state, but there are no tests for that and I didn't consider it important.

I have started thinking of size() as part of the public API. This makes the line between public methods and abstract operations a bit fuzzier.

reference-implementation/lib/writable-stream.js
```diff
@@ -656,6 +650,18 @@ function WritableStreamDefaultWriterWrite(writer, chunk) {
...
 
+  const controller = stream._writableStreamController;
+
+  const chunkSize = WritableStreamDefaultControllerGetChunkSize(controller, chunk);
+
+  if (stream !== writer._ownerWritableStream) {
+    return Promise.reject(defaultWriterLockException('write to'));
+  }
+
+  if (stream._state === 'closing') {
```
domenic: I guess this could be consolidated with line 666.

domenic:
Spec changes LGTM too. Will let you coordinate landing WPTs and then this.

reference-implementation/lib/writable-stream.js
```diff
@@ -656,20 +650,26 @@ function WritableStreamDefaultWriterWrite(writer, chunk) {
 
   assert(stream !== undefined);
 
+  const controller = stream._writableStreamController;
+
+  const chunkSize = WritableStreamDefaultControllerGetChunkSize(controller, chunk);
```
tyoshino: It's really unfortunate that we need to do this in the Writer's abstract op. Ideally, the detail of the operation should be hidden. But ok for now.

PR #688, opened 2017-03-07
Do not call sink abort() while sink start() is running

### LINKED ISSUE #683: sink.abort() can be called during sink.start()
This is inconsistent with our treatment of sink.write() and sink.close() as atomic, and is likely to be a source of logic errors in user code.

ricea:
Previously, calling abort() on a WritableStream or Writer would cause the sink
abort() to be called immediately if the sink start() was still executing. This
was inconsistent with the way write() and close() work and could be a cause of
bugs.

Postpone calling sink abort() until start() is complete. If start() errors, do
not call abort() at all.

Fixes #683.

### DIFF
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -799,12 +799,11 @@ function WritableStreamDefaultControllerStart(controller) {
 
       if (stream._state === 'errored') {
         WritableStreamRejectAbortRequestIfPending(stream);
-        return;
+      } else {
+        WritableStreamHandleAbortRequestIfPending(stream);
       }
 
-      WritableStreamHandleAbortRequestIfPending(stream);
-
-      // This is a no-op if the stream was errored by abort().
+      // This is a no-op if the stream was errored above.
       WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
     },
     r => {
--- a/reference-implementation/lib/writable-stream.js
+++ b/reference-implementation/lib/writable-stream.js
@@ -190,6 +190,18 @@ function WritableStreamAddWriteRequest(stream) {
   return promise;
 }
 
+function WritableStreamFinishAbortAndPendingAbortRequest(stream) {
+  WritableStreamFinishAbort(stream);
+
+  const abortRequest = stream._pendingAbortRequest;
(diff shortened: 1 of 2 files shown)


ricea:
I have not written the spec changes yet. Please review the reference implementation changes.

reference-implementation/lib/writable-stream.js
```diff
@@ -713,10 +713,34 @@ class WritableStreamDefaultController {
     Promise.resolve(startResult).then(
       () => {
         controller._started = true;
-        WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
+        const abortRequest = stream._pendingAbortRequest;
```
domenic: Is there any way to deduplicate this with the code in WritableStreamFinishPendingWrite ? They seem to serve rather similar purposes...

ricea:
I think I will wait for @tyoshino's #672 to land since that one already has spec text done. It will also change the way that deduplication needs to be done.

reference-implementation/lib/writable-stream.js
tyoshino: Yeah. They should be able to share the code.

ricea:
The code changes are ready for review. I still have to write the spec changes.

reference-implementation/lib/writable-stream.js
```diff
@@ -300,11 +300,7 @@ function WritableStreamFinishInFlightClose(stream) {
 }
 
 function WritableStreamFinishInFlightCloseInErroredState(stream) {
```
domenic: I would be OK with inlining WritableStreamFinishInFlightWriteInErroredState and WritableStreamFinishInFlightCloseInErroredState now that they are two-liners and each called twice. Up to you though, I guess they are reasonably semantic operations that can stand on their own.

ricea: I think on balance that the explanatory value they provide wins. They serve to highlight that we do the same thing on both code paths, which otherwise would be less obvious.

ricea:
This is now ready to go. Please review the standard text changes.

ricea:
Feel free to squash and rebase this if you don't want to wait until after my 3-day weekend to land it.

tyoshino:
the merged commit lgtm!

PR #689, opened 2017-03-08
Check precisely the way that transformer.transform() is called

ricea:
There is an incorrect possible optimisation to the call to
transformer.transform() in the TransformStreamTransform operation:

```javascript
let transformPromise;
try {
  transformPromise = Promise.resolve(transformer.transform(chunk, controller));
} catch (e) {
  if (transformer.transform === undefined) {
    transformPromise = TransformStreamDefaultTransform(chunk, controller);
  } else {
    transformPromise = Promise.reject(e);
  }
}
```

This optimisation is attractive because it moves the `undefined` check out of the
expected fast path. The flaw is the change in order of checking whether
transformer.transform is undefined is visible to user code.

Add a test to verify that the undefined check is performed before the method call.

Also test that transformer.transform is only looked up once, to protect against
flawed variations on this optimisation.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/general.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/general.js
@@ -6,6 +6,9 @@ if (self.importScripts) {
   self.importScripts('../resources/rs-utils.js');
 }
 
+const thrownError = new Error('bad things are happening!');
+thrownError.name = 'error1';
+
 test(() => {
   new TransformStream({ transform() { } });
 }, 'TransformStream can be constructed with a transform function');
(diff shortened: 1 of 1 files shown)


reference-implementation/to-upstream-wpts/transform-streams/general.js
```diff
@@ -300,4 +300,36 @@ promise_test(() => {
...
+  const transformer = {
+    transform() {
+      transformer.transform = undefined;
+      throw new TypeError();
+    }
+  };
+  const ts = new TransformStream(transformer);
+  return Promise.all([
+    promise_rejects(t, new TypeError(), ts.writable.getWriter().write('a'), 'write() should throw'),
+    promise_rejects(t, new TypeError(), ts.readable.getReader().read(), 'read() should throw')
```
domenic: It's a little unclear to me whether these throw because it's undefined or throw because you threw a TypeError. If you're testing that the code gets exercised, then maybe setting a boolean, or using the `error1` pattern, would be clearer.

reference-implementation/to-upstream-wpts/transform-streams/general.js
```diff
@@ -300,4 +300,36 @@ promise_test(() => {
...
+promise_test(() => {
+  let transformGetterCalls = 0;
+  const ts = new TransformStream({
+    get transform() {
+      ++transformGetterCalls;
+      return (chunk, controller) => controller.enqueue(chunk);
+    }
+  });
+  const writer = ts.writable.getWriter();
+  writer.write('a');
```
domenic: Maybe write more than one chunk to ensure it doesn't look it up once for each chunk?

ricea: It actually is looked up once per chunk.

But I'm starting to think that it shouldn't be. If we only looked it up once, in the constructor, then we could bypass Javascript completely in the case that it was undefined. However, this would be different from the way that sinks and sources work, and maybe too surprising?

reference-implementation/to-upstream-wpts/transform-streams/general.js
ricea: Someone might try to fix this broken optimisation by checking if the exception was TypeError. So the fact that it behaves exactly like transform() was undefined is WAI.

domenic: I'm sorry, I don't understand. What is this test testing then? Is it testing that transform() is called, or what?

reference-implementation/to-upstream-wpts/transform-streams/general.js
domenic: Oh, right. Yeah, that would be pretty different...

Ideally we'd add another test that writes two chunks and tests that it's looked up exactly twice, then. Otherwise this test is kind of confusing to me.

reference-implementation/to-upstream-wpts/transform-streams/general.js
ricea: Ah, I have a better idea. With the incorrect implementation, both transform() and the fallback will be called. So I can test for that.

ricea: Oh, that doesn't work because you can't read the queued chunk once the readable is errored. Maybe this isn't observable after all?

ricea: I confused myself. If we fallback there is no exception thrown at all. So it is testable. I switched to using thrownError instead of TypeError as I now have another test for implementations which try to be clever about checking the exception.

reference-implementation/to-upstream-wpts/transform-streams/general.js
ricea: Done.

ricea:
I filed #691 to discuss whether these tests should exist at all.

ricea:
Obsoleted by https://github.com/whatwg/streams/pull/860.

PR #698, opened 2017-03-16
Do not assume a BYOB reader when fulfilling a BYOB request

### LINKED ISSUE #686: Access to reader.readIntoRequests without checking reader type
In ReadableByteStreamControllerRespondInClosedState [1], ReadableStreamGetNumReadIntoRequests is called and checks reader.readIntoRequests, but it is not checked that the reader is a BYOBReader.

This could be checked based on firstDescriptor.readerType, but the purpose of the loop (step 4) may be the processing of all pendingPullIntos objects, in which case I'm not sure the test that involves the number of readIntoRequests is the right one (I may not get all the context here).

[1] https://streams.spec.whatwg.org/#readable-byte-stream-controller-respond-in-closed-state

domenic:
In particular, when autoAllocateChunkSize is in play, there may be a default reader instead. Fixes #686.

/cc @rombel. Let us know if you'd prefer a full name or similar in the acknowledgments!

@tyoshino I'm curious if this is the best solution. At first I tried to move the code into ReadableStreamClose, since ReadableStreamError has a branch on reader type and I thought ReadableStreamClose should be symmetric. But that caused failing tests, as I believe we want to keep the byobRequest around even if the stream gets closed. So this version seemed OK-ish.

This also needs tests before it can be merged. Let me know if there are any particular tests besides the one in https://github.com/whatwg/streams/issues/686#issuecomment-285368195 which might be good.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2412,9 +2412,10 @@ nothrow>ReadableByteStreamControllerRespondInClosedState ( <var>controller</var>
   1. Set _firstDescriptor_.[[buffer]] to ! TransferArrayBuffer(_firstDescriptor_.[[buffer]]).
   1. Assert: _firstDescriptor_.[[bytesFilled]] is *0*.
   1. Let _stream_ be _controller_.[[controlledReadableStream]].
-  1. Repeat the following steps while ! ReadableStreamGetNumReadIntoRequests(_stream_) > *0*,
-    1. Let _pullIntoDescriptor_ be ! ReadableByteStreamControllerShiftPendingPullInto(_controller_).
-    1. Perform ! ReadableByteStreamControllerCommitPullIntoDescriptor(_stream_, _pullIntoDescriptor_).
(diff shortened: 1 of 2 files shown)


rombel:
Ok for a full name then, thanks :) (Romain Bellessort)

domenic:
Tests added at https://github.com/w3c/web-platform-tests/pull/5398; acks tweaked; rebased on master. @tyoshino PTAL (although #721 is higher priority).

tyoshino:
Sorry for delay. The pull lgtm. It seems we need more work to handle the automatic buffer allocation cases correctly everywhere.

tyoshino:
> I'm curious if this is the best solution. At first I tried to move the code into ReadableStreamClose, since ReadableStreamError has a branch on reader type and I thought ReadableStreamClose should be symmetric. But that caused failing tests, as I believe we want to keep the byobRequest around even if the stream gets closed. So this version seemed OK-ish.

Right, though we might want to revisit necessity of this feature given the complexity.

PR #704, opened 2017-03-24
Use ResetQueue in WritableStreamDefaultControllerProcessWrite

ricea:
WritableStreamDefaultControllerProcessWrite contained the line "If wasErrored is
false, set controller.[[queue]] to an empty List."

Generally it is better to use ResetQueue() to keep the [[queue]] and
[[queueTotalSize]] slots in sink. Do that instead.


ricea:
I'm only 90% sure this is right, so please review carefully.

tyoshino:
lgtm

PR #705, opened 2017-03-24
Refactor writable stream erroring to be centralized

domenic:
@ricea's work in #704 got me thinking: why don't we ResetQueue() *every* time we error a writable stream?

In fact it seems like we have many different code paths for erroring a writable stream, some of which behave slightly differently. You can see me evolving them toward convergence in the sequential commits here, eventually ending up with a bunch of duplicated code which all gets factored into the single WritableStreamError abstract op.

There were two reasons that the erroring code was so different in different places:
- Often we omitted steps that were not necessary in a given scenario, which hid the underlying same-ness of the erroring operation.
- We decided that abort()ing during an in-flight close should error the stream. This was not a good idea: it was unlike every other erroring we did. And we recognized that it was weird, because we made abort() return a fulfilled promise. Instead we should treat this as a success and close the stream, both for the return value of abort() and for the stream's state.

This PR has a few minor issues which I will note in inline comments. It only updates the reference implementation for now, as well as the tests; it does not update the spec yet.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3652,7 +3652,6 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
   1. Let _stream_ be _controller_.[[controlledWritableStream]].
   1. Assert: _stream_.[[state]] is `"writable"`.
   1. Perform ! WritableStreamError(_stream_, _error_).
-  1. Perform ! WritableStreamRejectAbortRequestIfPending(_stream_).
 </emu-alg>
 
 <h2 id="ts">Transform Streams</h2>
--- a/index.bs
+++ b/index.bs
@@ -3515,8 +3515,6 @@ used polymorphically.
       1. Perform ! WritableStreamRejectAbortRequestIfPending(_stream_).
 </emu-alg>
 
-
-
 <h3 id="ws-default-controller-abstract-ops">Writable Stream Default Controller Abstract Operations</h3>
 
 <h4 id="is-writable-stream-default-controller" aoid="IsWritableStreamDefaultController"
--- a/index.bs
+++ b/index.bs
@@ -3492,7 +3492,6 @@ used polymorphically.
 <h5 id="ws-default-controller-private-abort" oldids="writable-stream-default-controller-abort">\[[AbortSteps]]()</h5>
 
 <emu-alg>
-  1. Perform ! ResetQueue(*this*).
   1. Let _sinkAbortPromise_ be ! PromiseInvokeOrNoop(*this*.[[underlyingSink]], `"abort"`, « _reason_ »).
   1. Return the result of <a>transforming</a> _sinkAbortPromise_ with a fulfillment handler that returns
      *undefined*.
--- a/index.bs
+++ b/index.bs
@@ -2709,7 +2709,7 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
      potential types in the future, without backward-compatibility concerns.</p>
   1. Set *this*.[[writableStreamController]] to ? Construct(`<a idl>WritableStreamDefaultController</a>`, « *this*,
      _underlyingSink_, _size_, _highWaterMark_ »).
-  1. Perform ? WritableStreamDefaultControllerStart(*this*.[[writableStreamController]]).
+  1. Perform ? *this*.[[writableStreamController]].[[StartSteps]]().
 </emu-alg>
 
 <h4 id="ws-prototype">Properties of the {{WritableStream}} Prototype</h4>
@@ -2800,18 +2800,30 @@ writable stream is <a>locked to a writer</a>.
   1. If _stream_.[[pendingAbortRequest]] is not *undefined*, return <a>a promise rejected with</a> _error_.
   1. Let _controller_ be _stream_.[[writableStreamController]].
   1. Assert: _controller_ is not *undefined*.
-  1. Let _writer_ be _stream_.[[writer]].
-  1. If _writer_ is not *undefined*,
-    1. If ! WritableStreamCloseQueuedOrInFlight(_stream_) is *false* and _stream_.[[backpressure]] is *true*,
-       <a>reject</a> _writer_.[[readyPromise]] with _error_.
-    1. Otherwise, set _writer_.[[readyPromise]] to <a>a promise rejected with</a> _error_.
-    1. Set _writer_.[[readyPromise]].[[PromiseIsHandled]] to *true*.
   1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false* and _controller_.[[started]] is *true*,
     1. Perform ! WritableStreamFinishAbort(_stream_).
-    1. Return ! WritableStreamDefaultControllerAbort(_controller_, _reason_).
+    1. Return ! _controller_.[[AbortSteps]](_reason_).
+  1. Let _writer_ be _stream_.[[writer]].
+  1. If _writer_ is not *undefined*,
+    1. Perform ! WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _error_).
   1. Let _promise_ be <a>a new promise</a>.
   1. Set _stream_.[[pendingAbortRequest]] to Record {[[promise]]: _promise_, [[reason]]: _reason_}.
   1. Return _promise_.
+</emu-alg>
+
+<h4 id="writable-stream-error" aoid="WritableStreamError" nothrow>WritableStreamError ( <var>stream</var>,
+<var>error</var> )</h4>
+
+<emu-alg>
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Set _stream_.[[storedError]] to _error_.
+  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
+  1. If _stream_.[[pendingAbortRequest]] is *undefined*,
+    1. Let _writer_ be _stream_.[[writer]].
+    1. If _writer_ is not *undefined*,
+      1. Perform ! WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _error_).
+  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*, perform !
+     WritableStreamRejectPromisesInReactionToError(_stream_).
 </emu-alg>
 
 <h4 id="writable-stream-finish-abort" aoid="WritableStreamFinishAbort" nothrow>WritableStreamFinishAbort (
@@ -2819,8 +2831,6 @@ writable stream is <a>locked to a writer</a>.
 
 <emu-alg>
   1. Let _error_ be a new *TypeError* indicating that the stream has been aborted.
-  1. Set _stream_.[[state]] to `"errored"`.
-  1. Set _stream_.[[storedError]] to _error_.
   1. Perform ! WritableStreamRejectPromisesInReactionToError(_stream_).
 </emu-alg>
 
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/writable-stream.js
```diff
@@ -163,13 +163,37 @@ function WritableStreamAbort(stream, reason) {
...
-function WritableStreamFinishAbort(stream) {
-  const error = new TypeError('Aborted');
-
+function WritableStreamError(stream, error) {
   stream._state = 'errored';
   stream._storedError = error;
 
-  WritableStreamRejectPromisesInReactionToError(stream);
+  // TODO: layering violation?
+  ResetQueue(stream._writableStreamController);
```
domenic: @tyoshino the stream is not supposed to reach into its controller like this, right? What do you suggest for fixing this?

reference-implementation/lib/writable-stream.js
```diff
@@ -332,15 +315,16 @@ function WritableStreamHandleAbortRequestIfPending(stream) {
...
-  WritableStreamFinishAbort(stream);
-
   const abortRequest = stream._pendingAbortRequest;
   stream._pendingAbortRequest = undefined;
   const promise = WritableStreamDefaultControllerAbort(stream._writableStreamController, abortRequest._reason);
   promise.then(
     abortRequest._resolve,
     abortRequest._reject
   );
+  stream._backpressure = false;
```
domenic: This feels a little hacky. It is necessary to make sure we go down the "reset the ready promise" path inside WritableStreamError, instead of the "reject the ready promise" path. Is there a better way?

The reason it doesn't feel terrible is that yeah, we're now processing an abort request, so there's not really any backpressure I guess...

reference-implementation/lib/writable-stream.js
domenic: Hmm, I guess we do similar things in WritableStreamAbort(). Maybe it is fine.

reference-implementation/lib/writable-stream.js
ricea: stream._backpressure is normally updated by WritableStreamUpdateBackpressure, so this feels like a bit of a rough edge to me too.

domenic: Hmm. That one also runs into the bad assert trying to update ready promise. Maybe something is very broken about ready promise once we're inside WritableStreamHandleAbortRequestIfPending.

reference-implementation/lib/writable-stream.js
ricea: It will probably lead to some extra churn when we add a ByteController, but as you say, we already have this issue in other places and in this case it's probably an easy fix (call stream.[[writableStreamController]].[[ResetQueue]] virtual method or similar).

tyoshino: We can introduce an interface and have different implementations for that. Yeah, virtual method as ricea said.

WritableStreamAbort() is touching the controller but through WritableStreamDefaultControllerAbort(). It's great if you abstract it into WritableStreamDefaultControllerError() e.g.

reference-implementation/lib/writable-stream.js
domenic: OK, I think I understand what's going on.

If there is a _pendingAbortRequest, WritableStreamError uses that as a signal to not mess with _readyPromise: it's already been rejected by the initial WritableStreamAbort() that set the pending abort request.

In the case we're worried about, WritableStreamHandleAbortRequestIfPending (which calls WritableStreamFinishAbort which calls WritableStreamError), we also don't want to mess with the ready promise, for the same reason---we were previously aborted, and that already rejected the ready promise. However, since WritableStreamHandleAbortRequestIfPending needs to manage the abort request itself, it clears out _pendingAbortRequest. So by the time WritableStreamError is reached, it thinks there is no _pendingAbortRequest, and goes to mess with the _readyPromise.

One reaction to this might be that attempting to use this _pendingAbortRequest signal to manage _readyPromise is not sound and we should use something _readyPromise-specific. That seems a little drastic though so I am not sure.

reference-implementation/lib/writable-stream.js
tyoshino: Or naming it ResetQueue() might just make sense too.

reference-implementation/lib/writable-stream.js
ricea: I was just reaching the same conclusion.

Can we pass a boolean into WritableStreamError called messWithReadyPromise or similar?

domenic: I think I have a fix that works by moving a tiny bit of the logic back out of WritableStreamError() and back to all the call sites except for WritableStreamFinishAbort. Uploading now-ish.

Upon further consideration it's not so bad to use _pendingAbortRequest as such a signal. It's basically a stand-in for a state between "writable" and "errored" ("aborting" perhaps). And such a state is a reasonable thing to consult when talking about messing with the ready promise.

reference-implementation/lib/writable-stream.js
domenic: Well right now WritableStreamDefaultControllerError does something rather different. I guess there is some collision in conventions for naming abtract ops in this area.

I think it is probably OK to leave as-is for now?

reference-implementation/lib/writable-stream.js
ricea: I just realised there was something else wrong with the code before you fixed it--it was doing work synchronously after calling into a sink function that might change the WritableStream internal state. It wouldn't necessarily have been a problem, but it's hard to be certain.

domenic:
Unfortunately the following very simple test fails with this patch, and was only caught by the transform stream tests, not any writable-stream tests:

```js
  close() {
    throw "x";
  }
});

writer.write("y");
writer.close();
```

It's again around the ready promise. Debugging help appreciated!

domenic:
Nevermind, I got it; patch coming :)

reference-implementation/lib/writable-stream.js
```diff
@@ -1029,13 +1026,15 @@ function defaultWriterReadyPromiseReset(writer) {
     writer._readyPromise_resolve = resolve;
     writer._readyPromise_reject = reject;
   });
+  writer._readyPromiseStatus = 'pending';
```
ricea: If we're going to have a [[readyPromiseStatus]] slot, it should be explicitly initialised in the constructor.

domenic: The plan is to use spec-ese to look at [[readyPromise]].[[PromiseStatus]]

domenic:
This should be ready for review. The corresponding tests PR is at https://github.com/w3c/web-platform-tests/pull/5224 .

reference-implementation/lib/writable-stream.js
tyoshino: I no longer can see the commit on that this comment was made, but maybe this is similar to the issue which I encountered and introduced the wasAborted variable for in #672 which I've removed finally.

reference-implementation/lib/writable-stream.js
```diff
@@ -5,6 +5,10 @@ const { InvokeOrNoop, PromiseInvokeOrNoop, ValidateAndNormalizeQueuingStrategy,
 const { rethrowAssertionErrorRejection } = require('./utils.js');
 const { DequeueValue, EnqueueValueWithSize, PeekQueueValue, ResetQueue } = require('./queue-with-sizes.js');
 
+const StartSteps = Symbol('[[Start]]');
```
ricea: Shouldn't the name of the Symbol also be '[[StartSteps]]'?

index.bs
```diff
@@ -3488,6 +3478,45 @@ WritableStreamDefaultController(<var>stream</var>, <var>underlyingSink</var>, <v
...
+      1. Otherwise, perform ! WritableStreamHandleAbortRequestIfPending(_stream_).
+      1. Perform ! WritableStreamDefaultControllerAdvanceQueueIfNeeded(*this*).
+    1. <a>Upon rejection</a> of _startPromise_ with reason _r_,
+      1. Assert: _stream_.[[state]] is `"writable"` or `"errored"`.
+      1. Perform ! WritableStreamDefaultControllerErrorIfNeeded(*this*, _r_).
+      1. Perform ! WritableStreamRejectAbortRequestIfPending(_stream_).
+</emu-alg>
+
+
+
```
ricea: Three blank lines is unusual in index.bs (I think this is the only place with 3).

index.bs
```diff
@@ -3488,6 +3478,43 @@ WritableStreamDefaultController(<var>stream</var>, <var>underlyingSink</var>, <v
...
+<h5 id="ws-default-controller-private-abort" oldids="writable-stream-default-controller-abort">\[[AbortSteps]]()</h5>
+
+<emu-alg>
+  1. Perform ! ResetQueue(*this*).
+  1. Let _sinkAbortPromise_ be ! PromiseInvokeOrNoop(*this*.[[underlyingSink]], `"abort"`, « _reason_ »).
+  1. Return the result of <a>transforming</a> _sinkAbortPromise_ with a fulfillment handler that returns
+     *undefined*.
+</emu-alg>
+
```
ricea: [[ErrorSteps]] missing?

reference-implementation/lib/writable-stream.js
```diff
@@ -736,16 +714,46 @@ class WritableStreamDefaultController {
...
+      r => {
+        assert(stream._state === 'writable' || stream._state === 'errored');
+        WritableStreamDefaultControllerErrorIfNeeded(this, r);
+        WritableStreamRejectAbortRequestIfPending(stream);
+      }
+    )
+    .catch(rethrowAssertionErrorRejection);
+  }
+
+  [AbortSteps](reason) {
```
ricea: Should these be in alphabetical order?

index.bs
```diff
@@ -3488,6 +3478,43 @@ WritableStreamDefaultController(<var>stream</var>, <var>underlyingSink</var>, <v
...
+The following are additional internal methods implemented by each {{WritableStreamDefaultController}} instance. The
+writable stream implementation will call into these.
+
+<p class="note">The reason these are in method form, instead of as abstract operations, is to make it clear that the
+writable stream implementation is decoupled from the controller implementation, and could in the future be expanded with
+other controllers, as long as those controllers implemented such internal methods. A similar scenario is seen for
+readable streams, where there actually are multiple controller types and as such the counterpart internal methods are
+used polymorphically.
+
+<h5 id="ws-default-controller-private-abort" oldids="writable-stream-default-controller-abort">\[[AbortSteps]]()</h5>
```
ricea: Is it intentional that virtual methods don't have throws/nothrow annotations?

domenic: For now, yes; they aren't abstract ops so the tooling won't let me :(

ricea:
lgtm

reference-implementation/lib/writable-stream.js
```diff
@@ -621,6 +596,15 @@ function WritableStreamDefaultWriterCloseWithErrorPropagation(writer) {
   return WritableStreamDefaultWriterClose(writer);
 }
 
+function WritableStreamDefaultWriterEnsureReadyPromiseRejected(writer, error) {
+  if (writer._readyPromiseState === 'pending') {
```
tyoshino: I think I've chosen to exclude the state of the promise intentionally from the condition to determine the action to take here on the promise so that we can see that all the state transitions are not broken (when there's anything wrong, the assertions on promise rejection helper would catch it).

I'm fine with switching to this way given much complexity in doing the promise-state-free calculation, but want you to be aware of the background.

reference-implementation/lib/writable-stream.js
```diff
@@ -226,25 +240,8 @@ function WritableStreamFinishInFlightWriteWithError(stream, reason) {
...
-      writer._readyPromise.catch(() => {});
-    }
-  } else {
-    stream._pendingAbortRequest._reject(reason);
-    stream._pendingAbortRequest = undefined;
-  }
-
-  WritableStreamRejectPromisesInReactionToError(stream);
+  WritableStreamError(stream, error);
+  WritableStreamRejectAbortRequestIfPending(stream);
```
tyoshino: hasn't this changed the order of promise rejection?

ready, closed
abort, closed

to

ready, closed
closed, abort

reference-implementation/lib/writable-stream.js
```diff
@@ -302,21 +290,8 @@ function WritableStreamFinishInFlightCloseWithError(stream, reason) {
...
-      writer._readyPromise.catch(() => {});
-    }
-  } else {
-    stream._pendingAbortRequest._reject(reason);
-    stream._pendingAbortRequest = undefined;
-  }
-
-  WritableStreamRejectClosedPromiseInReactionToError(stream);
+  WritableStreamError(stream, error);
+  WritableStreamRejectAbortRequestIfPending(stream);
```
tyoshino: ditto

reference-implementation/lib/writable-stream.js
tyoshino: hmm, the order checking unit tests might not be working well?

index.bs
```diff
@@ -2800,27 +2800,37 @@ writable stream is <a>locked to a writer</a>.
...
+      1. Perform ! WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _error_).
+  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*, perform !
+     WritableStreamRejectPromisesInReactionToError(_stream_).
+</emu-alg>
+
 <h4 id="writable-stream-finish-abort" aoid="WritableStreamFinishAbort" nothrow>WritableStreamFinishAbort (
 <var>stream</var> )</h4>
 
 <emu-alg>
   1. Let _error_ be a new *TypeError* indicating that the stream has been aborted.
```
tyoshino: remove this line

tyoshino: oh, no. L2834 in the new snapshot should be updated instead.

reference-implementation/lib/writable-stream.js
domenic: They were working. We had to update them slightly. If it is easy to switch them back though maybe that would be a good idea; let me check.

reference-implementation/lib/writable-stream.js
domenic: Yeah, I understand. It is indeed a good goal to be able to maintain the state out of band, but it just got really complex and error-prone. At least I personally was not able to figure it out.

tyoshino:
Reviewed reference impl. lgtm except for the comments!

reference-implementation/lib/writable-stream.js
domenic: Ah, it is not that easy, because WritableStreamRejectAbortRequestIfPending assumes the stream is already in the errored state. So I think it is an OK change. It would be slightly nicer for close to always be last but it still seems OK.

index.bs
tyoshino: Being fixed at https://github.com/whatwg/streams/pull/711

Issue #670, opened 2017-01-24
Assign names to the Records used in the standard


ricea:
As an implementer it is useful to be able to quickly find all the places a data structure is used in order to make decisions about how it should be structured. I had this issue even when implementing in a dynamically-typed language.

I think it would help to have names for the Records used in the standard.

For example, the record added to [[readRequests]] could be called "Read Request Record".

This doesn't completely solve the problem as Records are used in places where the fact that they are Records is not mentioned. But maybe we could rephrase things like

2. Let _readRequest_ be the first element of _reader_.[[readRequests]].

like this

2. Let _readRequest_ be the first Request Request Record from _reader_.[[readRequests]].

domenic:
> This doesn't completely solve the problem as Records are used in places where the fact that they are Records is not mentioned. But maybe we could rephrase things like

This is still unsolved. But I guess #1048 would tackle that.

PR #706, opened 2017-03-24
Resolve/reject [[pendingAbortRequest]].[[promise]]

ricea:
The [[pendingAbortRequest]] internal slot of WritableStream is a Record and so
cannot be resolved or rejected directly.

When resolving or rejecting [[pendingAbortRequest]], explicitly operate on the
[[promise]] slot.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2625,7 +2625,9 @@ Instances of {{WritableStream}} are created with the internal slots described in
   </tr>
   <tr>
     <td>\[[pendingAbortRequest]]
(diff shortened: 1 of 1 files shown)


ricea:
> Thank you for fixing! Can I ask you also update https://streams.spec.whatwg.org/#ws-internal-slots?

Updated. I'm not completely happy with the wording, please take a look.

domenic:
The phrasing seems good modulo #670 as a longer-term solution. Gotta rebase though.

PR #707, opened 2017-03-27
Editorial: change internal method naming

domenic:
This renames the [[Cancel]\]() and [[Pull]\]() internal methods on readable stream controllers to have the suffix "steps". This better emphasizes that they are the controller-specific steps as part of larger cancelation/pulling processes.

This is especially important for upcoming work for writable streams, in #705, where one of the methods will be named [[ErrorSteps]\](). Having an internal method of writable stream default controllers named [[Error]\](), which is distinct from the abstract operation WritableStreamDefaultControllerError(), would just be confusing.


ricea:
lgtm

PR #709, opened 2017-03-27
Clarify writer.desiredSize/ready semantics around aborts

### LINKED ISSUE #708: writer.desiredSize "will be null if the stream is errored"
https://streams.spec.whatwg.org/commit-snapshots/17c9fa6c075805891aad637d3395da96ab2b2d61#default-writer-desiredSize

> It will be null if the stream is errored,

It will also be null if abort() has been called but a write() or close() is still in progress. I'm not sure whether this matters enough to be more precise in the wording here.

domenic:
Fixes #708.


ricea:
lgtm (sorry for the delay, I somehow thought this had already landed).

tyoshino:
lgtm

PR #710, opened 2017-03-28
Remove the warning about writable stream spec stability

domenic:
As can be seen by the remaining issue count and few recent changes, we have stabilized the spec.


ricea:
lgtm

tyoshino:
lgtm!

PR #713, opened 2017-03-28
Update WebSocket examples

domenic:
I recently learned that WebSockets will tell you if they closed cleanly. This allows us to illustrate returning a potentially-rejected promise from the underlying sink's close() and abort() methods.

This also makes the writable streams error themselves if the WebSockets are closed by the server (as opposed to closed by the client), as this represents an unexpected cessation of the ability to write data.

Finally this synchronizes the examples so that they both send a status code of 1000 for closing successfully.

---

Please carefully check my semantics here as I have not used the web socket API in anger recently. I may not be holding it right.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4230,7 +4230,10 @@ promises returned by its <a lt="writable stream writer">writer</a>'s {{WritableS
 
     return new WritableStream({
       start(controller) {
-        ws.onerror = () => controller.error(new Error("The WebSocket errored!"));
+        ws.onerror = () => {
+          controller.error(new Error("The WebSocket errored!"));
+          ws.onclose = null;
+        };
         ws.onclose = () => controller.error(new Error("The server closed the connection unexpectedly!"));
         return new Promise(resolve => ws.onopen = resolve);
       },
@@ -4386,6 +4389,7 @@ source abstractions.
       this._ws.onclose = () => controller.error(new Error("The server closed the connection unexpectedly!"));
       this._ws.addEventListener("error", () => {
         controller.error(new Error("The WebSocket errored!"));
+        this._ws.onclose = null;
       });
 
       return new Promise(resolve => this._ws.onopen = resolve);
--- a/index.bs
+++ b/index.bs
@@ -4231,6 +4231,7 @@ promises returned by its <a lt="writable stream writer">writer</a>'s {{WritableS
     return new WritableStream({
       start(controller) {
         ws.onerror = () => controller.error(new Error("The WebSocket errored!"));
+        ws.onclose = () => controller.error(new Error("The server closed the connection unexpectedly!"));
         return new Promise(resolve => ws.onopen = resolve);
       },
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4241,19 +4242,26 @@ promises returned by its <a lt="writable stream writer">writer</a>'s {{WritableS
...
-          ws.close(4000, reason && reason.message);
-        });
-      }
+        return closeWS(4000, reason && reason.message);
+      },
     });
+
+    function closeWS(code, reasonString) {
+      return new Promise((resolve, reject) => {
+        ws.onclose = e => {
```
ricea: The onerror event will fire redundantly. I think it is okay to ignore it completely, so just unsetting the event handler it should work.

domenic: In the not-clean case? I believe you, but can't quite find it in the spec... all I see for error is

> If the user agent was required to fail the WebSocket connection, or if the the WebSocket connection was closed after being flagged as full, fire an event named error at the WebSocket object.

index.bs
```diff
@@ -4231,6 +4231,7 @@ promises returned by its <a lt="writable stream writer">writer</a>'s {{WritableS
     return new WritableStream({
       start(controller) {
         ws.onerror = () => controller.error(new Error("The WebSocket errored!"));
+        ws.onclose = () => controller.error(new Error("The server closed the connection unexpectedly!"));
```
ricea: If for example the remote server closes the connection abruptly, then onerror and onclose will both fire.

I don't actually know of a case where onerror fires without onclose also firing. Certainly Chrome's implementation never does it. So one way of simplifying this would just be to not set an onerror handler.

domenic: Just to check my understanding, the server could close it non-abruptly, right? In which case close would fire but not error.

index.bs
ricea: So "fail the WebSocket connection" points you to here https://tools.ietf.org/html/rfc6455#section-7.1.7 which tells you that the first thing we do it close the connection. Which then causes onclose to fire.

It might be that an implementation fires onerror much sooner than onclose, but Chrome doesn't do that. In fact, we tear down the connection synchronously when we detect a failure condition:

https://cs.chromium.org/chromium/src/net/websockets/websocket_channel.cc?q=%22Careful+study+of+RFC6455%22

index.bs
ricea: Correct.

index.bs
domenic: OK, but I believe it will fail it cleanly, right? Non-cleanly is reserved for cases where the server just goes offline or something, I thought?

index.bs
domenic: OK, so this definitely needs to be fixed since you can't call controller.error() twice. Not sure about the best way to do that while balancing:

- Giving as much information to the developer as possible (e.g. bad close vs good)
- Keeping the code simple enough for an example

Will noodle on it.

index.bs
ricea: Chrome definitely sets wasClean to false. 

For me the most important sentence is this from RFC6455 section 7.1.7:

   An endpoint MUST NOT continue to attempt to
   process data (including a responding Close frame) from the remote
   endpoint after being instructed to _Fail the WebSocket Connection_.

If we're prohibited from processing the Close frame from the server, there's no way to complete the closing handshake, and so it can't ever be clean.

This is actually quite subtle and it's possible other implementations have read the spec differently. There are a bunch of web-platform-tests where the test expects wasClean = true but Chrome sets it to false, but in my brief investigation it appears this is an interoperability issue with the test server, rather than being caused by any subtle issues.

I think from a standard's point of view, using wasClean in this way should totally work. For compatibility, you might want to have an additional wasOnErrorCalled flag. Personally I think the example should aim for brevity rather than maximum compatibility.

index.bs
```diff
@@ -4230,7 +4230,10 @@ promises returned by its <a lt="writable stream writer">writer</a>'s {{WritableS
 
     return new WritableStream({
       start(controller) {
-        ws.onerror = () => controller.error(new Error("The WebSocket errored!"));
+        ws.onerror = () => {
+          controller.error(new Error("The WebSocket errored!"));
+          ws.onclose = null;
```
ricea: That should work.

domenic: So does this overall LGTY now?

ricea: Yes. lgtm.

Issue #693, opened 2017-03-09
Inconsistency between abort() then error() and error() then abort(), with ongoing close()/write()


tyoshino:
From https://github.com/whatwg/streams/pull/672#discussion_r105079238

`writer.abort()` fails immediately when the writable stream is in the `"errored"` state regardless of whether or not the stream has an unfinished in-flight close()/write().

OTOH, when a writable stream is errored by `controller.error()` with a pending `writer.abort()` and an in-flight close()/write() on it, `writer.abort()` gets rejected only after the stream finishes the in-flight close()/write().

I found this inconsistent and raised a discussion in the PR.

I agree with ricea that this is a trivial less-important corner case, but sometimes inconsistent behavior confuses users or is working as a hidden source of code complexity. So, I'd like to another scan after landing the PR.

domenic:
I guess there are three alternatives:

1. Make consistent by making both write-abort-error and write-error-abort cause the abort to not reject until the write finishes
2. Make consistent by making both write-abort-error and write-error-abort reject the abort immediately
3. Leave as-is

I don't have a strong preference between them. I think it would be good to see if implementing either (1) or (2) would create code simplifications.

domenic:
I did some analysis on this and other related matters in https://docs.google.com/document/d/1rOe6CiEHxU7m3_YDh1lv334TxCR6i3X71Fjnem_XKCc/edit?usp=sharing (at the end). I ended up concluding that this is reasonable behavior even from first principles, and isn't (just) a result of our codebase and tests growing organically over time. So, I will close this tomorrow after verifying we have tests for it.

ricea:
This is reassuring. Developers will have to work with a simpler mental model than the real thing and obviously it is good if most behaviour is predicted by this model.

domenic:
This is tested in aborting.js by the two tests:

- writer.abort(), controller.error() while there is an in-flight write, and then finish the write
- controller.error(), writer.abort() while there is an in-flight write, and then finish the write

However I suppose we should also add tests for in-flight close.

tyoshino:
(1) would remove a few lines from WritableStreamAbort() while adding a few to WritableStreamFinishInFlightClose(), and move some lines in the other handlers, I guess. So, not much gain of simplification, right.

[...]

PR #715, opened 2017-03-28
Make writer.abort() fail immediately

tyoshino:


tyoshino:
A test impl for #693

ricea:
I would prefer not to reject abort() immediately on controller.error() as there may still be operations in progress. I would like there to be a strong guarantee for all of close(), abort() and .closed promises that when they resolve no sink methods are executing, and no further sink methods will be executed.

tyoshino:
I think the second sentence is what you said in https://github.com/whatwg/streams/pull/672#discussion_r104659281. As I said in https://github.com/whatwg/streams/pull/672#discussion_r104662189, it's not satisfied. Only (1) in Domenic's comment https://github.com/whatwg/streams/issues/693#issuecomment-285518838 would satisfy that.

Let me confirm the context. I agree with your point in https://github.com/whatwg/streams/pull/672#discussion_r104855103 but also wondered if correcting it can provide some real benefit in logic simplification and tried that.

Could you please summarize your thoughts again? I understand your point but the current status is that it's not satisfied.

ricea:
I wasn't thinking about it like that. You're right, since it is invisible to the user that controller.error() has been called they are not using abort() incorrectly and so can have a reasonable expectation that the invariant would hold.

You are right that I am contradicting what I said in https://github.com/whatwg/streams/pull/672#discussion_r104855103. I have been thinking about what I'd like to write in a "Design Principles" section, and this has convinced me that it is important to provide this guarantee.

But it worries me that this leads to a contradition with Domenic's conclusion in the [Writable stream state machine](https://docs.google.com/document/d/1rOe6CiEHxU7m3_YDh1lv334TxCR6i3X71Fjnem_XKCc/edit) document.

domenic:
The basic problem is figuring out what controller.error() means. Is it an immediate "transition to the errored state"? If so that's a bit unusual. But that's been our answer so far.

Or is it similar to abort(), in that it's a request to error when in-flight stuff finishes?

domenic:
In particular you end up with parallel questions, like, what if you call controller.error() while sink.close() is running, if sink.close() returns a fulfilled promise. If we switched to semantics similar to abort(), then we'd have to say that this is ignored, as by the time the request-to-error gets executed, the stream has already successfully transitioned to the closed state.

ricea:
> Or is it similar to abort(), in that it's a request to error when in-flight stuff finishes?

I think this an important question.

My understanding of controller.error():
1. It is for out-of-band errors that aren't tied to the execution of a particular operation
2. It is generally expected to be called at times when no operations are executing
3. If it is called when an operation *is* executing, we may have a race condition or conflicting information.

I am leaning towards the idea that if controller.error() happens simultaneously with close() or with abort() pending, then it is the close() or abort() that "wins" and determines the eventual state of the stream.

I think I want controller.error() to behave exactly like abort() (but still not return a promise).

Implementation idea: add a flag which takes over the signalling function of [[pendingAbortRequest]]. The flag signals that abort() or controller.error() has been called. [[pendingAbortRequest]] would still be used as storage for the promise returned by abort(). Alternatively, an extra state "erroring" could be added.

domenic:
I would be down for that change. It seems likely to simplify things. @tyoshino do you feel up for prototyping it?

domenic:
So this has been obsoleted by https://github.com/whatwg/streams/pull/721, I am pretty sure. Let me know if I am wrong!

tyoshino:
Agreed!

PR #716, opened 2017-03-29
Editorial: fix usage of RFC 2119 terms in non-normative notes

domenic:
Found using the neato new feature https://github.com/tabatkins/bikeshed/issues/958.

I didn't turn it on for streams right now because I plan to turn it on for all WHATWG specs very shortly. So I just added the "complain about" line, fixed all the issues, then removed the line.


ricea:
lgtm

PR #718, opened 2017-03-29
Early draft of WritableStream "Design Philosophy" section

ricea:
This is supposed to help implementors and others understand the design
principles.

Markup isn't done properly yet, so please evaluate the text only.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2568,6 +2568,43 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
   writeRandomBytesForever(myWritableStream).catch(e => console.error("Something broke", e));
   </code></pre>
 </div>
+
+<h3 id="ws-design-philosopy">Design Philosophy</h3>
+
+<div>While sharing the principles for streams in general, a number of additional principles have informed the design of
+the WritableStream class.</div>
+
+<ul>
+  <li><p>Only one sink method can ever be executing at a time.
+  <li><p>Sink methods are treated as atomic.
+    <ul>
+      <li><p>"Atomic" here means that the time between a sink method being called and the returned Promise resolving or
+        rejecting is treated as indivisible.
+      <li><p>A new sink method will never be called until the Promise from the previous one has resolved.
+      <li><p>State changes do not take effect until any in-flight sink method has completed.
+    </ul>
+  <li><p>Exception: If something has happened that will error the stream, for example writer.abort() has been called, then
+      new calls to writer.write() will start failing immediately. There's no user benefit in waiting for the current operation
+      to complete before informing the user that writer.write() has failed.
+  <li><p>The writer.ready promise and the value of writer.desiredSize reflect whether a write() performed right
+  now would be effective.
+      <ul>
+        <li>writer.ready and writer.desiredSize will change even in the middle of executing a sink method, as soon as we
+        know that calling writer.write() won't work any more.
+          <div class=note>Because promises are dispatched asynchronously, the state can still change between
+          writer.ready becoming fulfilled and write() being called.</div>
+        <li>The value of writer.desiredSize decreases synchronously with every call to writer.write(). This implies that
+        strategy.size() is executed synchronously.
+       </ul>
+   <li><p>The writer.closed promise and the promises returned by writer.close() and writer.abort() do not resolve or
+   reject until no sink methods are executing and no further sink methods will be executing.
+       <ul>
+         <li>If the user of the WritableStream wants to retry against the same underlying data item, it is important to
+         have confidence that all other operations have ceased.
+         <li>This principle also applies to the ReadableStream pipeTo() method.
+       </ul>
+</ul>
+
 <h3 id="ws-class" interface lt="WritableStream">Class <code>WritableStream</code></h3>
 
 <h4 id="ws-class-definition">Class Definition</h4>


ricea:
@domenic @tyoshino Suggestions, comments, radical changes all welcome.

I uploaded it as a whatwg/streams branch to make it easy to look at the preview, which should show up at
https://streams.spec.whatwg.org/branch-snapshots/writable-streams-design-philosophy/ eventually.

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
 </div>
+
+<h3 id="ws-design-philosopy">Design Philosophy</h3>
+
+<div>While sharing the principles for streams in general, a number of additional principles have informed the design of
+the WritableStream class.</div>
+
+<ul>
+  <li><p>Only one sink method can ever be executing at a time.
+  <li><p>Sink methods are treated as atomic.
```
domenic: IMO this can be one paragraph instead of sub-bullets

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+    <ul>
+      <li><p>"Atomic" here means that the time between a sink method being called and the returned Promise resolving or
+        rejecting is treated as indivisible.
```
domenic: This doesn't really clarify. What would divide them? I think the next bullet points actually state what atomic means.

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+      <li><p>A new sink method will never be called until the Promise from the previous one has resolved.
+      <li><p>State changes do not take effect until any in-flight sink method has completed.
```
domenic: "State changes" isn't too clear here, especially given the next two bullet points worth of exceptions.

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+    </ul>
+  <li><p>Exception: If something has happened that will error the stream, for example writer.abort() has been called, then
+      new calls to writer.write() will start failing immediately. There's no user benefit in waiting for the current operation
+      to complete before informing the user that writer.write() has failed.
+  <li><p>The writer.ready promise and the value of writer.desiredSize reflect whether a write() performed right
+  now would be effective.
+      <ul>
+        <li>writer.ready and writer.desiredSize will change even in the middle of executing a sink method, as soon as we
+        know that calling writer.write() won't work any more.
```
domenic: They also change in cases besides "won't work" though

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+          <div class=note>Because promises are dispatched asynchronously, the state can still change between
+          writer.ready becoming fulfilled and write() being called.</div>
+        <li>The value of writer.desiredSize decreases synchronously with every call to writer.write(). This implies that
+        strategy.size() is executed synchronously.
```
domenic: Given how we're not calling strategy.size as a method, I'd state this as "the queuing strategy's size() function is executed..."

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+       </ul>
+   <li><p>The writer.closed promise and the promises returned by writer.close() and writer.abort() do not resolve or
+   reject until no sink methods are executing and no further sink methods will be executing.
+       <ul>
+         <li>If the user of the WritableStream wants to retry against the same underlying data item, it is important to
```
domenic: "data item" = chunk?

"Retry against" is a bit confusing to me.

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+         have confidence that all other operations have ceased.
+         <li>This principle also applies to the ReadableStream pipeTo() method.
+       </ul>
+    <li><p>Promises resolve or reject in consistent order. In particular, writer.ready always resolves before
```
domenic: I think "fulfill" is more informative here (although both are technically accurate)

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+    writer.closed, even in cases where they both resolve "at the same time".
```
domenic: I'd state this as "where both are fulfilling in reaction to the same occurrence"

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
...
+</ul>
+
+<div>Some of these design decisions improve predictability, ease-of-use and safety for developers at the expense of
```
domenic: Oxford comma please (after "ease-of-use")

I know it's not markup-critiquing time yet but this `<div>` seems pretty unnecessary :)

domenic: I think this could move up to merge with the intro actually.

index.bs
```diff
@@ -2568,6 +2568,48 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
   writeRandomBytesForever(myWritableStream).catch(e => console.error("Something broke", e));
   </code></pre>
 </div>
+
+<h3 id="ws-design-philosopy">Design Philosophy</h3>
+
+<div>While sharing the principles for streams in general, a number of additional principles have informed the design of
+the WritableStream class.</div>
```
domenic: I wonder if we can make clearer the "scope" of this section. It's got great info but it's of a very specific kind. It's not about the philosophy behind writable streams in general (of the type of stuff covered by the "Model" section). It's more about the details of its API surface and the interactions between the sink, controller, and writer APIs. Not sure on the exact phrasing, but that might be a start.

ricea: I changed the name to "Design Of The State Machine" which may or may not be helpful.

index.bs
ricea: Should I use paragraphs for everything instead of sub-bullets? I quite like the sub-bullets to be honest.

index.bs
ricea: I just took this out.

index.bs
ricea: Reworded a bit.

index.bs
ricea: Reworded a bit.

index.bs
ricea: Done.

index.bs
ricea: I tried to clarify by using the example of a file, rather than try to come up with a general term for "named or otherwise identifiable entities which can be retried independently of the lifetime of any particular stream".

index.bs
ricea: Done.

index.bs
ricea: Great! Thanks.

index.bs
ricea: Moved it up and made it a paragraph.

ricea:
I added another item, "Queued calls to writer methods such as write() are not cancelled when writer.releaseLock() is called. This makes them easy to use in a "fire and forget" style." which raises the question of exactly how much detail belongs in this section.

Personally I'm happy with it being quite inclusive, but I don't think it should be an exhaustive list of every design decision we ever made.

Maybe I should rephrase it as "Fire-and-forget style usage is supported. Therefore..."

index.bs
```diff
@@ -2568,6 +2568,45 @@ nothrow>ReadableByteStreamControllerShouldCallPull ( <var>controller</var> )</h4
   writeRandomBytesForever(myWritableStream).catch(e => console.error("Something broke", e));
   </code></pre>
 </div>
+
+<h3 id="ws-design-of-the-state-machine">Design Of The State Machine</h3>
+
```
ricea: (Needs to be marked non-normative)

Ansiic112 (commented):
http://index.bs

PR #720, opened 2017-03-30
Require pipeTo() to properly use the destination queue

### LINKED ISSUE #653: pipeTo(): the stream queues should be used
An implementation of pipeTo() that doesn't make use of the queues in the streams is compliant with the standard. Consider

```javascript
let writable = new WritableStream(burstySink); // [1]
readable.pipeTo(writable);
```

versus

```javascript
let writable = new WritableStream(burstySink, new CountQueuingStrategy(100)); // [2]
readable.pipeTo(writable);
```

It is reasonable for content authors to expect that the extra queuing in [2] will stop the pipe from stalling due to the burstiness of the sink. However, user agents are permitted to treat these two cases as exactly the same. I think this hurts predictability of the platform.

Similarly, adding a queue to a ReadableStream should be effective in mitigating a bursty or high-latency source.

Two issues:
1. Balancing the requirement against "_WritableStreamDefaultWriterGetDesiredSize(writer) may be used to determine the flow rate heuristically, e.g. by delaying reads while it is judged to be "low" compared to the size of chunks that have been typically read._" is difficult.
2. Because of 1., phrasing it in a way that makes it testable is difficult.

domenic:
Fixes #653.

Tests: https://github.com/w3c/web-platform-tests/pull/5262 by @isonmad. I've confirmed these modifications to the reference implementation pass those tests.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -643,12 +643,11 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
          from _reader_.
        * If _reader_ is a <a>BYOB reader</a>, WritableStreamDefaultWriterGetDesiredSize(_writer_) should be used to
          determine the size of the chunks read from _reader_.
-       * After taking into account these backpressure signals, reading and writing should be done promptly; other
-         signals should not be used to delay the continual reading/writing.
-         <p class="example" id="example-bad-backpressure">An implementation that waits for each write to successfully complete before proceeding to
-         the next read/write operation violates this recommendation. In doing so, such an implementation makes the
-         <a>internal queue</a> of _dest_ useless, as it ensures _dest_ always contains at most one queued
-         <a>chunk</a>.</p>
+       * Reads or writes should not be delayed for reasons other than these backpressure signals.
+         <p class="example" id="example-bad-backpressure">An implementation that waits for each write to successfully
+         complete before proceeding to the next read/write operation violates this recommendation. In doing so, such an
+         implementation makes the <a>internal queue</a> of _dest_ useless, as it ensures _dest_ always contains at most
+         one queued <a>chunk</a>.</p>
      * <strong>Shutdown must stop all activity:</strong> if _shuttingDown_ becomes *true*, the user agent must not
        initiate further reads from _reader_ or writes to _writer_. (Ongoing reads and writes may finish.) In particular,
        the user agent must check the below conditions on *this*.[[state]] and _dest_.[[state]] before performing any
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -642,10 +642,15 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
          from _reader_.
        * If _reader_ is a <a>BYOB reader</a>, WritableStreamDefaultWriterGetDesiredSize(_writer_) should be used to
-         determine the size of the chunks read from _reader_.
-       * Otherwise, WritableStreamDefaultWriterGetDesiredSize(_writer_) may be used to determine the flow rate
-         heuristically, e.g. by delaying reads while it is judged to be "low" compared to the size of chunks that have
-         been typically read.
+         determine the size of the chunks read from _reader_. Otherwise, the desired size may be used to determine the
+         flow rate heuristically, e.g. by delaying reads while the desired size is judged to be "low" compared to the
+         size of <a>chunks</a> that have been typically read.
+       * After taking into account these backpressure signals, reading and writing should be done as fast as possible;
```
ricea: "as possible" seems ambiguous to me. It could have a wide variety of meanings from "you must optimise this at the machine code level for your target architecture" to "someone said we could maximise battery life by only forwarding data once a minute, so that's the fastest possible".

"promptly" feels right to me. "reading and writing should be done promptly". Not necessarily instantly, but not wasting time.

index.bs
```diff
@@ -642,10 +642,15 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+         other signals should not be used to delay the continual reading/writing.
```
ricea: I kinda do want to use "other signals". Let's say I have

```javascript
fetch(url).then(response => {
  response.body
    .pipeThrough(new TextDecoder())
    .pipeTo(tag.writable);
});
```

`tag.writable.desiredSize` is 4096 characters. The TextDecoder transform just counts objects, so it will happily pass through any chunk of bytes as a single object. The system as a whole has enough information to optimise the size of chunks read from `response.body`, but WritableStreamDefaultWriterGetDesiredSize(TextDecoder.writable) is just returning the number 1. 

In the general case, I'd like to use global information to optimise my pipes, and "other signals" might be too restrictive.

How about "reading should not be delayed for reasons other than optimizing chunk size" ?

domenic: Discussed offline---went in the direction of removing the heuristics instead.

index.bs
```diff
@@ -643,9 +643,12 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
          from _reader_.
        * If _reader_ is a <a>BYOB reader</a>, WritableStreamDefaultWriterGetDesiredSize(_writer_) should be used to
          determine the size of the chunks read from _reader_.
-       * Otherwise, WritableStreamDefaultWriterGetDesiredSize(_writer_) may be used to determine the flow rate
-         heuristically, e.g. by delaying reads while it is judged to be "low" compared to the size of chunks that have
-         been typically read.
+       * After taking into account these backpressure signals, reading and writing should be done promptly; other
```
ricea: We can probably say less here now that we don't have to argue with the previous sentence. How about:

"Reads or writes should not be delayed for reasons other than backpressure."

ricea:
lgtm

PR #721, opened 2017-03-30
Unified error handling for WritableStream

ricea:
Previously, controller.error() and errors from strategy size() could transition
the stream state to change to 'errored' while an underlying sink operation 
was in flight.

Only change to 'errored' while no underlying sink operations are in flight,
consistent with abort(). The first error to happen sets the [[storedError]]
internal slot, regardless of type.

Other behavioural changes:
- The writer.closed promise now doesn't reject until the underlying sink
  abort() method has completed.
- writer.close(), writer.abort() and writer.closed promises now consistently
  don't fulfill or reject until no underlying sink methods are executing or will
  be executed. Exceptions:
  - Duplicate calls to close() or abort() still reject immediately.
  - writer.closed still rejects when writer.releaseLock() is called.
- writer.write(), writer.close() and writer.abort() consistently return the
  status of the underlying sink operation, provided it executes.
- It is no longer possible to error the stream once underlying sink close()
  has been called.
- The controller parameter to the underlying sink close() method is no
  longer useful and has been removed.
- Duplicate calls to controller.error() are now a no-op.
- Calls to controller.error() while the stream is erroring, errored or
  closed are no a no-op.

Major implementation changes:
- There is a new "erroring" state. The state synchronously changes
  from "writable" to "erroring" when an error condition is triggered,
  then changes to "errored" once the underlying sink operation
  has finished.
- The stream.[[pendingAbortRequest]] Record has an additional
  boolean field [[wasAlreadyErrored]]. When it is true the underlying
  sink abort() method will not be called.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2679,7 +2679,7 @@ WritableStream(<var>underlyingSink</var> = {}, { <var>size</var>, <var>highWater
       <a>underlying sink</a>. It can return a promise to signal success or failure of the write operation. The stream
       implementation guarantees that this method will be called only after previous writes have succeeded, and never
       after <code>close</code> or <code>abort</code> is called.
-    <li> <code>close(controller)</code> is called after the producer signals that they are done writing chunks to the
+    <li> <code>close()</code> is called after the producer signals that they are done writing chunks to the
       stream, and all queued-up writes successfully complete. It can perform any actions necessary to finalize writes
       to the <a>underlying sink</a>, and release access to it. If this process is asynchronous, it can return a promise
       to signal success or failure. The stream implementation guarantees that this method will be called only after all
@@ -2799,43 +2799,18 @@ writable stream is <a>locked to a writer</a>.
   1. Let _state_ be _stream_.[[state]].
   1. If _state_ is `"closed"`, return <a>a promise resolved with</a> *undefined*.
   1. If _state_ is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
-  1. Assert: _state_ is `"writable"`.
   1. Let _error_ be a new *TypeError* indicating that the stream has been requested to abort.
   1. If _stream_.[[pendingAbortRequest]] is not *undefined*, return <a>a promise rejected with</a> _error_.
-  1. Let _controller_ be _stream_.[[writableStreamController]].
-  1. Assert: _controller_ is not *undefined*.
-  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false* and _controller_.[[started]] is *true*,
-    1. Perform ! WritableStreamFinishAbort(_stream_).
-    1. Return ! _controller_.[[AbortSteps]](_reason_).
-  1. Let _writer_ be _stream_.[[writer]].
-  1. If _writer_ is not *undefined*,
-    1. Perform ! WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _error_).
+  1. Assert: _state_ is `"writable"` or `"erroring"`.
+  1. Let _wasAlreadyErroring_ be *false*.
+  1. If _state_ is `"erroring"`,
+    1. Set _wasAlreadyErroring_ to *true*.
+    1. Set _reason_ to *undefined*.
   1. Let _promise_ be <a>a new promise</a>.
-  1. Set _stream_.[[pendingAbortRequest]] to Record {[[promise]]: _promise_, [[reason]]: _reason_}.
+  1. Set _stream_.[[pendingAbortRequest]] to Record {[[promise]]: _promise_, [[reason]]: _reason_,
+     [[wasAlreadyErroring]]: _wasAlreadyErroring_}.
+  1. If _wasAlreadyErroring_ is *false*, perform ! WritableStreamStartErroring(_stream_, _error_).
   1. Return _promise_.
-</emu-alg>
-
-<h4 id="writable-stream-error" aoid="WritableStreamError" nothrow>WritableStreamError ( <var>stream</var>,
-<var>error</var> )</h4>
-
-<emu-alg>
-  1. Set _stream_.[[state]] to `"errored"`.
-  1. Set _stream_.[[storedError]] to _error_.
-  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
-  1. If _stream_.[[pendingAbortRequest]] is *undefined*,
-    1. Let _writer_ be _stream_.[[writer]].
-    1. If _writer_ is not *undefined*,
-      1. Perform ! WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _error_).
-  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*, perform !
-     WritableStreamRejectPromisesInReactionToError(_stream_).
-</emu-alg>
-
-<h4 id="writable-stream-finish-abort" aoid="WritableStreamFinishAbort" nothrow>WritableStreamFinishAbort (
-<var>stream</var> )</h4>
-
-<emu-alg>
-  1. Let _error_ be a new *TypeError* indicating that the stream has been aborted.
-  1. Perform ! WritableStreamError(_stream_, _error_).
 </emu-alg>
 
 <h3 id="ws-abstract-ops-used-by-controllers">Writable Stream Abstract Operations Used by Controllers</h3>
@@ -2861,6 +2836,65 @@ visible through the {{WritableStream}}'s public API.
   1. Return _promise_.
 </emu-alg>
 
+<h4 id="writable-stream-deal-with-rejection" aoid="WritableStreamDealWithRejection"
+nothrow>WritableStreamDealWithRejection ( <var>stream</var>, <var>error</var> )</h4>
+
+<emu-alg>
+  1. Let _state_ be _stream_.[[state]].
+  1. If _state_ is `"writable"`,
+    1. Perform ! WritableStreamStartErroring(_stream_).
+    1. Return.
+  1. Assert: _state_ is `"erroring"`.
+  1. Perform ! WritableStreamFinishErroring(_stream_).
+</emu-alg>
+
+<h4 id="writable-stream-start-erroring" aoid="WritableStreamStartErroring" nothrow>WritableStreamStartErroring (
+<var>stream</var>, <var>reason</var> )</h4>
+
+<emu-alg>
+  1. Assert: _stream_.[[storedError]] is *undefined*.
+  1. Assert: _stream_.[[state]] is `"writable"`.
+  1. Let _controller_ be _stream_.[[writableStreamController]].
+  1. Assert: _controller_ is not *undefined*.
+  1. Set _stream_.[[state]] to `"erroring"`.
+  1. Set _stream_.[[storedError]] to _reason_.
+  1. Let _writer_ be _stream_.[[writer]].
+  1. If _writer_ is not *undefined*, perform !
+     WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _reason_).
+  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false* and _controller_.[[started]] is *true*, perform
+     ! WritableStreamFinishErroring(_stream_).
+</emu-alg>
+
+<h4 id="writable-stream-finish-erroring" aoid="WritableStreamFinishErroring" nothrow>WritableStreamFinishErroring
+( <var>stream</var> )</h4>
+
+<emu-alg>
+  1. Assert: _stream_.[[state]] is `"erroring"`.
+  1. Assert: ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*.
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
+  1. Let _storedError_ be _stream_.[[storedError]].
+  1. Repeat for each _writeRequest_ that is an element of _stream_.[[writeRequests]],
+    1. <a>Reject</a> _writeRequest_ with _storedError_.
+  1. Set _stream_.[[writeRequest]] to an empty List.
+  1. if _stream_.[[pendingAbortRequest]] is *undefined*,
+    1. Perform !  WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
+    1. Return.
+  1. Let _abortRequest_ be _stream_.[[pendingAbortRequest]].
+  1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. If _abortRequest_.[[wasAlreadyErroring]] is *true*,
+    1. <a>Reject</a> _abortRequest_.[[promise]] with _storedError_.
+    1. Perform ! WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
+    1. Return.
+  1. Let _promise_ be ! stream.[[writableStreamController]].[[AbortSteps]](_abortRequest_.[[reason]]).
+  1. <a>Upon fulfillment</a> of _promise_,
+    1. <a>Resolve</a> _abortRequest_.[[promise]] with *undefined*.
+    1. Perform ! WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
+  1. <a>Upon rejection</a> of _promise_ with reason _reason_,
+    1. <a>Reject</a> _abortRequest_.[[promise]] with _reason_.
+    1. Perform ! WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
+</emu-alg>
+
 <h4 id="writable-stream-finish-in-flight-write" aoid="WritableStreamFinishInFlightWrite"
 nothrow>WritableStreamFinishInFlightWrite ( <var>stream</var> )</h4>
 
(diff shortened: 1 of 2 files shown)


ricea:
Needs a proper commit message writing, but you can see the behaviour changes from the tests. PTAL.

The standard text changes are not written yet.

reference-implementation/lib/writable-stream.js
```diff
@@ -130,27 +136,17 @@ function WritableStreamAbort(stream, reason) {
...
   }
+  if (stream._pendingAbortRequest !== undefined) {
+    return Promise.reject(new TypeError('Aborting in progress'));
+  }
 
   assert(state === 'writable', 'state must be writable');
 
   const error = new TypeError('Requested to abort');
-  if (stream._pendingAbortRequest !== undefined) {
+  if (stream._pendingAbortRequest === true) {
```
domenic: Is it a boolean or a promise?

reference-implementation/lib/writable-stream.js
```diff
@@ -405,7 +383,7 @@ class WritableStreamDefaultWriter {
     const state = stream._state;
 
     if (state === 'writable') {
-      if (stream._pendingAbortRequest !== undefined) {
+      if (stream._pendingError === true) {
         const error = new TypeError('Requested to abort');
```
domenic: Error message needs tweaking. Or should we use _storedError? I think that would be consistent with e.g. write().

reference-implementation/lib/writable-stream.js
```diff
@@ -567,7 +542,7 @@ function WritableStreamDefaultWriterClose(writer) {
     stream._closeRequest = closeRequest;
   });
 
-  if (stream._backpressure === true) {
+  if (stream._backpressure === true && !stream._pendingError) {
```
domenic: Nit: === false to match spec text

reference-implementation/lib/writable-stream.js
```diff
@@ -627,16 +611,12 @@ function WritableStreamDefaultWriterRelease(writer) {
...
   WritableStreamDefaultWriterEnsureReadyPromiseRejected(writer, releasedError);
 
-  if (state === 'writable' || WritableStreamHasOperationMarkedInFlight(stream) === true) {
-    defaultWriterClosedPromiseReject(writer, releasedError);
-  } else {
-    defaultWriterClosedPromiseResetToRejected(writer, releasedError);
-  }
-  writer._closedPromise.catch(() => {});
+  // The state transitions to "errored" before the sink abort() method runs, but the writer.closed promise is not
+  // rejected until afterwards. This means that simply testing state will not work
```
domenic: Nit: period at end of sentence.

reference-implementation/lib/writable-stream.js
```diff
@@ -711,8 +691,12 @@ class WritableStreamDefaultController {
       throw new TypeError(
         'WritableStreamDefaultController.prototype.error can only be used on a WritableStreamDefaultController');
     }
+    const stream = this._controlledWritableStream;
+    if (stream._pendingError) {
```
domenic: Nit: === true

reference-implementation/lib/writable-stream.js
```diff
@@ -737,19 +720,13 @@ class WritableStreamDefaultController {
       () => {
         this._started = true;
 
-        if (stream._state === 'errored') {
-          WritableStreamRejectAbortRequestIfPending(stream);
-        } else {
-          WritableStreamHandleAbortRequestIfPending(stream);
-        }
-
         // This is a no-op if the stream was errored above.
```
domenic: Is this comment still accurate?

reference-implementation/lib/writable-stream.js
```diff
@@ -737,19 +720,13 @@ class WritableStreamDefaultController {
...
         WritableStreamDefaultControllerAdvanceQueueIfNeeded(this);
       },
       r => {
-        assert(stream._state === 'writable' || stream._state === 'errored');
-        WritableStreamDefaultControllerErrorIfNeeded(this, r);
-        WritableStreamRejectAbortRequestIfPending(stream);
+        assert(stream._state === 'writable');
```
domenic: Maybe this assert should be in both branches now?

reference-implementation/lib/writable-stream.js
ricea: It's a promise. Well, actually a record containing a promise and a reason. Great, now I need to work out whether this is a test gap or a no-op.

domenic: `npm run coverage` may help!

ricea: It was a no-op. There was a correct check above. I have removed the redundant check.

Coverage says the real check was being tested, but I can't find the test that does it. Maybe it is a piping one? Anyway, I added an explicit test for double aborting just to make it clear.

reference-implementation/lib/writable-stream.js
ricea: Gone.

reference-implementation/lib/writable-stream.js
ricea: It's no longer meaningful. I have removed it.

reference-implementation/lib/writable-stream.js
ricea: Done.

reference-implementation/lib/writable-stream.js
ricea: Done

reference-implementation/lib/writable-stream.js
```diff
@@ -12,6 +12,9 @@ const ErrorSteps = Symbol('[[ErrorSteps]]');
 class WritableStream {
   constructor(underlyingSink = {}, { size, highWaterMark = 1 } = {}) {
     this._state = 'writable';
+
+    // The error that will be reported by new method calls once the state becomes errored. Only set when [[state]] is
+    // 'errored' or [[pendingError]] is true. May be set to an undefined value.
```
domenic: Mentions "pendingError"

reference-implementation/lib/writable-stream.js
```diff
@@ -42,6 +45,9 @@ class WritableStream {
     // The backpressure signal set by the controller.
     this._backpressure = false;
 
+    // True if an error is pending on the stream.
+    this._pendingError = false;
```
domenic: No longer necessary

reference-implementation/lib/writable-stream.js
```diff
@@ -159,31 +151,74 @@ function WritableStreamAbort(stream, reason) {
...
+    WritableStreamSetPendingError(stream, error);
+  }
+
   return promise;
 }
 
-function WritableStreamError(stream, error) {
-  stream._state = 'errored';
-  stream._storedError = error;
+function WritableStreamSetPendingError(stream, reason) {
```
domenic: Maybe this should be renamed WritableStreamBecomeErroring or similar? Might be fine as-is.

reference-implementation/lib/writable-stream.js
ricea: Fixed.

reference-implementation/lib/writable-stream.js
ricea: Oops. Removed.

reference-implementation/lib/writable-stream.js
ricea: Looking at existing abstract operations, it seemed like "WritableStreamErroring" was the most consistent option, so I went with that.

domenic:
code changes lgtm, although @tyoshino always catches good bugs, so I'd really like his review.

domenic:
Oh, I guess changing controller.error() to no-op if erroring/errored would be a good idea actually (per my comment at https://github.com/w3c/web-platform-tests/commit/f196be0d7c18ae000fd5ae459be8b448eb71bc1b#commitcomment-21575312 )

ricea:
The standard changes are ready for review now.

domenic:
Can you rebase/resolve the test conflict so that a branch snapshot gets generated?

index.bs
```diff
@@ -2854,27 +2846,64 @@ visible through the {{WritableStream}}'s public API.
...
+  1. Set _stream_.[[state]] to `"erroring"`.
+  1. Set _stream_.[[storedError]] to _reason_.
+  1. Let _writer_ be _stream_.[[writer]].
+  1. If _writer_ is not *undefined*, perform !
+     WritableStreamDefaultWriterEnsureReadyPromiseRejected(_writer_, _reason_).
+  1. If ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false* and _controller_.[[started]] is *true*, perform
+     ! WritableStreamFinishError(_stream_).
+</emu-alg>
+
+<h4 id="writable-stream-finish-error" aoid="WritableStreamFinishError" nothrow>WritableStreamFinishError
```
domenic: Maybe the abstract ops should be named "WritableStreamStartErroring" and "WritableStreamFinishErroring"?

index.bs
```diff
@@ -2854,27 +2846,64 @@ visible through the {{WritableStream}}'s public API.
...
+( <var>stream</var> )</h4>
+
+<emu-alg>
+  1. Assert: _stream_.[[state]] is `"erroring"`.
+  1. Assert: ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*.
+  1. Set _stream_.[[state]] to `"errored"`.
+  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
+  1. Let _storedError_ by _stream_.[[storedError]];
```
domenic: Semicolon should be period. "by" should be "be"

index.bs
```diff
@@ -2854,27 +2846,64 @@ visible through the {{WritableStream}}'s public API.
...
+  1. Repeat for each _writeRequest_ that is an element of _stream_.[[writeRequests]],
+    1. <a>Reject</a> _writeRequest_ with _storedError_.
+  1. Set _stream_.[[writeRequest]] to an empty List.
+  1. if _stream_.[[pendingAbortRequest]] is *undefined*,
+    1. Perform !  WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
+    1. Return.
+  1. Let _abortRequest_ by _stream_.[[pendingAbortRequest]].
```
domenic: "by" should be "be". Might be worth running a regex for "Let ... by".

index.bs
```diff
@@ -2854,27 +2846,64 @@ visible through the {{WritableStream}}'s public API.
...
+  1. Set _stream_.[[pendingAbortRequest]] to *undefined*.
+  1. Let _promise_ be stream.[[writableStreamController]].[[AbortSteps]](_abortRequest_.[[reason]]).
```
domenic: Missing ! (although I admit it's unclear whether these internal methods return completion values).

index.bs
```diff
@@ -2854,27 +2846,64 @@ visible through the {{WritableStream}}'s public API.
...
+    1. <a>Upon fulfillment</a> of _promise_, <a>resolve</a> _abortRequest_.[[promise]] with *undefined*.
```
domenic: We're a bit inconsistent but we seem to lean toward not indenting "Upon fulfillment" and "Upon rejection" steps.

index.bs
```diff
@@ -3303,16 +3285,27 @@ nothrow>WritableStreamDefaultWriterCloseWithErrorPropagation ( <var>writer</var>
...
      <a>a promise resolved with</a> *undefined*.
   1. If _state_ is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
-  1. Assert: _state_ is `"writable"`.
+  1. Assert: _state_ is `"writable"` or `"erroring"`.
   1. Return ! WritableStreamDefaultWriterClose(_writer_).
 </emu-alg>
 
+<h4 id="writable-stream-default-writer-ensure-closed-promise-rejected"
+aoid="WritableStreamDefaultWriterEnsureClosedPromiseRejected"
+nothrow>WritableStreamDefaultWriterEnsureClosedPromiseRejected( <var>writer</var>, <var>error</var> )</h4>
```
domenic: It occurs to me we could abstract this out into a utility abstract op like EnsureStatePromiseRejected(promise, error). ("State" included to justify marking it as handled.)

Not sure if worth it.

index.bs
```diff
@@ -3582,10 +3568,13 @@ aoid="WritableStreamDefaultControllerAdvanceQueueIfNeeded" nothrow>WritableStrea
 
 <emu-alg>
   1. Let _stream_ be _controller_.[[controlledWritableStream]].
+  1. if _controller_.[[started]] is *false*, return.
```
domenic: Capitalize

domenic:
I am now confused about the relationship between stream.[[closeRequest]] and writer.[[closedPromise]]. In particular it seems like close() and closed are not returning the same promise anymore---close() returns closeRequest, whereas closed returns closedPromise?

domenic:
WritableStreamFinishAbort does not appear to be under the right heading based on the text

> The following abstract operations, unlike most in this specification, are meant to be generally useful by other specifications, instead of just being part of the implementation of this spec’s classes.

ricea:
I haven't checked the blame history, but I assume that closeRequest and closedPromise parted ways when getWriter() was added.

I simply failed to delete WritableStreamFinishAbort from the standard text. Now I need to go back and see what else I failed to delete.

index.bs
ricea: Makes sense. One that that bugs me is that StartErroring conditionally calls FinishErroring directly. This avoids duplicating the condition in 2 of the callers, and the condition is always true in WritableStreamDealWithRejection so it seems like the right thing, but I still find it confusing.

index.bs
ricea: Done.

index.bs
ricea: I found one of these myself which means I had it wrong in three places. 'y' and 'e' aren't even close to each other on the keyboard. Very odd.

index.bs
ricea: It makes sense to follow the ECMASCRIPT convention here. As long as the reader understands the "?" and "!" notation, they should be able to make sense of it.

tyoshino:
> I haven't checked the blame history, but I assume that closeRequest and closedPromise parted ways when getWriter() was added.

Right. It's 5a2753464bd6e34e3bc7a57fcaac24ea5153fbb6.

https://streams.spec.whatwg.org/commit-snapshots/5a2753464bd6e34e3bc7a57fcaac24ea5153fbb6/#writable-stream-default-writer-close
https://streams.spec.whatwg.org/commit-snapshots/8c89f6f54ed97b2bd01cf7881880ec713d4c317f/#ws-close

tyoshino:
I remember that we discussed long time ago a principle whether methods should return a new promise. But I'm not sure if this is related. I no longer remember the motivation. I guess making close() return `[[closedPromise]]` would also work.

reference-implementation/lib/writable-stream.js
```diff
@@ -204,44 +173,83 @@ function WritableStreamAddWriteRequest(stream) {
...
+  stream._state = 'errored';
+  stream._writableStreamController[ErrorSteps]();
+
+  const storedError = stream._storedError;
+  for (const writeRequest of stream._writeRequests) {
+    writeRequest._reject(storedError);
+  }
+  stream._writeRequests = [];
+
+  if (stream._pendingAbortRequest === undefined) {
```
tyoshino: Could you please update `'controller.error(), writer.abort() while there is an in-flight write, and then finish the write'` test to check that `sink.abort()` is not called for this (error() then abort()) sequence?

controller.error() calls WritableStreamErroring() to change the state to erroring. WritableStreamAbort() skips WritableStreamErroring() when the state is erroring, but it does set _pendingAbortRequest. It should be rejected (right?) when controller.error() happened before writer.abort(), but it looks it would get processed by the code below.

index.bs
ricea: It would be ugly in the reference implementation because of the way the state promises are manipulated by a set of functions there.

index.bs
ricea: Done.

index.bs
ricea: Fixed here and in [[StartSteps]].

reference-implementation/lib/writable-stream.js
ricea: Good catch. It was indeed broken. I have fixed it. PTAL.

ricea:
Remaining jobs for me to do: 

- [ ] Double-check the spec changes against the reference implementation changes.
- [ ] Write a more detailed commit message describing the semantic changes.

ricea:
Okay, ready for final reviews now.

reference-implementation/lib/writable-stream.js
```diff
@@ -339,41 +341,23 @@ function WritableStreamMarkFirstWriteRequestInFlight(stream) {
...
+  const writer = stream._writer;
+  if (writer !== undefined) {
+    defaultWriterClosedPromiseReject(writer, stream._storedError);
+    writer._closedPromise.catch(() => {});
+  }
 }
 
 function WritableStreamUpdateBackpressure(stream, backpressure) {
-  assert(stream._state === 'writable');
+  assert(stream._state === 'writable' || stream._state === 'erroring');
```
tyoshino: When does this get called with state set to erroring?

reference-implementation/lib/writable-stream.js
```diff
@@ -551,11 +535,8 @@ function WritableStreamDefaultWriterClose(writer) {
     return Promise.reject(new TypeError(
       `The stream (in ${state} state) is not in the writable state and cannot be closed`));
   }
-  if (stream._pendingAbortRequest !== undefined) {
-    return Promise.reject(new TypeError('Requested to abort'));
-  }
 
-  assert(state === 'writable');
+  assert(state === 'writable' || state === 'erroring');
```
tyoshino: Hmm, what was the rationale to have close() succeed while write() fails when the state is erroring?

reference-implementation/lib/writable-stream.js
ricea: Never. Fixed the assert.

reference-implementation/lib/writable-stream.js
ricea: writer.close() is useful to wait for the sink to become quiescent. You can also do this with writer.closed, but that would mean writing code like

```javascript
return writer.close().catch(() => writer.closed);
```

Also I want this to just work without the developer having to think about it. If there is a developer error like calling close() twice on the same stream then close() will reject immediately, but otherwise it consistently waits for no underlying sink methods to be running.

write() never provides any such guarantee, so rejecting immediately when we know there is no hope of the write() working is the best behaviour.

tyoshino: OK. Makes sense.

domenic:
Commit message nit: "un" should be "an".

Also let's please remember to delete the "(#721)" from the commit message title before merging. Gotta look good [on the twitters](https://twitter.com/streamsstandard) :).

LGTM!

ricea:
I find it convenient having the pull request linked from the commit title. Why does this look bad on the twitters?

domenic:
The pull request is always linked from the commit page, FWIW.

The problem is it references a mysterious number with no context for anyone consuming the commit messages outside the GitHub UI.

ricea:
Rebased, squashed and updated web-platform-tests submodule to the merged commit.

PR #722, opened 2017-04-04
Update for changes to HTML's structured cloning/transferring

### LINKED ISSUE #701: Update for HTML structured clone changes
https://github.com/whatwg/html/commit/97d644c97335956610a31e8ad98d1a388c063e84 changed structured cloning significantly. We need to:

- Not use StructuredClone(), but instead StructuredDeserialize(StructuredSerialize())
- Roll our own Transfer() for array buffers, as Transfer() is no longer defined in HTML.

domenic:
Closes #701.

---

This is basically editorial so if people are busy no need to review this; I'll merge it in a few days. But as always review is nice.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -331,7 +331,7 @@ copies.
 
   An important thing to note here is that the final <code>buffer</code> value is different from the
   <code>startingAB</code>, but it (and all intermediate buffers) shares the same backing memory allocation. At each
-  step, the buffer is <a abstract-op lt="Transfer">transferred</a> to a new {{ArrayBuffer}} object. The
+  step, the buffer is <a href="#transfer-array-buffer">transferred</a> to a new {{ArrayBuffer}} object. The
   <code>newView</code> is a new {{Uint8Array}}, with that {{ArrayBuffer}} object as its <code>buffer</code> property,
   the offset that bytes were written to as its <code>byteOffset</code> property, and the number of bytes that were
   written as its <code>byteLength</code> property.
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -797,11 +795,14 @@ readable stream is <a>locked to a reader</a>.
...
 stream">tee</a> a given readable stream.
 
-The second argument, <var>cloneForBranch2</var>, governs whether or not the data from the original stream will be <a
-abstract-op lt="StructuredClone">structured cloned</a> before appearing in the second of the returned branches. This is
+The second argument, <var>cloneForBranch2</var>, governs whether or not the data from the original stream will be cloned
+(using HTML's <a>serializable objects</a> framework) before appearing in the second of the returned branches. This is
 useful for scenarios where both branches are to be consumed in such a way that they might otherwise interfere with each
-other, such as by <a abstract-op lt="Transfer">transfering</a> their <a>chunks</a>. However, it does introduce a
-noticable asymmetry between the two branches. [[!HTML]]
+other, such as by <a href="#transfer-array-buffer">transfering</a> their <a>chunks</a>. However, it does introduce a
```
isonmad: This defines transfering by linking to TransferArrayBuffer, which asserts its argument has an [[ArrayBufferData]] internal slot, but this is about cloning arbitrary objects.

ricea: Should be "transferring"?

reference-implementation/lib/helpers.js
```diff
@@ -101,7 +101,7 @@ exports.PromiseInvokeOrPerformFallback = (O, P, args, F, argsF) => {
 };
 
 // Not implemented correctly
-exports.SameRealmTransfer = O => O;
+exports.TransferArrayBuffer = O => O;
```
ricea: How about using O => O.slice(0) instead? Or has this already been discussed?

ricea:
lgtm

index.bs
domenic: This sub-clause of the sentence is "such as", so yeah, it's just giving an example of how branches could interfere with each other.

reference-implementation/lib/helpers.js
domenic: In general this could probably be a lot better; that's a reasonable start.

index.bs
domenic: I'll link to HTML's "transferable objects" instead

PR #725, opened 2017-04-05
Update non-normative description of writer properties and methods

domenic:


domenic:
I guess .closed and close() no longer return the same promise, so I should fix that while I'm there.

ricea:
lgtm

PR #726, opened 2017-04-06
Fix pipeTo() to ensure all read chunks are written

### LINKED ISSUE #644: pipeTo() algorithm shuts down too early on readable close()?

domenic:
Previously, the spec wording allowed for immediate shutdown if the readable stream became closed, even if there were outstanding writes which did not yet get written. This fixes that, by requiring any already-read chunks to be written when possible.

The reference implementation mostly did not suffer from this, but did in the case where a read completes while a write is ongoing. This was fixed by making its "wait for writes to finish" functionality more robust.

Fixes #644.

Tests: https://github.com/w3c/web-platform-tests/pull/5270

### DIFF
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -188,13 +188,10 @@ class ReadableStream {
       });
 
       function waitForWritesToFinish() {
-        if (dest._state === 'writable' && WritableStreamCloseQueuedOrInFlight(dest) === false) {
-          // Another write may have started while we were waiting on this currentWrite, so we have to be sure to wait
-          // for that too.
-          const oldCurrentWrite = currentWrite;
-          return currentWrite.then(() => oldCurrentWrite !== currentWrite ? waitForWritesToFinish() : undefined);
-        }
-        return Promise.resolve();
+        // Another write may have started while we were waiting on this currentWrite, so we have to be sure to wait
+        // for that too.
+        const oldCurrentWrite = currentWrite;
+        return currentWrite.then(() => oldCurrentWrite !== currentWrite ? waitForWritesToFinish() : undefined);
       }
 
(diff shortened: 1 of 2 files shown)


domenic:
Note: the two steps (write any read chunks; wait for writes to finish) don't apply in all shutdown cases. In particular they only apply when the source errors or closes. This leads to some duplication. However the alternate approach of only putting those steps underneath their corresponding conditions ("Errors must be propagated forward" and "Closing must be propagated forward") meant you also needed to duplicate the shuttingDown set/check under those conditions. So this version seemed cleanest to me.

reference-implementation/lib/readable-stream.js
```diff
@@ -187,6 +187,16 @@ class ReadableStream {
         rethrowAssertionErrorRejection(err);
       });
 
+      function waitForWritesToFinish() {
+        if (dest._state === 'writable' && WritableStreamCloseQueuedOrInFlight(dest) === false) {
+          // Another write may have started while we were waiting on this currentWrite, so we have to be sure to wait
+          // for that too.
+          const before = currentWrite;
```
ricea: I find the name `before` a little unclear. I don't have strong feelings about it, but maybe `oldCurrentWrite` would be better?

ricea:
lgtm

ricea:
lgtm e22b3d8

PR #730, opened 2017-04-10
Use Promise.prototype.then brand check to detect real Promise

ricea:
ReadableStream pipeTo() was using instanceof to distinguish Promises from
non-promises. However, this can be fooled by Object.create(Promise). Use
the brand-check performed by Promise.prototype.then() for the check instead.

The test for this is in https://github.com/w3c/web-platform-tests/pull/5519.

### DIFF
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -85,8 +85,7 @@ class ReadableStream {
   pipeThrough({ writable, readable }, options) {
     const promise = this.pipeTo(writable, options);
 
-    // Can't actually do a promise brand check here, so instanceof is close enough.
-    if (typeIsObject(promise) && promise instanceof Promise) {
+    if (typeIsObject(promise) && hasPromiseIsHandledInternalSlot(promise)) {
       promise.catch(() => {});
     }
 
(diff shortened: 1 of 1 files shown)


reference-implementation/lib/readable-stream.js
```diff
@@ -1935,3 +1934,16 @@ function byteStreamControllerBrandCheckException(name) {
...
     `ReadableByteStreamController.prototype.${name} can only be used on a ReadableByteStreamController`);
 }
+
+// Helper function for ReadableStream pipeThrough
+
+function hasPromiseIsHandledInternalSlot(promise) {
+  try {
+    // This relies on the brand-check that is enforced by Promise.prototype.then(). As with the rest of the reference
+    // implementation, it doesn't attempt to do the right thing if someone has modified the global environment.
+    Promise.prototype.then.call(promise, undefined, () => {});
```
ricea: This is not side-effect free; it marks the promise as handled itself. Calling `then(undefined, undefined)` somehow prevents the promise from subsequently being marked as handled. I don't know whether that is standard behaviour or a bug in V8.

In the specific case where we are using it, this doesn't matter, because when this function runs true we always mark the promise as handled anyway.

domenic: It's standard behavior, so using undefined, undefined is going to be the better way to go.

ricea: Sorry, I did not explain it clearly. If we use undefined, undefined then test 'Piping through a transform errored on the writable end does not cause an unhandled promise rejection' fails.

domenic: Oh, I see, because it generates a new promise that is itself marked as unhandled, but that promise is never returned back for you to handle.

In that case IMO the best resolution is to rename this helper something like checkPromiseIsHandledInternalSlotAndSetItToTrueIfItExists, or something similar but less awkward, to make it clear it's effectively doing both.

ricea: I tried something different: I named the function ifIsObjectAndHasAPromiseIsHandledInternalSlotSetPromiseIsHandledToTrue and made it implement that whole line from the standard. PTAL.

PR #744, opened 2017-05-15
Allow aborting an ongoing pipe operation using AbortSignals

### LINKED ISSUE #446: pipeTo should return a cancelable promise
If I wanted to create:

``` js
const stream = merge([stream1, stream2, stream3]);
```

Being able to pipe each stream to an identity stream would make this easy and move a lot of the work out of JS, but the missing bit is being able to react to cancellation. If `stream` is cancelled while piping `stream1`, I want to know about that so I can cancel `stream2` and `stream3`.

domenic:
Closes #446.

---

Blocked on https://github.com/whatwg/dom/pull/437 finishing up, but it was important to write this up as a proof of concept; it shows that the AbortSignal design is good for our needs.

There's a bit of a namespace collision in how streams has chosen writer.abort() to mean one thing which is a bit different from signal.abort() on a pipe. Maybe we should put a bit more about that in the developer-facing domintro boxes. But I don't think it's avoidable.

/cc @annevk

***
<a href="https://whatpr.org/streams/744.html" title="Last updated on Nov 7, 2018, 9:23 PM GMT (8e0aaed)">Preview</a> | <a href="https://whatpr.org/streams/744/3197c7e...8e0aaed.html" title="Last updated on Nov 7, 2018, 9:23 PM GMT (8e0aaed)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -747,8 +747,8 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
   1. If ! IsWritableStream(_dest_) is *false*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. Set _preventClose_ to ! ToBoolean(_preventClose_), set _preventAbort_ to ! ToBoolean(_preventAbort_), and set
      _preventCancel_ to ! ToBoolean(_preventCancel_).
-  1. If _signal_ is not *undefined*, and _signal_ is not an instance of the {{AbortSignal}} interface, return <a>a
-     promise rejected with</a> a *TypeError* exception.
+  1. If _signal_ is not *undefined*, and _signal_ is not an instance of the `<a idl>AbortSignal</a>` interface, return
+     <a>a promise rejected with</a> a *TypeError* exception.
   1. If ! IsReadableStreamLocked(*this*) is *true*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. If ! IsWritableStreamLocked(_dest_) is *true*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. If ! IsReadableByteStreamController(*this*.[[readableStreamController]]) is *true*, let _reader_ be either !
@@ -758,6 +758,20 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
   1. Let _writer_ be ! AcquireWritableStreamDefaultWriter(_dest_).
   1. Let _shuttingDown_ be *false*.
   1. Let _promise_ be <a>a new promise</a>.
+  1. If _signal_ is not *undefined*,
+    1. Let _abortAlgorithm_ be the following steps:
+      1. Let _error_ be a new "`<a idl>AbortError</a>`" `<a idl>DOMException</a>`.
+      1. Let _actions_ be an empty <a>ordered set</a>.
+      1. If _preventAbort_ is *false*, <a for="set">append</a> the following action to _actions_:
+        1. If _dest_.[[_state_]] is `"writable"`, return ! WritableStreamAbort(_dest_, _error_).
+        1. Otherwise, return <a>a promise resolved with</a> *undefined*.
+      1. If _preventCancel_ is *false*, <a for="set">append</a> the following action action to _actions_:
+        1. If *this*.[[_state_]] is `"readable"`, return ! ReadableStreamCancel(*this*, _error_).
+        1. Otherwise, return <a>a promise resolved with</a> *undefined*.
+      1. <a href="#rs-pipeTo-shutdown-with-action">Shutdown with an action</a> consisting of <a>waiting for all</a>
+         of the actions in _actions_, and with _error_.
+    1. If _signal_'s <a for=AbortSignal>aborted flag</a> is set, perform _abortAlgorithm_ and return _promise_.
+    1. <a for="AbortSignal">Add</a> _abortAlgorithm_ to _signal_.
   1. <a>In parallel</a> <span class="XXX">but not really; see <a
      href="https://github.com/whatwg/streams/issues/905">#905</a></span>, using _reader_ and _writer_, read all
      <a>chunks</a> from *this* and write them to _dest_. Due to the locking provided by the reader and writer, the exact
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -700,6 +718,8 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
        an error _error_, which means to perform the following steps:
        1. Perform ! WritableStreamDefaultWriterRelease(_writer_).
        1. Perform ! ReadableStreamReaderGenericRelease(_reader_).
+       1. If _signal_ is not *undefined*, remove the above algorithm from _signal_'s <a for="AbortSignal">abort
+          algorithms</a>.
```
annevk: Maybe stuff the algorithm in a variable?

ricea:
Text changes roughly lgtm. I'm not sure about the name "signal" for the option. I'm asking myself "would I remember that?" and I don't know the answer. Generally the AbortController / AbortSignal design feels like it has a lot of cognitive overhead but I haven't come up with anything better.

Compared to that, having two slightly different concepts called "abort" doesn't bother me.

annevk:
The main reason we called it "signal" is that we might want to signal things other than abort too. And for Fetch at least the idea was those signals would all be on the same object.

domenic:
@bakulf @tschneidereit any chance on implementer support from Firefox on this? Although this is less important than fetch() cancelation, it's also simpler, so we were thinking of using it as the first motivating case for AbortSignal in Blink at least.

I realize Firefox doesn't have writable streams / pipeTo yet, so it might be hard for you to judge fully, but I'm hopeful the spec is straightforward enough to get multi-implementer support anyway :)

annevk:
Gecko implementation bug for abort things in general: https://bugzilla.mozilla.org/show_bug.cgi?id=1378342. I suspect we want this eventually, but it's highly likely we'll get `fetch()` done first.

tschneidereit:
@domenic apologies for the late reply. I still haven't implemented `pipeTo`, but I do have `WritableStream` implemented (but not yet published, @bakulf and me are focusing on RS right now).

The changes here look quite straight-forward to me, though, so I'm happy to give support for them, with the condition that @annevk is happy with the implications for `fetch`.

annevk:
As I understand it there won't be any implications for `fetch()` as it only uses `ReadableStream`. Maybe I'm missing something though.

tschneidereit:
> As I understand it there won't be any implications for fetch() as it only uses ReadableStream. Maybe I'm missing something though.

Sorry, that was confusing. What I meant was "as long as this satisfies use cases of specs that use streams, such as fetch, I'm happy to give support." That seems to be the case, but since I'm not familiar enough with the fetch spec to be quite certain on my own I wanted to flag this one more time.

domenic:
Yeah, fetch cancelability should be pretty independent of this. Sounds like we're good, once the DOM PR is merged!

domenic:
Yep, just gotta write some tests.

ricea:
Still lgtm. Do you want me to try adding it to the reference implementation and write some tests?

It looks like jsdom already has AbortController, so it should be pretty easy.

annevk:
@ricea I strongly suspect that'd be much appreciated!

domenic:
Yes please :). Sorry for letting this languish.

index.bs
```diff
@@ -733,13 +734,21 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
...
+  the respective options <code>preventCancel</code> or <code>preventAbort</code> are set.
+
 </div>
 
 <emu-alg>
   1. If ! IsReadableStream(*this*) is *false*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. If ! IsWritableStream(_dest_) is *false*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. Set _preventClose_ to ! ToBoolean(_preventClose_), set _preventAbort_ to ! ToBoolean(_preventAbort_), and set
      _preventCancel_ to ! ToBoolean(_preventCancel_).
+  1. If _signal_ is not *undefined*, and _signal_ is not an instance of the {{AbortSignal}} interface, return <a>a
```
ricea: I'm concerned that this condition may not be strong enough to guarantee we can execute the [add](https://dom.spec.whatwg.org/#abortsignal-add) algorithm.

Maybe we could use [language similar to WebIDL](https://heycam.github.io/webidl/#es-interface), for example "and _signal_ is not a platform object or does not implement _AbortSignal_, then ..."

index.bs
```diff
@@ -791,6 +800,16 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
...
          1. Otherwise, <a href="#rs-pipeTo-shutdown">shutdown</a> with _destClosed_.
+     * <strong>Abort signals must stop activity:</strong> if _signal_ is not *undefined*, the following algorithm
+       _abortAlgorithm_ must be <a for="AbortSignal">added</a> to _signal_:
+         1. Let _error_ be a new "`<a idl>AbortError</a>`" `<a idl>DOMException</a>`.
+         1. Let _actions_ be an empty <a>ordered set</a>.
+         1. If _preventAbort_ is *false*, <a for="set">append</a> the action of performing !
+            WritableStreamAbort(_dest_, _error_) to _actions_.
+         1. If _preventCancel_ is *false*, <a for="set">append</a> the action of performing !
+            ReadableStreamCancel(*this*, _error_) to _actions_.
+         1. <a href="#rs-pipeTo-shutdown-with-action">Shutdown with an action</a> consisting of <a>waiting for all</a>
```
ricea: The [waiting for all](https://www.w3.org/2001/tag/doc/promises-guide#waiting-for-all) algorithm is problematic because it uses `Promise.all()`, which is very sensitive to changes to the global object.

We need to do something else, or change the definition of "waiting for all" in the promises guide.

annevk: I think the promises guide is known to be broken and needs to be uplifted into IDL. But nobody has done the work...

index.bs
annevk: The language above is similar to what we use elsewhere though (for "brand" checks), but arguably none of that is precise enough (there's an issue against IDL to define a [[PlatformBrand]] internal slot we could use to formally define these things).

domenic: Yeah, I think this is the right thing for now.

index.bs
annevk: As far as I'm concerned we shouldn't block this on that, since there's presumably other dependencies that are similarly broken already? (I don't find it particularly great, but we seem to have no trouble adding dependencies on broken/unclear primitives all over the platform.)

ricea:
I noticed a couple more issues while writing tests:
* It doesn't work if the AbortSignal is already signalled when pipeTo() is called.
* What happens when there are other shutdown conditions is not defined; eg. calling `rs.pipeTo(ws, { signal })` where `rs` is already closed and `signal` is already signalled.

If we figure out the answer to the second question then I can solve the first.

ricea:
From the point of view of the standard text (and probably the implementation), it's neater to check the signal at the beginning before looking at the state of the streams.

I think this behaviour is logical, so unless anyone objects, I'll go ahead and implement it.

annevk:
That is what https://fetch.spec.whatwg.org/#dom-global-fetch does too, if I understand things correctly. We might need to document that pattern somehow more clearly so it's adopted by everyone consistently.

ricea:
I probably should have noticed this before, but when pipeTo() is aborted, the promise is fulfilled with an AbortError. I think it should be rejected instead, to match the behaviour of fetch().

ricea:
I think this is ready to land. PTAL.

domenic:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=902939. My understanding is other browsers don't have a pipeTo() implementation yet, so I didn't file bugs there, but let me know if folks want them.

Issue #746, opened 2017-05-22
Reference to CopyDataBlockBytes


rombel:
In ReadableByteStreamControllerFillPullIntoDescriptorFromQueue [1], CopyDataBlockBytes [2] is referenced at step 10.d:

> Perform ! CopyDataBlockBytes(headOfQueue.[[buffer]].[[ArrayBufferData]], headOfQueue.[[byteOffset]], pullIntoDescriptor.[[buffer]].[[ArrayBufferData]], destStart, bytesToCopy)

According to [2], operation CopyDataBlockBytes is as follows: 

> CopyDataBlockBytes(toBlock, toIndex, fromBlock, fromIndex, count)

Therefore, IIUC, it seems that step 10.d should be:

> Perform ! CopyDataBlockBytes(pullIntoDescriptor.[[buffer]].[[ArrayBufferData]], destStart, headOfQueue.[[buffer]].[[ArrayBufferData]], headOfQueue.[[byteOffset]], bytesToCopy)

[1] https://streams.spec.whatwg.org/#readable-byte-stream-controller-fill-pull-into-descriptor-from-queue 
[2] https://tc39.github.io/ecma262/#sec-copydatablockbytes

domenic:
Great catch, thank you! Are you up for submitting a pull request?

rombel:
Sure! I've just submitted one #747.

tyoshino:
Fixed by e286449d8fa1b0f9b368eb288512cd751eb51a7f. Thank you!

PR #747, opened 2017-05-23
Fix arguments order in CopyDataBlockBytes

rombel:
As described in #746, arguments of CopyDataBlockBytes are not in the right order. This commit fixes that.


tyoshino:
Confirmed that the ArrayBufferCopy function used in the reference implementation is already using the order for its argument list.

PR #749, opened 2017-06-30
Add demos of stream-to-DOM and stream-to-HTML

ricea:
Add three demos:
* "Append child WritableStream" demos a WritableStream which appends
elements to a DOM node.
* "Streaming element" demos a custom element which exposes a
WritableStream that can be used to write HTML into the element.
* "Streaming element with backpressure" is similar, but applies
backpressure so that the browser remains interactive as the data arrives.

These also demonstrate the streams-compatible TextDecoder interface via
a polyfill, and utilise a polyfill for TransformStream which is not
currently shipping.

Videos of the demos are included so that they can be viewed on browsers
that don't yet have WritableStream implementations.


ricea:
I chose to put the demos directly in the demos/ directory rather than creating a subdirectory. Either approach would work, but I think that since the resources are nested it's not too untidy even if we add more demos later.

Issue #752, opened 2017-07-21
ReadableByteStreamControllerPullInto should always transfer the given buffer


tschneidereit:
Currently, the buffer is not transferred if the bodies of steps 7 or 8 (sub-steps a or b) are hit.

For step 8, that's maybe not too bad, because the buffer's contents are copied and the buffer itself isn't exposed anyway. For step 7 we do expose the non-transferred buffer as the chunk value's underlying buffer.

In any case, it seems inconsistent for whether the buffer will be transferred or not to depend on the exact results of the pullInto operation.

Is there any reason not to just transfer the buffer before step 5?

cc @tyoshino @domenic

Issue #751, opened 2017-07-20
Callsites of TransferArrayBuffer don't all ensure the buffer isn't detached


tschneidereit:
`TransferArrayBuffer` asserts that the given buffer isn't detached, but its callsites don't ensure that this holds. In particular, [`ReadableByteStreamControllerEnqueue`](https://streams.spec.whatwg.org/#readable-byte-stream-controller-enqueue) passes in the chunk given to `enqueue` without checking if it's detached.

The best fix for this case might be to add
```
Step 6. If IsDetachedBuffer(chunk.[[ViewedArrayBuffer]]) is true, throw a TypeError exception.
```
to [`3.10.4.4. enqueue`](https://streams.spec.whatwg.org/#rbs-controller-enqueue).

I haven't fully verified the other callsites, but they might be ok because they deal with buffers that should have been verified earlier.

PR #753, opened 2017-07-21
A couple fixes for ArrayBuffer transferring

tschneidereit:
This implements what I proposed in #752: it moves the transfer to the beginning of `ReadableByteStreamControllerPullInto` and removes the other transfers.

All tests continue passing.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -101,7 +101,19 @@ exports.PromiseInvokeOrPerformFallback = (O, P, args, F, argsF) => {
 };
 
 // Not implemented correctly
-exports.TransferArrayBuffer = O => O.slice();
+exports.TransferArrayBuffer = O => {
+  const transferredIshVersion = O.slice();
+
+  // This is specifically to fool tests that test "is transferred" by taking a non-zero-length
+  // ArrayBuffer and checking if its byteLength starts returning 0.
+  Object.defineProperty(O, 'byteLength', {
+    get() {
+      return 0;
+    }
+  });
+
+  return transferredIshVersion;
+};
 
 exports.ValidateAndNormalizeHighWaterMark = highWaterMark => {
   highWaterMark = Number(highWaterMark);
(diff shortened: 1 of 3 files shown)


domenic:
THanks! Let me find a way to make this testable...

tschneidereit:
> THanks! Let me find a way to make this testable...
>

Yeah, that's why I didn't provide tests :) I guess the JS implementation
could postMessage the buffer to another global and back?

domenic:
I was thinking something hacky enough to pass whatever test we come up with. E.g. if we test that accessing [0] throws, we Object.defineProperty it to throw or something. (Not sure if it's configurable...)

Either that or I finally come up with a way to commit expected test results to this repository so that the reference implementation can fail them without failing our CI build.

domenic:
@tschneidereit sorry for the massive delay. I've updated the reference implementation with a hack like the one I described, and it's enough to pass the new tests at https://github.com/w3c/web-platform-tests/pull/7359 . Your review would be appreciated (although I think it'll be OK to accept one of the other editors' reviews too, since we're just polishing your existing proposal).

reference-implementation/lib/helpers.js
```diff
@@ -101,7 +101,19 @@ exports.PromiseInvokeOrPerformFallback = (O, P, args, F, argsF) => {
...
+exports.TransferArrayBuffer = O => {
+  const transferredIshVersion = O.slice();
+
+  // This is specifically to fool tests that test "is transferred" by taking a non-zero-length
+  // ArrayBuffer and checking if its byteLength starts returning 0.
+  Object.defineProperty(O, "byteLength", {
+    get() {
+      return 0;
+    }
+  });
```
tschneidereit: I think I'm a bit in awe with this hack :)

domenic: Wait until you see what I'm going to tack onto this PR for solving #751...

domenic:
I added a commit for fixing #751 here, and a commit for tests to https://github.com/w3c/web-platform-tests/pull/7359. (I didn't do separate PRs since they built on each other a bit.) Review appreciated!

tschneidereit:
Yes, yes I see what you meant earlier ;)

This looks like it gets the job done, though, so r+.

PR #759, opened 2017-07-28
Throw if undefined streams are passed to pipeThrough

### LINKED ISSUE #756: pipeThrough(): complain about undefined args?
Currently, according to the standard, `rs.pipeThrough({})` should do nothing. More precisely, `pipeTo()` will return a rejection which `pipeThrough()` will mark as handled and ignore.

This makes it hard to spot mistakes: if you write `rs.pipeThrough(writable, readable)` (ie. you miss off the `{}`) it will do nothing at all.

See https://github.com/whatwg/streams/issues/754 which brought this usability issue to my attention.

I am currently thinking that in order to make it easier to use we should special-case undefined values and throw an exception. I think we don't need to cater to values of `this` for which `pipeTo(undefined)` is a useful expression. Similarly, while returning `undefined` when `readable` wasn't specified is what the user asked for, it's unlikely to be what they meant.

This change would be pretty backwards compatible: we'd be changing behaviour that is useless anyway, and both Chrome and Safari currently throw a TypeError for an undefined `writable` (in Chrome's case this was a mistake). Useful uses of `pipeThrough()` would be interoperable between implementations with and without the change.

ricea:
pipeThrough() is intentionally flexible about what arguments it will
accept, but this can make it hard to diagnose when it's called
incorrectly. To catch common mistakes, throw an exception if the
readable or writable arguments are undefined.

Closes #756.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -570,9 +570,10 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
 <emu-alg>
   1. If _writable_ is *undefined* or _readable_ is *undefined*, throw a *TypeError* exception indicating that undefined
      values are not permitted.
-     <div class="note"><emu-val>undefined</emu-val> is special-cased here to make it easier to debug errors such as
-     <code>rs.pipeThrough(writable, readable)</code> which otherwise would silently do nothing. pipeThrough() is
-     otherwise agnostic about the values that it passes through.</div>
+     <p class="note"><emu-val>undefined</emu-val> is special-cased here to make it easier to debug errors such as
+     <code>rs.pipeThrough(writable, readable)</code> which otherwise would silently do nothing. <code><a
+     lt="pipeThrough()" for="ReadableStream">pipeThrough()</a></code> is
+     otherwise agnostic about the values that it passes through.</p>
   1. Let _promise_ be ? Invoke(*this*, `"pipeTo"`, « _writable_, _options_ »).
   1. If Type(_promise_) is Object and _promise_ has a [[PromiseIsHandled]] internal slot, set
      _promise_.[[PromiseIsHandled]] to *true*.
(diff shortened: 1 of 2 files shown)


ricea:
Some tests fail because they call `pipeThrough({})`. My plan is to split the wpt tests in https://github.com/w3c/web-platform-tests/pull/6651 into two parts, one of which makes the tests pass both with and without the new behaviour, and then a second one which actually tests the new behaviour.

index.bs
```diff
@@ -568,6 +568,8 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
 </div>
 
 <emu-alg>
+  1. If _writable_ is *undefined* or _readable_ is *undefined*, throw a *TypeError* exception indicating that undefined
+     values are not permitted.
```
domenic: _readable_ does not exist (use **this**). But that part is redundant with the next line, which will throw anyway. Implementations are free to do an identity check against undefined and throw a different TypeError if they want, since that's unobservable.

Will fix and merge.

domenic: Sorry, I was confused, readable obviously exists. And if `pipeTo` is overridden, it is observably different whether we throw an error for writable or not. Hmm.

ricea:
As it returns a Promise, pipeTo() shouldn't throw. pipeThrough() can throw because it is synchronous.

ricea:
I added a note explaining why the undefined checks are there. I'm not sure the markup is correct.

domenic:
I can take care of the markup tomorrow if you want. I should go to bed now though :).

ricea:
Yes, I will leave it to you. There's no hurry.

{{ReadableStream/pipeThrough()}} doesn't get substituted by bikeshed. Maybe because it's inside an emu-alg section?

domenic:
Ah, yeah, I think you have to use longhands there. I'll make it work.

PR #766, opened 2017-08-21
TransformStream: Make strategy arguments positional

ricea:
Change the calling convention for the TransformStream constructor from
`(transformer, {readableStrategy, writableStrategy})` to `(transformer,
writableStrategy, readableStrategy)`. This makes the calling convention
closer to the other Streams API constructors, and emphasises the
writableStrategy parameter.


ricea:
This implements the proposal from https://github.com/whatwg/streams/issues/190#issuecomment-323655800.

PR #772, opened 2017-08-22
Add brand checks to TransformStream tests

ricea:
Verify calls to IsTransformStream() and IsTransformStreamDefaultController()
abstract operations are correctly applied at public entry points.

Also add tests that the "readable" and "writable" properties pass the
IsReadableStream() and IsWritableStream() checks.

Also imports the latest version of test-utils.js from web-platform-tests and
adds a constructorThrowsForAll() test utility function.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/brand-checks.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/brand-checks.js
@@ -1,0 +1,79 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+  self.importScripts('../resources/test-utils.js');
+}
+
+const TransformStreamDefaultController = getTransformStreamDefaultControllerConstructor();
+
+function getTransformStreamDefaultControllerConstructor() {
+  return realTSDefaultController().constructor;
+}
+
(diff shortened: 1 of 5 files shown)


reference-implementation/to-upstream-wpts/transform-streams/brand-checks.js
```diff
@@ -0,0 +1,79 @@
...
+
+test(() => {
+  getterThrowsForAll(TransformStream.prototype, 'writable',
+                     [fakeTS(), realTSDefaultController(), undefined, null]);
+}, 'TransformStream.prototype.writable enforces a brand check');
+
+test(() => {
+  constructorThrowsForAll(TransformStreamDefaultController,
+                          [fakeTS(), realTS(), realTSDefaultController(), undefined, null]);
+}, 'TransformStreamDefaultConstructor enforces a brand check and doesn\'t permit independant construction');
```
domenic: Nit: "independ*e*nt". (I have trouble with this a lot so I imagine you probably copied this from a typo I made elsewhere.)

ricea: Fixed, thanks.

PR #773, opened 2017-08-22
Add tests for TransformStream backpressure

ricea:
Verify that backpressure is correctly applied when the readable side queue is
full, or when the transform() method blocks.

### DIFF
--- a/reference-implementation/to-upstream-wpts/.eslintrc.json
+++ b/reference-implementation/to-upstream-wpts/.eslintrc.json
@@ -56,7 +56,9 @@
     "flushAsyncEvents": true,
     "methodRejects": true,
     "readableStreamToArray": true,
+    "recordingIdentityTransformStream": true,
     "recordingReadableStream": true,
+    "recordingTransformStream": true,
     "recordingWritableStream": true,
     "sequentialReadableStream": true
   }
--- a/reference-implementation/to-upstream-wpts/.eslintrc.json
+++ b/reference-implementation/to-upstream-wpts/.eslintrc.json
@@ -56,7 +56,6 @@
     "flushAsyncEvents": true,
     "methodRejects": true,
     "readableStreamToArray": true,
-    "recordingIdentityTransformStream": true,
     "recordingReadableStream": true,
     "recordingTransformStream": true,
     "recordingWritableStream": true,
--- a/reference-implementation/to-upstream-wpts/resources/recording-streams.js
+++ b/reference-implementation/to-upstream-wpts/resources/recording-streams.js
@@ -90,3 +90,52 @@ self.recordingWritableStream = (extras = {}, strategy) => {
 
   return stream;
 };
+
+self.recordingTransformStream = (extras = {}, writableStrategy, readableStrategy) => {
+  let controllerToCopyOver;
+  const stream = new TransformStream({
+    start(controller) {
+      controllerToCopyOver = controller;
+
+      if (extras.start) {
+        return extras.start(controller);
+      }
+
+      return undefined;
+    },
+
+    transform(chunk, controller) {
+      stream.events.push('transform', chunk);
+
+      if (extras.transform) {
+        return extras.transform(chunk, controller);
+      }
+
+      return undefined;
+    },
+
+    flush(controller) {
+      stream.events.push('flush');
+
+      if (extras.flush) {
+        return extras.flush(controller);
+      }
+
+      return undefined;
+    }
+  }, writableStrategy, readableStrategy);
+
+  stream.controller = controllerToCopyOver;
+  stream.events = [];
+
+  return stream;
+};
+
+self.recordingIdentityTransformStream = (writableStrategy, readableStrategy) => {
+  const stream = self.recordingTransformStream({
+    transform(chunk, controller) {
+      controller.enqueue(chunk);
+    }
+  }, writableStrategy, readableStrategy);
+  return stream;
+};
--- a/reference-implementation/to-upstream-wpts/resources/recording-streams.js
+++ b/reference-implementation/to-upstream-wpts/resources/recording-streams.js
@@ -111,6 +111,8 @@ self.recordingTransformStream = (extras = {}, writableStrategy, readableStrategy
         return extras.transform(chunk, controller);
       }
 
+      controller.enqueue(chunk);
+
       return undefined;
     },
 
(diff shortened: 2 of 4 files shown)


reference-implementation/to-upstream-wpts/resources/recording-streams.js
```diff
@@ -90,3 +90,52 @@ self.recordingWritableStream = (extras = {}, strategy) => {
...
+  }, writableStrategy, readableStrategy);
+
+  stream.controller = controllerToCopyOver;
+  stream.events = [];
+
+  return stream;
+};
+
+self.recordingIdentityTransformStream = (writableStrategy, readableStrategy) => {
+  const stream = self.recordingTransformStream({
```
domenic: Nit: I'd just `return` this; no need to create a variable

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js', '../resources/recording-streams.js', '../resources/test-utils.js');
```
domenic: I'm not sure the wrapper-generator is smart enough to deal with the multi-arg form of importScripts. I think it might be better to be consistent and go with the single-arg form anyway.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
```diff
@@ -0,0 +1,112 @@
...
+
+promise_test(() => {
+  const ts = recordingIdentityTransformStream();
+  const writer = ts.writable.getWriter();
+  writer.write('a');
+  writer.write('b');
+  return delay(0).then(() => {
+    assert_array_equals(ts.events, ['transform', 'a'], 'transform should be called once');
+  });
+}, 'transform() should be called once with default identity transform');
```
domenic: I don't think this test name is really reflecting what is going on. Basically this is testing that transform() is called synchronously once and the second chunk gets delayed async. Right?

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
```diff
@@ -0,0 +1,112 @@
...
+
+promise_test(() => {
+  // Without a transform() implementation, recordingTransformStream() never enqueues anything.
```
domenic: Let's expand this, as for a few seconds I thought "doesn't that mean the pipe will just get clogged, so there _will_ be backpressure?"

I'd go with

> That is, we're just throwing away chunks so the readable side is never full enough to exert backpressure and stop transform() from being called.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
```diff
@@ -0,0 +1,112 @@
...
+  writer.write('a').then(() => events.push('a'));
+  writer.write('b').then(() => events.push('b'));
+  writer.close().then(() => events.push('closed'));
+  return flushAsyncEvents().then(() => {
+    assert_array_equals(events, [], 'no writes should have resolved yet');
+    return reader.read();
+  }).then(({ value, done }) => {
+    assert_false(done, 'done should not be true');
+    assert_equals('a', value, 'value should be "a"');
+    return delay(0);
```
domenic: I still don't know how to distinguish flushAsyncEvents() vs. delay(0) :-/. Could you remind me?

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
```diff
@@ -0,0 +1,112 @@
...
+  }).then(() => {
+    assert_array_equals(events, ['a', 'b', 'closed'], 'the second write and close should be resolved');
+    return reader.read();
+  }).then(({ done }) => {
+    assert_true(done, 'done should be true');
+  });
+}, 'writes should not resolve until backpressure clears');
+
+promise_test(() => {
+  const ts = new TransformStream({}, {}, { highWaterMark: 0 });
```
domenic: empty object for the second arg is interesting. It might be worth pointing out this is basically testing that size() is never called (for either of them)

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
ricea: I tend to lose track of which I'm supposed to be using too. They should probably have better names.

flushAsyncEvents() is used when making an assertion that something will not happen. It's equivalent to delay(Infinity) but with the benefit that the test actually completes :-)

delay(0) is used when making an assertion that something will happen after an unspecified number of microtasks have been executed.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
ricea: `{}` actually has the same effect as `undefined` here. I was trying to be clever and save characters, which was a mistake.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
domenic: Those are pretty great explanations.

reference-implementation/to-upstream-wpts/resources/recording-streams.js
ricea: I decided that having recordingTransformStream have different default behaviour from the equivalent TransformStream was confusing, and made it behave like a identity transform by default.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
ricea: Done.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
ricea: It must be confusing, because the second chunk is being blocked by backpressure.

reference-implementation/to-upstream-wpts/transform-streams/backpressure.js
ricea: Now that identity behaviour is the default for recordingTransformStream, it is explicit that this one is discarding chunks and so I've been able to reduce the need for explanation a bit.

ricea:
I have a follow-up change which adds a strategies.js test file that verifies that the strategies are correctly passed to the underlying streams. In that one I've specifically commented that I'm not testing the full behaviour of queuing, since we assume that the implementation is the same as for the normal ReadableStream and WritableStream.

reference-implementation/to-upstream-wpts/.eslintrc.json
```diff
@@ -56,7 +56,9 @@
     "flushAsyncEvents": true,
     "methodRejects": true,
     "readableStreamToArray": true,
+    "recordingIdentityTransformStream": true,
```
domenic: No longer necessary

PR #779, opened 2017-08-25
Revise abstract, introduction, and model

### LINKED ISSUE #769: Make the abstract shorter
Looking at https://spec.whatwg.org/, we're a clear outlier. It should probably just be

> The Streams Standard provides APIs for creating, composing, and consuming streams of data.

or maybe

> The Streams Standard provides APIs for creating, composing, and consuming streams of data that map efficiently to low-level I/O primitives, and allow easy composition with built-in backpressure and queuing.

Need to update it here and there.

The rest can go in the introduction section.

domenic:
Making the abstract shorter fixes #769. The introduction incorporates some of its former content, and also was updated in light of how streams and their usage have evolved over time. Finally, it also includes a brief bit about web-developer-created streams, which was curiously absent.

The model section was updated to tie the abstract concepts there to concrete JavaScript classes. Back when it was written, there was some idea of using duck-typing, but that has been abandoned, and this is much more helpful.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -123,7 +123,7 @@ synchronously, e.g. if it is held by the operating system's in-memory buffers, o
 read from disk. An example pull source is a file handle, where you seek to specific locations and read specific amounts.
 
 Readable streams are designed to wrap both types of sources behind a single, unified interface. The implementation
-details of a source are provided by an object containing certain methods and properties that is passed to the
+details of a source are provided by an object with certain methods and properties. It is passed to the
 {{ReadableStream()}} constructor.
 
 <a>Chunks</a> are enqueued into the stream by the stream's <a>underlying source</a>. They can then be read one at a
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -68,32 +64,40 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
 
 <em>This section is non-normative.</em>
 
+<div class="non-normative">
```
ricea: This is the only use of this class in the spec. Should we have a separate issue to add more?

index.bs
```diff
@@ -118,49 +122,64 @@ constantly being pushed from the OS level, at a rate that can be controlled by c
 synchronously, e.g. if it is held by the operating system's in-memory buffers, or asynchronously, e.g. if it has to be
 read from disk. An example pull source is a file handle, where you seek to specific locations and read specific amounts.
 
-Readable streams are designed to wrap both types of sources behind a single, unified interface.
+Readable streams are designed to wrap both types of sources behind a single, unified interface. The implementation
+details of a source are provided by an object containing certain methods and properties that is passed to the
```
ricea: "containing" may have a specific meaning here. If not I would prefer "having".

The mixture of plural and singular forms in this sentence requires the reader to backtrack to parse it. I'm not sure what the best fix would be, but maybe split it? For example,

> A source is implemented as an object having specific methods and properties. It is passed to the {{ReadableStream()}} constructor.

index.bs
```diff
@@ -118,49 +122,64 @@ constantly being pushed from the OS level, at a rate that can be controlled by c
...
-<a lt="locked to a reader">lock</a> the stream, making it no longer directly usable; however, it will create two new
-streams, called <dfn lt="branches of a readable stream tee">branches</dfn>, which can be consumed independently.
+Consumers can also <dfn lt="tee a readable stream">tee</dfn> a readable stream using its {{ReadableStream/tee()}}
+method. This will <a lt="locked to a reader">lock</a> the stream, making it no longer directly usable; however, it will
+create two new streams, called <dfn lt="branches of a readable stream tee">branches</dfn>, which can be consumed
+independently.
 
 For streams representing bytes, an extended version of the <a>readable stream</a> is provided to handle bytes
 efficiently, in particular by minimizing copies. The <a>underlying source</a> for such a readable stream is called
 a <dfn>underlying byte source</dfn>. A readable stream whose underlying source is an underlying byte source is sometimes
```
ricea: Should "a underlying byte source" be "an underlying byte source"?

index.bs
domenic: I went through and added a bunch more. It caught some cases of using RFC 2119 keywords incorrectly. So, yay!

PR #782, opened 2017-08-29
Remove TransformStream _readableClosed member

ricea:
Instead inspect the state of the ReadableStream directly using a new abstract
operation, ReadableStreamDefaultControllerIsErroredOrCloseRequested.

This avoids redundant state tracking.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1622,8 +1622,7 @@ desiredSize</h5>
 
 <emu-alg>
   1. If ! IsReadableStreamDefaultController(*this*) is *false*, throw a *TypeError* exception.
-  1. If *this*.[[closeRequested]] is *true*, throw a *TypeError* exception.
-  1. If *this*.[[controlledReadableStream]].[[state]] is not `"readable"`, throw a *TypeError* exception.
+  1. If ! ReadableStreamDefaultControllerCanCloseOrEnqueue(*this*) is *false*, throw a *TypeError* exception.
   1. Perform ! ReadableStreamDefaultControllerClose(*this*).
 </emu-alg>
 
@@ -1635,8 +1634,7 @@ desiredSize</h5>
 
 <emu-alg>
   1. If ! IsReadableStreamDefaultController(*this*) is *false*, throw a *TypeError* exception.
-  1. If *this*.[[closeRequested]] is *true*, throw a *TypeError* exception.
-  1. If *this*.[[controlledReadableStream]].[[state]] is not `"readable"`, throw a *TypeError* exception.
+  1. If ! ReadableStreamDefaultControllerCanCloseOrEnqueue(*this*) is *false*, throw a *TypeError* exception.
   1. Return ? ReadableStreamDefaultControllerEnqueue(*this*, _chunk_).
 </emu-alg>
 
@@ -1717,8 +1715,7 @@ nothrow>ReadableStreamDefaultControllerShouldCallPull ( <var>controller</var> )<
 
 <emu-alg>
   1. Let _stream_ be _controller_.[[controlledReadableStream]].
-  1. If _stream_.[[state]] is `"closed"` or _stream_.[[state]] is `"errored"`, return *false*.
-  1. If _controller_.[[closeRequested]] is *true*, return *false*.
+  1. If ! ReadableStreamDefaultControllerCanCloseOrEnqueue(_controller_) is *false*, return *false*.
   1. If _controller_.[[started]] is *false*, return *false*.
   1. If ! IsReadableStreamLocked(_stream_) is *true* and ! ReadableStreamGetNumReadRequests(_stream_) > *0*, return
      *true*.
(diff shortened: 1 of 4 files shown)


ricea:
ReadableStreamDefaultControllerIsErroredOrCloseRequested is a terrible name. Suggestions welcome.

tyoshino:
> close and enqueue. (But not error, for some reason?)

IIRC, this is because we wanted to allow erroring a readable stream already being closed.

tyoshino:
So, it might be an option to adjust the signal to be aligned with the current preconditions of ReadableStream.

Otherwise, strictly speaking, CanCloseOrEnqueue might be more correct? Hmm.

ricea:
@tyoshino Do you mean to change TransformStream so that is will also error a stream that it has already requested to be closed? So we'd add an IsReadableStreamReadable abstract operation and use that in TransformStreamErrorInternal() while in other places we'd use CanCloseOrEnqueue()?

tyoshino:
> Do you mean to change TransformStream so that is will also error a stream that it has already requested to be closed?

Right, though I haven't thought well to see if it's beneficial.

> So we'd add an IsReadableStreamReadable abstract operation and use that in TransformStreamErrorInternal() while in other places we'd use CanCloseOrEnqueue()?

Yes.

ricea:
With the latest commit you can now error the readable side after controller.close() just like with a ReadableStream. No existing tests were broken by this change. I added two which are sensitive to it. No behaviour change to ReadableStream. PTAL.

index.bs
```diff
@@ -840,6 +840,17 @@ readable stream is <a>locked to a reader</a>.
   1. Return *true*.
 </emu-alg>
 
+<h4 id="is-readable-stream-readable" aoid="IsReadableStreamReadable" nothrow>IsReadableStreamReadable (
```
domenic: This doesn't appear to be used in the spec, but only in the reference implementation.

Also, it's kind of a lame abstract operation? Is the idea just to preserve encapsulation in some way? Meh....

index.bs
```diff
@@ -1821,15 +1827,22 @@ Specifications should <em>not</em> use this on streams they did not create.
...
-</div>
+This abstract operation is used in the implementation of TransformStream.
 
 <emu-alg>
   1. If ! ReadableStreamDefaultControllerShouldCallPull(_controller_) is *true*, return *false*.
   1. Otherwise, return *true*.
 </emu-alg>
 
+<h4 id="readable-stream-default-controller-can-close-or-enqueue" aoid="ReadableStreamDefaultControllerCanCloseOrEnqueue"
+nothrow export>ReadableStreamDefaultControllerCanCloseOrEnqueue ( <var>controller</var> )</h4>
```
domenic: Probably don't want to export this one.

reference-implementation/lib/readable-stream.js
```diff
@@ -316,6 +318,16 @@ function IsReadableStreamLocked(stream) {
   return true;
 }
 
+function IsReadableStreamReadable(stream) {
+  assert(IsReadableStream(stream) === true, 'IsReadableStreamLocked should only be used on known readable streams');
```
domenic: Bad assertion message

reference-implementation/lib/readable-stream.js
```diff
@@ -1151,6 +1147,17 @@ function ReadableStreamDefaultControllerHasBackpressure(controller) {
   return true;
 }
 
+// This is used in the implementation of TransformStream.
```
domenic: Comment isn't in sync with the spec, since this is more widely used now.

index.bs
ricea: After reading your comment, I realised that pipeTo() looks at the internals of WritableStream directly, so abstraction here probably isn't appropriate.

index.bs
ricea: It may be useful for specs that don't want to trigger the asserts in ControllerClose or ControllerEnqueue but also don't want to track the closed state themselves. But we don't need it right now so I have removed the export.

reference-implementation/lib/readable-stream.js
ricea: Removed.

reference-implementation/lib/readable-stream.js
ricea: Thanks. Removed.

ricea:
PTAL.

ricea:
I hadn't rebased, which is why the tests were erroneously passing locally.

reference-implementation/lib/readable-stream.js
```diff
@@ -919,13 +920,8 @@ class ReadableStreamDefaultController {
...
 
-    if (this._closeRequested === true) {
-      throw new TypeError('The stream has already been closed; do not close it again!');
-    }
-
-    const state = this._controlledReadableStream._state;
-    if (state !== 'readable') {
-      throw new TypeError(`The stream (in ${state} state) is not in the readable state and cannot be closed`);
+    if (ReadableStreamDefaultControllerCanCloseOrEnqueue(this) === false) {
+      throw new TypeError('The stream is not in a state that can be closed');
```
tyoshino: can be closed -> permits close?

ricea: Done.

PR #783, opened 2017-08-29
Test that TransformStream strategies are used correctly

ricea:
Verify that the writableStrategy is used to configure the writable side and the
readableStrategy is used to configure the readable side.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/strategies.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/strategies.js
@@ -1,0 +1,61 @@
+'use strict';
+
+// Here we just test that the strategies are correctly passed to the readable and writable sides. We assume that
+// ReadableStream and WritableStream will correctly apply the strategies when they are being used by a TransformStream
+// and so it isn't necessary to repeat their tests here.
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+  self.importScripts('../resources/recording-streams.js');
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/transform-streams/strategies.js
```diff
@@ -0,0 +1,61 @@
...
+  let readableSizeCalled = false;
+  const ts = new TransformStream(
+    {
+      transform(chunk, controller) {
+        t.step(() => {
+          assert_true(writableSizeCalled, 'writableStrategy.size() should have been called');
+          assert_false(readableSizeCalled, 'readableStrategy.size() should not have been called');
+          controller.enqueue(chunk);
+          assert_true(readableSizeCalled, 'readableStrategy.size() should have been called');
+        });
```
domenic: Can you set a boolean to true inside this test and then check it before the test succeeds? Right now I believe the test will succeed if write() just does `return Promise.resolve()`.

domenic:
Oh, and I guess ESLint doesn't like some of the spacing; we should probably keep it happy. If all else fails declare separate object literals.

reference-implementation/to-upstream-wpts/transform-streams/strategies.js
ricea: Done.

ricea:
The older version of eslint I had installed wasn't bothered by it. I managed to find an indentation that eslint accepts that is tolerable to me.

Issue #777, opened 2017-08-25
Change default readableStrategy HWM to 0?


ricea:
Currently a TransformStream with the default strategies adds two chunks of buffering to any pipe. I'd like to reduce that as much as possible. We can't reduce the HWM of the writableStrategy to 0 because it would have permanent backpressure preventing the pipe from working. However, reducing the HWM of the readableStrategy to `0` will work.

Advantages:
- Reduces the amount of "bufferbloat" induced by adding an extra transform to a pipe, making backpressure propagate faster.
- Makes transform() be read-triggered by default, reducing unnecessary work. However, as soon as you pipe the readable to something transforms will start happening, so it doesn't actually make much difference.

Disadvantages:
- Breaks a surprising number of tests. I haven't looked at it in detail, but I don't think any of the breakage is fundamental. The plus side is that this will identify a set of tests whose behaviour changes with backpressure, and then we can split them into with- and without- backpressure versions.
- TransformStream ends up with a different default from ReadableStream, which may cause some surprise.

I've thought of three variations for how to treat the third argument of TransformStream:
1. **undefined** gets converted to `{highWaterMark: 0}`. Anything else is passed to the ReadableStream constructor unchanged.
2. **undefined** and empty objects get converted to `{highWaterMark: 0}`. Anything else is unchanged.
3. **undefined** gets converted to `{highWaterMark: 0}`. Anything else that doesn't already have a `highWaterMark` property gets one added with value `0`.

I think I prefer the first one.

domenic:
+1 to changing this. The disadvantages seem fine. I don't think we need the same defaults for readable vs. transform since we already know the readable strategy is very nich for transform streams.

The mechanics of how to do this are harder, hmm. I'm no longer convinced we're doing the right thing for non-transform streams, actually :(. With fresh eyes, I think I would have preferred that only `undefined` for the whole argument triggers defaults, and if it's not undefined, you have to specify both size and highWaterMark, otherwise you get errors.

Right now instead, providing `undefined` is the same as providing `{}`, and we have piece-wise defaults for size and highWaterMark. Oh well. So, how do we be consistent with that?

I think we do `{ size, highWaterMark } = {}`, and if highWaterMark is undefined, we change it to 0 before passing it to ReadableStream. So we don't necessarily pass the exact object through to the readable stream.

Practically I think this looks like your 3.

ricea:
I like your suggestion better so that is what I am working on.

I have run into an interesting problem with this test:

```javascript
[...]

PR #790, opened 2017-09-08
Clarify some of the ECMAScript conventions applied

### LINKED ISSUE #775: Normative requirements for constructor names?
Is `WritableStream.name` required to return "WritableStream"?

Is `new WritableStream().getWriter().constructor.name` required to return "WritableStreamDefaultWriter"?

domenic:
Closes #775.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -34,6 +34,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
     text: the typed array constructors table; url: #table-49; type: dfn
     text: TypeError; url: #sec-native-error-types-used-in-this-standard-typeerror; type: exception
     text: Invoke; url: #sec-invoke; type: abstract-op
+    text: DestructuringAssignmentEvaluation; url: #sec-runtime-semantics-destructuringassignmentevaluation; type: abstract-op
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4548,15 +4548,22 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+as <a href="https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects">built-in objects</a>. For example,
+their <code>name</code> and <code>length</code> properties are derived as described by that specification, as are the
+default property descriptor values.
+
+We also depart from the [[!ECMASCRIPT]] conventions in the following ways, mostly for brevity. It is hoped (and vaguely
+planned) that the conventions of ECMAScript itself will evolve in these ways.
 
 <ul>
   <li> We use destructuring notation in function and method declarations, and assume that the destructuring assignment
     procedure was performed before the algorithm starts.
```
ricea: Should we say "algorithm" rather than procedure? I was surprised when I was trying to reimplement it for #777 that there are some subtleties involved. (I have since abandoned the approach of trying to reimplement it in spec language).

index.bs
```diff
@@ -4548,15 +4548,22 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
   <li> We similarly use the default argument notation <code>= {}</code> in a couple of cases.
```
ricea: Maybe link to the algorithm? I found the interaction in the TransformStream constructor surprising when I looked at it. ie. writableStrategy and readableStrategy are defaulted to undefined, which is not an acceptable value for destructuring assignment, and then when the {Readable,Writable}Stream constructors are called those undefined values are converted to {} by the default arguments there, making the destructuring assignment work.

index.bs
```diff
@@ -4548,15 +4548,22 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
 
-This specification uses algorithm conventions very similar to those of [[!ECMASCRIPT]]. However, it deviates in the
-following ways, mostly for brevity. It is hoped (and vaguely planned) that eventually the conventions of ECMAScript
-itself will evolve in these ways.
+This specification uses algorithm conventions very similar to those of [[!ECMASCRIPT]], whose rules should be used to
+interpret it (apart from the exceptions enumerated below). In particular, the objects specified here should be treated
+as <a href="https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects">built-in objects</a>. For example,
+their <code>name</code> and <code>length</code> properties are derived as described by that specification, as are the
+default property descriptor values.
+
```
ricea: I think it would be nice to explicitly mention the fact that missing and surplus arguments are treated as per ECMASCRIPT function conventions. Although this is implied, it clearly surprises some people.

index.bs
domenic: We can do better; we'll just link to https://tc39.github.io/ecma262/#sec-runtime-semantics-destructuringassignmentevaluation

index.bs
domenic: This one isn't as easy to link to... I think it's buried in https://tc39.github.io/ecma262/#sec-functiondeclarationinstantiation . Maybe I'll just expand on it.

ricea:
Nice!

PR #792, opened 2017-09-08
Update acknowledgments and editorship

domenic:
As time goes on, the distinction between "special thanks" and "would like to thank" begins to feel a bit unfair and arbitrary. "Special thanks" is skewed toward those who were involved in the beginning. Their contributions were crucial for getting us started, but others, currently left in the "would like to thank" category, have contributed a lot too. The cutoff is not objective, but at the discretion of the editor at the time, which seems a bit unfair. As such, this commit collapses the two lists into a single one, without listing the details of individual contributions for "special thanks". This matches other WHATWG specifications. However, we also add an extra sentence to emphasize the degree to which community contributions have shaped this specification; Streams truly is special in that regard.

Additionally, this formally recognizes Adam as an editor, as he's doing most of the work these days.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4566,18 +4566,23 @@ values, and all arithmetic operations performed on them must be done in the stan
 
 <h2 id="acks" class="no-num">Acknowledgments</h2>
 
-The editor would like to thank
-Adam Rice,
+The editors would like to thank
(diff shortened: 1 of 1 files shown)


ricea:
Yay! lgtm, obviously.

PR #793, opened 2017-09-08
TransformStream: test enqueue(), close(), error() in bad states

ricea:
Add tests to very that controller.enqueue() and controller.close() throw
exceptions when the stream is already closed.

Also test that controller.error() throws when the stream is already errored.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/errors.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/errors.js
@@ -232,4 +232,52 @@ promise_test(t => {
   });
 }, 'abort should set the close reason for the writable when it happens first during underlying sink write');
 
+test(() => {
+  new TransformStream({
+    start(controller) {
+      controller.error(thrownError);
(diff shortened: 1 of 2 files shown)


reference-implementation/to-upstream-wpts/transform-streams/errors.js
```diff
@@ -232,4 +232,52 @@ promise_test(t => {
...
+  let controller;
+  const ts = new TransformStream({
+    start(c) {
+      controller = c;
+    }
+  });
+  const cancelPromise = ts.readable.cancel();
+  assert_throws(new TypeError(), () => controller.error(), 'error() should throw');
+  return cancelPromise;
+}, 'controller.error() throws after readable.cancel()');
```
domenic: "and (but?) the cancel succeeds"? (same for other `return cancelPromise` tests

ricea: Done.

PR #797, opened 2017-09-15
Don't be dismissive about the controller/stream division

### LINKED ISSUE #741: "At this point in time the division of work betw..."
https://streams.spec.whatwg.org/commit-snapshots/3193db4f3fa34c52d726055ab99a623f66c4a623#ws-abstract-ops-used-by-controllers

> At this point in time the division of work between the stream and its controller may seems somewhat arbitrary, but centralizing much of the logic in the controller is a useful structure for the future\.

This sentence is maybe overly ambiguous? I feel we've actually been quite careful in designing this to enable byte streams, but that doesn't come across at all.

Based on https://github.com/w3ctag/design-reviews/issues/163#issuecomment-297909069 by @dbaron.

domenic:
Closes #741.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2919,11 +2919,9 @@ writable stream is <a>locked to a writer</a>.
 
 <h3 id="ws-abstract-ops-used-by-controllers">Writable Stream Abstract Operations Used by Controllers</h3>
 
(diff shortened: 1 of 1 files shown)


ricea:
lgtm

PR #799, opened 2017-09-15
Detect if the TransformStream errored inside write() and reject

ricea:
If a TransformStream errored while the TransformStreamDefaultSink write() method
was waiting for backpressure to be relieved, it would assert inside
the TransformStreamStream operation.

Since it is not useful to call transform() once the stream is errored, return a
rejection from the write() method instead.

Also add a test for this case.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -318,7 +318,12 @@ class TransformStreamDefaultSink {
 
     if (transformStream._backpressure === true) {
       return transformStream._backpressureChangePromise
-          .then(() => TransformStreamTransform(transformStream, chunk));
+          .then(() => {
+            if (transformStream._errored === true) {
+              return Promise.reject(transformStream._storedError);
+            }
+            return TransformStreamTransform(transformStream, chunk);
+          });
     }
 
     return TransformStreamTransform(transformStream, chunk);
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -318,7 +318,12 @@ class TransformStreamDefaultSink {
 
     if (transformStream._backpressure === true) {
       return transformStream._backpressureChangePromise
-          .then(() => TransformStreamTransform(transformStream, chunk));
+          .then(() => {
+            if (transformStream._errored === true) {
+              return Promise.reject(transformStream._storedError);
```
domenic: I think this will be better as `throw transformStream._storedError`? They're equivalent, but thinking of how we'll write it in spec language, I think we'll say "transforming by a function that throws ... if ..., and otherwise returns ..."

domenic: Reviewing https://github.com/whatwg/streams/pull/800 I see there is a lot of pre-existing code that does return Promise.reject() from inside fulfillment/rejection handlers. So maybe not worth fixing just this one-off, but instead doing a greater alignment when or slightly before you do the spec translation.

ricea: It will be easier to do as one big cleanup in a separate pass, so if you don't mind I will do that. I have filed https://github.com/whatwg/streams/issues/804 to track this.

PR #800, opened 2017-09-19
Remove _errored and _storedError from TransformStream

### LINKED ISSUE #796: TransformStream: remove [[errored]] and [[storedError]] slots
They appear to be unneeded. If the TransformStream is errored, then the readable side will always be errored (or closed, but this is okay), and the writable side will always be errored or erroring.

[[storedError]] is only used in two places:

1. If enqueue() throws, to make sure we throw the first error that happened, not necessarily the one from enqueue. Since we check we aren't errored before we call enqueue, and enqueue is synchronous, I don't think it's even possible for this to be different.
2. In the rejection returned by TransformStreamDefaultSink when flush() fails, or when it succeeds but the stream was errored in the meantime.

The first can be removed and the second can use the readable sides' storedError instead.

The benefit of removing them is simplification of the spec and reduction in the amount of state needed by implementations.

ricea:
These slots are mostly redundant with the same state on the readable
side, so use the ReadableStream [[state]] and [[storedError]] instead to
reduce complexity.

It is possible for the readable to be "closed" where previously _errored
was set, but this doesn't make any practical difference as the
implementation threw a TypeError in both cases anyway.

TransformStreamError is renamed to TransformStreamDefaultControllerError to
better reflect its new semantics.

Closes #796.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -13,9 +13,6 @@ class TransformStream {
   constructor(transformer = {}, writableStrategy = undefined, readableStrategy = undefined) {
     this._transformer = transformer;
 
-    this._errored = false;
-    this._storedError = undefined;
-
     this._writableController = undefined;
     this._readableController = undefined;
     this._transformStreamController = undefined;
@@ -48,13 +45,6 @@ class TransformStream {
     const startResult = InvokeOrNoop(transformer, 'start',
                           [transformStream._transformStreamController]);
     startPromise_resolve(startResult);
-    startPromise.catch(e => {
-      // The underlyingSink and underlyingSource will error the readable and writable ends on their own.
-      if (transformStream._errored === false) {
-        transformStream._errored = true;
-        transformStream._storedError = e;
-      }
-    });
   }
 
   get readable() {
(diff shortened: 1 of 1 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -277,7 +240,11 @@ class TransformStreamDefaultController {
       throw defaultControllerBrandCheckException('error');
     }
 
-    TransformStreamError(this._controlledTransformStream, reason);
+    if (this._controlledTransformStream._readable._state !== 'readable') {
+      throw new TypeError('TransformStream is not in a state that can be errored');
+    }
+
+    TransformStreamDefaultControllerError(this, reason);
```
domenic: Since TransformStreamDefaultControllerError is so short, can we just inline it here?

reference-implementation/lib/transform-stream.js
```diff
@@ -150,28 +131,10 @@ function TransformStreamEnqueueToReadable(transformStream, chunk) {
...
-}
-
-function TransformStreamErrorIfNeeded(transformStream, e) {
-  if (transformStream._errored === false) {
-    TransformStreamErrorInternal(transformStream, e);
-  }
-}
-
+// This is a no-op if both sides are already errored.
 function TransformStreamErrorInternal(transformStream, e) {
```
domenic: This should probably lose the -Internal suffix now?

reference-implementation/lib/transform-stream.js
ricea: I was thinking it might be useful as a hook for other standards, but since they are more likely to just throw from transform(), maybe not?

reference-implementation/lib/transform-stream.js
ricea: That makes sense.

reference-implementation/lib/transform-stream.js
domenic: Oh, I see, in parallel to the ones for other controllers. Yeah, maybe keep it, in which case LGTM for the whole patch.

PR #805, opened 2017-09-22
Rename "transformStream" variables to "stream"

ricea:
For consistency with Readable- and WritableStream, and brevity, rename
all "transformStream" parameters and local variables to just "stream".

Rename _transformStream members in Source and Sink to
_ownerTransformStream which is hopefully slightly clearer.

No functional changes.


domenic:
Hmm maybe also "transformStreamController" -> controller when used as a variable.

Issue #774, opened 2017-08-22
What should the behaviour of controller.close() be?


ricea:
I think the following test should pass:

```javascript
promise_test(t => {
  const ts = recordingIdentityTransformStream();
  const rs = new ReadableStream({
    start(controller) {
      controller.enqueue(0);
      controller.enqueue(1);
      controller.enqueue(2);
      controller.close();
    }
  });
  const pipeToPromise = rs.pipeTo(ts.writable);
  return delay(0).then(() => {
    assert_array_equals(ts.events, ['transform', 0], 'transform() should have seen one chunk');
    ts.controller.close();
    return promise_rejects(t, new TypeError(), pipeToPromise, 'enqueuing the second chunk should fail');
  }).then(() => {
    assert_array_equals(ts.events, ['transform', 0, 'transform', 1], 'transform() should have seen two chunks');
  });
}, 'closing readable should permit remaining chunks to be processed');
```

Currently `pipeToPromise` never becomes settled because the second chunk is never written.

I think this should work. Having a method which blocks the pipe permanently if you call it when there is backpressure doesn't seem very useful to me.

domenic:
If you'll indulge me, I'm kind of unclear on the semantics of a transform controller's close().

I guess the situation is when you want to discard any further incoming data. You're signaling that everything you've seen so far is enough, and now you're successfully done. E.g. you could imagine a findFirstPrime transformer: [1, 4, 6, 9, 7, 10, 2] (length = 7) -> [false, false, false, false, true] (length = 5).

In that scenario do we say that attempting to transform the 10 is a failure? Probably, just like attempting to write to any other closed writable stream is a failure. OK, that makes sense. But the readable side never gets marked as errored, right?

In this situation you have [0, 1, 2] -> [0]. Do we count the pipe as failed, or succeeded? Your test indicates failed, which I guess is pretty baked in to the semantics of pipeTo: if the writable side errors, everything blows up. But it seems like disallowing early-quitting in this way is a bit bad.

ricea:
> If you'll indulge me, I'm kind of unclear on the semantics of a transform controller's close().

Me too.

What I'm imagining is that you already have code to discard chunks you aren't interested in. For example:

```javascript
new TransformStream({
  start() {
    this.sawPrime = false;
  },
  transform(chunk, controller) {
    if (this.sawPrime) {
      return;
    }
    controller.enqueue(chunk);
    if (isPrime(chunk)) {
      controller.close();
      this.sawPrime = true;
[...]

PR #811, opened 2017-09-26
Standard text for TransformStream

ricea:
Add text to the standard for TransformStream.

***
[Preview](https://streams.spec.whatwg.org/branch-snapshots/transform-stream-standard-text/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/streams/10469ac...cfd14a0.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3897,7 +3897,21 @@ TransformStream(<var>transformer</var> = {}, writableStrategy = undefined, { <va
 </div>
 
 <emu-alg>
-<!-- TODO: write algo steps -->
+  1. Set *this*.[[transformer]] to _transformer_.
+  1. Set *this*.[[transformStreamController]] to *undefined*.
+  1. Set *this*.[[backpressure]] and *this*.[[backpressureChangePromise]] to *undefined*.
+  1. Set *this*.[[transformStreamController]] to ? Construct(`<a idl>TransformStreamDefaultController</a>`, « *this* »).
+  1. Let _startPromise_ be <a>a new promise</a>.
+  1. Let _source_ be ! Construct(`<a idl>TransformStreamDefaultSource</a>`, « *this*, _startPromise_ »).
+  1. Let _readableStrategy_ be a new object.
+  1. Set _readableStrategy_["size"] to _size_.
+  1. Set _readableStrategy_["highWaterMark"] to _highWaterMark_.
+  1. Set *this*.[[readable]] to ? Construct(`<a idl>ReadableStream</a>`, « _source_, _readableStrategy_ »).
+  1. Let _sink_ be ! Construct(`<a idl>TransformStreamDefaultSink</a>`, « *this*, _startPromise_ »).
+  1. Set *this*.[[writable]] to ? Construct(`<a idl>WritableStream</a>`, « _sink_, _writableStrategy_ »).
+  1. Perform ! TransformStreamSetBackpressure(*this*, *true*).
+  1. Let _startResult_ be ? InvokeOrNoop(_transformer_, 'start', « *this*.[[transformStreamController]] »).
+  1. Resolve _startPromise_ with _startResult_.
 </emu-alg>
 
 <h4 id="ts-prototype">Properties of the {{TransformStream}} Prototype</h4>
--- a/index.bs
+++ b/index.bs
@@ -4235,7 +4235,39 @@ lt="TransformStreamDefaultSink(stream, startPromise)">new TransformStreamDefault
     1. A rejection handler that, when called with argument _r_, performs the following steps:
       1. Perform ! TransformStreamError(_stream_, _r_).
       1. Throw _readable_.[[storedError]].
+</emu-alg>
 
+<h3 id="ts-default-sink-abstract-ops">Transform Stream Default Sink Abstract Operations</h3>
+
+<h4 id="transform-stream-default-sink-invoke-transform" aoid="TransformStreamDefaultSinkInvokeTransform"
+throws>TransformStreamDefaultSinkInvokeTransform ( <var>stream</var>, <var>chunk</var> )</h4>
+
+<emu-alg>
+  1. Let _controller_ be _stream_.[[transformStreamController]].
+  1. Let _transformer_ be _stream_.[[transformer]].
+  1. Let _method_ be _transformer_.transform.
+  1. If _method_ is *undefined*,
+    1. Perform ? TransformStreamDefaultControllerEnqueue(_controller_, _chunk_).
+    1. Return *undefined*.
+  1. Return ? Call(_method_, _transformer_, « _chunk_, _controller_ »).
+</emu-alg>
+
+<h4 id="transform-stream-default-sink-transform" aoid="TransformStreamDefaultSinkTransform"
+nothrow>TransformStreamDefaultSinkTransform ( <var>sink</var>, <var>chunk</var> )</h4>
+
+<emu-alg>
+  1. Let _stream_ be _sink_.[[ownerTransformStream]].
+  1. Assert: _stream_.[[readable]].[[state]] is not `"errored"`.
+  1. Assert: _stream_.[[backpressure]] is *false*.
+  1. Let _transformPromise_ be *undefined*.
+  1. Let _transformResult_ be TransformStreamDefaultSinkInvokeTransform(_stream_, _chunk_).
+  1. If _transformResult_ is an abrupt completion, set _transformPromise_ to <a>a promise rejected with</a>
+     _transformResult_.[[Value]].
+  1. Otherwise, set _transformPromise_ to <a>a promise resolved with</a> _transformResult_.[[Value]].
+  1. Return the result of <a>transforming</a> _transformPromise_ with a rejection handler that, when called with
+     argument _e_, performs the following steps:
+    1. Perform ! TransformStreamError(_stream_, _e_).
+    1. Throw _e_.
 </emu-alg>
 
 <h3 id="ts-default-source-class" interface lt="TransformStreamDefaultSource">Class
--- a/index.bs
+++ b/index.bs
@@ -3993,9 +3993,16 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
 <var>e</var> )</h4>
 
 <emu-alg>
-  1. Perform ! WritableStreamDefaultControllerErrorIfNeeded(_stream_.[[writable]].[[writableStreamController]], _e_).
   1. If _stream_.[[readable]].[[state]] is `"readable"`, perform !
      ReadableStreamDefaultControllerError(_stream_.[[readable]].[[readableStreamController]], _e_).
+  1. Perform ! TransformStreamErrorWritableAndUnblockWrite(_stream_, _e_).
+</emu-alg>
+
+<h4 id="transform-stream-error-writable-and-unblock-write" aoid="TransformStreamErrorWritableAndUnblockWrite"
+nothrow>TransformStreamErrorWritableAndUnblockWrite ( <var>stream</var>, <var>e</var> )</h4>
+
+<emu-alg>
+  1. Perform ! WritableStreamDefaultControllerErrorIfNeeded(_stream_.[[writable]].[[writableStreamController]], _e_).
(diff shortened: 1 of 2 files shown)


ricea:
https://streams.spec.whatwg.org/branch-snapshots/transform-stream-standard-text/

ricea:
This is now complete and ready for review.

If possible, I would like both @tyoshino and @domenic to review the formatted text. Almost all the changes are in the [transform streams model](https://streams.spec.whatwg.org/branch-snapshots/transform-stream-standard-text/#ts-model) section and the [Transform Streams](https://streams.spec.whatwg.org/branch-snapshots/transform-stream-standard-text/#ts) section. All of the issues I've fixed in my own checks have been visible in the formatted version.

I think we need one volunteer to pick through all the new markup for errors, and another volunteer to double-check that the steps in the text match the steps in the reference implementation. I think it's best to coordinate so we don't duplicate this grungy work.

domenic:
I can certainly volunteer for markup review for now.

domenic:
I did a first pass focused on non-normative text. Of note, I tried to eliminate talking about "output" and "input", given the confusion in #174. Instead I introduced "writable side" and "readable side" and tried to use those as much as possible.

I think we should also add some nice examples to the appendix. Ideas:

- Wrapping a TextEncoder or TextDecoder
- A small suite of simple transforms such as uppercase , double the output, etc.
- Wrapping web cypto?

I'll try to do more review tomorrow.

index.bs
```diff
@@ -3763,41 +3775,54 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+
+  <pre><code class="lang-javascript">
+    const { writable, readable } = new TransformStream();
+    fetch("...", { body: readable }).then(response => /* ... */);
+
+    writable.write(new Uint8Array([0x73, 0x74, 0x72, 0x65, 0x61, 0x6D, 0x73, 0x21]));
+    writable.close();
+  </code></pre>
+
+  Another use of identity transform streams is to add additional buffering to a <a>pipe</a>. In this example we add
```
ricea: Is it possible to have a separate id for the buffering example without disrupting the readability of this section too much?

index.bs
```diff
@@ -3871,30 +3896,36 @@ TransformStream(<var>transformer</var> = {}, writableStrategy = undefined, { <va
...
-      This permits a single input chunk to result in zero or multiple output chunks. The stream implementation
-      guarantees that <code>transform</code> will be called only after previous transformations have succeeded, and
-      never before <code>start</code> has completed or after <code>flush</code> is called. When no
-      <code>transform</code> method is supplied the identity transform is used, which copies chunks unchanged from the
-      input to the output.
-    <li> <code>flush(controller)</code> is called after all input chunks have been transformed. It is typically used to
-      append a suffix to the output data. If this process is asynchronous, it can return a promise
-      to signal success or failure. On success, the {{readable}} side will be closed.
+    <li><p><code>start(controller)</code> is called immediately, and is typically used to enqueue prefix <a>chunks</a>
+      that appear in the <a>readable side</a> but doesn't depend on any writes to the <a>writable side</a>. If this
```
ricea: "... **don't** depend on ..."

"appear" is awkward here. Maybe "will be read from"?

index.bs
domenic: Sure, we can just make it another `<div>`, not a big deal I think.

index.bs

ricea:
98157cd lgtm. Explicitly defining "readable side" and "writable side" solves a lot of the wording problems I struggled with.

index.bs
```diff
@@ -3763,10 +3772,653 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
 
 <h2 id="ts">Transform Streams</h2>
 
-Transform streams have been developed in the testable implementation, but not yet re-encoded in spec language.
-We are waiting to validate their design before doing so. In the meantime, see <a
-href="https://github.com/whatwg/streams/blob/master/reference-implementation/lib/transform-stream.js">reference-implementation/lib/transform-stream.js</a>.
+<h3 id="ts-intro">Using Transform Streams</h3>
+
+<div class="example" id="example-basic-pipe-through">
+  The natural way to use a transform stream is place it in a <a lt="piping">pipe</a> between a <a>readable stream</a>
```
tyoshino: place -> placing?

index.bs
tyoshino: +1

index.bs
```diff
@@ -3763,10 +3772,653 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+
+<emu-alg>
+  1. Set *this*.[[transformer]] to _transformer_.
+  1. Set *this*.[[transformStreamController]] to *undefined*.
+  1. Set *this*.[[backpressure]] and *this*.[[backpressureChangePromise]] to *undefined*.
+  1. Let _readableType_ be ? GetV(_transformer_, `"readableType"`).
+  1. If _readableType_ is not *undefined*, throw a *RangeError* exception.
+  1. Let _writableType_ be ? GetV(_transformer_, `"writableType"`).
+  1. If _writableType_ is not *undefined*, throw a *RangeError* exception.
+  1. Let _controller_ be  ? Construct(`<a idl>TransformStreamDefaultController</a>`, « *this* »).
```
tyoshino: extra space between be and ?

index.bs
```diff
@@ -3763,10 +3772,653 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+  1. Set *this*.[[transformStreamController]] to _controller_.
+  1. Let _startPromise_ be <a>a new promise</a>.
+  1. Let _source_ be ! Construct(`<a idl>TransformStreamDefaultSource</a>`, « *this*, _startPromise_ »).
+  1. Let _readableStrategy_ be ! ObjectCreate(<a>%ObjectProperty%</a>).
+  1. Perform ! CreateDataProperty(_readableStrategy_, `"size"`, _size_).
+  1. Perform ! CreateDataProperty(_readableStrategy_, `"highWaterMark"`, _highWaterMark_).
+  1. Set *this*.[[readable]] to ? Construct(`<a idl>ReadableStream</a>`, « _source_, _readableStrategy_ »).
+  1. Let _sink_ be ! Construct(`<a idl>TransformStreamDefaultSink</a>`, « *this*, _startPromise_ »).
+  1. Set *this*.[[writable]] to ? Construct(`<a idl>WritableStream</a>`, « _sink_, _writableStrategy_ »).
```
tyoshino: missing _writableStrategy_ creation

index.bs
ricea: We could use "by placing it" or "to place it". I think I prefer "to place it".

index.bs
ricea: It comes directly from the constructor arguments. I can add a note to explain the asymmetry between readableStrategy and writableStrategy.

tyoshino: Oops, sorry. I should've checked the arguments list.

Yeah, it might be helpful, thought it's just because I'm not catching up with the changes. Is https://github.com/whatwg/streams/commit/fd17cf00133784f5a8ed251e8a0fe633967875a8 the only background of the asymmetry?

ricea: Yes, and issue #777 which discusses it.

ricea:
I made some fixes in https://github.com/whatwg/streams/commit/4d2d14e367a5af7bc98a7b3c2571161e821bcb43 which is on this branch but inexplicably missing from this PR. If it doesn't show up by tomorrow I will do a rebase to force it.

domenic:
OK, going to work on checking that the steps match up now.

index.bs
```diff
@@ -3763,10 +3772,656 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+  1. Return *true*.
+</emu-alg>
+
+<h4 id="transform-stream-error" aoid="TransformStreamError" nothrow>TransformStreamError ( <var>stream</var>, <var>e</var> )</h4>
+
+<emu-alg>
+  1. Perform ! WritableStreamDefaultControllerErrorIfNeeded(_stream_.[[writable]].[[writableStreamController]], _e_).
+  1. If _stream_.[[readable]].[[state]] is `"readable"`, perform !
+     ReadableStreamDefaultControllerError(_stream_.[[readable]].[[readableStreamController]], _e_).
+  1. If _stream_.[[backpressure]] is *true*, perform ! TransformStreamSetBackpressure(_stream_, *false*).
```
domenic: This has a comment in the source that seems like it might be useful to include in the standard as a note. However it references things I'm not sure about, like write()/pull()/enqueue(). Are those on the transform stream source/sink? Maybe it can be rephrased to capture the important points without the details; not sure.

index.bs
```diff
@@ -3763,10 +3772,656 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+  1. Assert: _stream_.[[backpressure]] is *true*.
+  1. Assert: _stream_.[[backpressureChangePromise]] is not *undefined*.
+  1. Perform ! TransformStreamSetBackpressure(_stream_, *false*).
+  1. Return _stream_.[[backpressureChangePromise]].
+</emu-alg>
+
+<h5 id="ts-default-source-cancel" method for="TransformStreamDefaultSource">cancel(<var>reason</var>)</h5>
+
+<emu-alg>
+  1. Perform ! TransformStreamError(*this*.[[ownerTransformStream]], _reason_).
```
domenic: Just noticing this... is this correct behavior? I'd have thought it'd be more like terminate than error.

ricea: It actually does behave like terminate, because the readable side is already closed when cancel() is called. I added a note to make this clearer.

domenic: Is there a reason to call TransformStreamError instead of TransformStreamDefaultControllerTerminate? I guess layering, hmm... seems to argue for factoring out a TransformStreamTerminate, if I am understanding correctly.

ricea: We'd end up with TypeError('TransformStream terminated') as the storedError in the writable. This seems less desirable to me than retaining whatever was passed to cancel().

However... maybe factoring out the "bottom half" of TransformStreamDefaultControllerTerminate and TransformStreamError is the way to go. I will try it.

ricea: I made factored out a new abstract op named "TransformStreamErrorWritableAndUnblockWrite". PTAL.

ricea:
I reviewed the comments in the reference implementation to see if any would be useful as notes in the standard. I added a couple more notes. To be honest, I feel that inline notes can make the algorithm harder to follow. It would be nice to have a separate non-normative "algorithm overview" section where we explain how it works. But we don't have anything like that for ReadableStream and WritableStream (and it might be prohibitively complicated to have an up-to-date explanation for those).

Anyway, I believe that implementers will find out how it works in the process of implementing it. Ordinary web developers don't need to know why it works. The mental model of "transform is called when there is no backpressure" is super-easy to understand and the subtleties involved in delivering that are not their problem.

domenic:
Made some tweaks, ready to merge when you are!

ricea:
1a26588 lgtm

ricea:
Rebased.

ricea:
I will go ahead and land this since @tyoshino is busy. We can always fix any issues he finds later.

index.bs
```diff
@@ -3763,10 +3771,678 @@ nothrow>WritableStreamDefaultControllerError ( <var>controller</var>, <var>error
...
+
+<pre><code class="lang-javascript">
+  class TransformStreamDefaultController {
+    constructor(stream)
+
+    get desiredSize()
+
+    enqueue(chunk)
+    error(reason)
+    terminate()
```
TejasQ: I find it interesting that we have `cancel()`, `terminate()`, and `close()`.

ricea: Yes, they all have different semantics. So based on the principle of "different things should look different", we use different names for them. See #774 for the discussion where we added `terminate()`. The difference between `cancel()` and `close()` is mentioned in the FAQ: https://github.com/whatwg/streams/blob/master/FAQ.md#why-dont-errors-that-occur-while-cancelling-put-the-readable-stream-in-an-error-state

TejasQ:
I'm building a demo of this as soon as support lands in Canary/any nightly of any browser.

index.bs
TejasQ: That's awesome. Definitely a lesson in there for me. Thanks, @ricea.

PR #817, opened 2017-10-02
TransformStream: Verify behaviour of abort() during write()

ricea:
A writer.write() call that was waiting for backpressure should reject if
the writable is aborted via abort() before it completes. Add a test to
verify this.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/errors.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/errors.js
@@ -314,12 +314,15 @@ promise_test(t => {
     const writer = ts.writable.getWriter();
     // write should start synchronously
     const writePromise = writer.write(0);
+    // The underlying sink's abort() is not called until the write() completes.
     const abortPromise = writer.abort();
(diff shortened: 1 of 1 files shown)


reference-implementation/to-upstream-wpts/transform-streams/errors.js
```diff
@@ -308,4 +308,19 @@ promise_test(t => {
...
 }, 'erroring during write with backpressure should result in the write failing');
 
+promise_test(t => {
+  const ts = new TransformStream({}, undefined, { highWaterMark: 0 });
+  return delay(0).then(() => {
+    const writer = ts.writable.getWriter();
+    // write should start synchronously
+    const writePromise = writer.write(0);
+    const abortPromise = writer.abort();
+    return promise_rejects(t, new TypeError(), ts.readable.getReader().read(), 'read() should reject')
```
domenic: Why not one Promise.all covering all three at once?

ricea: I think I was trying to make it clearer why the read() was there. But it didn't work, because I couldn't remember why it was needed. I have added comments instead.

PR #818, opened 2017-10-02
Add TransformStreamDefaultController terminate() method

### LINKED ISSUE #774: What should the behaviour of controller.close() be?

ricea:
Replace close() with terminate(). In addition to closing the readable
side, terminate() also errors the writable side. This stops data from
being produced after we are no longer interested in it.

Modify existing tests to handle the new method, and add tests for the new
functionality.

Closes #774.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/terminate.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/terminate.js
@@ -1,0 +1,85 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+  self.importScripts('../resources/recording-streams.js');
+  self.importScripts('../resources/test-utils.js');
+}
+
+promise_test(t => {
+  const ts = recordingTransformStream({}, undefined, { highWaterMark: 0 });
+  const rs = new ReadableStream({
+    start(controller) {
+      controller.enqueue(0);
+    }
+  });
+  let pipeToRejected = false;
+  const pipeToPromise = promise_rejects(t, new TypeError(), rs.pipeTo(ts.writable), 'pipeTo should reject').then(() => {
+    pipeToRejected = true;
+  });
+  return delay(0).then(() => {
+    assert_array_equals(ts.events, [], 'transform() should have seen no chunks');
+    assert_false(pipeToRejected, 'pipeTo() should not have rejected yet');
+    ts.controller.terminate();
+    return pipeToPromise;
+  }).then(() => {
+    assert_array_equals(ts.events, [], 'transform() should still have seen no chunks');
+    assert_true(pipeToRejected, 'pipeToRejected must be true');
+  });
+}, 'controller.terminate() should error pipeTo()');
+
+promise_test(t => {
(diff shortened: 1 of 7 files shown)


reference-implementation/to-upstream-wpts/transform-streams/terminate.js
```diff
@@ -0,0 +1,85 @@
...
+  const ts = new TransformStream({
+    start(controller) {
+      controller.enqueue(0);
+      controller.terminate();
+      controller.error(error1);
+    }
+  });
+  return Promise.all([
+    promise_rejects(t, new TypeError(), ts.writable.abort(), 'abort() should reject with a TypeError'),
+    promise_rejects(t, error1, ts.readable.cancel(), 'cancel() should reject with error1')
```
domenic: Should we test more than .cancel() to test if it's errored properly?

reference-implementation/to-upstream-wpts/transform-streams/terminate.js
```diff
@@ -0,0 +1,85 @@
...
+  ]);
+}, 'controller.error() after controller.terminate() with queued chunk should error the readable');
+
+test(() => {
+  new TransformStream({
+    start(controller) {
+      controller.terminate();
+      assert_throws(new TypeError(), () => controller.error(error1), 'error() should throw');
```
domenic: Opened https://github.com/whatwg/streams/issues/821.

In general, have we tested/are we OK with the behavior for when controller methods throw? I think a guiding principle is that you shouldn't throw if it's not "your fault", e.g. if someone using the public API of the stream can transition the state in some way as to cause an exception. Does that sound familiar from previous discussions? Are we following it here, e.g. if the writable side is aborted or the readable side canceled?

ricea: Thanks for bringing this up. I had forgotten about it actually.

I filed #822 to track the changes to transform stream.

reference-implementation/to-upstream-wpts/transform-streams/terminate.js
ricea: .cancel() is actually pretty specific. I added reader.closed to the tests just to be sure. I don't want to get into testing Readablestream itself here.

PR #819, opened 2017-10-02
throw instead of returning Promise.reject() from handlers

### LINKED ISSUE #804: TransformStream: use throw instead of Promise.reject() in promise handlers
In a few places the TransformStream implementation returns a rejection from a a promise handler. Per https://github.com/whatwg/streams/pull/799#discussion_r140462821 it would be better to throw in these cases instead.

ricea:
Where a fulfillment or rejection handler returned Promise.reject(), throw
instead. This should result in simpler standard text.

Also remove one unneeded return of Promise.resolve() from a fulfillment
handler.

Closes #804.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -253,7 +253,7 @@ class TransformStreamDefaultSink {
             const writable = stream._writable;
             const state = writable._state;
             if (state === 'erroring') {
-              return Promise.reject(writable._storedError);
+              throw writable._storedError;
             }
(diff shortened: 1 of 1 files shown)


ricea:
TransformStreamDefaultSinkDefaultTransform has to explicitly return a promise due to the defintion of the PromiseInvokeOrPerformFallback operation. However, since PromiseInvokeOrPerformFallback is only used to call `transform()`, I think I will inline and remove it. Then TransformStreamDefaultSinkDefaultTransform can also go away.

PR #820, opened 2017-10-02
Add an extension point for byte transform streams

ricea:
Prohibit non-undefined values of readableType and writableType properties
on the transformer to reserve them for future use defining byte
transforms.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -19,6 +19,18 @@ class TransformStream {
     this._backpressureChangePromise = undefined;
     this._backpressureChangePromise_resolve = undefined;
 
+    const readableType = transformer.readableType;
+
+    if (readableType !== undefined) {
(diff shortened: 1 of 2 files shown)


domenic:
Maybe do `type` too in case we end up wanting to just control both at once?

ricea:
I think having `type` alongside `readableType` and `writableType` would be confusing. Asymmetrical transform streams appear to be useful, with TextEncoder being the canonical example. In which case, we need `readableType: 'bytes'` which would mean not having `type`. We need more experience with byte streams before we can be certain.

Since we don't know what is needed yet, maybe we should just leave no extension points at all? Or maybe I'm misunderstanding?

domenic:
I guess it slipped my mind that asymmetrical transform streams were obviously necessary. I was envisioning a potential future where everything was symmetrical, but that's not realistic. So indeed this seems like the way to go.

PR #823, opened 2017-10-03
Remove TransformStreamDefaultSinkDefaultTransform

ricea:
TransformStreamDefaultSinkDefaultTransform was confusing and not very
useful. Replace it with TransformStreamDefaultSinkInvokeTransform, which
replaces the functionality of both that and
PromiseInvokeOrPerformFallback. Modify
TransformStreamDefaultSinkTransform to call the new function inside a try
block.

Remove PromiseInvokeOrPerformFallback as it no longer has any callers.

No semantic changes.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -299,14 +299,14 @@ function TransformStreamDefaultSinkInvokeTransform(stream, chunk) {
   const controller = stream._transformStreamController;
   const transformer = stream._transformer;
 
-  const method = transformer.transform; // throws
+  const method = transformer.transform; // can throw
 
   if (method === undefined) {
-    TransformStreamDefaultControllerEnqueue(controller, chunk); // throws
-    return undefined; // explicit "undefined" to keep eslint happy.
+    TransformStreamDefaultControllerEnqueue(controller, chunk); // can throw
+    return undefined;
   }
 
-  return Call(method, transformer, [chunk, controller]); // throws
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -295,9 +295,18 @@ class TransformStreamDefaultSink {
...
-  return Promise.resolve();
+function TransformStreamDefaultSinkInvokeTransform(stream, chunk) {
+  const controller = stream._transformStreamController;
+  const transformer = stream._transformer;
+
+  const method = transformer.transform; // throws
+
+  if (method === undefined) {
+    TransformStreamDefaultControllerEnqueue(controller, chunk); // throws
+    return undefined; // explicit "undefined" to keep eslint happy.
```
domenic: I don't think we need this comment. We do the same thing in lots of places.

reference-implementation/lib/transform-stream.js
```diff
@@ -295,9 +295,18 @@ class TransformStreamDefaultSink {
...
 }
 
-function TransformStreamDefaultSinkDefaultTransform(chunk, controller) {
-  TransformStreamDefaultControllerEnqueue(controller, chunk);
-  return Promise.resolve();
+function TransformStreamDefaultSinkInvokeTransform(stream, chunk) {
+  const controller = stream._transformStreamController;
+  const transformer = stream._transformer;
+
+  const method = transformer.transform; // throws
```
domenic: Although I realize they are emulating the spec annotations, these comments would be clearer as "can throw", IMO.

PR #829, opened 2017-10-06
Remove inaccurate note about ReadableStreamDefaultControllerEnqueue

### LINKED ISSUE #824: Note on ReadableStreamDefaultControllerEnqueue is inaccurate
https://streams.spec.whatwg.org/commit-snapshots/cd10872861aa35c0490f5d4ff46db7593e866a87#readable-stream-default-controller-enqueue

> In this case we allow the controller’s enqueue method to be called and silently do nothing

This hasn't been true since [e601d69](https://github.com/whatwg/streams/commit/e601d69f3ba40e6f4b5506e4bb1fbe3f3eabf393) in Dec 2015. The change was documented in the change description, and discussed in #424, so it was clearly intentional. The note was just never removed.

I prefer the new behaviour for `enqueue()` since I think it might be harmful to continue to run code on the assumption that the chunk has been queued. Plus this behaviour has shipped in browsers so changing it now could break running code. I propose removing the note.

ricea:
ReadableStreamDefaultControllerEnqueue has a note indicating that it
could be called when the stream was closed without an exception being
thrown. This hasn't been true since December 2015. Remove the inaccurate
note.

Fixes #824.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1851,6 +1851,13 @@ nothrow>ReadableStreamDefaultControllerCanCloseOrEnqueue ( <var>controller</var>
   1. Otherwise, return *false*.
 </emu-alg>
 
+<div class="note">
+  The case where <var>stream</var>.\[[closeRequested]] is <emu-val>false</emu-val>, but <var>stream</var>.\[[state]] is
+  not <code>"readable"</code>, happens when the stream is errored via {{ReadableStreamDefaultController/error(e)}}, or
+  when it is closed without its controller's <code>close</code> method ever being called: e.g., if the stream was closed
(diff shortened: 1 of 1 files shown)


domenic:
I think we should move the first half of this note to ReadableStreamDefaultControllerCanCloseOrEnqueue, as it remains a tricky situation. Something like

> The case where stream.[[closeRequested]] is false, but stream.[[state]] is not "readable", will happen if the stream was closed without its controller's close method ever being called: i.e., if the stream was closed by a call to cancel(reason).

ricea:
I tried to also cover the errored case in the note. PTAL.

PR #830, opened 2017-10-06
Switch from "assert" to "better-assert" in reference implementation

### LINKED ISSUE #807: Some assert()s have no description
This always causes me problems when I trip them, because I can't tell which assert fired.

Manually writing descriptions for all the asserts in the reference implementation would be very labour intensive, but a simple translation from assert(_condition_) to assert(_condition_, \`_condition_\`) can be done automatically. This is what I propose to do.

ricea:
The "better-assert" implementation of assert() includes the failed
predicate and stack trace in the output. This makes it much easier to
track down assertion failures.

No functional changes.

Resolves #807.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -1,5 +1,5 @@
 'use strict';
-const assert = require('assert');
+const assert = require('better-assert');
 
 const isFakeDetached = Symbol('is "detached" for our purposes');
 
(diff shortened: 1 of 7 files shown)


domenic:
(Rebased for you.) Do you think it's worth removing some now-redundant descriptions? Or is that not worth it? Seems like it'd be nice to keep things clean, but I can appreciate just wanting to make the change and be done.

LGTM, either way.

ricea:
I'm going to remove the descriptions in a separate PR so we can consider that change in isolation.

PR #831, opened 2017-10-06
Use "debug" package for logging

### LINKED ISSUE #499: Use some logging framework
https://github.com/whatwg/streams/pull/488#issuecomment-236500833

Such as https://www.npmjs.com/package/debug

ricea:
Use the "debug" package https://www.npmjs.com/package/debug for logging.

Debugging can be enabled by setting the DEBUG environment variable, for example

  DEBUG=streams:* npm test

This change converts existed commented-out log statements in
transform-streams.js to use the debug package. No new logging statements are
included in this change.

It is anticipated that new logging statements will be added as and when a need
for them is found.

Closes #499.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -1,5 +1,9 @@
 'use strict';
 const assert = require('better-assert');
+
+// Calls to verbose() are purely for debugging the reference implementation and tests. They are not part of the standard
+// and do not appear in the standard text.
+const verbose = require('debug')('streams:transform-stream:verbose');
 const { Call, InvokeOrNoop, PromiseInvokeOrNoop, typeIsObject } = require('./helpers.js');
 const { ReadableStream, ReadableStreamDefaultControllerClose, ReadableStreamDefaultControllerEnqueue,
         ReadableStreamDefaultControllerError, ReadableStreamDefaultControllerGetDesiredSize,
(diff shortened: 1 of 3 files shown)


reference-implementation/lib/transform-stream.js
```diff
@@ -334,7 +338,7 @@ class TransformStreamDefaultSource {
   }
 
   pull() {
-    // console.log('TransformStreamDefaultSource.pull()');
+    verbose('TransformStreamDefaultSource.pull()');
```
domenic: Can we name this TransformStreamDefaultSource.prototype.pull()?

reference-implementation/lib/transform-stream.js
```diff
@@ -241,7 +245,7 @@ class TransformStreamDefaultSink {
   }
 
   write(chunk) {
-    // console.log('TransformStreamDefaultSink.write()');
+    verbose('TransformStreamDefaultSink.write()');
```
domenic: Please let's not use Class.method() for non-static methods. Use .prototype. as appropriate.

reference-implementation/lib/transform-stream.js
```diff
@@ -88,7 +92,7 @@ function TransformStreamError(stream, e) {
 }
 
 function TransformStreamSetBackpressure(stream, backpressure) {
-  // console.log(`TransformStreamSetBackpressure(${backpressure})`);
+  verbose(`TransformStreamSetBackpressure(${backpressure})`);
```
domenic: Maybe `(stream, ${backpressure})` or `() [backpressure = ${backpressure}]` or similar? Right now the mismatch seems a bit confusing.

ricea: I went with the latter as it seems odd to have this method log argument names when the others don't.

reference-implementation/lib/transform-stream.js
ricea: Done.

reference-implementation/lib/transform-stream.js
ricea: Done.

ricea:
I added a "Diagnostics" section to README.md. PTAL.

PR #833, opened 2017-10-10
Editorial: address minor nits in the FAQ

annevk:
Also, instead of canceling, should we talk about aborting?


domenic:
Thanks!

Regarding canceling vs. aborting, we're making the not-super-great distinction that canceling means "loss of interest" and is not an error, whereas aborting signals an error. Readable streams can be canceled by their consumer, but not aborted.

PR #839, opened 2017-10-12
Initialise readableStrategy without using an object literal

ricea:
The strategy argument passed to the ReadableStream constructor within the
TransformStream argument used an object literal. The standard text only
uses object literal syntax in default arguments. In order to align the
reference implementation with the standard text, create the
`readableStrategy` object in explicit algorithm steps before passing it
to the constructor.

No functional changes.

### DIFF
--- a/reference-implementation/lib/transform-stream.js
+++ b/reference-implementation/lib/transform-stream.js
@@ -43,8 +43,10 @@ class TransformStream {
     });
 
     const source = new TransformStreamDefaultSource(this, startPromise);
-
-    this._readable = new ReadableStream(source, { size, highWaterMark });
+    const readableStrategy = {};
+    readableStrategy.size = size;
+    readableStrategy.highWaterMark = highWaterMark;
+    this._readable = new ReadableStream(source, readableStrategy);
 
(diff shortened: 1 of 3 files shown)


domenic:
I think you're actually going to want to use [CreateDataProperty](https://tc39.github.io/ecma262/#sec-createdataproperty), both here and in the standard text. Otherwise you'll trigger any setters on Object.prototype.

See e.g.

ricea:
Good point. I will change the standard text on Monday if you don't get to it first. I assume doing the same steps as https://tc39.github.io/ecma262/#sec-createiterresultobject would be appropriate?

Would the right way to do this in the reference implementation be to use Object.defineProperty()?

domenic:
Yes and yes. (Although you could also do ObjectCreate(null); it's unobservably different in this case.)

domenic:
Can we add a test to prevent any implementers from making this mistake?

ricea:
Test added. PTAL.

An oddity I noticed is that the test doesn't test anything under nodejs because transform-stream.js doesn't share Object.prototype with the page under test.

PR #842, opened 2017-10-16
Meta: Use pr-preview to preview pull requests

### LINKED ISSUE #836: [META] Integrate with pr-preview
[pr-preview](https://github.com/tobie/pr-preview) would automatically provide diffs of the visible HTML for standard changes. Example output: https://s3.amazonaws.com/pr-preview/whatwg/encoding/ee357f3...6bc288b.html

Unfortunately it doesn't appear to support emu-algify so the integration would be non-trivial. It looks like we'd need a web service to do the translation?

ricea:
Use the [pr-preview tool](https://tobie.github.io/pr-preview/) to automatically generate previews for pull requests.

Closes #836.

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,19 @@
+{
+    "src_file": "index.bs",
+    "type": "bikeshed",
+    "params": {
+        "md-status": "LS-COMMIT",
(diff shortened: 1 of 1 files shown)


ricea:
It appears that the emu-algify step is not run.

ricea:
That was just a bug in the preview tool. Landing it so that I can see if it works.

PR #843, opened 2017-10-16
Add SyncOneToOneTransformFunctionAdaptor example

ricea:
An example of a simple adaptor which wrap a synchronous 1:1 transform function
into a transformer object ready to construct a TransformStream.

***
[Preview](https://s3.amazonaws.com/pr-preview/ricea/streams/transform-stream-trivial-adaptor-example.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/streams/3c0a2ab...ricea:8414c24.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5150,52 +5150,6 @@ for the [promise]." But we don't have to worry about that when writing the <code
 function, since the stream implementation guarantees that the <a>underlying sink</a>'s <code>write</code> method will
 not be called until any promises returned by previous calls have fulfilled!
 
-<h3 id="example-ts-sync-mapper">A transform stream created from a sync mapper function</h3>
-
-The following function allows creating new {{TransformStream}} instances from synchronous "mapper" functions, of the
-type you would normally pass to {{Array.prototype/map|Array.prototype.map}}. It demonstrates that the API is concise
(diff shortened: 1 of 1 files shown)


ricea:
This is an idea for an example I had to demonstrate that doing simple things with the API is simple. It wouldn't be good as the first TransformStream example, as that would imply that you should always wrap the API :-)

It includes a couple of nice simple 1:1 transforms.

Open questions:

- Can we come up with a less terrible name for the adaptor?
- Should the adaptor return a TransformStream rather than returning a transformer?
- Should we include this example at all?

domenic:
This is a great example. I like the points you highlighted in the text: it shows that the API is simple even for trivial transforms, and it talks about backpressure vs. wastage, and how errors work.

I updated it a bit, most notably trying to play off the Array.prototype.map analogy. Let me know what you think. I also did indeed make the adapter return a TransformStream, as it seemed to clean things up in this example at least.

ricea:
dffdc10 lgtm.

I didn't notice the parallel with Array.prototype.map. That really makes everything a lot clearer.

domenic:
This is ready to merge IMO (after transform stream does), so assigning to you as a way of tracking that.

PR #844, opened 2017-10-17
Add a Random Values Generator Demonstration

abdulhannanali:
Adding a random values readable stream that generates a finites amount of random values using WebCrypto's getRandomValues method. This is a small demo, any suggestions to improve this demo are welcomed.

### DIFF
--- a/demos/index.html
+++ b/demos/index.html
@@ -8,14 +8,17 @@
 <dl>
   <dt><a href="append-child.html">Append child WritableStream</a>
   <dd>Piping through a series of transforms to a custom writable stream that appends children to
-  the DOM.
+  the DOM.</dd>
 
-  <dt><a href="streaming-element.html">Streaming element</a>
-  <dd>Using custom element to stream into a given DOM location.
+  <dt><a href="streaming-element.html">Streaming element</a></dt>
+  <dd>Using custom element to stream into a given DOM location.</dd>
 
   <dt><a href="streaming-element-backpressure.html">Streaming element with backpressure</a>
   <dd>A variation on the streaming element demo that adds backpressure to the element's writable
-  stream, decreasing jank.
+  stream, decreasing jank.</dd>
+
+  <dt><a href="./random-values-stream/">Random Values Stream</a></dt>
+  <dd>Piping a readable stream generating random values through a transform stream and then eventually writing it to the UI</dd>
 </dl>
 
 <p>Feel free to submit more demos by sending a pull request to the
--- a/demos/random-values-stream/index.js
+++ b/demos/random-values-stream/index.js
@@ -1,0 +1,217 @@
+/**
+ * Creates random values with the help of WebCrypto API
+ */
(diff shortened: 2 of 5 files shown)


abdulhannanali:
I am also working on another demo right now, and would want to give a demos section a rehaul arranging the demos in a more friendlier, extensible and helpful format.

demos/random-values-stream/index.js
```diff
@@ -0,0 +1,217 @@
...
+        start(controller) {
+            logStatusText('`start` method of the readable stream called')
+            this.startValueInterval(controller);
+        },
+
+        /**
+         * Starting the random values generation again after a certain period
+         * @param {*} controller 
+         */
+        async pull(controller) {
```
ricea: This is unnecessarily complicated. You don't need to track `desiredSize` yourself. As long as you have called `enqueue()` you can rely on the ReadableStream to call you back when there is no backpressure.

It would be simpler to do something like

```javascript
pull(controller) {
  if (maxValues && this.totalEnqueuedItemsCount >= maxValues) {
    controller.close();
    return;
  }
  let resolve;
  const promise = new Promise(r => {
    resolve = r;
  });
  setTimeout(() => {
    controller.enqueue(randomValuesUint8Array(20));
    ++this.totalEnqueuedItemsCount;
    resolve();
  }, valueInterval);
  return promise;
}
```

(untested).

demos/random-values-stream/index.js
```diff
@@ -0,0 +1,217 @@
...
+            logStatusText('Closing the stream');
+            console.log('Stream closed');
+        }
+    });
+
+    /**
+     * Writes a chunk to the span and appends it to the parent element
+     * @param {*} chunk 
+     */
+    async function writeChunk(chunk) {
```
ricea: This function doesn't need to be async as it performs no asynchronous operations. Also, as it is only called within write(), it might as well just be included in that function.

demos/random-values-stream/index.js
```diff
@@ -0,0 +1,217 @@
...
+        highWaterMark: 3,
+        size() { return 1; }
+    }
+
+    const writable = new WritableStream({
+        async write(chunk, controller) {
+            try {
+                await writeChunk(chunk);
+                return;
+            } catch (error) {
```
ricea: Not necessary to catch the exception. If write() throws, the stream will errored automatically.

demos/index.html
```diff
@@ -8,14 +8,17 @@
 <dl>
   <dt><a href="append-child.html">Append child WritableStream</a>
   <dd>Piping through a series of transforms to a custom writable stream that appends children to
-  the DOM.
+  the DOM.</dd>
```
ricea: The end tag is not needed, see https://html.spec.whatwg.org/multipage/grouping-content.html#the-dd-element.

I find it tidier to leave it out.

lastthyme (approved):

abdulhannanali:
Hi @ricea 

Thank you for reviewing this PR. I'll make the changes, I should also have that emoji stream converter on Github somewhere. I'll make a PR with that too.

PR #846, opened 2017-10-20
Add an example of a transform stream that replaces template tags

ricea:
Add an example of a TransformStream which replaces tags in a streaming template
according to the values of variables passed to the constructor.

***
[Preview](https://s3.amazonaws.com/pr-preview/ricea/streams/transform-stream-example-lipfuzz.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/streams/7457dd9...ricea:c942426.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5252,6 +5252,83 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
 
 </div>
 
+<h3 id="example-ts-lipfuzz">A transform stream that replaces template tags</h3>
+
+It's often useful to substitute tags with variables on a stream of data, where the parts that need to be replaced are
+small compared to the overall data size. This example presents a simple way to do that. It maps strings to strings,
+transforming a template like <code>"Time: \{{time}} Message: \{{message}}"</code> to <code>"Time: 15:36 Message:
+hello"</code> assuming that <code>{ time: "15:36", message: "hello" }</code> was passed in the
+<code>substitutions</code> parameter to <code>LipFuzzTransformer</code>.
+
+This example also demonstrates one way to deal with a situation where a chunk contains partial data that cannot be
+transformed until more data is received. In this case, a partial template tag will be accumulated in the
+<code>partialChunk</code> instance variable until either the end of the tag is found or the end of the stream is
+reached.
+
+<pre><code class="lang-javascript">
+  class LipFuzzTransformer {
+    constructor(substitutions) {
+      this.substitutions = substitutions;
+      this.partialChunk = '';
+      this.lastIndex = undefined;
+    }
+
+    transform(chunk, controller) {
+      chunk = this.partialChunk + chunk;
+      this.partialChunk = '';
+      // lastIndex is the index of the first character after the last substitution.
+      this.lastIndex = 0;
+      chunk = chunk.replace(/\{\{([a-zA-Z0-9_-]+)\}\}/g, this.replaceTag.bind(this));
+      // Regular expression for an incomplete template at the end of a string.
+      const partialAtEndRegexp = /\{(\{([a-zA-Z0-9_-]+(\})?)?)?$/g;
+      // Avoid looking at any characters that have already been substituted.
+      partialAtEndRegexp.lastIndex = this.lastIndex;
+      this.lastIndex = undefined;
+      const match = partialAtEndRegexp.exec(chunk);
+      if (match) {
+        this.partialChunk = chunk.substring(match.index);
+        chunk = chunk.substring(0, match.index);
+      }
+      controller.enqueue(chunk);
+    }
+
+    flush(controller) {
+      if (this.partialChunk.length > 0) {
+        controller.enqueue(this.partialChunk);
+      }
+    }
+
+    replaceTag(match, p1, offset) {
+      let replacement = this.substitutions[p1];
+      if (replacement === undefined) {
+        replacement = '';
+      }
+      this.lastIndex = offset + replacement.length;
+      return replacement;
+    }
+  }
+</code></pre>
+
+In this case we define the <a>transformer</a> to be passed to the {{TransformStream}} constructor as a class. This is
+useful when there is instance data to track.
+
+The class would be used in code like:
+
+<pre><code class="lang-javascript">
+  const data = { userName, displayName, icon, date };
+  const ts = new TransformStream(new LipFuzzTransformer(data));
+
+  fetchEvent.respondWith(
+    fetch(url).then(response => {
+      const transformedBody = response.body.pipeThrough(ts);
+      return new Response(transformedBody);
+    )
+  );
+</code></pre>
+
+<div class="warning">For simplicity, <code>LipFuzzTransformer</code> performs unescaped text substitutions. In real
+applications, a template system that performs context-aware escaping is good practice for security and robustness.</div>
+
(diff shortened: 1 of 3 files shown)


ricea:
Another proposed example for TransformStream. I like this, it's clearly useful and it demonstrates the use of flush() and how to deal with edge cases. But it has a few drawbacks:
- It's sufficiently tricky that it needs its own tests. We don't have a good place to put the tests. I've included them in this PR so you can inspect them.
- Some parts of the implementation may be tricky to understand.
- It's sufficiently useful that it needs a note warning you not to use it in production.

index.bs
```diff
@@ -4572,6 +4572,67 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
 
 </div>
 
+<h3 id="example-ts-lipfuzz">A transform stream that replaces template tags</h3>
+
+It's often useful to substitute tags with variables on a stream of data, where the parts that need to be replaced are
+small compared to the overall data size. This example presents a simple way to do that. It maps strings to strings,
+replacing a template like <code>"Time: <span>{</span>{time}} Message: <span>{</span>{message}}"</code> with <code>"Time:
```
domenic: "replacing a template" doesn't seem like the right verb. "transforming"? "interpolating"?

index.bs
```diff
@@ -4572,6 +4572,67 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+15:36 Message: hello"</code> assuming that <code>{ time: "15:36", message: "hello" }</code> was passed in the
+<code>substitutions</code> parameter to <code>LipFuzzTransform(substitutions)</code>.
+
+This example also demonstrates one way to deal with a situation where a chunk contains partial data that cannot be
+transformed until more data is received. In this case, a partial template tag will be accumulated in the
+<code>partialChunk</code> variable until either the end of the tag is found or the end of the stream is reached.
+
+<pre><code class="lang-javascript">
+  function LipFuzzTransform(substitutions) {
```
domenic: I wonder if we should demonstrate using a class for the transformer here, and storing the partialChunk as an instance variable? Then replaceFunc could also be a method, although it'd need to be .bind()ed unfortunately.

index.bs
```diff
@@ -4572,6 +4572,67 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+          return replacement;
+        }
+        chunk = chunk.replace(/\{\{([a-zA-Z0-9_-]+)\}\}/g, replaceFunc);
+        // Regular expression for an incomplete template at the end of a string.
+        const partialAtEndRegexp = /\{(\{([a-zA-Z0-9_-]+(\})?)?)?$/g;
+        // Avoid looking at any characters that have already been substituted.
+        partialAtEndRegexp.lastIndex = lastIndex;
+        const match = partialAtEndRegexp.exec(chunk);
+        if (match) {
+          partialChunk = chunk.substr(match.index);
```
domenic: substr() is "Annex B" (kinda-ish deprecated); stick to substring(), which in these cases will be equivalent I believe.

index.bs
domenic: Functions should be lowercase if they're not classes. Also, great name (it took me a minute :)

index.bs
```diff
@@ -4572,6 +4572,67 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+        }
+      }
+    });
+  }
+</code></pre>
+
+This would be used in code like:
+<pre><code class="lang-javascript">
+  fetchEvent.respondWith(new Response(fetch(url).then(response =>
+    response.body.pipeThrough(LipFuzzTransform({ userName, displayName, icon, date })))));
```
domenic: Give this code some room!! Also I don't think you can pass a promise to Response(). Maybe

```js
const data = { userName, displayName, icon, date };
const ts = lipFuzzTransform(data);

fetchEvent.respondWith(
  fetch(url).then(response => {
    const transformedBody = response.body.pipeThrough(ts);
    return new Response(transformedBody);
  )
);
```

reference-implementation/to-upstream-wpts/transform-streams/lipfuzz.js
```diff
@@ -0,0 +1,161 @@
...
+
+    flush(controller) {
+      if (partialChunk.length > 0) {
+        controller.enqueue(partialChunk);
+      }
+    }
+  });
+}
+
+const cases = [
```
domenic: This is fun :D

index.bs
```diff
@@ -4572,6 +4572,67 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+</code></pre>
+
+<div class="note">For simplicity, <code>LipFuzzTransform()</code> performs unescaped text substitutions. In real
```
domenic: Maybe class="warning"

index.bs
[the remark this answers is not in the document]
ricea: To prevent {{time}} and {{message}} from being expanded by Bikeshed. I couldn't find another way to do it.

ricea:
I tried making LipFuzzTransformer a class in a8c8c55 but I'm not sure the extra verbiage is adding anything. I've only changed lipfuzz.js for now, not the version in index.bs. What do you think?

index.bs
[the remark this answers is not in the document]
domenic: Huh, I would have thought backslashes would have helped. @tabatkins?

domenic:
Hmm. After seeing it I'm ambivalent on what's more readable. But, I think it's a kind of nice demonstration of class-based transformers, which inclines me toward keeping it. (It's the only such example we're currently slated to include, right?)

In the spec at least, we could omit `lipFuzzTransform` and just use `TransformStream(new LipFuzzTransformer(substitutions))` as a further demonstration of different styles.

Happy to let you make the call.

ricea:
We don't have an async transform stream example yet. And I think we need one. No good ideas have struck me yet. Once we have one we can look again at which fits the class-based transformer pattern better.

index.bs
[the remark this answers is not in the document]
ricea: When I tried backslashes I ended up with `\time` and `\message` in the output, which was not what I wanted. Probably something to do with the context.

domenic:
For an async one, an easy idea that comes to mind is expanding the mapper to allow async functions, e.g. `async transform(chunk, controller) { controller.enqueue(await mapper(chunk)); }`.

ricea:
I'd like to find something that compellingly demonstrates one-chunk-to-many behaviour, as we don't have much of that yet.

index.bs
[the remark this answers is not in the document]
tabatkins: No, I just don't have consistent support for backslashes in these things.  Let me go fix that.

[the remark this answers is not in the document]
tabatkins: Fixed in fda391df, thanks for reporting.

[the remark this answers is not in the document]
ricea: @tabatkins Works great, thank you.

ricea: I went with "transforming".

index.bs

index.bs
ricea: Resolved by replacing the function with a class :-)

index.bs
ricea: I often get confused between stream and promise-for-stream. Which inspired me to write the second example in #847.

index.bs

PR #847, opened 2017-10-27
Add examples of using identity transforms to manipulate streams

ricea:
Add examples concatenateReadables and promiseToReadable, both of which combine
pipeTo and TransformStream to synthesize a readable stream from something else.

The examples are similar so they appear in a single section.

Tests are included to verify the correct behaviour of the examples.

***
<a href="https://whatpr.org/streams/847.html" title="Last updated on Aug 26, 2020, 3:35 PM UTC (19604c5)">Preview</a> | <a href="https://whatpr.org/streams/847/f2be831...19604c5.html" title="Last updated on Aug 26, 2020, 3:35 PM UTC (19604c5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6863,6 +6863,67 @@ reader.read().catch(e => console.error(e));
 writer.write("{}").catch(e => console.error(e));
 </xmp>
 
+<h3 id="example-identity-transform-usages">Using an identity transform stream as a primitive to
+create new readable streams</h3>
+
+Combining an [=identity transform stream=] with {{pipeTo()}} is a powerful way to manipulate
+streams. This section contains a couple of examples of this general technique.
+
+It's sometimes natural to treat a promise for a [=readable stream=] as if it were a readable stream.
+A simple adapter function is all that's needed:
+
+<xmp highlight="js">
+function promiseToReadable(promiseForReadable) {
+  const ts = new TransformStream();
+
+  promiseForReadable
+      .then(readable => readable.pipeTo(ts.writable))
+      .catch(reason => ts.writable.abort(reason))
+      .catch(() => {});
+
+  return ts.readable;
+}
+</xmp>
+
+Here, we pipe the data to the [=writable side=] and return the [=readable side=]. If the pipe
+errors, we [=abort a writable stream|abort=] the writable side, which automatically propagates the
+error to the returned readable side. If the writable side had already been errored by
+{{ReadableStream/pipeTo()}}, then the {{WritableStream/abort()}} call will return a rejection, which
+we can safely ignore.
+
+A more complex extension of this is concatenating multiple readable streams into one:
+
+<xmp highlight="js">
+function concatenateReadables(readables) {
+  const ts = new TransformStream();
+  let promise = Promise.resolve();
+
+  for (const readable of readables) {
+    promise = promise.then(
+     () => readable.pipeTo(ts.writable, { preventClose: true }),
+     reason => {
+       return Promise.all([
+         ts.writable.abort(reason),
+         readable.cancel(reason)
+       ]);
+     }
+   );
+  }
+
+  promise.then(() => ts.writable.close(),
+               reason => ts.writable.abort(reason))
+         .catch(() => {});
+
+  return ts.readable;
+}
+</xmp>
+
+The error handling here is subtle because canceling the concatenated stream has to cancel all the
+input streams. However, the success case is simple enough. We just pipe each stream in the
+<code>readables</code> iterable one at a time to the [=identity transform stream=]'s [=writable
+side=], and then close it when we are done. The [=readable side=] is then a concatenation of all the
+chunks from all of of the streams. We return it from the function. Backpressure is applied as usual.
+
 <h2 id="acks" class="no-num">Acknowledgments</h2>
 
 The editors would like to thank


ricea:
Another proposed example.

This one probably needs more work to provide motivation for the examples.

index.bs
```diff
@@ -4572,6 +4572,55 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
 
 </div>
 
+<h3 id="example-ts-promise-to-readable">Using an identity transform stream as a primitive when manipulating streams</h3>
+
+Combining an identity transform stream with {{pipeTo()}} is a powerful way to manipulate streams. For example,
+concatenating multiple readable streams into one:
+
+<pre><code class="lang-javascript">
+function concatenateReadables(readables) {
```
domenic: Although we haven't used async functions much so far, in this case they give a pretty clear benefit:

```js
async function concatenateReadables(readables) {
  const ts = new TransformStream();
  
  for (const readable of readables) {
    try {
      await readable.pipeTo(ts.writable, { preventClose: true });
    } catch (e) {
      await Promise.all([
        ts.writable.abort(e),
        readable.cancel(e)
      ]);
      return;
    }
  }
  
  await ts.writable.getWriter().close();
  return ts.readable;
}
```

This departs from yours slightly in some error cases in ways that I think are probably improvements?

I'm also unsure that the catch() is necessary. Perhaps reading the tests will illuminate why you included it.

index.bs
```diff
@@ -4572,6 +4572,55 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+  }
+  promise.then(() => ts.writable.getWriter().close(),
+               reason => ts.writable.abort(reason))
+      .catch(() => {});
+  return ts.readable;
+}
+</code></pre>
+
+The error handling here is subtle because cancelling the concatenated stream has to cancel all the intput
+streams. However, the success case is simple enough. We just pipe each stream in the <code>readables</code> array one at
```
domenic: I think it might be better to not cancel them, but leave them for someone else to consume. Which may eliminate the catch block, or at least part of it.

Especially in an example.

index.bs
```diff
@@ -4572,6 +4572,55 @@ and similarly, closing or aborting the <code>writable</code> side will implicitl
...
+applied as usual.
+
+It's sometimes natural to treat a promise for a readable as if it were a readable. A simple adaptor function is all
+that's needed:
+
+<pre><code class="lang-javascript">
+function promiseToReadable(promiseForReadable) {
+  const ts = new TransformStream();
+  promiseForReadable.then(readable => readable.pipeTo(ts.writable))
+      .catch(reason => ts.writable.abort(reason))
```
domenic: Since we didn't pass preventAbort: true, why is this necessary?

reference-implementation/to-upstream-wpts/transform-streams/examples.js
```diff
@@ -0,0 +1,173 @@
...
+      controller.enqueue(2);
+      controller.error();
+    }
+  });
+  const readable = concatenateReadables([arrayToReadableStream([1]), erroredReadable]);
+  const reader = readable.getReader();
+  return reader.read().then(({ value, done }) => {
+    assert_false(done, 'done should be false');
+    assert_equals(value, 1, 'value should be 1');
+    // Error is not passed through due to WritableStreamAbort algorithm step 4.
```
domenic: Now I'm questioning whether abort not passing through the error is a good idea :(. I think this kind of example was one of the primary reasons I originally designed it to pass through. Oh well...

reference-implementation/to-upstream-wpts/transform-streams/examples.js
```diff
@@ -0,0 +1,173 @@
...
+
+promise_test(t => {
+  const erroredReadable = new ReadableStream({
+    pull() {
+      throw new Error('badness');
+    }
+  });
+  const promise = new Promise(resolve => resolve(erroredReadable));
+  const readable = promiseToReadable(promise);
+  return promise_rejects(t, new TypeError(), readable.getReader().read(), 'read() should reject');
```
domenic: Again sad we lost the error :(

index.bs
ricea: It can fail if `readable` doesn't implement pipeTo(), or if pipeTo() is implemented but throws, or readable is locked... probably some more cases too.

index.bs
ricea: It's not really helpful to leave some of them open, as the caller has no way to detect which ones have been closed.

Since this is an important primitive, I'd like to have a fully correct implementation, or no implementation at all. I expect in the future we will have a static method called `ReadableStream.concatenate()` or similar that does this exact thing.

index.bs
ricea: Using "async function" makes it return a promise. We could of course wrap it in a call to promiseToReadable(), but that would be poor for ergonomics.

reference-implementation/to-upstream-wpts/transform-streams/examples.js
ricea: Yes. I am also regretting making `writable.abort()` throw when the stream is errored, as it forces us to stick in empty catch blocks.

domenic:
Want to revisit this PR now that we've fixed a lot of the issues that make the code annoying?

ricea:
Will do.

ricea:
lgtm. We seem to have lost the tests somewhere along the way, but I am not concerned.

PR #849, opened 2017-10-31
Meta: use the common deploy.sh (with POST_BUILD_STEP)

foolip:
A deploy key is added and installed on both old and new servers, since
previously a password was used.


foolip:
Cool, it worked on the first try, https://streams.spec.whatwg.org/branch-snapshots/use-common-deploy/ is similar enough to https://streams.spec.whatwg.org/ source-wise that the emu-algify thing must have worked.

foolip:
(It's also been deployed to marquee, but of course it's not being served anywhere.)

foolip:
Ugh, there are still problems though.

demos/ was being copied before, although only to the root. I can fix that.

manifest.json, readable-stream.svg and tests/ also exist on the old server, but aren't part of the old deploy script. readable-stream.svg once existed in this repo, the others not. @domenic, what to do?

domenic:
Those can all be removed. tests/ used to be a primitive version of w3c-test.org for our non-WPT test format; I think it was a build product of the deploy. It was since removed but I guess I didn't clean up the server.

readable-stream.svg is a diagram we used to include but removed as it became outdated. Again I didn't clean the server.

And manifest.json was a failed experiment.

domenic:
Please also delete service-worker.js from this repository as it is now handled by the common script.

If it's OK, let's hold off on merging this until @ricea does a round of merging on transform stream stuff. (A big new feature that's about to land.)

ricea:
I've merged the transform stream stuff, so feel free to proceed.

foolip:
OK, I've made it so that the demos directory is copied, also to snapshots because that's easier. Ready for review and merging.

Issue #850, opened 2017-11-01
Constructing ReadableStream and WritableStream in TransformStream constructor


ricea:
Since we use internal abstract ops on the readable and writable sides we need to call **the original version** of the ReadableStream and WritableStream constructors in the TransformStream constructor.
- Do we need explicit spec language for this?
- If we do, what does it look like?

I also need to add a test for this.

domenic:
We've been implicit about this so far; we do it in lots of places. Tests would be good. The real fix for this case is #813, but for the other cases we should maybe add something to the conventions section.

ricea:
I thought the TransformStream constructor was the only place where we construct something that is exposed on the global object, but actually ReadableStreamTee has the same issue.

Would it be okay to use %ReadableStream% and %WritableStream% as is done for ES constructors?

domenic:
Ah, I see, you're specifically concerned about ones exposed on the global object.

We could use the %% business, but it's kind of a strange layer of indirection, so I want to think a bit more about if there's a nicer approach. I've always kind of hoped that by linking directly to the spec section in question it'd be clear we meant "that one", but that's pretty implicit.

ricea:
You mean that `%ReadableStream%` would have to link to something that says "the original value of ReadableStream", and then you'd have to follow that link to find the actual definition? That would indeed be bad.

Given that it only happens four times in total, I'd be content just to append "... where ReadableStream is the original value of the ReadableStream constructor" to the algorithm steps where it happens.

domenic:
Note that these occurrences are decreasing as #857 happens.

ricea:
Since #857 we explicitly use the initial value of FooStream.prototype when we create an object, so this issue is fixed.

Issue #851, opened 2017-11-02
Move transform stream tests to web platform tests


domenic:
It's time!

PR #852, opened 2017-11-02
Test that TransformStream does not use global constructors

ricea:
The TransformStream constructor constructs a ReadableStream and
WritableStream. It should use the original constructors rather than what
currently exists on the global object.

Add tests to verify that it does.

Related to issue #850.

### DIFF
--- a/reference-implementation/to-upstream-wpts/transform-streams/patched-global.js
+++ b/reference-implementation/to-upstream-wpts/transform-streams/patched-global.js
@@ -20,7 +20,7 @@ test(() => {
   assert_not_equals(new TransformStream(), null, 'constructor should work');
 }, 'TransformStream constructor should not call setters for highWaterMark or size');
 
-test(() => {
+test(t => {
   /* eslint-disable no-native-reassign */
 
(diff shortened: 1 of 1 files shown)


ricea:
I can move this to w3c if #851 happens first.

reference-implementation/to-upstream-wpts/transform-streams/patched-global.js
```diff
@@ -20,4 +20,34 @@ test(() => {
...
+
+  const ts = new TransformStream();
+
+  // Just to be sure, ensure the readable and writable pass brand checks.
+  assert_not_equals(getReader.call(ts.readable), undefined,
+                    'getReader should work when called on ts.readable');
+  assert_not_equals(getWriter.call(ts.writable), undefined,
+                    'getWriter should work when called on ts.writable');
+
+  ReadableStream = oldReadableStream;
```
domenic: These should probably be in a `t.cleanup(() => { ... })` at the top, just to be safe.

ricea: That's better, thanks.

PR #856, opened 2017-11-16
Prohibit setting a strategy size for byte streams

### LINKED ISSUE #729: new ReadableStream({type: 'bytes'}, new CountQueuingStrategy({highWaterMark: 5})) is weird
If you see CountQueuingStrategy when constructing a ReadableStream, you might expect it to be setting the queue size in chunks. But if `type` is `"bytes"`, it sets the queue size in bytes instead.

I was discussing this with @domenic and we concluded that new ReadableStream() should probably throw if the strategy passed to a byte stream has a size function.

ricea:
Readable byte streams always measure chunks in bytes, so specifying a size()
function in the strategy never did anything. This is confusing, so throw an
exception when size() is specified along with type: 'bytes'.

Update editorial descriptions to match the new behaviour.

Tests are in https://github.com/w3c/web-platform-tests/pull/8279.

Fixes #729.

***
[Preview](https://s3.amazonaws.com/pr-preview/ricea/streams/byte-streams-no-size-function.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/streams/99ffba7...ricea:64bdd25.html)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -245,8 +245,11 @@ For readable streams, an underlying source can use this desired size as a backpr
 generation so as to try to keep the desired size above or at zero. For writable streams, a producer can behave
 similarly, avoiding writes that would cause the desired size to go negative.
 
-Concretely, a queueing strategy is given by any JavaScript object with a <code>highWaterMark</code> property and a
-<code>size()</code> method. <!-- TODO: https://github.com/whatwg/streams/issues/427 -->
+Concretely, a queuing strategy is given by any JavaScript object with a <code>highWaterMark</code> property. For byte
+streams the <code>highWaterMark</code> always has units of bytes. For other streams the default unit is <a>chunks</a>,
+but a <code>size()</code> function can be included in the strategy object which returns the size for a given chunk. This
+permits the <code>highWaterMark</code> to be specified in arbitrary floating-point units.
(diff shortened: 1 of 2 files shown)


ricea:
@domenic The algorithmic changes are trivial, but the editorial parts are quite tricky.

index.bs
```diff
@@ -245,8 +245,11 @@ For readable streams, an underlying source can use this desired size as a backpr
 generation so as to try to keep the desired size above or at zero. For writable streams, a producer can behave
 similarly, avoiding writes that would cause the desired size to go negative.
 
-Concretely, a queueing strategy is given by any JavaScript object with a <code>highWaterMark</code> property and a
-<code>size()</code> method. <!-- TODO: https://github.com/whatwg/streams/issues/427 -->
+Concretely, a queuing strategy is given by any JavaScript object with a <code>highWaterMark</code> property. For byte
+streams the <code>highWaterMark</code> always has units of bytes. For other streams the default unit is <a>chunks</a>,
+but a <code>size()</code> function can be included in the strategy object which returns the size for a given chunk. This
```
domenic: Good correction, from method to function.

ricea: This wasn't actually intentional: I delete the text and rewrote it :-)

Issue #691, opened 2017-03-09
How much latitude should implementations have in looking up transformer.transform()?


ricea:
Pull request #689 verifies that transformer.transform is looked up in exactly the same way as the reference implementation. But this puts pretty severe restrictions on possible optimisations. For example, `new TransformStream()` can be optimised to skip Javascript altogether, but `new TransformStream({})` cannot. Here are some use cases for a TransformStream with no transform() method:

1. A TransformStream which adds extra buffering to a pipe.
2. A TransformStream which adds a header and/or footer to the stream but does nothing else.

I see two ways we could permit optimisation here:
1. Make the standard ambiguous about when (and how?) the "transform" attribute of transformer is looked up.
2. Change the reference implementation (and hence the standard) to look up the "transform" attribute in the constructor only.

I think for developers 2. is better than 1. because it will at least be consistent between implementations even if it's different from the way sources and sinks work.

The semantics we provide for sources and sinks are that we treat them exactly like an object, including expected Javascript behaviours like dynamically changing prototypes.

My feeling is that complying with this expectation is important, even though only a tiny proportion of developers may take advantage of the flexibility. On the other hand, falling off a performance cliff because you wrote `new TransformStream({})` instead of `new TransformStream()` is suboptimal. Making identity transforms always be slower just to avoid surprising performance behaviour would also be bad.

Thoughts?

tschneidereit:
I agree that we shouldn't require implementations to do a double lookup for
`transformer.transform`. Retaining the ability to easily perform stream
transforms without JS involvement is important.

I also agree that your proposed change 2) is better. Yes, it'd be different
from the source and sink handling, but it'd be in a very obvious way. That
is, as soon as you tried changing the transformer's `transform` method (be
it through a straight property set or a prototype change), you'd realize
that it doesn't have an effect.

Additionally, there isn't really a loss of flexibility: you can always e.g.
have the `transform` method forward to another function that's under your
control, or check a flag on the receiver.

domenic:
I'm loathe to make the constructor design inconsistent between the three classes. On the other hand, it seems like a low-risk change to cache all the methods on all the constructors, so if this is important, I think we should do that.

It's a bit weird from a JS perspective IMO. If we're treating these things as methods, the usual syntax for method calling is `foo.bar()`, not `previouslySavedBar.call(foo)`. But few people will notice.

We don't want to go back to just treating them as functions, as that breaks cases like https://streams.spec.whatwg.org/#example-both.

[...]

PR #860, opened 2017-11-30
Lookup methods in constructors

ricea:
Lookup methods on underlyingSource, underlyingSink and transformer in
the constructor. The main benefit is that identity transforms can be
simply implemented without touching Javascript. Other less critical
optimisation opportunities are unlocked for other stream types.

The algorithm construction for used-supplied methods is changed so that
a different algorithm is constructed depending on whether or not the
method was supplied. Two new abstract operations are added:

* GetMethod works like GetV but throws if the value is neither undefined
  nor a function.
* PromiseInvoke works like PromiseInvokeOrNoop but doesn't check for
  undefined.

PromiseInvokeOrNoop is no longer used and has been removed.

Constructors now throw for invalid method parameters, which results in
some test expectation changes. Changing the methods or prototype on the
underlying object after construction no longer does anything. Other than
these edge cases there are no web-developer visible changes.

Fixes https://github.com/whatwg/streams/issues/691.

***
<a href="https://whatpr.org/streams/860.html" title="Last updated on Dec 6, 2017, 9:32 AM GMT (b90b687)">Preview</a> | <a href="https://whatpr.org/streams/860/5b9f39d...b90b687.html" title="Last updated on Dec 6, 2017, 9:32 AM GMT (b90b687)">Diff</a>

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -83,7 +83,7 @@ exports.InvokeOrNoop = (O, P, args) => {
   return Call(method, O, args);
 };
 
-function PromiseInvoke(F, V, args) {
+function PromiseCall(F, V, args) {
   try {
     return Promise.resolve(Call(F, V, args));
   } catch (value) {
@@ -103,7 +103,7 @@ exports.CreateAlgorithmWithNoParametersFromUnderlyingMethod = (underlyingObject,
   const method = exports.GetMethod(underlyingObject, methodName);
   if (method !== undefined) {
     return () => {
-      return PromiseInvoke(method, underlyingObject, args);
+      return PromiseCall(method, underlyingObject, args);
     };
   }
   return () => Promise.resolve();
@@ -113,7 +113,7 @@ exports.CreateAlgorithmWithOneParameterFromUnderlyingMethod = (underlyingObject,
   const method = exports.GetMethod(underlyingObject, methodName);
   if (method !== undefined) {
     return arg => {
-      return PromiseInvoke(method, underlyingObject, [arg].concat(extraArgs));
+      return PromiseCall(method, underlyingObject, [arg].concat(extraArgs));
     };
   }
   return () => Promise.resolve();
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -83,13 +83,13 @@ exports.InvokeOrNoop = (O, P, args) => {
   return Call(method, O, args);
 };
 
-exports.PromiseInvoke = (F, V, args) => {
+function PromiseInvoke(F, V, args) {
   try {
     return Promise.resolve(Call(F, V, args));
   } catch (value) {
     return Promise.reject(value);
   }
-};
+}
 
 exports.GetMethod = (V, methodName) => {
   const method = V[methodName];
@@ -97,6 +97,26 @@ exports.GetMethod = (V, methodName) => {
     throw new TypeError(`${methodName} is not a function`);
   }
   return method;
+};
+
+exports.CreateAlgorithmWithNoParametersFromUnderlyingMethod = (underlyingObject, methodName, args) => {
+  const method = exports.GetMethod(underlyingObject, methodName);
+  if (method !== undefined) {
+    return () => {
+      return PromiseInvoke(method, underlyingObject, args);
+    };
+  }
+  return () => Promise.resolve();
+};
+
+exports.CreateAlgorithmWithOneParameterFromUnderlyingMethod = (underlyingObject, methodName, extraArgs) => {
+  const method = exports.GetMethod(underlyingObject, methodName);
+  if (method !== undefined) {
+    return arg => {
+      return PromiseInvoke(method, underlyingObject, [arg].concat(extraArgs));
+    };
+  }
+  return () => Promise.resolve();
 };
 
 // Not implemented correctly
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -70,6 +70,41 @@ function Call(F, V, args) {
 
 exports.Call = Call;
 
+exports.CreateAlgorithmFromUnderlyingMethod = (underlyingObject, methodName, algoArgCount, extraArgs) => {
+  assert(underlyingObject !== undefined);
+  assert(IsPropertyKey(methodName));
+  assert(algoArgCount === 0 || algoArgCount === 1);
+  assert(Array.isArray(extraArgs));
+  const method = exports.GetMethod(underlyingObject, methodName);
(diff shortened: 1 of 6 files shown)


ricea:
Only the reference implementation changes are done so far.

The algorithm construction is very repetitive. Part of the problem is that the controller is not available at the time we construct the algorithms. Suggestions welcome.

reference-implementation/lib/helpers.js
```diff
@@ -83,15 +83,20 @@ exports.InvokeOrNoop = (O, P, args) => {
   return Call(method, O, args);
 };
 
-exports.PromiseInvokeOrNoop = (O, P, args) => {
-  assert(O !== undefined);
-  assert(IsPropertyKey(P));
-  assert(Array.isArray(args));
+exports.PromiseInvoke = (F, V, args) => {
```
domenic: This is no longer an "Invoke", but instead a "Call". "Invokes" take (object, property name) pairs.

reference-implementation/lib/readable-stream.js
```diff
@@ -1156,12 +1156,22 @@ function SetUpReadableStreamDefaultControllerFromUnderlyingSource(stream, underl
     return InvokeOrNoop(underlyingSource, 'start', [stream._readableStreamController]);
   }
 
-  function pullAlgorithm() {
-    return PromiseInvokeOrNoop(underlyingSource, 'pull', [stream._readableStreamController]);
+  // eslint-disable-next-line func-style
```
domenic: I see what you mean about the controller not existing yet being a problem for algorithm construction. Can we just move the single line

```js
const controller = Object.create(ReadableStreamDefaultController.prototype);
```

before the algorithm creation? Then perhaps pass that as a parameter to the SetUpXYZ() abstract ops? I guess maybe also move the line

```js
stream._readableStreamController = controller;
```

and then we can pass in `controller` instead of `stream` to the SetUpXYZ() ops.

reference-implementation/lib/helpers.js
```diff
@@ -83,15 +83,20 @@ exports.InvokeOrNoop = (O, P, args) => {
...
-    return Promise.resolve(exports.InvokeOrNoop(O, P, args));
-  } catch (returnValueE) {
-    return Promise.reject(returnValueE);
+    return Promise.resolve(Call(F, V, args));
+  } catch (value) {
+    return Promise.reject(value);
+  }
+};
+
+exports.GetMethod = (V, methodName) => {
```
domenic: This might end up inlined into whatever helper is used for deduping the algorithm creation code.

reference-implementation/lib/readable-stream.js
```diff
@@ -1156,12 +1156,22 @@ function SetUpReadableStreamDefaultControllerFromUnderlyingSource(stream, underl
...
+  const pullMethod = GetMethod(underlyingSource, 'pull');
+  if (pullMethod !== undefined) {
+    pullAlgorithm = () => {
+      return PromiseInvoke(pullMethod, underlyingSource, [stream._readableStreamController]);
+    };
   }
 
-  function cancelAlgorithm(reason) {
-    return PromiseInvokeOrNoop(underlyingSource, 'cancel', [reason]);
+  // eslint-disable-next-line func-style
```
domenic: We should probably just disable this lint in the eslintrc, given that it's hurting us repeatedly in this way. I don't like one-off lint disables like this in general.

reference-implementation/lib/transform-stream.js
```diff
@@ -259,30 +259,44 @@ function SetUpTransformStreamDefaultController(stream, transformAlgorithm, flush
 }
 
 function SetUpTransformStreamDefaultControllerFromTransformer(stream, transformer) {
-  function transformAlgorithm(chunk) {
```
domenic: That got a lot simpler, yay.

reference-implementation/lib/helpers.js
ricea: Thanks, fixed.

reference-implementation/lib/helpers.js
ricea: It turns out TransformStream still needs it to look up the "transform" method.

reference-implementation/lib/readable-stream.js
ricea: This worked, thanks. I chose to pass both stream and controller and still leave tying them together to SetUpXYZ().

reference-implementation/lib/readable-stream.js
ricea: Done.

reference-implementation/lib/transform-stream.js
ricea: Not as simple as I'd like, but all the other algorithm creation steps are looking great now.

ricea:
There's still a linking error for GetV. Apart from that, I think it's pretty much done.

ricea:
I think this is ready to merge.

The wording "Let _fullArgs_ be a List consisting of _arg_ followed by the elements of _extraArgs_ in order." could possibly be improved. I couldn't find an appropriate operation defined in the ecmascript standard and I didn't want to write it out in terms of low-level element manipulations.

domenic:
Oh, hmmm. I realized we have our own GetMethod which is now different from the ES spec's. We should remove it, and I can do that from the spec side, but this is a slight normative change in that it allows `null` as equivalent to undefined. I don't really like that change, but it's probably better to be consistent with ES than to not be. We'll need to update the tests a bit.

domenic:
I wonder if we should also be consistent and allow `size` to be null? Gah, this kind of sucks... What do you think we should do?

domenic:
I see a few options:

- Back out my GetMethod changes, pretend for a few days that we didn't notice this problem, and solve it ASAP.
- Back out my GetMethod changes, rename our GetMethod to GetMethod2, open an issue to fix it, add a note in the spec pointing to that issue, and then merge this PR using GetMethod2
- Keep my GetMethod changes, and fix everything in this PR, probably including `size`. (The way I'd do that is by no longer doing the destructuring for the strategies, then maybe creating a mode for CreateAlgorithmFromUnderlyingMethod or a second abstract op, which can be used for both `start` and `size`.)

WDYT?

ricea:
I would be happier if I knew why the ES version of GetMethod squashes `null` to `undefined`. I can't make sense of it.
```javascript
new TransformStream({
  transform: null
});
```
doesn't look like a sensible way to create an identity transform, it just looks like nonsense to me. It's not like we intend for people to write `{ transform: undefined }`, there's just no point in disallowing it.

So, in the absence of a compelling justification for squashing `null` to `undefined`, I'd prefer to keep our existing semantics. I plan to proceed with option 4: inline GetMethod as it's only used in two places anyway.

PR #862, opened 2017-12-01
Change quote style in lipfuzz example

ricea:
All the examples use "" for string quotes. Change the lipfuzz example to
match.


domenic:
Although in other specs there is sometimes intentional style divergence in examples to show diversity, I too like conformity :)

PR #870, opened 2018-01-22
Throw type error in ReadableStreamBYOBRequest ctor if view isn't a non-detached ArrayBuffer view object

tschneidereit:
***
<a href="https://whatpr.org/streams/870.html" title="Last updated on Feb 9, 2018, 3:13 PM GMT (926dbc6)">Preview</a> | <a href="https://whatpr.org/streams/870/98b9be7...926dbc6.html" title="Last updated on Feb 9, 2018, 3:13 PM GMT (926dbc6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2041,7 +2041,10 @@ ReadableByteStreamController()</h4>
     1. Let _view_ be ! Construct(<a idl>%Uint8Array%</a>, « _firstDescriptor_.[[buffer]],
        _firstDescriptor_.[[byteOffset]] + _firstDescriptor_.[[bytesFilled]], _firstDescriptor_.[[byteLength]] −
        _firstDescriptor_.[[bytesFilled]] »).
-    1. Set *this*.[[byobRequest]] to ! Construct(`<a idl>ReadableStreamBYOBRequest</a>`, « *this*, _view_ »).
+    1. Let _byobRequest_ be ObjectCreate(the original value of `<a idl>ReadableStreamBYOBRequest</a>`'s `prototype`
+      property).
+    1. Perform ? SetUpReadableStreamBYOBRequest(_byobRequest_, _controller_, _view_).
+    1. Set *this*.[[byobRequest]] to _byobRequest_.
   1. Return *this*.[[byobRequest]].
 </emu-alg>
 
@@ -2199,9 +2202,7 @@ lt="ReadableStreamBYOBRequest(controller, view)">new
 ReadableStreamBYOBRequest(<var>controller</var>, <var>view</var>)</h4>
 
 <emu-alg>
-  1. If ! IsReadableByteStreamController(_controller_) is *false*, throw a *TypeError* exception.
-  1. Set *this*.[[associatedReadableByteStreamController]] to _controller_.
-  1. Set *this*.[[view]] to _view_.
+  1. Throw a *TypeError* exception.
 </emu-alg>
 
 <h4 id="rs-byob-request-prototype">Properties of the {{ReadableStreamBYOBRequest}} Prototype</h4>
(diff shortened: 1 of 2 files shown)


ricea:
I think the IsDetachedBuffer() check is unnecessary, because the ReadableByteStreamController doesn't actually care about the view contained inside the ReadableStreamBYOBRequest, and respond() will simply throw if IsDetachedBuffer() is true.

However, we have bigger problems if user code creates a ReadableStreamBYOBRequest. ReadableByteStreamControllerRespond does 

> Assert: controller.[[pendingPullIntos]] is not empty.

I suspect there are other, less obvious problems that could occur. I propose we change the constructor to always throw so that user code cannot create a ReadableStreamBYOBRequest for which  IsReadableStreamBYOBRequest() is true. Then we can have an abstract op we use internally to create one.

What do you think?

tschneidereit:
> However, we have bigger problems if user code creates a ReadableStreamBYOBRequest. ReadableByteStreamControllerRespond does

I realized that a while after opening this PR, yes. I'm reworking things to follow the pattern by which the controller is created from under a Stream's constructor, i.e. using `ObjectCreate` followed by `SetUpReadableStreamBYOBRequest`. Will update the PR with that once I'm done. (But might not be before Friday because tc39.)

tschneidereit:
> However, we have bigger problems if user code creates a ReadableStreamBYOBRequest. ReadableByteStreamControllerRespond does

I realized that a while after opening this PR, yes. I'm reworking things to follow the pattern by which the controller is created from under a Stream's constructor, i.e. using `ObjectCreate` followed by `SetUpReadableStreamBYOBRequest`. Will update the PR once I'm done on Friday or Monday.

tschneidereit:
@ricea, @domenic this new version does the `SetUpReadableStreamBYOBRequest` thing I mentioned above and should solve the issue in a much better way. Does this look good to you?

ricea:
lgtm. I don't think I'll have time to do the reference implementation changes today. I will add them tomorrow unless someone else does it first.

ricea:
Sorry, I'm behind schedule. I'm not going to manage to do the reference implementation changes today either.

index.bs
```diff
@@ -2041,7 +2041,10 @@ ReadableByteStreamController()</h4>
     1. Let _view_ be ! Construct(<a idl>%Uint8Array%</a>, « _firstDescriptor_.[[buffer]],
        _firstDescriptor_.[[byteOffset]] + _firstDescriptor_.[[bytesFilled]], _firstDescriptor_.[[byteLength]] −
        _firstDescriptor_.[[bytesFilled]] »).
-    1. Set *this*.[[byobRequest]] to ! Construct(`<a idl>ReadableStreamBYOBRequest</a>`, « *this*, _view_ »).
+    1. Let _byobRequest_ be ObjectCreate(the original value of `<a idl>ReadableStreamBYOBRequest</a>`'s `prototype`
+      property).
+    1. Perform ? SetUpReadableStreamBYOBRequest(_byobRequest_, _controller_, _view_).
```
ricea: Should be '!' not '?' as SetUpReadableStreamBYOBRequest performs no operations that could throw. Should be \*this*, not \_controller_.

I will fix.

ricea:
@domenic I think this is ready once the wpt change https://github.com/w3c/web-platform-tests/pull/9457 is landed . PTAL.

domenic:
Changes LGTM, but can we also add tests for the various bad cases we're trying to fix here? In particular tests that the constructor throws, and maybe if it's possible to get down this code path via allocated-by-the-implementation BYOB requests, a test that allocated-by-the-implementation BYOB requests don't suffer the same failure modes. (The latter might already be tested though.)

ricea:
@domenic Good point. I will write the tests next week. Do you want to wait for the tests to land this?

domenic:
Nah we can land this now, as long as we don't forget :). Since the spec was just broken before, better to get it fixed sooner. (Whereas if this was a normative change from one working state to another, we should wait.)

PR #871, opened 2018-01-26
Export "readable stream", "writable stream" and "chunk"

ricea:
Add export annotations to the definitions for the concepts "readable
stream", "writable stream" and "chunk" so that they can be used from
other standards.

***
<a href="https://whatpr.org/streams/871.html" title="Last updated on Jan 26, 2018, 8:38 AM GMT (6de2ade)">Preview</a> | <a href="https://whatpr.org/streams/871/98b9be7...6de2ade.html" title="Last updated on Jan 26, 2018, 8:38 AM GMT (6de2ade)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -85,14 +85,13 @@ ecosystem to grow around these shared and composable interfaces.
 
 <h2 id="model">Model</h2>
 
-A <dfn>chunk</dfn> is a single piece of data that is written to or read from a stream. It can be of any type; streams
-can even contain chunks of different types. A chunk will often not be the most atomic unit of data for a given stream;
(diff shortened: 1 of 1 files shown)


ricea:
I need these for the encoding standard (https://github.com/whatwg/encoding/pull/127). I think "readable stream" and "writable stream" are perfectly reasonable to export, but I'm not sure about "chunk". Opinions welcome.

domenic:
LGTM. I haven't heard anyone else use "chunk" on the platform, so we can claim it un-namespaced. (If we were to namespace it's a simple matter of putting for=stream on it.)

PR #872, opened 2018-01-26
Add more cross-links

surma:
**This PR does not contain functional changes.**

I found the spec to be a bit frustrating to browse when I was trying to figure out which methods are on a stream, a stream controller or the underlying sink/source. I am linkifying a couple of things to make the spec more navigational.

Please mind that this is my first work with bikeshed so I might not be doing it right, don’t hesitate to make me correct things.

PTAL :)

***
<a href="https://whatpr.org/streams/872.html" title="Last updated on Jan 31, 2018, 5:32 PM GMT (322ad72)">Preview</a> | <a href="https://whatpr.org/streams/872/1ea1c84...322ad72.html" title="Last updated on Jan 31, 2018, 5:32 PM GMT (322ad72)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -95,7 +95,7 @@ A <dfn export>readable stream</dfn> represents a source of data, from which you 
 <em>out</em> of a readable stream. Concretely, a readable stream is an instance of the {{ReadableStream}} class.
 
 Although a readable stream can be created with arbitrary behavior, most readable streams wrap a lower-level I/O source,
-called the <dfn>underlying source</dfn> ([[#rs-constructor|interface description]]). There are two types of underlying source: push sources and pull sources.
+called the <dfn>underlying source</dfn>. There are two types of underlying source: push sources and pull sources.
 
 <dfn lt="push source">Push sources</dfn> push data at you, whether or not you are listening for it. They may also
 provide a mechanism for pausing and resuming the flow of data. An example push source is a TCP socket, where data is
@@ -105,9 +105,9 @@ constantly being pushed from the OS level, at a rate that can be controlled by c
 synchronously, e.g. if it is held by the operating system's in-memory buffers, or asynchronously, e.g. if it has to be
 read from disk. An example pull source is a file handle, where you seek to specific locations and read specific amounts.
 
-Readable streams are designed to wrap both types of sources behind a single, unified interface. For
-web developer–created streams, the implementation details of a source are provided by an object with certain methods and
-properties that is passed to the {{ReadableStream()}} constructor.
+Readable streams are designed to wrap both types of sources behind a single, unified interface. For web
+developer–created streams, the <a href="#rs-constructor">implementation details of a source</a> are provided by an
+object with certain methods and properties that is passed to the {{ReadableStream()}} constructor.
 
 <a>Chunks</a> are enqueued into the stream by the stream's <a>underlying source</a>. They can then be read one at a
 time via the stream's public interface, in particular by using a <a>readable stream reader</a> acquired using the
@@ -137,12 +137,12 @@ A <dfn export>writable stream</dfn> represents a destination for data, into whic
 goes <em>in</em> to a writable stream. Concretely, a writable stream is an instance of the {{WritableStream}} class.
 
 Analogously to readable streams, most writable streams wrap a lower-level I/O sink, called the
-<dfn>underlying sink</dfn> ([[#ws-constructor|interface description]]). Writable streams work to abstract away some of the complexity of the underlying sink, by
+<dfn>underlying sink</dfn>. Writable streams work to abstract away some of the complexity of the underlying sink, by
 queuing subsequent writes and only delivering them to the underlying sink one by one.
 
 <a>Chunks</a> are written to the stream via its public interface, and are passed one at a time to the stream's
-<a>underlying sink</a>. For web developer-created streams, the implementation details of the sink are provided by an
-object with certain methods that is passed to the {{WritableStream()}} constructor.
+<a>underlying sink</a>. For web developer-created streams, the <a href="#ws-constructor">implementation details of the
+sink</a> are provided by an object with certain methods that is passed to the {{WritableStream()}} constructor.
 
 Code that writes into a writable stream using its public interface is known as a <dfn>producer</dfn>.
 
@@ -385,15 +385,16 @@ like
 
 <pre><code class="lang-javascript">
   class ReadableStream {
-    <l>[[#rs-constructor|constructor]]</l>(underlyingSource = {}, { size, highWaterMark } = {})
+    <a href="#rs-constructor">constructor</a>(<a href="#rs-constructor">underlyingSource</a> = {}, { size, highWaterMark
+    } = {})
 
-    get <l>[[#rs-locked|locked]]</l>()
+    get <a href="#rs-locked">locked</a>()
 
-    <l>[[#rs-cancel|cancel]]</l>(reason)
-    <l>[[#rs-get-reader|getReader]]</l>()
-    <l>[[#rs-pipe-through|pipeThrough]]</l>({ writable, readable }, options)
-    <l>[[#rs-pipe-to|pipeTo]]</l>(dest, { preventClose, preventAbort, preventCancel } = {})
-    <l>[[#rs-tee|tee]]</l>()
+    <a href="#rs-cancel">cancel</a>(reason)
+    <a href="#rs-get-reader">getReader</a>()
+    <a href="#rs-pipe-through">pipeThrough</a>({ writable, readable }, options)
+    <a href="#rs-pipe-to">pipeTo</a>(dest, { preventClose, preventAbort, preventCancel } = {})
+    <a href="#rs-tee">tee</a>()
   }
 </code></pre>
 
@@ -422,7 +423,7 @@ Instances of {{ReadableStream}} are created with the internal slots described in
   <tr>
     <td>\[[reader]]
     <td class="non-normative">A {{ReadableStreamDefaultReader}} or {{ReadableStreamBYOBReader}} instance, if the stream
-      is <a>locked to a reader</a>, or <emu-val>undefined</emu-val> if it is not
+    is <a>locked to a reader</a>, or <emu-val>undefined</emu-val> if it is not
   </tr>
   <tr>
     <td>\[[state]]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -444,11 +444,11 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
   govern how the constructed stream instance behaves:
 
   <ul>
-    <li><p><code>start(controller)</code> is called immediately, and is typically used to adapt a <a>push
+    <li><p><code>start({{ReadableStreamDefaultController|controller}})</code> is called immediately, and is typically used to adapt a <a>push
```
surma: This one is a bit questionable because it could also be a BYOB, but I think having it clickable is worth the simplification.

index.bs
```diff
@@ -95,7 +95,7 @@ A <dfn export>readable stream</dfn> represents a source of data, from which you
 <em>out</em> of a readable stream. Concretely, a readable stream is an instance of the {{ReadableStream}} class.
 
 Although a readable stream can be created with arbitrary behavior, most readable streams wrap a lower-level I/O source,
-called the <dfn>underlying source</dfn>. There are two types of underlying source: push sources and pull sources.
+called the <dfn>underlying source</dfn> ([[#rs-constructor|interface description]]). There are two types of underlying source: push sources and pull sources.
```
domenic: This isn't good, because we're currently talking about the general concept of underlying sources, not the JS objects that sometimes (for web-developer-created streams) embody them. Such a link belongs later, where we say "For web developer–created streams, the implementation details of a source are provided by an object with certain methods and properties that is passed to the ReadableStream() constructor." But, that already links to the constructor, so I'm not sure if there's anything to do here.

index.bs
```diff
@@ -137,7 +137,7 @@ A <dfn export>writable stream</dfn> represents a destination for data, into whic
 goes <em>in</em> to a writable stream. Concretely, a writable stream is an instance of the {{WritableStream}} class.
 
 Analogously to readable streams, most writable streams wrap a lower-level I/O sink, called the
-<dfn>underlying sink</dfn>. Writable streams work to abstract away some of the complexity of the underlying sink, by
+<dfn>underlying sink</dfn> ([[#ws-constructor|interface description]]). Writable streams work to abstract away some of the complexity of the underlying sink, by
```
domenic: Same problem as above.

index.bs
```diff
@@ -385,15 +385,15 @@ like
 
 <pre><code class="lang-javascript">
   class ReadableStream {
-    constructor(underlyingSource = {}, { size, highWaterMark } = {})
+    <l>[[#rs-constructor|constructor]]</l>(underlyingSource = {}, { size, highWaterMark } = {})
```
domenic: Personally I find this more confusing than just doing `<a href="#rs-constructor">constructor</a>`. Does that work, or is only `<l>` allowed inside `<pre>` blocks?

index.bs
domenic: I'm not really OK with linking variable names to their types in general, here or elsewhere. (Especially, as you say, when their types are not accurate.) If you think it's unclear what the type of the controller is, we should add a more explicit sentence explaining.

This occurs several times below; I'll avoid noting it each time.

index.bs
```diff
@@ -385,15 +385,15 @@ like
 
 <pre><code class="lang-javascript">
```
domenic: The introduction of links into these blocks is a great improvement; thank you!

index.bs
surma: I see your point. The problem that I’m trying to solve here is the following flow:

1. Go to stream spec
1. Click “ReadableStream” in ToC
1. Get confused, maybe click `ReadableStream` again?
1. Okay, I’m at the class, I see `underlyingSource`, but I can’t click it. The paragraph has a clickable “underlying Source”, tho. So let’s click that.
1. I was already here...

Maybe we can turn `underlyingSource` in the class definition into a link to the note? Would that be acceptable?

index.bs
surma: Just tested it. Works as well.

index.bs
surma: That’s fair. I’ll take a stab at rephrasing the text so I can add some links to that instead.

surma:
Fixed line length and moved some of the links away from the code blocks into the descriptive text. PTAL :)

index.bs
```diff
@@ -4028,15 +4031,18 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
...
+      Those chunks will be read from the <a>readable side</a> but don't depend on any writes to the <a>writable
+      side</a>. If this process is asynchronous, it can return a promise to signal success or failure.
 
     <li>
       <p><code>transform(chunk, controller)</code> is called when a new chunk originally written to the writable side is
       ready to be transformed. It can return a promise to signal success or failure of the transformation. The results
       of the transformation can be enqueued to the readable side using the
-      {{ReadableStreamDefaultController/enqueue(chunk)|controller.enqueue()}} method. This permits a single chunk
-      written to the writable side to result in zero or multiple chunks on the readable side.
+      {{TransformStreamDefaultController|controller}}.{{TransformStreamDefaultController/enqueue(chunk)|enqueue()}}
```
surma: What do you think about splitting the links to contain both the class definition as well as the method definition?

index.bs
```diff
@@ -385,15 +385,16 @@ like
 
 <pre><code class="lang-javascript">
   class ReadableStream {
-    constructor(underlyingSource = {}, { size, highWaterMark } = {})
+    <a href="#rs-constructor">constructor</a>(<a href="#rs-constructor">underlyingSource</a> = {}, { size, highWaterMark
```
surma: I realize linking `underlyingSource` to `ReadableStream()` could be seen as questionable. This is just as suggestion. My thinking was that if someone’s looking at the class definition and the constructor, they are most likely a web developer and looking for an interface description. There’s multiple links to `underlyingSource` in the note, so there’s still a path to the definition.

index.bs
surma: Additionally, was it intentional that this linked to `ReadableStreamDefaultController.enqueue()` instead of `TransformStreamDefaultController.enqueue()`

ricea: I'm 99% sure that linking to `ReadableStreamDefaultController.enqueue()` was a mistake here. Thanks for catching it.

Splitting the links looks okay to me, but I will defer to @domenic who knows much more about correct markup fro WhatWG standards.

index.bs
```diff
@@ -105,9 +105,9 @@ constantly being pushed from the OS level, at a rate that can be controlled by c
 synchronously, e.g. if it is held by the operating system's in-memory buffers, or asynchronously, e.g. if it has to be
 read from disk. An example pull source is a file handle, where you seek to specific locations and read specific amounts.
 
-Readable streams are designed to wrap both types of sources behind a single, unified interface. For
-web developer–created streams, the implementation details of a source are provided by an object with certain methods and
-properties that is passed to the {{ReadableStream()}} constructor.
+Readable streams are designed to wrap both types of sources behind a single, unified interface. For web
+developer–created streams, the <a href="#rs-constructor">implementation details of a source</a> are provided by an
```
domenic: Still not sure this helps that much compared to just clicking on the already-existing "ReadableStream() constructor" link, but I'm OK keeping it if you think so.

index.bs
```diff
@@ -385,15 +385,15 @@ like
 
 <pre><code class="lang-javascript">
   class ReadableStream {
-    constructor(underlyingSource = {}, { size, highWaterMark } = {})
+    <a href="#rs-constructor">constructor</a>(<a href="#rs-constructor">underlyingSource</a> = {}, { size, highWaterMark } = {})
```
domenic: Again the second link to the exact same place feels pretty redundant.

Maybe what this is pointing toward is the need for a dedicated section describing the underlying source objects, instead of putting it into the constructor's intro note? I guess same for the strategies too? I'd be happy to work on that as a follow-up. In the meantime I'm indifferent as to whether we include these redundant links or not.

index.bs
```diff
@@ -2778,12 +2778,13 @@ like
 
 <pre><code class="lang-javascript">
   class WritableStream {
-    constructor(underlyingSink = {}, { size, highWaterMark = 1 } = {})
+    <a href="#ws-constructor">constructor</a>(<a href="#ws-constructor">underlyingSink</a> = {}, { size, highWaterMark =
+    1 } = {})
```
domenic: I forgot that line wrapping is not a good idea inside `<pre>` blocks; sorry, we should undo this.

index.bs
```diff
@@ -4028,15 +4030,18 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
 
   <ul>
     <li><p><code>start(controller)</code> is called immediately, and is typically used to enqueue prefix <a>chunks</a>
-      that will be read from the <a>readable side</a> but don't depend on any writes to the <a>writable side</a>. If
-      this process is asynchronous, it can return a promise to signal success or failure.
+      using
+      {{TransformStreamDefaultController|controller}}.{{TransformStreamDefaultController/enqueue(chunk)|enqueue()}}.
```
domenic: I'd rather not have separate links for the controller since that's just a variable name, and it is described below. So just changing the linking text from enqueue() to controller.enqueue() is probably best here.

index.bs
surma: It mostly because of the point below, that as a reader it’s not obvious that the underlying source is defined at the constructor. I’d prefer leaving this link for now until we give `underlyingSource` its own section.

index.bs
surma: The more I work on this, the more I think that’s the actual crux: Something as integral to developing streams shouldn’t just be defined in a note. So I’m up for formalizing `underlyingSource` and `underlyingSink`, but defo need some guidance what the right formalism is here.

I’ll remove the links.

index.bs
```diff
@@ -4028,15 +4029,18 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
...
+      Those chunks will be read from the <a>readable side</a> but don't depend on any writes to the <a>writable
+      side</a>. If this process is asynchronous, it can return a promise to signal success or failure.
 
     <li>
       <p><code>transform(chunk, controller)</code> is called when a new chunk originally written to the writable side is
       ready to be transformed. It can return a promise to signal success or failure of the transformation. The results
       of the transformation can be enqueued to the readable side using the
-      {{ReadableStreamDefaultController/enqueue(chunk)|controller.enqueue()}} method. This permits a single chunk
-      written to the writable side to result in zero or multiple chunks on the readable side.
+      {{TransformStreamDefaultController|controller}}.{{TransformStreamDefaultController/enqueue(chunk)|enqueue()}}
```
domenic: Also change this one?

Issue #874, opened 2018-01-31
"await writer.write(bytes);" in example seems wrong


domenic:
https://streams.spec.whatwg.org/#example-manual-write-with-backpressure

We're already awaiting writer.ready. Awaiting writer.write() seems counterproductive, as it means that even if the writable stream has a large high water mark, we won't be writing multiple chunks to fill the queue, instead waiting until every individual chunk has been processed sequentially.

ricea:
Sorry, I really should have responded to this sooner. Because the `write()` always uses all of `desiredSize`, desiredSize will go to 0 and stay at 0 until the write finishes. So `writer.ready` cannot fulfill before `writer.write()`.

Additionally, if we don't wait on `writer.write()`, there will be an uncaught rejection, which is not ideal. My preferred way to write the loop is:

```javascript
  await writer.ready; // ensure desiredSize > 0
  while (true) {
    const bytes = new Uint8Array(writer.desiredSize);
    window.crypto.getRandomValues(bytes);

    await writer.write(bytes);
  }
```

Except now it's not a good example of `writer.ready`. Okay, how about

```javascript
  while (true) {
    await writer.ready;
    const bytes = new Uint8Array(1024);
    window.crypto.getRandomValues(bytes);

    writer.write(bytes).catch(() => {});
  }
```

?  Oh, but we're also trying to demonstrate `desiredSize`. Never mind, let's just use `write().catch()` instead of `await write()`.

By the way, why is it `window.crypto.getRandomValues()`? If it was just `crypto.getRandomValues()` it would work in a worker without modification.

domenic:
Oh, right, in this particular example it is the same, good point. Hmm. I'd still like to not encourage waiting on writer.write() as except in cases where you're explicitly using desiredSize, it's not a good idea.

This may benefit from two examples, one to demonstrate desiredSize and one to demonstrate how if you don't use desiredSize, ready and write() will get out of sync.

And yeah, we should get rid of the `window`.

PR #875, opened 2018-01-31
A variety of clarity and documentation improvements

domenic:
/cc @lgrahl @surma who inspired these changes.

Proofreading and editorial suggestions definitely welcome; this took more work than I anticipated and I'm sure I missed some things.

***
<a href="https://whatpr.org/streams/875.html" title="Last updated on Feb 22, 2018, 9:27 PM GMT (031d1e9)">Preview</a> | <a href="https://whatpr.org/streams/875/634b181...031d1e9.html" title="Last updated on Feb 22, 2018, 9:27 PM GMT (031d1e9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2729,7 +2729,7 @@ nothrow>SetUpReadableStreamBYOBRequest ( <var ignore>request</var>, <var>control
   <pre><code class="lang-javascript">
     function writeArrayToStream(array, writableStream) {
       const writer = writableStream.getWriter();
-      array.forEach(chunk => writer.write(chunk));
+      array.forEach(chunk => writer.write(chunk).catch(() => {}));
 
       return writer.close();
     }
@@ -2738,6 +2738,11 @@ nothrow>SetUpReadableStreamBYOBRequest ( <var ignore>request</var>, <var>control
       .then(() => console.log("All done!"))
       .catch(e => console.error("Error with the stream: " + e));
   </code></pre>
+
+  Note how we use <code>.catch(() => {})</code> to suppress any rejections from the
+  {{WritableStreamDefaultWriter/write()}} method; we'll be notified of any fatal errors via a rejection of the
+  {{WritableStreamDefaultWriter/close()}} method, and leaving them un-caught would cause potential
+  {{unhandledrejection}} events and console warnings.
 </div>
 
 <div class="example" id="example-manual-write-with-error-handling">
@@ -2757,7 +2762,8 @@ nothrow>SetUpReadableStreamBYOBRequest ( <var ignore>request</var>, <var>control
 
   What "success" means is up to a given stream instance (or more precisely, its <a>underlying sink</a>) to decide. For
   example, for a file stream it could simply mean that the OS has accepted the write, and not necessarily that the
-  chunk has been flushed to disk.
+  chunk has been flushed to disk. Some streams might not be able to give such a signal at all, in which case the
+  returned promise will fulfill immediately.
 </div>
 
 <div class="example" id="example-manual-write-with-backpressure">
@@ -2775,14 +2781,53 @@ nothrow>SetUpReadableStreamBYOBRequest ( <var ignore>request</var>, <var>control
       await writer.ready;
 
       const bytes = new Uint8Array(writer.desiredSize);
-      window.crypto.getRandomValues(bytes);
+      crypto.getRandomValues(bytes);
 
-      await writer.write(bytes);
+      // Purposefully don't await; awaiting writer.ready is enough.
+      writer.write(bytes).catch(() => {});
     }
   }
 
   writeRandomBytesForever(myWritableStream).catch(e => console.error("Something broke", e));
   </code></pre>
+
+  Note how we don't <code>await</code> the promise returned by {{WritableStreamDefaultWriter/write()}}; this would be
+  redundant with <code>await</code>ing the {{WritableStreamDefaultWriter/ready}} promise. Additionally, similar to <a
+  href="#example-manual-write-batch">a previous example</a>, we use the <code>.catch(() => {})</code> pattern on the
+  promises returned by {{WritableStreamDefaultWriter/write()}}; in this case we'll be notified about any failures
+  <code>await</code>ing the {{WritableStreamDefaultWriter/ready}} promise.
+</div>
+
+<div class="example" id="example-manual-write-dont-await">
+  To further emphasize how it's a bad idea to <code>await</code> the promise returned by
+  {{WritableStreamDefaultWriter/write()}}, consider a modification of the above example, where we continue to use the
+  {{WritableStreamDefaultWriter}} interface directly, but we don't control how many bytes we have to write at a given
+  time. In that case, the <a>backpressure</a>-respecting code looks the same:
+
+  <pre><code class="lang-javascript">
+  async function writeSuppliedBytesForever(writableStream, getBytes) {
+    const writer = writableStream.getWriter();
+
+    while (true) {
+      await writer.ready;
+
+      const bytes = getBytes();
+      writer.write(bytes).catch(() => {});
+    }
+  }
+  </code></pre>
+
+  Unlike the previous example, where—because we were always writing exactly
+  {{WritableStreamDefaultWriter/desiredSize|writer.desiredSize}} bytes each time—the
+  {{WritableStreamDefaultWriter/write()}} and {{WritableStreamDefaultWriter/ready}} promises were synchronized, in this
+  case it's quite possible that the {{WritableStreamDefaultWriter/ready}} promise fulfills before the one returned by
+  {{WritableStreamDefaultWriter/write()}} does. Remember, the {{WritableStreamDefaultWriter/ready}} promise fulfills
+  when the <a lt="desired size to fill a stream's internal queue">desired size</a> becomes positive, which might be
+  before the write succeeds (especially in cases with a larger <a>high water mark</a>).
+
+  In other words, <code>await</code>ing the return value of {{WritableStreamDefaultWriter/write()}} means you never
+  queue up writes in the stream's <a>internal queue</a>, instead only executing a write after the previous one succeeds,
+  which can result in low throughput.
 </div>
 
 <h3 id="ws-class" interface lt="WritableStream">Class <code>WritableStream</code></h3>
@@ -5531,6 +5576,7 @@ Jake Archibald,
 Jake Verbaten,
 Janessa Det,
 Jens Nockert,
+Lennart Grahl,
 Mangala Sadhu Sangeet Singh Khalsa,
 Marcos Caceres,
 Marvin Hagemeister,
--- a/index.bs
+++ b/index.bs
@@ -111,8 +111,8 @@ synchronously, e.g. if it is held by the operating system's in-memory buffers, o
 read from disk. An example pull source is a file handle, where you seek to specific locations and read specific amounts.
 
 Readable streams are designed to wrap both types of sources behind a single, unified interface. For web
-developer–created streams, the <a href="#rs-constructor">implementation details of a source</a> are provided by an
-object with certain methods and properties that is passed to the {{ReadableStream()}} constructor.
+developer–created streams, the implementation details of a source are provided by <a href="#underlying-source-api">an
+object with certain methods and properties</a> that is passed to the {{ReadableStream()}} constructor.
 
 <a>Chunks</a> are enqueued into the stream by the stream's <a>underlying source</a>. They can then be read one at a
 time via the stream's public interface, in particular by using a <a>readable stream reader</a> acquired using the
@@ -146,15 +146,15 @@ Analogously to readable streams, most writable streams wrap a lower-level I/O si
 queuing subsequent writes and only delivering them to the underlying sink one by one.
 
 <a>Chunks</a> are written to the stream via its public interface, and are passed one at a time to the stream's
-<a>underlying sink</a>. For web developer-created streams, the <a href="#ws-constructor">implementation details of the
-sink</a> are provided by an object with certain methods that is passed to the {{WritableStream()}} constructor.
+<a>underlying sink</a>. For web developer-created streams, the implementation details of the sink are provided by <a
+href="#underlying-sink-api">an object with certain methods</a> that is passed to the {{WritableStream()}} constructor.
 
 Code that writes into a writable stream using its public interface is known as a <dfn>producer</dfn>.
 
 Producers also have the ability to <dfn lt="abort a writable stream">abort</dfn> a writable stream, using its
 {{WritableStream/abort()}} method. This indicates that the producer believes something has gone wrong, and that future
 writes should be discontinued. It puts the stream in an errored state, even without a signal from the <a>underlying
-sink</a>.
+sink</a>, and it discards all writes in the stream's <a>internal queue</a>.
 
 <h3 id="ts-model">Transform Streams</h3>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2854,61 +2907,114 @@ Instances of {{WritableStream}} are created with the internal slots described in
...
+    custom handling.</p>
+
+    <p>If the shutdown process is asynchronous, the function can return a promise to signal success or failure; the
+    result will be communicated via the return value of the called <code>abort()</code> method. Throwing an exception is
+    treated the same as returning a rejected promise. Regardless, the stream will be errored with a new {{TypeError}}
+    indicating that it was aborted.</p>
+  </dd>
+</dl>
+
+The <code>controller</code> argument passed to {{underlying sink/start()}} and {{underlying sink/write()}} is an
```
surma: I just stumbled over this. It’s pretty easy to miss. Could we move it to the top or maybe even duplicate it to the individual methods that take `controller`? I know this is a spec and not developer documentation, so you can make the call here, just giving a developer’s perspective on this.

index.bs
```diff
@@ -502,6 +469,85 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
   1. Otherwise, throw a *RangeError* exception.
 </emu-alg>
 
+<h4 id="underlying-source-api">Underlying Source API</h4>
+
+The {{ReadableStream()}} constructor accepts as its first argument a JavaScript object representing the <a>underlying
+source</a>. Such objects may contain any of the following properties:
+
+<dl>
+  <dt><dfn method for="underlying source">start(<var>controller</var>)</dfn></dt>
```
surma: `controller` is not defined in this section. At the very least there should have a paragraph similar to the one in “Underlying Sink API”.

index.bs
```diff
@@ -4087,6 +4163,66 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
   1. <a>Resolve</a> _startPromise_ with _startResult_.
 </emu-alg>
 
+<h4 id="transformer-api">Transformer API</h4>
```
surma: This should also be directly linked to by “2.3. Transform Streams” imo

index.bs
domenic: Hmm, I think it's best to leave as-is; I don't feel comfortable talking about the methods before they are defined, and duplicating it is not good.

index.bs
domenic: It's defined in the section on `type`. I'll move it down to the bottom, to be symmetric with writable streams

domenic:
@ricea do you want to review this? If you're too busy that's no problem, and I can just go ahead and merge it.

lgrahl:
I'll try take a look at this later this week (hopefully).

index.bs
```diff
@@ -2757,12 +2810,19 @@ throws>SetUpReadableByteStreamControllerFromUnderlyingSource ( <var>stream</var>
       const bytes = new Uint8Array(writer.desiredSize);
       window.crypto.getRandomValues(bytes);
 
-      await writer.write(bytes);
+      // Purposefully don't await; awaiting writer.ready is enough.
+      writer.write(bytes);
```
ricea: How about adding `.catch(() => {})` to avoid uncaught rejections?

index.bs
```diff
@@ -2757,12 +2810,19 @@ throws>SetUpReadableByteStreamControllerFromUnderlyingSource ( <var>stream</var>
...
   }
 
   writeRandomBytesForever(myWritableStream).catch(e => console.error("Something broke", e));
   </code></pre>
+
+  Note how we don't <code>await</code> the promise returned by {{WritableStreamDefaultWriter/write()}}. Doing so would
+  prevent us from executing further writes until the previous one succeeded, which defeats the point of the stream's
+  <a>internal queue</a>. Instead, we <code>await</code> the {{WritableStreamDefaultWriter/ready}} promise. That promise
+  will fulfill when the <a lt="desired size to fill a stream's internal queue">desired size</a> becomes positive, which
+  might be before the write succeeds (especially in cases with a larger <a>high water mark</a>).
```
ricea: Sadly this comment is not correct, because `write()` always fills the queue in one go. See longer explanation at https://github.com/whatwg/streams/issues/874#issuecomment-364120545. Maybe just leave it out?

index.bs
```diff
@@ -2854,61 +2914,114 @@ Instances of {{WritableStream}} are created with the internal slots described in
...
 </div>
 
 <emu-alg>
   1. Perform ! InitializeWritableStream(_this_).
   1. Let _type_ be ? GetV(_underlyingSink_, `"type"`).
   1. If _type_ is not *undefined*, throw a *RangeError* exception. <p class="note">This is to allow us to add new
      potential types in the future, without backward-compatibility concerns.</p>
+  1. Let _size_ be ? GetV(_readableStrategy_, `"size"`).
   1. Let _sizeAlgorithm_ be ? MakeSizeAlgorithmFromSizeFunction(_size_).
+  1. Let _highWaterMark_ be ? GetV(_readableStrategy_, `"highWaterMark"`).
```
ricea: Because MakeSizeAlgorithmFromSizeFunction() will throw if size is not a function, placing this GetV() after that operation changes behaviour.

index.bs
```diff
@@ -436,58 +438,24 @@ Instances of {{ReadableStream}} are created with the internal slots described in
...
+  The <code>strategy</code> argument represents the stream's <a>queuing strategy</a>, as described in [[#qs-api]]. If it
+  is not provided, the default behavior will be the same as a {{CountQueuingStrategy}} with a <a>high water mark</a>
+  of 1.
 </div>
 
 <emu-alg>
   1. Perform ! InitializeReadableStream(*this*).
   1. Let _type_ be ? GetV(_underlyingSource_, `"type"`).
   1. Let _typeString_ be ? ToString(_type_).
+  1. Let _size_ be ? GetV(_readableStrategy_, `"size"`).
```
ricea: Maybe move these up so that existing behaviour doesn't change?

index.bs
```diff
@@ -2854,61 +2914,114 @@ Instances of {{WritableStream}} are created with the internal slots described in
...
+    <p>If the shutdown process is asynchronous, the function can return a promise to signal success or failure; the
+    result will be communicated via the return value of the called
+    {{WritableStreamDefaultWriter/close()|writer.close()}} method. Additionally, a rejected promise will error the
+    stream, instead of letting it close successfully. Throwing an exception is treated the same as returning a rejected
+    promise.</p>
+  </dd>
+
+  <dt><dfn method for="underlying sink">abort(<var>reason</var>)</dfn></dt>
+  <dd>
+    <p>A function that is called after the <a>producer</a> signals, via {{WritableStream/abort()|stream.abort()}} or
```
ricea: Maybe mention `pipeTo()`?

index.bs
ricea: `controller.error()` should be rarely used, so I don't want to draw undue attention to it. I think this text is fine as-is.

index.bs
```diff
@@ -4087,6 +4170,65 @@ readableStrategy)">new TransformStream(<var>transformer</var> = {}, <var>writabl
...
+    <p>If no {{transformer/transform()}} is supplied, the identity transform is used, which enqueues chunks unchanged
+    from the writable side to the readable side.</p>
+  </dd>
+
+  <dt><dfn method for="transformer">flush(<var>controller</var>)</dfn></dt>
+  <dd>
+    <p>A function called after all <a>chunks</a> written to the <a>writable side</a> have been transformed by
+    successfully passing through {{transformer/transform()}}, and the writable side is about to be closed.</p>
+
+    <p>Typically this is used to enqueue suffix chunks to the <a>readable side</a>, before that too becomes closed. An
```
ricea: Do you think we should mention that there's no need to explicitly call `controller.close()` inside `flush()`? I made this mistake in my early text encoder stream implementation.

index.bs
```diff
@@ -4501,7 +4643,42 @@ aoid="TransformStreamDefaultSourcePullAlgorithm" nothrow>TransformStreamDefaultS
...
+The {{ReadableStream()}}, {{WritableStream()}}, and {{TransformStream()}} constructors all accept at least one argument
+representing an appropriate <a>queuing strategy</a> for the stream being created. Such objects contain the following
+properties:
+
+<dl>
+  <dt><dfn method for="queuing strategy">size(<var>chunk</var>)</dfn> (non-byte streams only)</dt>
+  <dd>
+    <p>A function that computes and returns the size of the given <a>chunk</a> value.</p>
+
+    <p>The result is used to determine <a>backpressure</a>, manifesting via
```
ricea: I find this sentence hard to parse, but I don't have a concrete suggestion for how to improve it at the moment.

index.bs
```diff
@@ -4895,6 +5072,21 @@ writable stream:
     .catch(e => console.error("Something went wrong!", e));
 </code></pre>
 
+<div class="note" id="note-web-socket-wrapping-examples">
+  This specific style of wrapping a web socket interprets web socket messages directly as <a>chunks</a>.
+  This can be convenient in some cases, for example when <a>piping</a> to a <a>writable stream</a> or <a>transform
```
ricea: How about saying "This can be a convenient abstraction, ..." instead.

The phrasing "in some cases" sounds a bit too negative to my ears.

index.bs
```diff
@@ -4895,6 +5072,21 @@ writable stream:
...
+  stream</a> for which each web socket message makes sense as a chunk to consume or transform.
+
+  However, often when people talk about "adding streams support to web sockets", they are hoping instead for the ability
```
ricea: How about "for a new capability" instead of "for the ability"? To emphasise the distinction between wrapping an existing API, as we are doing here, and increasing the capabilities of the platform, as we'd like to do.

index.bs
```diff
@@ -502,6 +470,91 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
   1. Otherwise, throw a *RangeError* exception.
 </emu-alg>
 
+<h4 id="underlying-source-api">Underlying Source API</h4>
```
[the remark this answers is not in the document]
domenic: It's meant to be a series of statements of fact, with no "must"s requiring implementation work, but also no contradictions with other normative text. How does that sound?

index.bs
```diff
@@ -502,6 +470,91 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
...
+
+<dl>
+  <dt><dfn method for="underlying source">start(<var>controller</var>)</dfn></dt>
+  <dd>
+    <p>A function that is called immediately upon creation of the {{ReadableStream}}.</p>
+
+    <p>Typically this is used adapt a <a>push source</a> by setting up relevant event listeners, as in the example of
+    [[#example-rs-push-no-backpressure]], or to acquire access to a <a>pull source</a>, as in [[#example-rs-pull]].</p>
+
+    <p>If this setup process is asynchronous, it can return a promise to signal success or failure; a rejected promise
```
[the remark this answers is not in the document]
domenic: As in previous discussions, I think we can leave off mentioning thenables here. Agreed on mentioning pull vs. start ordering.

index.bs
domenic: Sure, seems reasonable now that we have all this extra space.

index.bs
domenic: I gave it a potential upgrade; thanks for flagging.

domenic:
@ricea thanks as always for your careful review!! I pushed several commits that I think fix all the issues you identified. I still hope to rearrange this PR into three distinct commits before merging, but for now I added the changes on top of the original commits, for easier reviewing.

index.bs
```diff
@@ -2755,14 +2820,52 @@ throws>SetUpReadableByteStreamControllerFromUnderlyingSource ( <var>stream</var>
...
+
+      const bytes = getBytes();
+      writer.write(bytes).catch(() => {});
+    }
+  }
+  </code></pre>
+
+  Unlike the previous example, where—because we were always writing exactly
+  {{WritableStreamDefaultWriter/desiredSize|writer.desiredSize}} bytes each time—the
+  {{WritableStreamDefaultWriter/write()}} and {{WritableStreamDefaultWriter/ready}} promises were synchronized, in this
```
lgrahl: This example (and the one above) assumes that the underlying sink returns such a promise to the writer. It might be worth adding that assumption.

index.bs
```diff
@@ -2755,14 +2820,52 @@ throws>SetUpReadableByteStreamControllerFromUnderlyingSource ( <var>stream</var>
...
+  case it's quite possible that the {{WritableStreamDefaultWriter/ready}} promise fulfills before the one returned by
+  {{WritableStreamDefaultWriter/write()}} does. Remember, the {{WritableStreamDefaultWriter/ready}} promise fulfills
+  when the <a lt="desired size to fill a stream's internal queue">desired size</a> becomes positive, which might be
+  before the write succeeds (especially in cases with a larger <a>high water mark</a>).
+
+  In other words, <code>await</code>ing the return value of {{WritableStreamDefaultWriter/write()}} means you never
+  queue up writes in the stream's <a>internal queue</a>, instead only executing a write after the previous one succeeds.
```
lgrahl: Maybe add this at the end of the sentence:

> ... which may result in low throughput.

index.bs
```diff
@@ -2854,61 +2957,116 @@ Instances of {{WritableStream}} are created with the internal slots described in
...
+    write. Throwing an exception is treated the same as returning a rejected promise.</p>
+
+    <p>Note that such signals are not always available; compare e.g. [[#example-ws-no-backpressure]] with
+    [[#example-ws-backpressure]]. In such cases, it's best to not return anything.</p>
+
+    <p>The promise potentially returned by this function also governs whether the given chunk counts as written for the
+    purposes of computed the <a lt="desired size to fill a stream's internal queue">desired size to fill the stream's
+    internal queue</a>. That is, during the time it takes the promise to settle,
+    {{WritableStreamDefaultWriter/desiredSize|writer.desiredSize}} will stay at its previous value, only increasing to
+    signal the desire for more chunks once the write succeeds.</p>
```
lgrahl: 👍

index.bs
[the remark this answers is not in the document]
ricea: But now it's normative. So an implementer could implement this and say that they were standard-compliant.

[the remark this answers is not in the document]
domenic: No, there's no "must" or even "should" implementation requirement here.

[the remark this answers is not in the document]
ricea: I wouldn't interpret the absence of the words "must" or "should" to mean that a statement in a normative section was not a requirement.

Our algorithms don't say "must" or "should" either. In fact, we seem to be missing a bit of boilerplate somewhere that would make them a "must". The [[INFRA]](https://infra.spec.whatwg.org/#algorithms) standard has something like that, but we're not using it.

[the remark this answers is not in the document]
domenic: Yeah, we should be using Infra.

In any case, it's clear this isn't as straightforward for at least you as a reader as it is for me as the writer, so I guess we can explicitly make these sections non-normative, even though per my reading they do not contain any normative text anyway.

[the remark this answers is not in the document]
ricea: That's my preference. Different people can disagree over whether a particular statement constitutes a requirement or not.

I like that the standard unambiguously defines consistent behaviour, even when it's wrong.

But I'm just concerned, not dead against it. If there's a benefit to making these sections normative then we can do that.

I can add the reference to infra if you are busy. I assume we just need one sentence in the "Conventions" section at the end?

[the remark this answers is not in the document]
domenic: Well, hopefully different people don't disagree; that seems like a recipe for trouble :-/.
I'll add the Infra reference too

index.bs
domenic: I think the advice holds even in the absence of that assumption, so I'll leave it out.

domenic:
@ricea any further thoughts, or shall I squash/rebase/merge this?

ricea:
b74def3 lgtm. Sorry for not responding sooner.

PR #883, opened 2018-02-15
Suppress Bikeshed warnings about <vars> used once

### LINKED ISSUE #869: Bikeshed warnings
There's quite a number of variables that are reportedly only used once. It would help https://github.com/whatwg/whatwg.org/issues/176 to address these so we can eventually make Bikeshed warnings fatal.

ricea:
The arguments to methods and abstract operations are marked up with
Bikeshed `<var>` tags, but the actual algorithms are marked up using
emu-algify. Bikeshed does not recognize the uses of the arguments within
the emu-algify blocks, and so warns that the `<var>`s were only used once.

Use the "ignore" parameter on affected `<var>` tags to suppress the
warning. It is only used on `<var>` tags that actually trigger the
warning, as the benefit of having the other `<var>` tags checked outweighs
the potential benefit of consistency.

This change eliminates all current Bikeshed warnings.

No functional or visible changes.

Fixes #869.

***
<a href="https://whatpr.org/streams/883.html" title="Last updated on Feb 15, 2018, 1:53 PM GMT (5e39846)">Preview</a> | <a href="https://whatpr.org/streams/883/d918645...5e39846.html" title="Last updated on Feb 15, 2018, 1:53 PM GMT (5e39846)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -534,7 +534,7 @@ ReadableStream(<var>underlyingSource</var> = {}, { <var>size</var>, <var>highWat
   1. Return ! ReadableStreamCancel(*this*, _reason_).
 </emu-alg>
 
-<h5 id="rs-get-reader" method for="ReadableStream">getReader({ <var>mode</var> } = {})</h5>
+<h5 id="rs-get-reader" method for="ReadableStream">getReader({ <var ignore>mode</var> } = {})</h5>
 
 <div class="note">
   The <code>getReader</code> method creates a reader of the type specified by the <code>mode</code> option and <a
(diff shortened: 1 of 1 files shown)


domenic:
Hmm. I was hoping for a better fix here, thus not addressing it previously, but there's nothing easy.

Do you think adding `ignore` is the way to go, or should we instead use the `Ignored Vars: x, y, z, ...` metadata in the header? The latter seems a bit less disruptive if we eventually do get things straightened out.

ricea:
I wish I'd known about `Ignored Vars`... it took ages to track them down one by one. Having said that, I feel `<var ignore>` is slightly better because it is localised to the place where the problem exists. For example, if we rename a variable it will update automatically. 

If we find a real fix for this, then removing all the `ignore` annotations can be easily done with search-and-replace, so I'm not worried about that.

PR #884, opened 2018-02-15
Fix pipe not correctly waiting for writes to finish

### LINKED ISSUE #738: pipeTo() text appears to say all queued writable chunks must be written on readable error
https://streams.spec.whatwg.org/commit-snapshots/037ec544adc7a7677ada2ba140849fb5549b09fd#rs-pipeTo-shutdown-with-action

> If any chunks have been read but not yet written, write them to dest\.
> Wait until every chunk that has been read has been written

It appears to me that this requires all queued chunks to be written when the ReadableStream becomes errored. This is not what the reference implementation actually implements, and I am not sure it is desirable.

https://github.com/w3c/web-platform-tests/pull/5636 contains a test which illustrates the behaviour that I think the standard text is requiring.

### LINKED ISSUE #882: Pipe completes before all writes are finished with preventClose = true
I believe there's a bug in the reference implementation for piping with `preventClose = true`. The implementation does not correctly wait for all writes to finish before completing the pipe operation.

## Test
The following web platform test is derived from the existing test _"Closing must be propagated forward: shutdown must not occur until the final write completes; preventClose = true"_ ([`streams/piping/close-propagation-forward.js`](https://github.com/w3c/web-platform-tests/blob/840362217deb3f9b82fe18a619d0bf21ddc55781/streams/piping/close-propagation-forward.js#L427)).

The only change is that the readable stream is closed is **asynchronously** after enqueuing, whereas in the original test `enqueue()` and `close()` happen in the same tick.

This test fails on the assert _"the pipe must not be complete"_. That is: the pipe completes **before** the write promise is resolved, which violates the specification.
```js
promise_test(() => {

  let resolveWritePromise;
  const ws = recordingWritableStream({
    write() {
      return new Promise(resolve => {
        resolveWritePromise = resolve;
      });
    }
  });

  let pipeComplete = false;
  const pipePromise = rs.pipeTo(ws, { preventClose: true }).then(() => {
    pipeComplete = true;
  });

  rs.controller.enqueue('a');

  return flushAsyncEvents().then(() => {
    rs.controller.close();

    // Flush async events and verify that no shutdown occurs.
    return flushAsyncEvents();
  }).then(() => {
    assert_array_equals(ws.events, ['write', 'a'],
      'the chunk must have been written, but close must not have happened yet');
    assert_equals(pipeComplete, false, 'the pipe must not be complete');

    resolveWritePromise();

    return pipePromise;
  });

}, 'Closing must be propagated forward: shutdown must not occur until the final write completes; preventClose = true, delayed close');
```

## Root cause
From my investigation, it seems the bug is in [`pipeLoop`](https://github.com/whatwg/streams/blob/d91864509e66fc292f8cdba34ca00bfc09e2565e/reference-implementation/lib/readable-stream.js#L125):
```js
// Using reader and writer, read all chunks from this and write them to dest
// - Backpressure must be enforced
// - Shutdown must stop all activity
function pipeLoop() {
  currentWrite = Promise.resolve();

  if (shuttingDown === true) {
    return Promise.resolve();
  }

  return writer._readyPromise.then(() => {
    return ReadableStreamDefaultReaderRead(reader).then(({ value, done }) => {
      if (done === true) {
        return;
      }

      currentWrite = WritableStreamDefaultWriterWrite(writer, value).catch(() => {});
    });
  })
  .then(pipeLoop);
}
```
Removing the first line (`currentWrite = Promise.resolve()`) solves the problem, which I believe is a correct fix for the problem. When `pipeLoop` recurses through `.then(pipeLoop)`, only the read has resolved, and `currentWrite` is still pending. Therefore, `currentWrite` **should not** be reset when starting a new iteration, it should only be replaced by a new write (which is guaranteed to resolve after the previous write).

MattiasBuelens:
Fixes #738.
Fixes #882.

### DIFF
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -126,8 +126,6 @@ class ReadableStream {
       // - Backpressure must be enforced
       // - Shutdown must stop all activity
       function pipeLoop() {
-        currentWrite = Promise.resolve();
-
         if (shuttingDown === true) {
           return Promise.resolve();
         }


MattiasBuelens:
This will need additional web platform tests. We already have one [for #738](https://github.com/whatwg/streams/issues/882#issuecomment-366084939) and one [for #882](https://github.com/whatwg/streams/issues/882#issue-297274918), but we probably need a few more.

There's already w3c/web-platform-tests#5636, but it's marked "do not merge yet". Should that PR be updated?

ricea:
This change looks good to me, but I want to make sure we have tests landed in web-platform-tests before we land it here. I need to take another look at the pipeTo() implementation and make sure everything that might be affected by this is covered. But I'm not going to get to that until next week.

If you want to contribute some tests too that would be awesome.

w3c/web-platform-tests#5636 was just waiting for you to come along and fix this. It's the same test as #738. I applied your fixes to it and I or someone else will land it once it is reviewed.

MattiasBuelens:
@ricea Your WPT PR looks good to me!

As I said before, I'll try to find some time to write more tests. When I do, I'll submit a PR to WPT and let you know - so we don't end up writing the same test twice. 😛

domenic:
I've pushed a commit to this PR that rolls the web platform tests to latest. Were we hoping to write more, or is this ready to merge?

domenic:
Builds are currently failing due to some new changes to the HTML checker which isn't compatible with modern JS syntax used in the demos... @sideshowbarker has been alerted and will hopefully fix that soon.

MattiasBuelens:
Looking at it again, I think we can better align the close and error propagation tests. Right now, we have the following variants for propagating close forward:
1. Closing must be propagated forward: shutdown must not occur until the final write completes
2. Closing must be propagated forward: shutdown must not occur until the final write completes; preventClose = true
3. Closing must be propagated forward: shutdown must not occur until the final write completes; becomes closed after first write
4. Closing must be propagated forward: shutdown must not occur until the final write completes; becomes closed after first write; preventClose = true

For propagating errors forward, we only have:
1. Errors must be propagated forward: shutdown must not occur until the final write completes
2. Errors must be propagated forward: abort should not happen until all queued writes complete

I suggest we rework these error tests to be more similar to their closing counterparts, so we end up with:
1. Errors must be propagated forward: shutdown must not occur until the final write completes
2. Errors must be propagated forward: shutdown must not occur until the final write completes; preventAbort = true
3. Errors must be propagated forward: shutdown must not occur until the final write completes; becomes errored after first write
4. Errors must be propagated forward: shutdown must not occur until the final write completes; becomes errored after first write; preventAbort = true

Should be simple enough to rename the existing tests and add two more variants. WPT PR coming right up! 😉

sideshowbarker:
> Builds are currently failing due to some new changes to the HTML checker which isn't compatible with modern JS syntax used in the demos... @sideshowbarker has been alerted and will hopefully fix that soon.

Fixed now (with sincere apologies for breaking the build)

MattiasBuelens:
w3c/web-platform-tests#9591 is merged, so I think we've got enough tests now. For me, this is ready to merge.

PR #893, opened 2018-03-04
Add fetch-progress examples to demos page

AnthumChris:
Adds link to the "fetch-progress" repository, which provides examples for showing progress indicators with Streams, Fetch, and Service Workers


ricea:
Looks good.

The examples would be cleaner with TransformStream, but unfortunately no browser is shipping it yet. You could be really concise, something like

```javascript
fetchEvent.respondWith(new Response(response.body.pipeThrough(new TransformStream({
    loaded += chunk.length;
    progress({loaded, total});
    controller.enqueue(chunk);
  }
})), response));
```

(totally untested).

It would be great if you could add web-platform-tests for the incompatibilities between browsers that you found. This would allow the various browsers to track their compatibility issues. I think the issue with `abort()` not being called when a fetch is cancelled in Chrome is being tracked here: https://bugs.chromium.org/p/chromium/issues/detail?id=638494.

AnthumChris:
TransformStream looks great.  I made a note to come back to it:
https://github.com/AnthumChris/fetch-progress-indicators/issues/8

Haven't used web-platform-tests before, but I'm open to writing some.

PR #895, opened 2018-03-07
Do not throw when calling RS's controller.error()

### LINKED ISSUE #821: Is it bad that readableController.error() throws but writableController.error() no-ops in non-readable/writable states?
Compare:

- https://streams.spec.whatwg.org/commit-snapshots/6ad0e3d59cd8ff0223f178752f5f9a4f0d7c073f/#ws-default-controller-error
- https://streams.spec.whatwg.org/commit-snapshots/6ad0e3d59cd8ff0223f178752f5f9a4f0d7c073f/#rs-default-controller-error

I remember us consciously changing this for writable streams. Did we just forget to change it for readable streams, or is there an intentional distinction?

If it's intentional, we should capture it in the spec in a note, I think.

domenic:
Previously, we would throw an exception when calling controller.error() for a readable stream that was not readable. As noted in #821, this is inconsistent with the behavior introduced for writable and transform streams. The fact that the spec is constantly checking whether it's in the right state, e.g. via ReadableStreamDefaultControllerCallPullIfNeeded, is also a strong indicator that we should not throw.

This simplifies various abstract operations by centralizing state checks and removing ReadableStreamDefaultControllerCallPullIfNeeded.

Closes #821.

Tests incoming, will edit.

***
<a href="https://whatpr.org/streams/895.html" title="Last updated on Mar 8, 2018, 5:27 AM GMT (2824abc)">Preview</a> | <a href="https://whatpr.org/streams/895/a5f51ae...2824abc.html" title="Last updated on Mar 8, 2018, 5:27 AM GMT (2824abc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1045,8 +1045,8 @@ noticeable asymmetry between the two branches, and limits the possible <a>chunks
   1. Set _branch2_ to CreateReadableStream(_startAlgorithm_, _pullAlgorithm_, _cancel2Algorithm_).
   1. <a>Upon rejection</a> of _reader_.[[closedPromise]] with reason _r_,
     1. If _closedOrErrored_ is *false*, then:
-      1. Perform ! ReadableStreamDefaultControllerErrorIfNeeded(_branch1_.[[readableStreamController]], _r_).
-      1. Perform ! ReadableStreamDefaultControllerErrorIfNeeded(_branch2_.[[readableStreamController]], _r_).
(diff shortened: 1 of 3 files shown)


ricea:
Nit: In the description you wrote "ReadableStreamDefaultControllerCallPullIfNeeded"; it should be "ReadableStreamDefaultControllerErrorIfNeeded" instead. Otherwise looks good. I'm going to give it another read-through for good measure.

PR #898, opened 2018-03-07
Fix rethrowAssertionErrorRejection

### LINKED ISSUE #865: rethrowAssertionErrorRejection is broken
84c584c replaced the default Node `assert` with the `better-assert` module, in order to resolve #807. However, this refactoring accidentally broke the `rethrowAssertionErrorRejection` helper, since [it uses `assert.AssertionError`](https://github.com/whatwg/streams/blob/a480a4fd23c1e3f0cea94cbbdf224f352ae60ab2/reference-implementation/lib/utils.js#L8).

While `assert` does indeed export an `AssertionError` class, `better-assert` does not. Therefore, `assert.AssertionError` is `undefined`, and `e.constructor === assert.AssertionError` always evaluates to `false`. As such, the helper never re-throws an assertion error.

One possibility would be to use `require('assert').AssertionError`, but this assumes `better-assert` will never introduce a different error class. If necessary, we could detect the error class with something like:
```js
const AssertionError = (() => {
  try {
    assert(false);
  } catch (e) {
    return e.constructor;
  }
})();
```
But maybe let's not go *that* far. 😛 

Thoughts? I can make a quick PR to use `require('assert').AssertionError` again.

MattiasBuelens:
Fixes #865

### DIFF
--- a/reference-implementation/lib/utils.js
+++ b/reference-implementation/lib/utils.js
@@ -5,7 +5,7 @@ exports.rethrowAssertionErrorRejection = e => {
   // Used throughout the reference implementation, as `.catch(rethrowAssertionErrorRejection)`, to ensure any errors
   // get shown. There are places in the spec where we do promise transformations and purposefully ignore or don't
   // expect any errors, but assertion errors are always problematic.
-  if (e && e.constructor === assert.AssertionError) {
+  if (e && e instanceof assert.AssertionError) {
     setTimeout(() => {
       throw e;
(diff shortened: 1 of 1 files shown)


MattiasBuelens:
Fun fact: `instanceof` throws a `TypeError` if the right-hand side is not an object. Therefore, if this were ever refactored incorrectly again and `assert.AssertionError` no longer existed, we would get an error instead. 😄

PR #899, opened 2018-03-07
Refactor exports

MattiasBuelens:
I'm working on an updated polyfill for streams over at [MattiasBuelens/web-streams-polyfill](https://github.com/MattiasBuelens/web-streams-polyfill). I use Rollup to convert the reference implementation from CommonJS to ES2015 modules, and then bundle it in a nice UMD wrapper.

This works great, but unfortunately the [Rollup CommonJS plugin](https://github.com/rollup/rollup-plugin-commonjs) fails to optimally transform some parts of the code. For example, it has trouble with this function in `helpers.js`:
```js
exports.createDataProperty = (o, p, v) => {
  assert(exports.typeIsObject(o));
  Object.defineProperty(o, p, { value: v, writable: true, enumerable: true, configurable: true });
};
```
Because `exports.typeIsObject` might be re-assigned externally, the plugin must retain the `exports` object and wrap the whole module with a `createCommonJS` call. This prevents further optimizations such as tree-shaking.

This PR changes all instances of `exports.foo = ...` with `const foo = ...; module.exports = { foo };` so that internal functions can use `foo` instead of `exports.foo`. This prevents externally re-assigned exports from changing the behavior of internal functions, and allows Rollup to transform the CommonJS module into a "proper" ES2015 module.

I realize that this is not immediately relevant for the reference implementation, but it helps to shave off a couple of bytes in the polyfill. I could keep this in my fork of the reference implementation, but I'd prefer not to need a fork. 😅

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -7,24 +7,24 @@ function IsPropertyKey(argument) {
   return typeof argument === 'string' || typeof argument === 'symbol';
 }
 
-exports.typeIsObject = x => (typeof x === 'object' && x !== null) || typeof x === 'function';
+const typeIsObject = x => (typeof x === 'object' && x !== null) || typeof x === 'function';
 
-exports.createDataProperty = (o, p, v) => {
-  assert(exports.typeIsObject(o));
+const createDataProperty = (o, p, v) => {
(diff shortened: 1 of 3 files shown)


MattiasBuelens:
Fair enough, it was worth a shot.

PR #901, opened 2018-03-08
Have WS's abort() fulfill, instead of reject, under error conditions

### LINKED ISSUE #900: Should writableStream.abort() reject on already-errored streams?
As noted in #847, this can lead to lots of redundant `.catch(() => {})`s.

I think it's especially strange that this rejects given that we just resolve with undefined if the stream is already _closed_. If anything that's the case where you failed to abort. If you're already _errored_, then you've basically aborted the stream already.

So I'm inclined to change this. Any thoughts?

domenic:
* If the stream is errored, abort() now returns a promise resolved with undefined, instead of returning a promise rejected with the stored error.
* If the stream has a pending abort, abort() now returns that pending abort's promise (which will eventually resolve with undefined, unless aborting fails), instead of returning a promise rejected with a TypeError.

Fixes #900.

Tests: https://github.com/w3c/web-platform-tests/pull/9917

***
<a href="https://whatpr.org/streams/901.html" title="Last updated on Mar 8, 2018, 3:20 PM GMT (0db3090)">Preview</a> | <a href="https://whatpr.org/streams/901/ec2a957...0db3090.html" title="Last updated on Mar 8, 2018, 3:20 PM GMT (0db3090)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3205,10 +3205,9 @@ writable stream is <a>locked to a writer</a>.
 
 <emu-alg>
   1. Let _state_ be _stream_.[[state]].
-  1. If _state_ is `"closed"`, return <a>a promise resolved with</a> *undefined*.
-  1. If _state_ is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
+  1. If _state_ is not `"closed"` or `"errored"`, return <a>a promise resolved with</a> *undefined*.
+  1. If _stream_.[[pendingAbortRequest]] is not *undefined*, return _stream_.[[pendingAbortRequest]].[[promise]].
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -3205,10 +3205,9 @@ writable stream is <a>locked to a writer</a>.
 
 <emu-alg>
   1. Let _state_ be _stream_.[[state]].
-  1. If _state_ is `"closed"`, return <a>a promise resolved with</a> *undefined*.
-  1. If _state_ is `"errored"`, return <a>a promise rejected with</a> _stream_.[[storedError]].
+  1. If _state_ is not `"closed"` or `"errored"`, return <a>a promise resolved with</a> *undefined*.
```
ricea: There shouldn't be a "not" here.

PR #902, opened 2018-03-08
Clarify realm and task-queuing situation in pipeTo()

### LINKED ISSUE #845: Need to define the current Realm for pipeTo operations
https://streams.spec.whatwg.org/#rs-pipe-to defines things like calling ReadableStreamReaderGenericRelease during shutdown.  ReadableStreamReaderGenericRelease step 4 needs to create a "promise rejected with", which means it needs to know what global to use for that.

Presumably the global is that of the current Realm.  But it's not clear to me that the current Realm is well-defined on all the paths that can lead to shutdown in the parallel section of pipeTo.  In particular, nothing really indicates that `if dest.[[state]] is or becomes "errored"` has to happen while there is a current Realm set up.

It would be good to make this clearer, ideally by explicitly saying what Realm should be used here or having a note explaining how the existence of a current Realm is guaranteed.

We should probably also have tests for this, if we don't already.

domenic:
Closes #845.

@bzbarsky, does this address your concerns?

(I'm always a bit skeptical of the DOM manipulation task source, but it appears to be the default...)

***
<a href="https://whatpr.org/streams/902.html" title="Last updated on Mar 9, 2018, 10:33 AM GMT (47573e8)">Preview</a> | <a href="https://whatpr.org/streams/902/239b679...47573e8.html" title="Last updated on Mar 9, 2018, 10:33 AM GMT (47573e8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -747,10 +747,11 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
   1. Let _writer_ be ! AcquireWritableStreamDefaultWriter(_dest_).
   1. Let _shuttingDown_ be *false*.
   1. Let _promise_ be <a>a new promise</a>.
-  1. <a>In parallel</a>, using _reader_ and _writer_, read all <a>chunks</a> from *this* and write them to _dest_. Due
-     to the locking provided by the reader and writer, the exact manner in which this happens is not observable to
-     author code, and so there is flexibility in how this is done. The following constraints apply regardless of the
-     exact algorithm used:
+  1. <a>In parallel</a> <span class="XXX">but not really; see <a
+     href="https://github.com/whatwg/streams/issues/905">#905</a></span>, using _reader_ and _writer_, read all
+     <a>chunks</a> from *this* and write them to _dest_. Due to the locking provided by the reader and writer, the exact
+     manner in which this happens is not observable to author code, and so there is flexibility in how this is done. The
+     following constraints apply regardless of the exact algorithm used:
      * <strong>Public API must not be used:</strong> while reading or writing, or performing any of the operations
        below, the JavaScript-modifiable reader, writer, and stream APIs (i.e. methods on the appropriate prototypes)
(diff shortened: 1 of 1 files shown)


ricea:
I'm concerned that this change is user-visible, and that it will be problematic to implement in Chrome. Take this example:

```javascript
async function ObserveTaskQueue() {
  let pipeToComplete = false;
    async start(controller) {
      controller.close();
      for (let i = 0; i < 1000; ++i) {
        if (pipeToComplete) {
          console.log('done');
          return;
        }
        await Promise.resolve();
      }
      console.log('fail');
    }
  });
  const ws = new WritableStream();
  await rs.pipeTo(ws);
  pipeToComplete = true;
}
```

With Chrome's current implementation, this function prints "done". I believe that Chrome's implementation is compliant with the standard in this respect. I think an implementation is also permitted to print "fail".

However, if I understand correctly, no task from the DOM manipulation task source can run while microtasks are executing. If this change to the standard is made, all implementations must print "fail".

bzbarsky:
> However, if I understand correctly, no task from the DOM manipulation task source can run while microtasks are executing.

No task at all can run while microtasks are executing.   I don't see how the code pasted above can ever print "done" if there is a task being queued from the close() call that needs to run to resolve the promise returned by pipeTo.

domenic:
After thinking about this more, I realized the current "in parallel" block is a bit bonkers. You could do all that work in parallel unobservably for UA-provided streams, but for developer-created streams, reading and writing will call back into the JS underlying source/sink, and we haven't noted how or whether to queue a task in those cases. So, that's pretty wrong right now.

As such, I'll scope this PR down to just object creation, and add a red "XXX" box around "in parallel" linking to a new issue for straightening that out as a second step.

ricea:
Looks okay as a stopgap.

domenic:
Gotta remove the pointer to the TypeError since we fixed that while this PR was out.

PR #903, opened 2018-03-08
Set stored error from |reason| in abort(reason)

### LINKED ISSUE #896: Propagate errors from writableStream.abort(reason)
This was removed from the spec, in favor of making the [[storedError]] after abort always a TypeError. I believe this was because previously sometimes we had to synthesize a TypeError anyway, and we thought it would be best to be uniform. We should dig into the commit history to find out.

We should reverse this decision, and add back error propagation. This is crucial, in particular, for transform streams. https://github.com/whatwg/streams/pull/847 contains some examples whose results are surprising without this propagation.

ricea:
Previously, calling abort(reason) on a WritableStream would set the
stored error to a TypeError, meaning that future operations on that
WritableStream would reject with a TypeError. This resulted in losing
the original reason when an error happened in a pipe. Instead, retain
the abort reason in the stored error slot.

Similarly, aborting the writable side of a TransformStream would set the
stored error on the readable side to a TypeError. Make it use the reason
as well to reflect the new behaviour of the writable side.

Closes #896.

***
<a href="https://whatpr.org/streams/903.html" title="Last updated on Mar 9, 2018, 5:27 AM GMT (30f8d5c)">Preview</a> | <a href="https://whatpr.org/streams/903/8e99a69...30f8d5c.html" title="Last updated on Mar 9, 2018, 5:27 AM GMT (30f8d5c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3207,7 +3207,6 @@ writable stream is <a>locked to a writer</a>.
   1. Let _state_ be _stream_.[[state]].
   1. If _state_ is `"closed"` or `"errored"`, return <a>a promise resolved with</a> *undefined*.
   1. If _stream_.[[pendingAbortRequest]] is not *undefined*, return _stream_.[[pendingAbortRequest]].[[promise]].
-  1. Let _error_ be a new *TypeError* indicating that the stream has been requested to abort.
   1. Assert: _state_ is `"writable"` or `"erroring"`.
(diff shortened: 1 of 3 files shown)


ricea:
Tests are in https://github.com/w3c/web-platform-tests/pull/9926.

PR #904, opened 2018-03-08
Readable byte stream must call pull if needed after receiving new chunk

### LINKED ISSUE #877: Readable byte stream must call pull after receiving new chunk
After a readable byte stream receives a new chunk from the underlying source, the stream must call `pull` (if needed). Otherwise, the stream may get stuck (see test below).

For regular streams, this is already done correctly: `ReadableStreamDefaultController` correctly calls `ReadableStreamDefaultControllerCallPullIfNeeded`. However, byte streams do not call the corresponding `ReadableByteStreamControllerCallPullIfNeeded` in response to `enqueue`, `respond` or `respondWith`.

## Test
The following test (written as a Web platform test) demonstrates the problem. Here, a readable byte stream is constructed that will enqueue a new chunk every time `pull` is called, and closes the stream after 3 chunks. (This test is derived from the existing test _"ReadableStream with byte source: Respond to pull() by enqueue() asynchronously"_, but enqueues its chunks in separate pulls instead.)

This test *times out* with the current reference implementation. The first read resolves, but the other reads remain pending. `pull` is only called 1 time, rather than the expected 4 times (3 times to enqueue a chunk, 1 last time to close the source).

However, if the test is changed to use regular readable streams instead of byte streams, the test passes (see highlighted line).
```
promise_test(() => {
  let pullCount = 0;

  let byobRequest;
  const desiredSizes = [];

  const stream = new ReadableStream({
    pull(c) {
      byobRequest = c.byobRequest;
      desiredSizes.push(c.desiredSize);

      if (pullCount < 3) {
        c.enqueue(new Uint8Array(1));
      } else {
        c.close();
      }

      ++pullCount;
    },
    type: 'bytes' // <<< removing this line makes the test pass
  }, {
    highWaterMark: 256
  });

  const reader = stream.getReader();

  const p0 = reader.read();
  const p1 = reader.read();
  const p2 = reader.read();

  assert_equals(pullCount, 0, 'No pull as start() just finished and is not yet reflected to the state of the stream');

  return Promise.all([p0, p1, p2]).then(result => {
    assert_equals(pullCount, 4, 'pullCount after completion of all read()s');

    assert_equals(result[0].done, false, 'result[0].done');
    assert_equals(result[0].value.byteLength, 1, 'result[0].value.byteLength');
    assert_equals(result[1].done, false, 'result[1].done');
    assert_equals(result[1].value.byteLength, 1, 'result[1].value.byteLength');
    assert_equals(result[2].done, false, 'result[2].done');
    assert_equals(result[2].value.byteLength, 1, 'result[2].value.byteLength');
    assert_equals(byobRequest, undefined, 'byobRequest should be undefined');
    assert_equals(desiredSizes[0], 256, 'desiredSize on pull should be 256');
    assert_equals(desiredSizes[1], 256, 'desiredSize after 1st enqueue() should be 256');
    assert_equals(desiredSizes[2], 256, 'desiredSize after 2nd enqueue() should be 256');
    assert_equals(desiredSizes[3], 256, 'desiredSize after 3rd enqueue() should be 256');
  });
}, 'ReadableStream with byte source: Respond to multiple pull() by separate enqueue()');
```

## Proposed solution
I believe the fix should be to add calls to `ReadableByteStreamControllerCallPullIfNeeded` at the end of:
* `ReadableByteStreamControllerEnqueue` (handles `enqueue`)
* `ReadableByteStreamControllerRespondInternal` (handles `respond` and `respondWith`)

If you want, I can make a pull request that does just that. The above test can be added directly to the Web platform tests, but I'd also need to make similar tests for `respond` and `respondWith`.

MattiasBuelens:
This implements the proposed solution for #877.

Two tests are currently failing, but I believe those tests are actually wrong ([see comment](https://github.com/whatwg/streams/issues/877#issuecomment-363941374)). I'll update the web platform tests and submit a PR, then we can further discuss.

Fixes #877.

***
<a href="https://whatpr.org/streams/904.html" title="Last updated on Mar 20, 2018, 4:21 PM GMT (c3e2e96)">Preview</a> | <a href="https://whatpr.org/streams/904/293de26...c3e2e96.html" title="Last updated on Mar 20, 2018, 4:21 PM GMT (c3e2e96)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2428,6 +2428,7 @@ nothrow>ReadableByteStreamControllerEnqueue ( <var>controller</var>, <var>chunk<
     1. Assert: ! IsReadableStreamLocked(_stream_) is *false*.
     1. Perform ! ReadableByteStreamControllerEnqueueChunkToQueue(_controller_, _transferredBuffer_, _byteOffset_,
        _byteLength_).
+  1. Perform ! ReadableByteStreamControllerCallPullIfNeeded(_controller_).
 </emu-alg>
 
 <h4 id="readable-byte-stream-controller-enqueue-chunk-to-queue" aoid="ReadableByteStreamControllerEnqueueChunkToQueue"
@@ -2654,6 +2655,7 @@ throws>ReadableByteStreamControllerRespondInternal ( <var>controller</var>, <var
   1. Otherwise,
     1. Assert: _stream_.[[state]] is `"readable"`.
     1. Perform ? ReadableByteStreamControllerRespondInReadableState(_controller_, _bytesWritten_, _firstDescriptor_).
+  1. Perform ! ReadableByteStreamControllerCallPullIfNeeded(_controller_).
 </emu-alg>
 
 <h4 id="readable-byte-stream-controller-respond-with-new-view" aoid="ReadableByteStreamControllerRespondWithNewView"
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -1671,6 +1671,8 @@ function ReadableByteStreamControllerRespondInternal(controller, bytesWritten) {
 
     ReadableByteStreamControllerRespondInReadableState(controller, bytesWritten, firstDescriptor);
   }
+
+  ReadableByteStreamControllerCallPullIfNeeded(controller);
 }
 
 function ReadableByteStreamControllerShiftPendingPullInto(controller) {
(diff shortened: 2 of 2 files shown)


MattiasBuelens:
Updated tests are in [this branch](https://github.com/MattiasBuelens/web-platform-tests/tree/readable-byte-stream-pull-after-new-chunk) ([diff](https://github.com/MattiasBuelens/web-platform-tests/compare/MattiasBuelens:master...readable-byte-stream-pull-after-new-chunk)).

We definitely need more tests, but I think @ricea first [wanted to try out the new behavior for himself](https://github.com/whatwg/streams/issues/877#issuecomment-365865223). I'll wait for his feedback first.

ricea:
Tests look good.I suggest you send a pull request so we can land those tests and then think about what other tests we need.

MattiasBuelens:
Tests have landed. Updated PR to latest WPT.

Issue #778, opened 2017-08-25
ReadableStream should be an async iterable


mariusGundersen:
(I'm adding this mostly as a tracking bug after I [asked about it in the wrong place](https://github.com/whatwg/fetch/issues/586)) 

NodeJS is [looking into supporting async iterables](https://github.com/tc39/proposal-async-iteration/issues/74) as a way to stream data, and it would be great if fetch (or the readable stream part of fetch) supported the same interface. This would make easier to move code between nodejs and the browser.

jakearchibald:
https://jakearchibald.com/2017/async-iterators-and-generators/#making-streams-iterate - article on this, including basic implementation.

mariusGundersen:
Excellent! I really look forward to this!

guest271314:
@jakearchibald Getting two errors at jsbin 

`Uncaught SyntaxError: Unexpected token *` 

when both `async` and `*` are used, and

`Uncaught SyntaxError: Unexpected reserved word`

at `await` before `for..of` loop at Chromium 60 with `--js-flags=--harmony-async-iteration` flag set.

Can you provide link documentations relevant to `--js-flags=--harmony-async-iteration` flag, which does not appear to be listed at [List of Chromium Command Line Switches](https://peter.sh/experiments/chromium-command-line-switches/)?

`await` can be used both before `iterable` at `(variable of iterable)` and within `for..of` loop `statement`. If expected result is for iteration to wait for current `Promise` value following `await`. For example


```
// wait for each `iterable` before reaching `statement`
(async() => {
  for (let prop of [await new Promise(resolve => 
      setTimeout(resolve, Math.floor(Math.random() * 10000), 1)), await new Promise(resolve => 
      setTimeout(resolve, Math.floor(Math.random() * 5000), 2))]
      ) {
    console.log(prop);
  };
 })();
```

```
// use generator function for `iterable`
(async(n) => {
  function* gen() {
    while(++n < 5) yield new Promise(resolve => setTimeout(resolve, 1000, {value:n, done:false}))
  }

  for (let prop of gen()) {
    console.log(await prop)
  }
})(0);
```

```
// pass an array of functions to be called within `statement`,
// next `iterable` awaits fulfillment of `Promise` following `await` 
const fn = async(res, ...props) => {
  for (let prop of props) {
    res.push(await prop());
    console.log(res);
  };
  return res
}

const arr = [
  () =>
    new Promise((resolve, reject) => 
      setTimeout(() =>
        resolve(0)
[...]

PR #911, opened 2018-03-27
Add two demos using streams for PNG manipulation

ksm2:
I added two demos I wrote for MDN to the list of your demos, if you'd like!


ricea:
Great stuff!

- I'd like `GrayscalePNGSource` to be a transform stream. Wrapping a ReadableStream inside another ReadableStream is a pattern I'm trying to discourage. I realise the fact that no browser has shipped TransformStream yet makes this inconvenient! If it's not possible to change this, could we link only to "Unpack chunks of a PNG" for the time being?
- In `logReadableStream()`, it would be much easier to create a WritableStream and then pipe to it. Basically the contents of `pump()` would instead be in the `write()` method, except for the `if (done)` part which would go in the `close()` method. Try it, I think you'll like it!

As a general rule, any time you're using `getReader()` or `getWriter()`, you should think about whether there would be a simpler way to do it with `pipeTo()` or `pipeThrough()`.

General comments:

- It looks like `GrayscalePNGSource` won't work properly unless the input is an rgba PNG. Maybe it should throw in other cases?
- Using
```javascript
const {value, done} = await reader.read();
```
inside a loop in an async function is much easier and cleaner that using a `pump()` function. I think all the browsers that have or will have ReadableStream already support async/await.

It took me a while to understand what "Unpack chunks of a PNG" was doing. It would probably benefit from a bit more documentation.

ksm2:
I have seen TransformStream in the spec, but without the browser support I felt not encouraged enough to build the whole demo against it if I cannot test the result in at least Chrome. Is there some kind of shim I could use to patch TransformStreams to browsers which support ReadableStream and WritableStream?

You could not make me happier with your statement about `async`/`await`! Actually, I wrote the whole code with async functions first as I am a huge fan of the concept – but MDN disapproved it fo being to complex to newbies of ES6+. (Actually, in the case of streams, I hope for [for-await-of](https://github.com/tc39/proposal-async-iteration) to enter ECMAScript as this would be the easiest way to iterate over a ReadableStream IMHO.)

Should I make the code changes here in the spec's repository in the demo directory? I also just filed a PR against MDN's repository. I am not shure they support changes there.

ricea:
> I have seen TransformStream in the spec, but without the browser support I felt not encouraged enough to build the whole demo against it if I cannot test the result in at least Chrome. Is there some kind of shim I could use to patch TransformStreams to browsers which support ReadableStream and WritableStream?

It's available in Chrome with the experimental web-platform-features flag enabled: chrome://flags/#enable-experimental-web-platform-features

It will hopefully be turned on by default in version 67.

https://streams.spec.whatwg.org/demos/transforms/transform-stream-polyfill.js is the outdated but mostly compatible polyfill used by the demos linked from the standard.

> Actually, in the case of streams, I hope for for-await-of to enter ECMAScript as this would be the easiest way to iterate over a ReadableStream IMHO.

Hopefully you'll be able to use this sooner rather than later: https://github.com/whatwg/streams/issues/778

> Should I make the code changes here in the spec's repository in the demo directory? I also just filed a PR against MDN's repository. I am not shure they support changes there.

If we cannot find a compromise with MDN then that would be my preference, yes.

ksm2:
> If we cannot find a compromise with MDN then that would be my preference, yes.

Hey @ricea, the folks at MDN merged my updates now using TransformStreams!

demos/index.html
```diff
@@ -25,6 +25,12 @@ <h1>Streams Demos</h1>
 
   <dt><a href="https://fetch-progress.anthum.com/">Progress indicators with Service Workers and Fetch</a> | <a href="https://github.com/AnthumChris/fetch-progress-indicators">(Source Code)</a>
   <dd>Streams intercept Service Worker <code>FetchEvent</code> and <code>fetch()</code> statements to show progress indicators. See <a href="https://github.com/AnthumChris/fetch-progress-indicators#browser-support">browser support</a> notes.
+
+  <dt><a href="http://mdn.github.io/dom-examples/streams/grayscale-png/">Grayscale a PNG</a> | <a href="https://github.com/mdn/dom-examples/tree/master/streams/grayscale-png">(Source Code)</a>
+  <dd>A PNG gets fetched to receive a ReadableStream on it, than the PNG data is converted to black and white and enqueued into another ReadableStream.
```
ricea: Could you update this comment to indicate that it uses TransformStream?

Also, in the demo itself, it would nice if it displayed a message in the page when TransformStream is not defined, to avoid confusion.

ricea: We don't need to block this on the changing http://mdn.github.io/dom-examples/streams/grayscale-png/. The MDN people might not agree to that anyway.

Issue #907, opened 2018-03-09
Consider better styling for API intro notes


domenic:
Compare the green notes at https://streams.spec.whatwg.org/#rs-prototype to those right below https://dom.spec.whatwg.org/#event-path.

We're different than them in that we only do one API per section, so at first glance that list style isn't that necessary. But the "domintro" style shown there is a bit better than ours in a few ways:

- Nice "For web developers (non-normative)" label, instead of "Note"
- Not everything is italic.
- They don't repeat the API name (E.g. instead of "The pipeThrough method provides..." we'd just say "Provides...")
- They give very bare-bones usage examples and notation, which is sometimes nice. For example, we could use `[, x]` to denote optionality, or spell out dictionaries even if the standard treats them as a single variable, or show that `tee()`s return value is usually used with destructuring, or insert `await` before promise-returning method calls... Could be nice.

Should we adapt? Should we do anything different than most specs do, because of our one-API-per-section nature? Thoughts appreciated. Once we have a plan, I'd like to turn this into a "good first bug".

ricea:
I think "Not everything is italic" is the biggest selling point for me :smile:. We'd need at least two types, because some of our notes are for implementers.

The DOM spec example looks more like our "Class definitions" than our "Note" sections. Moving usage notes into the "Class definition" and rewriting it in that style might be a way forward.

domenic:
My latest take on this is in https://domenic.github.io/async-local-storage/; see e.g. https://domenic.github.io/async-local-storage/#storagearea-set.

I'm undecided on whether we should have this in the method definitions, or maybe consolidate them all into one place per class. A nice feature of the last is that it allows us to nuke the class definitions, which have caused significant confusion (#45, #732, and some recent IRC conversations).

surma:
I like the way it looks in your async-local-storage draft. 

You still have a [class definition in the current draft](https://domenic.github.io/async-local-storage/#storagearea). I think they should not get nuked as they make the spec very navigational (provided that they get linked up in the same way I’ve done in #872 and #913).

domenic:
That's a bit different, as there it's used to normatively create the class.

I think if all the domintros were consolidated into one block (see e.g. below the IDL block at https://dom.spec.whatwg.org/#interface-abortcontroller) then having the class definition nearby would be just redundant. So that's why we'd get to nuke the class definition.

[...]

PR #913, opened 2018-03-28
Linkify all class definitions

surma:
As a follow-up to #872, I linkified all the other class definitions as well.

***
<a href="https://whatpr.org/streams/913.html" title="Last updated on Apr 9, 2018, 9:06 PM GMT (05a4927)">Preview</a> | <a href="https://whatpr.org/streams/913/cd63477...05a4927.html" title="Last updated on Apr 9, 2018, 9:06 PM GMT (05a4927)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1240,13 +1240,13 @@ would look like
 
 <pre><code class="lang-javascript">
   class ReadableStreamDefaultReader {
-    constructor(stream)
+    <a href="#default-reader-constructor">constructor(stream)</a>
 
-    get closed()
+    get <a href="#default-reader-closed">closed</a>()
 
-    cancel(reason)
-    read()
-    releaseLock()
+    <a href="#default-reader-cancel">cancel</a>(reason)
+    <a href="#default-reader-read">read</a>()
+    <a href="#default-reader-release-lock">releaseLock</a>()
   }
 </code></pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3539,7 +3539,7 @@ lt="WritableStreamDefaultWriter(stream)">new WritableStreamDefaultWriter(<var>st
   1. Return *this*.[[closedPromise]].
 </emu-alg>
 
-<h5 id="default-writer-desiredSize" attribute for="WritableStreamDefaultWriter" lt="desiredSize">get desiredSize</h5>
+<h5 id="default-writer-desired-size" attribute for="WritableStreamDefaultWriter" lt="desiredSize">get desiredSize</h5>
```
surma: I changed this to have consistent spelling with the other IDs and couldn’t find any other references to it. I am assuming I’m not breaking anything herE?

index.bs
```diff
@@ -1240,13 +1240,13 @@ would look like
 
 <pre><code class="lang-javascript">
   class ReadableStreamDefaultReader {
-    constructor(stream)
+    <a href="#default-reader-constructor">constructor(stream)</a>
```
ricea: This is desirable for readability of the output, but it sure makes the source unreadable.

@tabatkins Can Bikeshed help our markup look nicer here?

surma: I had a different markup in #872 but the consensus was that simple a tags were preferred. I think these links could be automatically generated, so I am in favor of somehow using bikeshed for this.

domenic: If we end up doing what @ricea suggested in #907 we'd align more with existing specs so I wouldn't suggest adding anything to Bikeshed for this definition.

That said, the domintro blocks (as proposed in #907) are also not very readable in the source either. See e.g. https://github.com/domenic/async-local-storage/blob/master/spec.bs#L239-L249 for my latest attempt. (Explicit `<p>`s needed because of https://github.com/tabatkins/bikeshed/issues/1203.)

Personally I'm not too worried about the source readability here.

surma:
🎉🎉🎉

Issue #917, opened 2018-03-30
ReadableStream constructor: standard text and reference implementation doesn't match


ricea:
In the standard text, underlyingSource.type is looked up before strategy.size, but the reference implementation still uses destructuring assignment in the constructor arguments and so looks up "size" first.

This difference is observable, but none of our tests are observing it.

ricea:
I'm going to work on it, and testing the ordering of lookups in all the constructors.

ricea:
Turns out the WritableStream constructor in the reference implementation doesn't match either.

Some of these orderings are a little odd. For example, the WritableStream constructor looks up type, size and highWaterMark, and only then checks whether type is undefined or not. If I'm going to write a test for this, I'd rather the behaviour be something I can explain. @domenic would it be okay to reorder things so that parameters are in general checked right after being looked up?

domenic:
I think the idea was to make things easier to implement using destructuring or Web IDL. In those cases, lookups all happen first, then inspection of the extracted values.

ricea:
It's easy enough for a JS implementation to switch from destructuring to explicitly lookups. Blink's implementation has switched back and forth at various times. So I'm not worried about that.

However, for C++ implementations, trying to unpack an options object directly from C++ can be cumbersome and buggy, as Blink's RequestInit constructor demonstrates: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/fetch/RequestInit.cpp?q=RequestInit::RequestInit.

If you use WebIDL then you end up with this algorithm:

https://heycam.github.io/webidl/#es-dictionary

On the positive side, most validation ends up interleaved with lookup this way, which is how I want to do it. On the negative

- WebIDL wants us to [convert](https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value) the arguments in order. This is problematic now that we've split out the digestion of the underlying object into SetUp\*DefaultControllerFromUnderlying\*, which always happens after processing the strategy.
- WebIDL wants us to lookup dictionary members in lexicographical order. This is doable but annoying.
- The check that highWaterMark is >=0 will happen in a later step from the conversion of highWaterMark to a number, which is really weird from the point of view of our standard text.

Having said that, I haven't written a C++ implementation and I don't know how big of a problem this is in practice. @tschneidereit, do you have any input?

I'm thinking the solution may be to add a normative note to each constructor saying something like "All these lookups, type conversions, and validations must be performed, but implementations may change the order".

[...]

Issue #921, opened 2018-04-12
Numeric conversion is not done on autoAllocateChunkSize


ricea:
https://streams.spec.whatwg.org/commit-snapshots/6f97246dd5577b2b15a8991b5bc5f63939c8f59d/#set-up-readable-byte-stream-controller-from-underlying-source

> If \! IsInteger\(autoAllocateChunkSize\) is false

In other cases when we accept a number, we call ToNumber() on it. For example, you can do new ReadableStream({}, { highWaterMark: "7" }) and it will do what you'd expect. But we don't do this for autoAllocateChunkSize. We probably should.

@domenic Do you know of any reason not to do this?

domenic:
This is just a bug; great find.

ricea:
Fixed in #922.

PR #922, opened 2018-04-13
Revise property access order

ricea:
Make the order that properties are accessed on the underlying
source, underlying sink, transformer and strategy objects more
consistent. Also make the order that properties are validated more
consistent.

The general order of lookups is now size, highWaterMark, type,
methods. In general values are validated immediately after lookup, but
size and highWaterMark are exceptions. They are always looked up, but
the validation needed depends on the value of "type", so lookup and
validation are separated. This applies even to WritableStream and
TransformStream, to allow for the addition of byte types to those in
future.

The order of lookups in the reference implementation now matches the
standard.

Also convert autoAllocateChunkSize to a number before checking whether
it is integer, for consistency with the way other numeric input is
handled.

Fixes: #917, #921

***
<a href="https://whatpr.org/streams/922.html" title="Last updated on Apr 24, 2018, 9:56 AM GMT (827d8a9)">Preview</a> | <a href="https://whatpr.org/streams/922/c3a23ee...827d8a9.html" title="Last updated on Apr 24, 2018, 9:56 AM GMT (827d8a9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -457,19 +457,19 @@ ReadableStream(<var>underlyingSource</var> = {}, <var>strategy</var> = {})</h4>
 
 <emu-alg>
   1. Perform ! InitializeReadableStream(*this*).
-  1. Let _type_ be ? GetV(_underlyingSource_, `"type"`).
   1. Let _size_ be ? GetV(_strategy_, `"size"`).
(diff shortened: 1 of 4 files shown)


ricea:
Tests are in https://github.com/w3c/web-platform-tests/pull/10461.

PR #924, opened 2018-05-02
Changes to Example 8.9: Fix code issue, add const, add text decode step

ksm2:
Hi WHATWG team, I want to propose the following changes to example 8.9 in order to clarify it to the reader:
- There was a closing brace missing which I've added.
- The `url` variable is used out of nowhere; I would recommend to take the one coming from the `fetchEvent` which comes close to a real world scenario when using a ServiceWorker.
- The example won't work in browsers as the response body is binary-encoded but the `LipFuzzTransformer` requires string chunks. Therefor, I would recommend adding a text decoding step before applying the former. Hence, the reader is able to try the example out in her/his browser.

***
<a href="https://whatpr.org/streams/924.html" title="Last updated on May 15, 2018, 9:22 PM GMT (dfb3701)">Preview</a> | <a href="https://whatpr.org/streams/924/af48a18...dfb3701.html" title="Last updated on May 15, 2018, 9:22 PM GMT (dfb3701)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5692,25 +5692,16 @@ useful when there is instance data to track.
 The class would be used in code like:
 
 <pre><code class="lang-javascript">
-  function textDecoderStream() {
-    const td = new TextDecoder();
-    return new TransformStream({
-      transform(chunk, controller) {
-        controller.enqueue(td.decode(chunk));
-      }
-    });
-  }
-
   const data = { userName, displayName, icon, date };
   const ts = new TransformStream(new LipFuzzTransformer(data));
-  const { url } = fetchEvent.request;
 
   fetchEvent.respondWith(
-    fetch(url).then(response => {
-      // Decode the binary-encoded response to string
-      const decodedBody = response.body.pipeThrough(textDecoderStream());
-      // Apply the LipFuzzTransformer
-      const transformedBody = decodedBody.pipeThrough(ts);
+    fetch(fetchEvent.request.url).then(response => {
+      const transformedBody = response.body
+        // Decode the binary-encoded response to string
+        .pipeThrough(new TextDecoderStream())
+        // Apply the LipFuzzTransformer
+        .pipeThrough(ts);
       return new Response(transformedBody);
     })
   );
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5692,14 +5692,27 @@ useful when there is instance data to track.
 The class would be used in code like:
 
 <pre><code class="lang-javascript">
+  function textDecoderStream() {
```
ricea: I would much rather use the TextDecoderStream class that is planned to be added to the encoding standard (https://github.com/whatwg/encoding/issues/72), even though it isn't available yet. It avoids spending a lot of space on this mostly-unrelated topic, and avoids data corruption when characters are split between chunk boundaries.

index.bs
```diff
@@ -5692,14 +5692,27 @@ useful when there is instance data to track.
...
+    return new TransformStream({
+      transform(chunk, controller) {
+        controller.enqueue(td.decode(chunk));
+      }
+    });
+  }
+
   const data = { userName, displayName, icon, date };
   const ts = new TransformStream(new LipFuzzTransformer(data));
+  const { url } = fetchEvent.request;
```
ricea: Might be better not to have a `url` intermediate value, but just extract the value directly in the call to fetch().

index.bs
```diff
@@ -5692,14 +5692,27 @@ useful when there is instance data to track.
...
 
   fetchEvent.respondWith(
     fetch(url).then(response => {
-      const transformedBody = response.body.pipeThrough(ts);
+      // Decode the binary-encoded response to string
+      const decodedBody = response.body.pipeThrough(textDecoderStream());
```
ricea: It's better to not create a `decodedBody` variable, but just chain the pipes like this

```javascript
const transformedBody = response.body
  .pipeThrough(new TextDecoderStream())
  .pipeThrough(ts);
```
I don't think it's necessary to comment each operation separately, but you can if you think it is clearer.

ricea:
@ksm2 I can apply my recommended changes if you do not have time. Please let me know.

ksm2:
@ricea sorry for letting you wait! If it's okay, I can apply the changes this evening (CEST)!

ksm2:
@ricea Sorry again for letting you wait!
I kept the comments around `pipeThrough` statements, though – from my experience in explaining the Streams API to collegues you cannot have enough documentation to describe what is happening where! :smile:

PR #926, opened 2018-05-18
ShouldCallPull: Assert desiredSize is not null

### LINKED ISSUE #919: Assert that desiredSize is non-null in ReadableStreamDefaultControllerShouldCallPull
https://streams.spec.whatwg.org/commit-snapshots/c3a23ee03a5242b9c7517006c849b114ae121cfc/#readable-stream-default-controller-should-call-pull

> Let desiredSize be ReadableStreamDefaultControllerGetDesiredSize\(controller\)\.
> If desiredSize &gt; 0, return true\.
> 

There should probably be an assert that desiredSize is non-null here.

ricea:
ReadableStreamDefaultControllerShouldCallPull implicitly assumes that
desiredSize is not null when it compares it to 0. This is guaranteed by
the check of
ReadableStreamDefaultControllerCanCloseOrEnqueue(controller), but it's
not obvious. Add an assert to make it explicit.

Also do the same for ReadableByteStreamControllerShouldCallPull.

No functional changes.

Closes #919.

***
<a href="https://whatpr.org/streams/926.html" title="Last updated on May 23, 2018, 7:05 AM GMT (edcef0a)">Preview</a> | <a href="https://whatpr.org/streams/926/d98f243...edcef0a.html" title="Last updated on May 23, 2018, 7:05 AM GMT (edcef0a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1826,6 +1826,7 @@ nothrow>ReadableStreamDefaultControllerShouldCallPull ( <var>controller</var> )<
   1. If ! IsReadableStreamLocked(_stream_) is *true* and ! ReadableStreamGetNumReadRequests(_stream_) > *0*, return
      *true*.
   1. Let _desiredSize_ be ReadableStreamDefaultControllerGetDesiredSize(_controller_).
+  1. Assert: _desiredSize_ is not *null*.
   1. If _desiredSize_ > *0*, return *true*.
   1. Return *false*.
 </emu-alg>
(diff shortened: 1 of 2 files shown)


ricea:
@domenic What do you think?

index.bs
```diff
@@ -1826,6 +1826,7 @@ nothrow>ReadableStreamDefaultControllerShouldCallPull ( <var>controller</var> )<
   1. If ! IsReadableStreamLocked(_stream_) is *true* and ! ReadableStreamGetNumReadRequests(_stream_) > *0*, return
      *true*.
   1. Let _desiredSize_ be ReadableStreamDefaultControllerGetDesiredSize(_controller_).
```
domenic: While here can you add a `!`?

ricea: Done, thanks.

PR #928, opened 2018-05-29
Meta: Review Draft infrastructure

annevk:
See https://github.com/whatwg/whatwg.org/issues/197 and https://github.com/whatwg/meta/issues/92 for details.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,5 +14,5 @@ max_line_length = 120
 [.gitmodules]
 indent_style = tab
 
-[deploy.sh]
-indent_size = 4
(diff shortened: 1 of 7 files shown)


annevk:
So most of the stuff said about `make` in https://github.com/whatwg/meta/pull/86 does not appear to be true for Streams. How do you want to resolve that conflict?

domenic:
Let me see if I can make it work.

ricea:
I think it would make it easier for contributors if we added "local" and "deploy" targets to match the other standards.

PR #930, opened 2018-06-07
Add a text encoding step to example 8.9

MattiasBuelens:
In #924, a text decoding step was added to example 8.9 so it could run in the browser. However, the reverse needs to happen in order to construct a valid `Response` with the transformed stream. Specifically, in step 3 of "read all bytes from a ReadableStream" from [the Fetch specification](https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream), the stream must produce `Uint8Array` objects for it to be consumed correctly.

This PR fixes this by adding an extra pipe through a `TextEncoderStream`.

***
<a href="https://whatpr.org/streams/930.html" title="Last updated on Jun 7, 2018, 7:51 PM GMT (6a14699)">Preview</a> | <a href="https://whatpr.org/streams/930/08145be...6a14699.html" title="Last updated on Jun 7, 2018, 7:51 PM GMT (6a14699)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5703,7 +5703,9 @@ The class would be used in code like:
         // Decode the binary-encoded response to string
         .pipeThrough(new TextDecoderStream())
         // Apply the LipFuzzTransformer
-        .pipeThrough(ts);
+        .pipeThrough(ts)
+        // Encode the transformed string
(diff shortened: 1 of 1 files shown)


ricea:
lgtm

PR #936, opened 2018-07-02
Don't create iterators with Object.prototype for internal use

### LINKED ISSUE #933: Setting Object.prototype.then permits interfering with pipeTo() internals
The following test fails in Chromium:
```javascript
promise_test(async () => {
  const intercepted = [];

    start(controller) {
      controller.enqueue('a');
      controller.close();
    }
  });

  let callCount = 0;
  Object.prototype.then = function(resolver) {
    if (!this.done) {
      intercepted.push(this.value);
    }
    const retval = Object.create(null);
    retval.done = ++callCount === 3;
    retval.value = callCount;
    resolver(retval);
  };

  await rs.pipeTo(ws);
  delete Object.prototype.then;

  assert_array_equals(intercepted, [], 'nothing should have been intercepted');
  assert_array_equals(ws.events, ['write', 'a', 'close'], 'written chunk should be "a"');
}, 'piping should not be observable');
```
The actual observed values of `intercepted` and `ws.events` are `["a"]` and `["write", 1, "write", 2, "close"]` respectively. This violates the design principle that the internals of pipeTo() should not be observable.

The test passes in the reference implementation, but for the wrong reason: a side-effect of the way the tests are run in Node.js is that the iterator object is created in the wrong realm. I am pretty sure from the definition of [CreateIterResultObject()](https://tc39.github.io/ecma262/#sec-createiterresultobject) that the iterator object is supposed to be created in the realm of the page.

The reason the test works/fails is this step in the standard:
* [Resolve](https://www.w3.org/2001/tag/doc/promises-guide/#resolve-promise) _readRequest_.[[promise]] with ! CreateIterResultObject().

"Resolve" ends up calling https://tc39.github.io/ecma262/#sec-promise-resolve-functions which looks up `"then"` on the prototype chain, ending up at `Object.prototype.then`. This provides a hook where the page can execute arbitrary Javascript, including reading and writing the iterator object. The hook is normally harmless, except in pipeTo().

I think it would be perfectly valid for an implementation of pipeTo() to pass this test even if CreateIterResultObject() created Objects in the current realm. I don't think pipeTo() is required to call "Resolve" or "CreateIterResultObject" at all. An optimised implementation would probably skip those steps. However, setting `Object.prototype.then` makes the optimisation observable.

I think it's important that implementations can optimise pipeTo() without it being observable to the page. I can think of a couple of ways to fix this:
1. Specify pipeTo() as not calling "resolve" or CreateIterResultObject().
2. Use a special variant of CreateIterResultObject() that sets a `null` prototype on the iterator object.

I prefer the second option. I think changing pipeTo() would add a lot of complexity to the standard text, and restrict implementation choice.

ricea:
An iterator object which inherits from Object.prototype can make calls
to Promise.resolve observable to use code. This is inappropriate when
the promise is being resolved internally, so make the behaviour be
controlled by a parameter that defaults to false.

Fixes #933.

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -24,14 +24,6 @@ exports.ArrayBufferCopy = (dest, destOffset, src, srcOffset, n) => {
   new Uint8Array(dest).set(new Uint8Array(src, srcOffset, n), destOffset);
 };
 
-exports.CreateIterResultObject = (value, done) => {
-  assert(typeof done === 'boolean');
-  const obj = {};
-  Object.defineProperty(obj, 'value', { value, enumerable: true, writable: true, configurable: true });
-  Object.defineProperty(obj, 'done', { value: done, enumerable: true, writable: true, configurable: true });
-  return obj;
-};
-
 exports.IsFiniteNonNegativeNumber = v => {
   if (exports.IsNonNegativeNumber(v) === false) {
     return false;
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -1,7 +1,7 @@
 'use strict';
 const assert = require('better-assert');
-const { ArrayBufferCopy, CreateAlgorithmFromUnderlyingMethod, CreateIterResultObject, IsFiniteNonNegativeNumber,
-        InvokeOrNoop, IsDetachedBuffer, TransferArrayBuffer, ValidateAndNormalizeHighWaterMark, IsNonNegativeNumber,
+const { ArrayBufferCopy, CreateAlgorithmFromUnderlyingMethod, IsFiniteNonNegativeNumber, InvokeOrNoop,
+        IsDetachedBuffer, TransferArrayBuffer, ValidateAndNormalizeHighWaterMark, IsNonNegativeNumber,
         MakeSizeAlgorithmFromSizeFunction, createArrayFromList, typeIsObject } = require('./helpers.js');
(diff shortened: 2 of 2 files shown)


ricea:
Only the reference implementation changes are done so far.

ricea:
I need better names for the `useObjectPrototype` parameter and the `CreateIterResultObjectPrototype` abstract operation.

domenic:
This isn't too bad; thanks for proving it out.

For naming, hmm... `CreateStreamIterResultObject` maybe? Or `ReadableStreamCreateIterResultObject`? Or maybe depart from the "iter-result" paradigm and go with something like `CreateChunkObject` or `CreateReadResult` or similar (optionally with `ReadableStream` prefix).

I think `useObjectPrototype` is a fine name for the parameter, but an alternate direction would be to talk about the reasoning instead of the implementation details, e.g. `exposedToAuthorCode` or `exposedToJS` or `internal` or similar.

We'll definitely need a note explaining/apologizing for this operation. I'm happy to help with the wording on that.

ricea:
Okay, functionally I'm happy with this. The note on ReadableStreamCreateReadResult could probably use some improvement to convey the proper degree of apology. Feel free to make changes directly on the branch.

Maybe some additional commentary is needed on the changed operations, such as DefaultReaderRead? I don't know.

ricea:
Tests at https://github.com/web-platform-tests/wpt/pull/11767.

domenic:
Oh, commit message should probably say { value, done } objects instead of "iterators". Iterators is a different thing.

Issue #932, opened 2018-06-18
Permit freeing underlying source / sink / transformer after close


ricea:
Once a stream is closed or errored, none of the methods on the underlying source / sink / transformer will be called any more. As a quality-of-implementation enhancement some browsers might decide to drop their references at that point so that the underlying object can be garbage collected even if the stream is still referenced.

This is currently a valid optimisation, however if someone adds weak refs to Javascript it will become observable.

I propose that we explicitly drop the references in the standard to ensure this optimisation is valid. Implementations still won't be required to drop the references because they're not required to perform GC at any particular time or on any particular object.

domenic:
Great catch, and strong +1. Weird to think about weak references making this sort of thing observable :-/.

ricea:
It's actually tricky to follow through all the operations to find the necessary and sufficient places to clear the algorithms slots, so I decided I'd do TransformStream, ReadableStream and WritableStream separately to make it easier.

ricea:
As @domenic mentioned in #944, we also need to clear [[strategySizeAlgorithm]].

ricea:
All done.

PR #937, opened 2018-07-02
TransformStream: clear algorithms once they will no longer be called

ricea:
Clear the flushAlgorithm and transformAlgorithm slots when they will no
longer be called. This allows resources such as the transformer object
to be freed.

The case where transformAlgorithm returns a rejection isn't currently
handled because it can't be done efficiently as the moment. Once
finally() is added to the WritableStream underlying sink interface this
will be trivial.

Part of #932.

***
<a href="https://whatpr.org/streams/937.html" title="Last updated on Jul 3, 2018, 4:03 AM GMT (18e3c12)">Preview</a> | <a href="https://whatpr.org/streams/937/eff8161...18e3c12.html" title="Last updated on Jul 3, 2018, 4:03 AM GMT (18e3c12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4431,6 +4431,7 @@ algorithms do not need to check stream states when responding to an error condit
 nothrow>TransformStreamErrorWritableAndUnblockWrite ( <var>stream</var>, <var>e</var> )</h4>
 
 <emu-alg>
+  1. Perform ! TransformStreamDefaultControllerClearAlgorithms(_stream_.[[transformStreamController]]);
   1. Perform ! WritableStreamDefaultControllerErrorIfNeeded(_stream_.[[writable]].[[writableStreamController]], _e_).
   1. If _stream_.[[backpressure]] is *true*, perform ! TransformStreamSetBackpressure(_stream_, *false*).
 </emu-alg>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -4621,6 +4622,18 @@ throws>SetUpTransformStreamDefaultControllerFromTransformer ( <var>stream</var>,
   1. Perform ! SetUpTransformStreamDefaultController(_stream_, _controller_, _transformAlgorithm_, _flushAlgorithm_).
 </emu-alg>
 
+<h4 id="transform-stream-default-controller-clear-algorithms" aoid="TransformStreamDefaultControllerClearAlgorithms"
+nothrow>TransformStreamDefaultControllerClearAlgorithms ( <var>controller</var> )</h4>
+
+This abstract operation is called once the stream is closed or errored and the algorithms will not be executed any more.
+By removing the algorithm references it permits the <a>transformer</a> object to be garbage collected even if the
+{{TransformStream}} itself is still referenced.
```
domenic: Maybe add

```html
<p class="note">The results of this algorithm are not currently observable, but could become so if JavaScript eventually adds <a href="https://github.com/tc39/proposal-weakrefs/">weak references</a>. But even without that factor, implementations will likely want to include similar steps.</p>
```

or some similar phrasing?

PR #940, opened 2018-07-11
WritableStream: clear algorithms once they will no longer be called

ricea:
Clear the [[writeAlgorithm]], [[closeAlgorithm]] and [[abortAlgorithm]]
slots when they will no longer be called. This allows resources such as
the underlying sink object to be freed.

Add [[FinishSteps]] polymorphic operation so that the algorithms will be
cleared correctly even for a future controller that has different slots.
[[FinishSteps]] is only used when the clear operation is initiated from
the WritableStream -- in the case of abort() and close() the clear
operation is initiated by the controller itself and doesn't indirect via
[[FinishSteps]].

Part of #932.

***
<a href="https://whatpr.org/streams/940.html" title="Last updated on Jul 13, 2018, 7:17 AM GMT (6f1ac85)">Preview</a> | <a href="https://whatpr.org/streams/940/a27a1fd...6f1ac85.html" title="Last updated on Jul 13, 2018, 7:17 AM GMT (6f1ac85)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4026,9 +4026,9 @@ By removing the algorithm references it permits the <a>underlying sink</a> objec
 
 <p class="note">The results of this algorithm are not currently observable, but could become so if JavaScript eventually
 adds <a href="https://github.com/tc39/proposal-weakrefs/">weak references</a>. But even without that factor,
-implementations will likely want to include similar steps.
+implementations will likely want to include similar steps.</p>
 
-This operation will be performed multiple times in some edge cases. After the first time it will do nothing.</p>
+<p class="note">This operation will be performed multiple times in some edge cases. After the first time it will do nothing.</p>
 
 <emu-alg>
   1. Set _controller_.[[writeAlgorithm]] to *undefined*.
--- a/index.bs
+++ b/index.bs
@@ -3334,24 +3334,21 @@ nothrow>WritableStreamDealWithRejection ( <var>stream</var>, <var>error</var> )<
   1. Assert: _stream_.[[state]] is `"erroring"`.
   1. Assert: ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*.
   1. Set _stream_.[[state]] to `"errored"`.
-  1. Let _controller_ be _stream_.[[writableStreamController]].
-  1. Perform ! _controller_.[[ErrorSteps]]().
+  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
   1. Let _storedError_ be _stream_.[[storedError]].
   1. Repeat for each _writeRequest_ that is an element of _stream_.[[writeRequests]],
     1. <a>Reject</a> _writeRequest_ with _storedError_.
   1. Set _stream_.[[writeRequests]] to an empty List.
-  1. if _stream_.[[pendingAbortRequest]] is *undefined*,
-    1. Perform ! _controller_[[FinishSteps]]().
+  1. If _stream_.[[pendingAbortRequest]] is *undefined*,
     1. Perform ! WritableStreamRejectCloseAndClosedPromiseIfNeeded(_stream_).
     1. Return.
   1. Let _abortRequest_ be _stream_.[[pendingAbortRequest]].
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -3291,21 +3291,24 @@ nothrow>WritableStreamDealWithRejection ( <var>stream</var>, <var>error</var> )<
...
   1. Set _stream_.[[state]] to `"errored"`.
-  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
+  1. Let _controller_ be _stream_.[[writableStreamController]].
+  1. Perform ! _controller_.[[ErrorSteps]]().
   1. Let _storedError_ be _stream_.[[storedError]].
   1. Repeat for each _writeRequest_ that is an element of _stream_.[[writeRequests]],
     1. <a>Reject</a> _writeRequest_ with _storedError_.
   1. Set _stream_.[[writeRequests]] to an empty List.
   1. if _stream_.[[pendingAbortRequest]] is *undefined*,
+    1. Perform ! _controller_[[FinishSteps]]().
```
domenic: Missing dots here and below.

index.bs
```diff
@@ -3291,21 +3291,24 @@ nothrow>WritableStreamDealWithRejection ( <var>stream</var>, <var>error</var> )<
...
   1. Assert: ! WritableStreamHasOperationMarkedInFlight(_stream_) is *false*.
   1. Set _stream_.[[state]] to `"errored"`.
-  1. Perform ! _stream_.[[writableStreamController]].[[ErrorSteps]]().
+  1. Let _controller_ be _stream_.[[writableStreamController]].
+  1. Perform ! _controller_.[[ErrorSteps]]().
   1. Let _storedError_ be _stream_.[[storedError]].
   1. Repeat for each _writeRequest_ that is an element of _stream_.[[writeRequests]],
     1. <a>Reject</a> _writeRequest_ with _storedError_.
   1. Set _stream_.[[writeRequests]] to an empty List.
   1. if _stream_.[[pendingAbortRequest]] is *undefined*,
```
domenic: While here, maybe fix this lowercase "if"

ricea:
> In the two cases, i.e. steps 9 and 12 of WritableStreamFinishedErroring---is it possible there was an earlier point where the controller could know that its algorithms are clearable?

I managed to get this working. PTAL.

index.bs
```diff
@@ -4015,6 +4017,25 @@ throws>SetUpWritableStreamDefaultControllerFromUnderlyingSink ( <var>stream</var
...
+
+This abstract operation is called once the stream is closed or errored and the algorithms will not be executed any more.
+By removing the algorithm references it permits the <a>underlying sink</a> object to be garbage collected even if the
+{{WritableStream}} itself is still referenced.
+
+<p class="note">The results of this algorithm are not currently observable, but could become so if JavaScript eventually
+adds <a href="https://github.com/tc39/proposal-weakrefs/">weak references</a>. But even without that factor,
+implementations will likely want to include similar steps.</p>
+
+<p class="note">This operation will be performed multiple times in some edge cases. After the first time it will do nothing.</p>
```
domenic: Nit: line too long

reference-implementation/lib/writable-stream.js
```diff
@@ -919,6 +929,9 @@ function WritableStreamDefaultControllerProcessWrite(controller, chunk) {
       WritableStreamDefaultControllerAdvanceQueueIfNeeded(controller);
     },
     reason => {
+      if (stream._state === 'writable') {
```
MattiasBuelens: This check does not appear [in the spec text](https://streams.spec.whatwg.org/commit-snapshots/323223e65455e1f22be656b01fe2d810902db2ac/#writable-stream-default-controller-process-write). I believe it should be removed from the reference implementation?

ricea: Great catch! Actually this check is required, and I failed to add it to the spec text. Without this check, if a write() rejects while an abort() is already in progress, the abortAlgorithm will be cleared too early and so when the time comes to call the underlying source abort() it will fail.

PR #943, opened 2018-07-14
Add missing check for WritableStream...ClearAlgorithms

ricea:
When an underlying sink write is rejected but an abort is already in
process, the underlying sink abort() method still needs to be called, so
the algorithms cannot be cleared yet.

This was correct in the reference implementation but the corresponding
condition was missing from the standard text.

Add the missing check.

Thanks to @MattiasBuelens for pointing out the issue.

***
<a href="https://whatpr.org/streams/943.html" title="Last updated on Jul 14, 2018, 3:06 PM GMT (0cd0a3d)">Preview</a> | <a href="https://whatpr.org/streams/943/323223e...0cd0a3d.html" title="Last updated on Jul 14, 2018, 3:06 PM GMT (0cd0a3d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4141,7 +4141,7 @@ nothrow>WritableStreamDefaultControllerProcessWrite ( <var>controller</var>, <va
       1. Perform ! WritableStreamUpdateBackpressure(_stream_, _backpressure_).
     1. Perform ! WritableStreamDefaultControllerAdvanceQueueIfNeeded(_controller_).
   1. <a>Upon rejection</a> of _sinkWritePromise_ with _reason_,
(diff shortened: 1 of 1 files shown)


domenic:
Do we have coverage for this? I.e. if we remove the check from the reference implementation, will some tests fail.

ricea:
Many tests fail without the check. '.closed should not resolve before rejected write(); write() error should not overwrite abort() error' in [writable-streams/aborting.js](https://github.com/web-platform-tests/wpt/blob/6bbdc36d277bf24bef261034660e8e6a8aaee58f/streams/writable-streams/aborting.js) is probably the most specific one.

PR #956, opened 2018-10-10
Add explainer for transferable streams

ricea:


MattiasBuelens:
This sounds great! 😄 

After a `ReadableStream` or `WritableStream` is transferred, the original thread must no longer be able to use it. I guess that means it should be considered "locked" from the perspective of that thread? That is: `locked` equals `true`, and `getReader()` and `getWriter()` throw an "already locked" error?

> It's possible to emulate this behaviour by using postMessage() directly. Existing techniques for moving work off the main thread often resemble a subset of this functionality. However, it is clumsy and hard to work with.

Can confirm, very clumsy. That didn't stop me from trying though: [remote-web-streams](https://github.com/MattiasBuelens/remote-web-streams). 😛

ricea:
> After a ReadableStream or WritableStream is transferred, the original thread must no longer be able to use it. I guess that means it should be considered "locked" from the perspective of that thread?

Yes. I should probably mention that in the explainer, but I think I will do it in a follow-up PR.

ricea:
@MattiasBuelens If you have a chance, could you try out the implementation in Chrome Canary (requires --enable-experimental-web-platform-features flag)? I'm interested in how it performs compared to your remote-web-streams implementation.

MattiasBuelens:
@ricea I've already tried it out a bit in Canary, but I haven't done any performance comparisons yet. I'll quickly add a variant using native transferable streams to [my examples](https://mattiasbuelens.github.io/remote-web-streams/examples/) and see how they compare. 🙂

PR #966, opened 2018-11-22
Add a brand check to pipeThrough()

### LINKED ISSUE #961: Make pipeThough() non-generic?
Currently `pipeThrough()` operates generically on **this**. It performs no brand check, and will attempt to call `this.pipeTo()` regardless of whether **this** is a ReadableStream or not.

Blink is currently changing its streams implementation to use IDL, and there's no way to directly express this in IDL. We will need some workaround to pass the tests.

Having `pipeThrough()` be generic complicates the standard (because we have to check if the return value of `pipeTo()` is a Promise), and complicates IDL-based implementations. I think it doesn't provide much value, and the easiest thing to do would be to add a brand check and make it call an abstract PipeTo operation internally.

@domenic, what do you think?

ricea:
Require pipeThrough() to be called on a ReadableStream. Stop delegating to
this.pipeTo().

Also throw exceptions if *this* or _writable_ are not streams or they are
locked, or if *signal* is not an AbortSignal. Also require _readable_ to be a
ReadableStream.

Add a new abstract operation, ReadableStreamPipeTo(), which is used internally
by pipeTo() and pipeThrough() and also can be referenced from other
standards.

Closes #961.


index.bs
```diff
@@ -661,31 +661,31 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
 </div>
 
 <h5 id="rs-pipe-through" method for="ReadableStream" lt="pipeThrough(transform, options)">pipeThrough({
-<var ignore>writable</var>, <var ignore>readable</var> }, <var ignore>options</var>)</h5>
+<var ignore>writable</var>, <var ignore>readable</var> }, { <var>preventClose</var>, <var>preventAbort</var>,
+<var>preventCancel</var>, <var>signal</var> } = {})</h5>
```
domenic: Also update the class definition to use this style?

index.bs
```diff
@@ -661,31 +661,31 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
...
-  1. If Type(_promise_) is Object and _promise_ has a [[PromiseIsHandled]] internal slot, set
-     _promise_.[[PromiseIsHandled]] to *true*.
+  1. If ! IsReadableStream(*this*) is *false*, throw a *TypeError* exception.
+  1. If ! IsWritableStream(_writable_) is *false*, throw a *TypeError* exception.
+  1. If ! IsReadableStream(_readable_) is *false*, throw a *TypeError* exception
+  1. Set _preventClose_ to ! ToBoolean(_preventClose_), set _preventAbort_ to ! ToBoolean(_preventAbort_), and set
+     _preventCancel_ to ! ToBoolean(_preventCancel_).
+  1. If _signal_ is not *undefined*, and _signal_ is not an instance of the `<a idl>AbortSignal</a>` interface, throw a
+     *TypeError* exception.
+  1. If ! IsReadableStreamLocked(*this*) is *true*, throw a *TypeError* exception.
```
domenic: Interesting that these now throw sync exceptions instead of returning a rejected promise that gets ignored. This is probably the right thing to do...

index.bs

index.bs
ricea: The alternative would be to error `readable`. That would have the benefit that the whole pipe chain would be errored in the normal case, and so you'd get a rejection rather than an exception. But it wouldn't work properly if `readable` was locked. It also would be a big departure from our behaviour up until now of passing through `readable` unchanged.

On balance, I think throwing an exception here is the best approach.

ricea:
> When merging this we should be sure to file browser bugs; see https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests

I filed https://bugs.chromium.org/p/chromium/issues/detail?id=908747 for Chrome. Edge and Firefox don't have implementations. It would probably not be helpful to file issues saying "implement pipeThrough". Safari appears to have a semi-functional pipeThrough that delegates to a non-functional pipeTo. I've no idea what to do about that.

domenic:
LGTM, merge at will after the tests. I forgot about the implementation status, so yeah, probably it's OK to not do the bugs...

PR #968, opened 2018-11-24
Fix memory leak in pipe loop reference implementation

MattiasBuelens:
The reference implementation of `ReadableStream.pipeTo` leaks memory when piping a lot of chunks.

## Steps to reproduce

1. Go to the project root
1. Run `npm i heapdump`
1. Create a file called `leak.js` containing the code snippet below
1. Run `node --expose_gc leak.js`
1. Open the two created snapshots (`before.heapsnapshot` and `after.heapsnapshot`) in Chrome DevTools
    1. Click the Memory tab
    1. Right-click in the Profiles sidebar
    1. Select Load
    1. Open the heap snapshot file
1. Select the `after` profile and switch to the Comparison view

<details>
<summary>leak.js</summary>

```js
'use strict';

const { ReadableStream } = require('./reference-implementation/lib/readable-stream.js');
const { WritableStream } = require('./reference-implementation/lib/writable-stream.js');
const heapdump = require('heapdump');

const delay = ms => new Promise(resolve => setTimeout(resolve, ms));

(async () => {
  let controller;
    start(c) {
      controller = c;
    }
  });
  const writable = new WritableStream(); // no-op sink
  const promise = readable.pipeTo(writable);

  gc();
  heapdump.writeSnapshot('./before.heapsnapshot');

  for (let i = 0; i < 1e4; i++) {
    controller.enqueue(i);
  }
  await delay(0);

  gc();
  heapdump.writeSnapshot('./after.heapsnapshot');

  controller.close();
  await promise;
})();

```
</details>

## Expected results
* The difference in total memory usage is small.
* The difference in amount of `Promise` instances (`# Delta`) is zero.

## Actual results
* The difference in total memory usage is ~1 MB.
* The difference in amount of `Promise` instances is 10,000.

diff-leak

## Root cause analysis

The cause of this memory leak is due to [`pipeLoop` creating a long promise chain using recursion](https://github.com/whatwg/streams/blob/46c3b89dd3aff28b2fc381dd1d397c12b4fb8a16/reference-implementation/lib/readable-stream.js#L171):
```js
function pipeLoop() {
  if (shuttingDown === true) {
    return Promise.resolve();
  }

  return writer._readyPromise.then(() => {
      // snipped
  })
  .then(pipeLoop); // <<< recursion!
}
```
Every time a chunk is piped through, the promise chain is *extended*. This chain keeps a ton of promises alive, and cannot be garbage collected until the pipe ends (i.e. when either end closes or errors).

## Proposed solution
This PR fixes the memory leak by creating a single `Promise` for the return value of `pipeLoop`, and manually resolving/rejecting it.

With this fix, the leak is gone:
diff-fixed

## Environment
* Windows 10 1803
* Node 10.13.0

### DIFF
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -310,20 +310,34 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
     // - Backpressure must be enforced
     // - Shutdown must stop all activity
     function pipeLoop() {
+      return new Promise((resolveLoop, rejectLoop) => {
+        function next(done) {
+          if (done) {
+            resolveLoop();
+          } else {
+            pipeStep().then(next, rejectLoop);
+          }
+        }
+
+        next(false);
+      });
+    }
+
+    function pipeStep() {
       if (shuttingDown === true) {
-        return Promise.resolve();
+        return Promise.resolve(true);
       }
 
       return writer._readyPromise.then(() => {
         return ReadableStreamDefaultReaderRead(reader).then(({ value, done }) => {
           if (done === true) {
-            return;
+            return true;
           }
 
           currentWrite = WritableStreamDefaultWriterWrite(writer, value).catch(() => {});
+          return false;
         });
-      })
-          .then(pipeLoop);
+      });
     }
 
     // Errors must be propagated forward


MattiasBuelens:
I know the reference implementation isn't meant to be very optimized, but I still think it shouldn't leak memory like this. (And I don't really want to maintain a fork just to fix this leak for my stream polyfill. 😛 )

Chrome's implementation is not affected. [Its internal `pump`](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/streams/ReadableStream.js?l=225&rcl=c6ab3c2387e3d30bb4dd3a7df6f496fdbcdb84b8) does not use the promises returned by `thenPromise` to create a chain. I don't think any other browser has already implemented `pipeTo`, so they cannot be affected yet.

ricea:
Excellent analysis. Fix looks good to me. I'm going to wait and see if @domenic has any comment.

MattiasBuelens:
Side note: this will definitely have conflicts with #966. You can merge #966 first, I'll rebase this PR afterwards.

PR #969, opened 2018-11-26
Treat forAuthorCode as an internal slot of reader object.

### LINKED ISSUE #964: forAuthorCode can be a bit per reader
Eight different algorithms take a forAuthorCode argument, and it is passed to a lot of places. But if I understand the spec correctly, forAuthorCode is meant to be true iff the stream's current [[reader]] is visible to author code.

So it seems possible to give each reader a [[forAuthorCode]] field instead, and set it to false in AcquireReadableStreamDefaultReader.

This seems more direct, and might even be a nice simplification; if you agree, I'd like to contribute a patch.

jorendorff:
This editorial change doesn't affect any observable behavior.

Closes #964.

***
<a href="https://whatpr.org/streams/969.html" title="Last updated on Nov 26, 2018, 8:34 PM GMT (be7d6b0)">Preview</a> | <a href="https://whatpr.org/streams/969/7f88f1d...be7d6b0.html" title="Last updated on Nov 26, 2018, 8:34 PM GMT (be7d6b0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -625,9 +625,9 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
 
 <emu-alg>
   1. If ! IsReadableStream(*this*) is *false*, throw a *TypeError* exception.
-  1. If _mode_ is *undefined*, return ? AcquireReadableStreamDefaultReader(*this*).
+  1. If _mode_ is *undefined*, return ? Construct(`<a idl>ReadableStreamDefaultReader</a>`, « *this* »).
   1. Set _mode_ to ? ToString(_mode_).
-  1. If _mode_ is `"byob"`, return ? AcquireReadableStreamBYOBReader(*this*).
+  1. If _mode_ is `"byob"`, return ? Construct(`<a idl>ReadableStreamBYOBReader</a>`, « *this* »).
   1. Throw a *RangeError* exception.
 </emu-alg>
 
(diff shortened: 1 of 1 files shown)


jorendorff:
This does seem to reduce the footprint of forAuthorCode in the spec, from being mentioned in 30 places to 8.

Reader object creation could stand to be cleaned up a bit, but I wanted to keep the patch minimal.

I think a nice way to verify that this doesn't affect observable behavior might be to update the reference implementation, but its tests don't pass for me currently, even without any code changes.

MattiasBuelens:
> I think a nice way to verify that this doesn't affect observable behavior might be to update the reference implementation, but its tests don't pass for me currently, even without any code changes.

Huh, that's weird, they should just work. What kind of errors are you getting? Did you run `npm install` and `npm run sync-wpt` inside `./reference-implementation` to update the tests and the test runner?

ricea:
> Huh, that's weird, they should just work. What kind of errors are you getting? Did you run `npm install` and `npm run sync-wpt` inside `./reference-implementation` to update the tests and the test runner?

I had to run `npm install` a couple of times before it correctly picked up the new wpt-runner.

MattiasBuelens:
@jorendorff Are you still interested in updating the reference implementation? Or can I have a go at it? 🙂

jorendorff:
Go for it. Sorry for the delay, I kept thinking I'd get around to it...

ricea:
Superseded by #984.

PR #972, opened 2018-12-05
[Demos] Update TextDecoderStream polyfill

ricea:
Update to final TextDecoderStream by replacing uses of the TextDecoder
constructor with TextDecoderStream. Update to the later version of the
TextDecoderStream polyfill.

### DIFF
--- a/demos/append-child.html
+++ b/demos/append-child.html
@@ -78,7 +78,7 @@ async function fetchThenJSONToDOM() {
 
   const table = createTable(targetDiv);
   return response.body
-    .pipeThrough(new TextDecoder())
+    .pipeThrough(new TextDecoderStream())
     .pipeThrough(splitStream('\n'))
     .pipeThrough(parseJSON())
     .pipeThrough(jsonToElementTransform)
--- a/demos/streaming-element-backpressure.html
+++ b/demos/streaming-element-backpressure.html
@@ -48,7 +48,7 @@ async function fetchDirectlyIntoDOM() {
                              });
 
   await response.body
-      .pipeThrough(new TextDecoder())
+      .pipeThrough(new TextDecoderStream())
       .pipeTo(targetDiv.writable);
 }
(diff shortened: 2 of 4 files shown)


ricea:
@domenic This appears to have broken the live site. It looks like transforms/text-encode-transform.js was updated as expected, but we still have the old versions of the HTML files. I've verified that it's not just a problem with my browser cache.

domenic:
Eek. I think I've found the issue, but am not sure I am able to dive into fixing it while on vacation. Here are the moving parts:

- [Our Makefile](https://github.com/whatwg/streams/blob/a403a222be585e4d03af414b4f4558aa0d9b2e67/Makefile#L9) sets `EXTRA_FILES="demos/**/*"`
- The [deploy.sh script](https://github.com/whatwg/whatwg.org/blob/e4f1fd147f387483fc14cacf215fa7adbc3d5d6d/resources.whatwg.org/build/deploy.sh#L47-L54) runs `rsync --relative $EXTRA_FILES "$1"` to copy the extra files into the various output directories
- Note that we use `rsync --relative` per this commit: https://github.com/whatwg/whatwg.org/commit/aabdd6698ab653c05c71e4e2140687dfd899ab20#diff-e9607901cbff22c83689e88cb1c1eca6
- You can reproduce the problem by running `make deploy` on your machine, which will "deploy" into a local directory.

So I am pretty sure the issue is that `rsync` does not interpret `demos/**/*` in the way we expected. It seems to only include files under subdirectories of demos, not files inside the top level of demos itself.

I think one fix (untested) would be:

- Update deploy.sh to pass `--recursive` to rsync as well.
- Update our Makefile to pass `demos` instead of `demos/**/*` as `EXTRA_FILES`.

However we should be sure to test that this doesn't break other WHATWG repos that use EXTRA_FILES. A representative (perhaps exhaustive?) list of those would be whatwg/encoding and whatwg/console.

PR #973, opened 2018-12-07
Update Makefile to copy top-level demos files

ricea:
Top-level files in the top-level "demos/" directory were not being
copied as they did not match the expression "demos/**/*". Change the
expression to "demos/* demos/**/*" to cause all the files to be copied.

This solution is ugly and causes (harmless) "skipping directory"
warnings to be output by rsync.

It may be better to add the --recursive flag to rsync as a long-term
solution.


ricea:
@domenic This is not great, but it works locally and avoids doing anything that might impact other repos.

PR #974, opened 2018-12-07
Demos: Add lang="en" to html files

ricea:
The Nu Html Checker really wants <html lang="en"> to be included in all
HTML files, and will exit with an error if it is not present. The deploy
script will not proceed if the Html Checker exits with an error.

Add an <html lang="en"> tag to all the demo HTML files to keep the Html
Checker happy.


ricea:
@domenic Looks like I need to land this before deployment will work properly.

PR #977, opened 2018-12-13
Editorial changes to transferable streams explainer

ricea:
* Make some language clearer.
* Add a section about end-user benefit.

### DIFF
--- a/transferable-streams-explainer.md
+++ b/transferable-streams-explainer.md
@@ -43,7 +43,10 @@ Note that `w.postMessage(rs)` would not work. Streams can only be _transferred_,
 stream will throw a DataCloneError.
 
 Once a stream has been transferred with `postMessage()` the original stream is locked and cannot be read or written.
-This is similar to how ArrayBuffers are neutered after they are transferred.
+This is similar to how ArrayBuffers are neutered after they are transferred. However, the code of the underlying source
+or sink is still running is still running in the original context. The benefits to user experience can be seen in [this
+demo of streaming digits of PI](https://glitch.com/edit/#!/streaming-pi?path=pi.js:1:0) (if you have a browser that
+supports transferable streams, you can see it live at https://streaming-pi.glitch.me/).
 
 Transferable streams are also useful in constructing responses for a service worker. See
 https://gist.github.com/domenic/ea5ebedffcee27f552e103963cf8585c/ for an example.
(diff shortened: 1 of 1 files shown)


ricea:
@torgo Sorry for the delay. How does the latest version look?

travisleithead:
I was a little concerned that Transferable Streams would make it pretty easy to route data around the cross-origin restriction, but I don't think this feature makes it much worse than it already is--namely, today an origin A can postMessage the received chunks from a ReadableStream to origin B's message handler. With this feature, only one postMessage is needed to pass the Stream, and then origin B can drink from the pipe for as long as it likes.

I do think the explainer needs to support a better use-case than the general "workers make things easier", or "delegation is good". After all, the workaround for not having this feature is really trival--just make the request for Stream in the Worker/Frame instead (so the transfer is unnecessary). What are the use-cases where the Worker/Frame would be unable to implement that workaround? Why? (The only cases I can imagine are to enable easier cross-origin information sharing on the client.)

MattiasBuelens:
> After all, the workaround for not having this feature is really trival--just make the request for Stream in the Worker/Frame instead (so the transfer is unnecessary). What are the use-cases where the Worker/Frame would be unable to implement that workaround? Why?

Sure, since `fetch` is also available inside a worker, you could just as well have the worker request the data. However, for some use cases, you *have* to touch the main thread to access certain functionality:

* The stream may be coming from an input device only accessible on the main thread. You could use `MediaRecorder` to capture the audio of a user's microphone and/or the video of a user's camera, pipe the captured data through a off-thread `TransformStream` to transcode it into a hypothetical new experimental media format and then upload the resulting stream to a server.
* The stream may need to be displayed on the web page. You could invert the previous example: you download a stream of a video file encoded in an experimental media format, you transcode it to a natively supported format in a worker, and finally you transfer the resulting stream to the main thread to play back the video using `<video>` and `MediaSource`.
* You could have a video editing application as a pure web application, which basically combines the ideas from the two previous examples. You load a local video file or record video from your camera, transcode it in a worker and then store the result to a local file.

(Disclaimer: if I sound _slightly_ biased towards video-related use cases, that's probably because I work for a company in the online video industry. 😛 )

ricea:
> I was a little concerned that Transferable Streams would make it pretty easy to route data around the cross-origin restriction

It's equivalent in power to a MessageChannel, so it doesn't weaken any security guarantees.

> I do think the explainer needs to support a better use-case than the general "workers make things easier", or "delegation is good".

It is currently only an ergonomics win, although I hope to add optimisations in future exploiting the fact that the platform understands the semantics of the transfer. I think it opens up the benefits of offloading work to a larger audience, and that's where the end-user benefit comes from.

I will try to incorporate @MattiasBuelens's examples into the explainer.

transferable-streams-explainer.md
```diff
@@ -66,13 +69,23 @@ https://gist.github.com/domenic/ea5ebedffcee27f552e103963cf8585c/ for an example
 ## Use cases
 
 *   Performing expensive transformations off the main thread. Transcoding, for example.
-*   Synthesizing responses from a service worker.
+*   Synthesizing responses from a service worker. For example, generating a PDF from data in the DOM and streaming it to
+    the service worker where it can then be downloaded as a file.
+*   Processing a stream of data from an input device only accessible on the main thread. For example, uou could use
```
MattiasBuelens: uou -> you

transferable-streams-explainer.md
```diff
@@ -43,7 +43,10 @@ Note that `w.postMessage(rs)` would not work. Streams can only be _transferred_,
 stream will throw a DataCloneError.
 
 Once a stream has been transferred with `postMessage()` the original stream is locked and cannot be read or written.
-This is similar to how ArrayBuffers are neutered after they are transferred.
+This is similar to how ArrayBuffers are neutered after they are transferred. However, the code of the underlying source
+or sink is still running is still running in the original context. The benefits to user experience can be seen in [this
```
MattiasBuelens: "is still running" appears twice.

ricea: > Spotted a few typos. Looks good though!

Thanks for proofreading!

PR #984, opened 2019-02-09
Treat forAuthorCode as an internal slot of reader object

### LINKED ISSUE #964: forAuthorCode can be a bit per reader

MattiasBuelens:
This continues @jorendorff's initial work in #969.

Closes #964.

***
<a href="https://whatpr.org/streams/984.html" title="Last updated on Feb 9, 2019, 6:46 PM UTC (84d4af6)">Preview</a> | <a href="https://whatpr.org/streams/984/bf2cac8...84d4af6.html" title="Last updated on Feb 9, 2019, 6:46 PM UTC (84d4af6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -617,7 +617,7 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
 
 <emu-alg>
   1. If ! IsReadableStream(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _reader_ be ? AcquireReadableStreamDefaultReader(*this*).
+  1. Let _reader_ be ? Construct(`<a idl>ReadableStreamDefaultReader</a>`, « *this* »).
   1. Let _iterator_ be ! ObjectCreate(`<a idl>ReadableStreamAsyncIteratorPrototype</a>`).
   1. Set _iterator_.[[asyncIteratorReader]] to _reader_.
   1. Set _iterator_.[[preventCancel]] to ! ToBoolean(_preventCancel_).
@@ -870,7 +870,7 @@ prototype, are created with the internal slots described in the following table:
     1. Let _done_ be ? Get(_result_, `"done"`).
     1. Assert: Type(_done_) is Boolean.
     1. If _done_ is *true*, perform ! ReadableStreamReaderGenericRelease(_reader_).
-    1. Return ! ReadableStreamCreateReadResult(_value_, _done_, *true*).
+    1. Return ! ReadableStreamCreateReadResult(_value_, _done_, _reader_).
 </emu-alg>
 
 <h4 id="rs-asynciterator-prototype-return" method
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -648,9 +648,9 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
 
 <emu-alg>
   1. If ! IsReadableStream(*this*) is *false*, throw a *TypeError* exception.
-  1. If _mode_ is *undefined*, return ? AcquireReadableStreamDefaultReader(*this*).
+  1. If _mode_ is *undefined*, return ? Construct(`<a idl>ReadableStreamDefaultReader</a>`, « *this* »).
```
MattiasBuelens: I'm not really a fan of this. I think I'll change it back to use the abstract operation, and add an optional `forAuthorCode` argument to that operation.

index.bs
```diff
@@ -617,7 +617,7 @@ option. If <code><a for="underlying source">type</a></code> is set to <code>unde
 
 <emu-alg>
   1. If ! IsReadableStream(*this*) is *false*, throw a *TypeError* exception.
-  1. Let _reader_ be ? AcquireReadableStreamDefaultReader(*this*).
+  1. Let _reader_ be ? Construct(`<a idl>ReadableStreamDefaultReader</a>`, « *this* »).
```
MattiasBuelens: This is a bit of a problem...
* If we create a reader with `forAuthorCode = false`, then `ReadableStreamCreateReadResult` inside `it.next()` will create an object with `null` prototype, which breaks the contract.
* If we create a reader with `forAuthorCode = false`, then `ReadableStreamDefaultReaderRead` inside `it.next()` will also access `Object.prototype.then` for the internal read result.

I think I'll have to rework this a bit. `ReadableStreamCreateReadResult` should take a `forAuthorCode` argument, rather than a `reader`. In most cases, we'll just pass `reader.[[forAuthorCode]]` as the argument, but in the case of `it.next()` we should explicitly pass `true`, even though `it.[[asyncIteratorReader]].[[forAuthorCode]]` will be `false`.

MattiasBuelens:
Reworked. It's much simpler now! 😄

ricea:
@jorendorff, how would you like to be credited in the "Acknowledgements" section? Is "Jason Orendorff" okay?

jorendorff:
> @jorendorff, how would you like to be credited in the "Acknowledgements" section? Is "Jason Orendorff" okay?

Yes, that's fine! Sorry for the slow response.

ricea:
> Yes, that's fine! Sorry for the slow response.

Good because I already added you :smile:.

PR #990, opened 2019-02-18
Initialize ReadableByteStreamController.[[byobRequest]] to undefined

MattiasBuelens:
`SetUpReadableByteStreamController` calls `ReadableByteStreamControllerClearPendingPullIntos`, which in turn calls `ReadableByteStreamControllerInvalidateBYOBRequest`. This last function returns early if `controller.[[byobRequest]]` is already `undefined`, which holds **even if we don't explicitly set it** to `undefined`. This means the internal slot is never really "initialized" to `undefined`, so if I understand correctly looking up the slot's value would mean going through the prototype chain to find out that the slot does not exist. (At least, until the first BYOB request comes in and gives it an actual value.)

This change explicitly initializes `ReadableByteStreamController.[[byobRequest]]` to `undefined`. We already do this for other internal slots (such as `ReadableStream.[[storedError]]` or `WritableStream.[[closeRequest]]`), so it seemed appropriate to also do it for `[[byobRequest]]`.

***
<a href="https://whatpr.org/streams/990.html" title="Last updated on Feb 18, 2019, 11:25 PM UTC (0560f36)">Preview</a> | <a href="https://whatpr.org/streams/990/ddfba5f...0560f36.html" title="Last updated on Feb 18, 2019, 11:25 PM UTC (0560f36)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2247,7 +2247,8 @@ Instances of {{ReadableByteStreamController}} are created with the internal slot
   </tr>
   <tr>
     <td>\[[byobRequest]]
-    <td class="non-normative">A {{ReadableStreamBYOBRequest}} instance representing the current BYOB pull request
+    <td class="non-normative">A {{ReadableStreamBYOBRequest}} instance representing the current BYOB pull request,
+      or <emu-val>undefined</emu-val> if there are no pending requests
   </tr>
   <tr>
     <td>\[[cancelAlgorithm]]
@@ -2965,6 +2966,7 @@ throws>SetUpReadableByteStreamController ( <var>stream</var>, <var>controller</v
     1. Assert: _autoAllocateChunkSize_ is positive.
   1. Set _controller_.[[controlledReadableByteStream]] to _stream_.
   1. Set _controller_.[[pullAgain]] and _controller_.[[pulling]] to *false*.
+  1. Set _controller_.[[byobRequest]] to *undefined*.
   1. Perform ! ReadableByteStreamControllerClearPendingPullIntos(_controller_).
   1. Perform ! ResetQueue(_controller_).
   1. Set _controller_.[[closeRequested]] and _controller_.[[started]] to *false*.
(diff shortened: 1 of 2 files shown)


ricea:
I couldn't find where the behaviour of internal slots is specified, but I don't think they follow the prototype chain. I think reading from an uninitialised slot is implementation-specific behaviour. An implementation could crash if it wanted to.

If you find any other cases like this, fixes or bug reports are welcome!

MattiasBuelens:
> I couldn't find where the behaviour of internal slots is specified, but I don't think they follow the prototype chain. I think reading from an uninitialised slot is implementation-specific behaviour. An implementation could crash if it wanted to.

I found [the following paragraph in the JavaScript spec](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-object-internal-methods-and-internal-slots):
> Internal slots correspond to internal state that is associated with objects and used by various ECMAScript specification algorithms. Internal slots are not object properties and they are not inherited. Depending upon the specific internal slot specification, such state may consist of values of any ECMAScript language type or of specific ECMAScript specification type values. Unless explicitly specified otherwise, internal slots are allocated as part of the process of creating an object and may not be dynamically added to an object. Unless specified otherwise, the initial value of an internal slot is the value **undefined**. Various algorithms within this specification create objects that have internal slots. However, the ECMAScript language provides no direct way to associate internal slots with an object.

Indeed, they do not follow the prototype chain, they must exist on the object itself. And they are implicitly initialized to `undefined`, so the explicit initialization is *technically speaking* not needed.

Still, I prefer being explicit, even if it only makes a difference for the reference implementation. 🙂

ricea:
> I found [the following paragraph in the JavaScript spec](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-object-internal-methods-and-internal-slots):

Thanks! That was very helpful.

PR #994, opened 2019-02-21
In ReadableStreamTee, replace _closedOrErrored_ with _closed_

### LINKED ISSUE #991: Semantics of |closedOrErrored| variable in ReadableStreamTee are odd
https://streams.spec.whatwg.org/commit-snapshots/bf2cac8a52664df3e6da7a48755890e87b00953a/#ref-for-upon-rejection

> Upon rejection of reader\.\[\[closedPromise\]\] with reason r,
> If closedOrErrored is false, then:​
> Perform \! ReadableStreamDefaultControllerError\(branch1\.\[\[readableStreamController\]\], r\)\.
> Perform \! ReadableStreamDefaultControllerError\(branch2\.\[\[readableStreamController\]\], r\)\.
> Set closedOrErrored to true\.

|closedOrErrored| can't actually be true when this is called, because it is only set to true when the stream is closed or errored, but if the stream is closed then it won't be errored.

Similarly, when |closedOrErrored| is set because |done| was true, we know that it can't have been set due to an error, because the call to DefaultReaderRead would have rejected in that case.

My suggestion is to rename |closedOrErrored| to just |closed| and only look at or modify it for the closed case. But it may be possible to eliminate the variable altogether.

ricea:
DefaultReaderRead will never resolve after the [[closedPromise]] has rejected,
therefore setting _closedOrErrored_ on rejection makes no difference. Rename the
variable to just _closed_ and only use it to signal that a read has returned
with "done" set to true.

Also change the lookups of "done" and "value" in ReadableStreamTee from "? Get"
to "! Get" as they can never throw, and pullAlgorithm returning an abrupt
completion is not supported anyway.

Rearrange the steps slightly to reduce redundant work.

Closes #991.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1047,7 +1047,7 @@ noticeable asymmetry between the two branches, and limits the possible <a>chunks
   1. Assert: ! IsReadableStream(_stream_) is *true*.
   1. Assert: Type(_cloneForBranch2_) is Boolean.
   1. Let _reader_ be ? AcquireReadableStreamDefaultReader(_stream_).
-  1. Let _closedOrErrored_ be *false*.
+  1. Let _closed_ be *false*.
   1. Let _canceled1_ be *false*.
   1. Let _canceled2_ be *false*.
   1. Let _reason1_ be *undefined*.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1058,17 +1058,18 @@ noticeable asymmetry between the two branches, and limits the possible <a>chunks
   1. Let _pullAlgorithm_ be the following steps:
     1. Return the result of <a>transforming</a> ! ReadableStreamDefaultReaderRead(_reader_) with a fulfillment handler
        which takes the argument _result_ and performs the following steps:
+      1. If _closed_ is *true*, return.
       1. Assert: Type(_result_) is Object.
-      1. Let _value_ be ? Get(_result_, `"value"`).
-      1. Let _done_ be ? Get(_result_, `"done"`).
+      1. Let _done_ be ! Get(_result_, `"done"`).
```
MattiasBuelens: Interesting! If I understand correctly, this works because we know `result` has a `null` prototype, so we always go through [`Get`](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-get-o-p) -> [`[[Get]]`](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver) -> [`OrdinaryGet`](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-ordinaryget) -> [`[[GetOwnProperty]]`](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-ordinary-object-internal-methods-and-internal-slots-getownproperty-p) -> [`OrdinaryGetOwnProperty`](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-ordinarygetownproperty), which cannot throw.

For the same reason, I think we can also change `? Get` to `! Get` in [`ReadableStreamAsyncIteratorPrototype.next()`](https://streams.spec.whatwg.org/commit-snapshots/a441f59db468c3a40515cc2e7590abe381632f8d/#rs-asynciterator-prototype-next)?

ricea: > For the same reason, I think we can also change `? Get` to `! Get` in [`ReadableStreamAsyncIteratorPrototype.next()`](https://streams.spec.whatwg.org/commit-snapshots/a441f59db468c3a40515cc2e7590abe381632f8d/#rs-asynciterator-prototype-next)?

Yes. Feel free to send a PR! :smile:

MattiasBuelens: ➡️ #995

PR #997, opened 2019-02-23
ReadableStreamTee: do not read when already pulling

### LINKED ISSUE #996: ReadableStreamTee pulls too many chunks
While reviewing #991, [I stumbled on a bug in `tee()`](https://github.com/whatwg/streams/issues/991#issuecomment-465814861).

For a single readable stream, we guarantee that we won't call `pull()` until the previous pull promise has resolved. But in the case of `tee()`, we use *the same pull algorithm* for both branches, so there's nothing stopping one branch from calling that function while the promise from a previous call by the other branch is still pending. Since both branches are constructed with a default high water mark of 1, they'll start out calling `pull()`, so we will start **2 read operations instead of just 1**.

## Reproduction code
```js
let remainingChunks = ['a', 'b', 'c'];
  pull(c) {
    console.log('enqueue', remainingChunks[0]);
    c.enqueue(remainingChunks.shift());
    if (remainingChunks.length === 0) {
      console.log('close');
      c.close();
    }
  }
}, {highWaterMark: 0});
```
## Expected output
```
enqueue a
```
Both branches have HWM = 1, so `tee()` should read one chunk and enqueue it on both branches.

## Actual output
```
enqueue a
enqueue b
```
Both branches call `pull()` at the same time, so we read two chunks and enqueue them on both branches. This pushes them (unnecessarily) over their HWM = 1.

MattiasBuelens:
Currently, when one branch calls `pull()`, it is possible that we call `read()` - even though a `read()` is still pending from a previous `pull()` call by the other branch. This can lead to reading too many chunks from the original streams, and unnecessarily filling up the queues of the branches past their high water mark.

This PR fixes that issue by *only* starting a new `read()` if the previous `read()` was resolved. As a bonus, there will now be only one `read()` that resolves with `{ done: true }` when the original stream closes, so we no longer need to keep track of whether we already closed the two branches.

WPT PR: web-platform-tests/wpt#16167

Fixes #996

***
<a href="https://whatpr.org/streams/997.html" title="Last updated on Apr 2, 2019, 5:36 AM UTC (71ccbcd)">Preview</a> | <a href="https://whatpr.org/streams/997/2c8f35e...71ccbcd.html" title="Last updated on Apr 2, 2019, 5:36 AM UTC (71ccbcd)">Diff</a>

### DIFF
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -560,7 +560,7 @@ function ReadableStreamTee(stream, cloneForBranch2) {
 
     pulling = true;
 
-    return ReadableStreamDefaultReaderRead(reader).then(result => {
+    const pullPromise = ReadableStreamDefaultReaderRead(reader).then(result => {
       pulling = false;
 
       assert(typeIsObject(result));
@@ -595,6 +595,13 @@ function ReadableStreamTee(stream, cloneForBranch2) {
         ReadableStreamDefaultControllerEnqueue(branch2._readableStreamController, value2);
       }
     });
+
+    // If pullPromise rejects with an AssertionError, but the branch is already closed or errored,
+    // the rejection will be silently ignored by ReadableStreamDefaultControllerError.
+    // We have to manually rethrow any AssertionError to make sure they are not ignored.
+    pullPromise.catch(rethrowAssertionErrorRejection);
+
+    return pullPromise;
   }
 
   function cancel1Algorithm(reason) {
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -540,6 +540,7 @@ function ReadableStreamTee(stream, cloneForBranch2) {
 
   const reader = AcquireReadableStreamDefaultReader(stream);
 
+  let pulling = false;
   let closed = false;
   let canceled1 = false;
   let canceled2 = false;
@@ -554,7 +555,15 @@ function ReadableStreamTee(stream, cloneForBranch2) {
   });
 
   function pullAlgorithm() {
+    if (pulling === true) {
+      return Promise.resolve();
+    }
+
+    pulling = true;
+
     return ReadableStreamDefaultReaderRead(reader).then(result => {
+      pulling = false;
+
       if (closed === true) {
         return;
       }
--- a/reference-implementation/lib/readable-stream.js
+++ b/reference-implementation/lib/readable-stream.js
@@ -560,7 +560,7 @@ function ReadableStreamTee(stream, cloneForBranch2) {
 
     reading = true;
 
-    const pullPromise = ReadableStreamDefaultReaderRead(reader).then(result => {
+    const readPromise = ReadableStreamDefaultReaderRead(reader).then(result => {
       reading = false;
 
       assert(typeIsObject(result));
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/readable-stream.js
```diff
@@ -592,6 +595,13 @@ function ReadableStreamTee(stream, cloneForBranch2) {
         ReadableStreamDefaultControllerEnqueue(branch2._readableStreamController, value2);
       }
     });
+
+    // If pullPromise rejects with an AssertionError, but the branch is already closed or errored,
+    // the rejection will be silently ignored by ReadableStreamDefaultControllerError.
+    // We have to manually rethrow any AssertionError to make sure they are not ignored.
+    pullPromise.catch(rethrowAssertionErrorRejection);
```
MattiasBuelens: Concretely: if the implementation has a bug that causes an assert failure in `ReadableStreamDefaultControllerClose` or `ReadableStreamDefaultControllerEnqueue`, this failure would be ignored if the branch was previously cancelled. This line ensures that an assert failure *always* causes an unhandled exception.

reference-implementation/lib/readable-stream.js
```diff
@@ -592,6 +595,13 @@ function ReadableStreamTee(stream, cloneForBranch2) {
...
+
+    return pullPromise;
```
MattiasBuelens: Technically speaking, we could also just return `Promise.resolve()` here. We already have a `catch` handler on `reader._closedPromise` to error the branches when the original stream errors, so we don't really *need* to return any rejections from `pull()`. Then it would look more like:

```js
function pullAlgorithm() {
  if (pulling === false) {
    pulling = true;
    ReadableStreamDefaultReaderRead(reader).then(result => {
      pulling = false;
      // ...
    }).catch(rethrowAssertionErrorRejection);
  }
  return Promise.resolve();
}
```
Thoughts?

reference-implementation/lib/readable-stream.js
```diff
@@ -554,10 +554,14 @@ function ReadableStreamTee(stream, cloneForBranch2) {
   });
 
   function pullAlgorithm() {
-    return ReadableStreamDefaultReaderRead(reader).then(result => {
-      if (closed === true) {
-        return;
-      }
+    if (pulling === true) {
```
ricea: You could use `ReadableStreamGetNumReadRequests(stream) > 0` instead of having a separate `pulling` variable to keep track of. However, using a `pulling` variable makes it more obviously correct, so on balance I prefer it.

Nit: I think `reading` is a better name than `pulling`, since it really reflects the duration of the call to ReadableStreamDefaultReaderRead.

reference-implementation/lib/readable-stream.js
ricea: Good point, I missed this in my implementation.

reference-implementation/lib/readable-stream.js
ricea: I like the idea of always returning Promise.resolve() as it makes the behaviour more symmetrical.

I don't think we can wrap the whole function in another if(), as it will increase the number of levels of indentation, which is bad for the readability of the standard text. We still have to do an early exit.

Of course, we still need abundant tests of error conditions to be confident it works.

ricea:
I think we need a bunch of tests like

* Errored with no pulls
* Errored with one pull
  * branch1
  * branch2
* Errored with both branches pulling (do we need variants for which branch started the `read`? Hopefully not).

reference-implementation/lib/readable-stream.js
MattiasBuelens: Okay, so then it'd become:
```js
function pullAlgorithm() {
  if (pulling === true) {
    return Promise.resolve();
  }
  pulling = true;
  ReadableStreamDefaultReaderRead(reader).then(result => {
    pulling = false;
    // ...
  }).catch(rethrowAssertionErrorRejection);
  return Promise.resolve();
}
```
I guess it's not too bad? 😛

reference-implementation/lib/readable-stream.js
MattiasBuelens: Fine by me, I'll rename it to `reading`.

MattiasBuelens:
WPT PR is up: web-platform-tests/wpt#16167. Sorry for the wait! 😅

ricea:
Thanks! I kept meaning to work on the tests, but other things got in the way.

PR #999, opened 2019-03-23
Make it clear that a chunk size has to be non-negative

surma:
***
<a href="https://whatpr.org/streams/999.html" title="Last updated on Mar 24, 2019, 12:11 PM UTC (493043c)">Preview</a> | <a href="https://whatpr.org/streams/999/a7f6210...493043c.html" title="Last updated on Mar 24, 2019, 12:11 PM UTC (493043c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5051,7 +5051,7 @@ properties:
 <dl>
   <dt><dfn method for="queuing strategy">size(<var>chunk</var>)</dfn> (non-byte streams only)</dt>
   <dd>
-    <p>A function that computes and returns the size of the given <a>chunk</a> value.</p>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5051,7 +5051,7 @@ properties:
 <dl>
   <dt><dfn method for="queuing strategy">size(<var>chunk</var>)</dfn> (non-byte streams only)</dt>
   <dd>
-    <p>A function that computes and returns the size of the given <a>chunk</a> value.</p>
+    <p>A function that computes and returns the non-negative size of the given <a>chunk</a> value.</p>
```
MattiasBuelens: If we want to be **really** precise, we could exclude `+Infinity` as well:
> returns the finite non-negative size

Issue #1004, opened 2019-07-03
Reconnecting a Stream after creating it - possible?


alvestrand:
Domenic said this might be a good place to ask...
I'm designing an API where I intend to use WHATWG Streams.
I have to make a choice between a model where I show the user a bunch of readable/writable/transferStreams and tell the user "connect them all", and a model where I present a connected network of streams, and tell the user "make the changes you want".

I experimented a little, and found that the code below didn't work - but I could find no obvious API to do what I wanted; once you have connected a stream, there seems to be no convenient API for reconnecting them, short of messing around with the readers and writers explicitly.

Is there something obvious I'm missing?

Thanks in advance!
```

var dest1 = new WritableStream({
  write(chunk) {
    console.log('Dest 1 got ' + chunk);
  }
});

var dest2 = new WritableStream({
  write(chunk) {
    console.log('Dest 2 got ' + chunk);
  }
});

var src = new ReadableStream({
  start(controller) {
    const producer = () => {
      controller.enqueue('chunk');
      setTimeout(producer, 1000);
    };
    setTimeout(producer, 1000);
  }
});

src.pipeTo(dest1);

const pipeTo1 = () => {
  console.log('Piping to 1');
  src.pipeTo(dest1);
}

const pipeTo2 = async () => {
  console.log('Piping to 2');
  // Here there's something missing
  src.pipeTo(dest2);
}
```

MattiasBuelens:
You can pass an `AbortSignal` to `pipeTo()` to stop an ongoing pipe. You can use `preventAbort` and `preventCancel` to prevent the aborted pipe from also cancelling the input and/or aborting the output.

Note that the pipe still waits for any pending writes to complete, and both ends remain locked until everything has been written. That means you still have to wait for the `pipeTo()` promise to settle (either resolve or reject) before you can start a new pipe.

```javascript
let currentPipePromise = Promise.resolve();
let currentPipeAbortController = new AbortController();

async function startNewPipe(dest) {
  // Abort the previous pipe
  currentPipeAbortController.abort();
  currentPipeAbortController = new AbortController();
  // Wait for previous pipe to complete
  await currentPipePromise.catch(() => {});
  // Start new pipe, but make sure we can abort it later
  currentPipePromise = src.pipeTo(dest, {
    preventAbort: true,
    preventCancel: true,
    signal: currentPipeAbortController.signal
  });
}

startNewPipe(dest1);

const pipeTo1 = () => {
  console.log('Piping to 1');
[...]

PR #1006, opened 2019-07-12
Fix pipe not aborting when both preventAbort and preventCancel are set

MattiasBuelens:
When both flags are set, the _abortAlgorithm_ passes an empty list of actions to ["wait for all"](https://www.w3.org/2001/tag/doc/promises-guide/#waiting-for-all). This helper should immediately resolve the returned promise when no promises are given, but instead the promise remained pending forever.

This fixes the reference implementation of the "wait for all" helper to resolve correctly when given an empty list of promises. This matches the spec change in w3ctag/promises-guide#58.

Originally reported in [this comment](https://github.com/whatwg/streams/issues/1004#issuecomment-511037767) on #1004.

WPT: web-platform-tests/wpt#17816

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -175,6 +175,10 @@ exports.WaitForAll = (promises, successSteps, failureSteps) => {
   let fulfilledCount = 0;
   const total = promises.length;
   const result = new Array(total);
+  if (total === 0) {
+    queueMicrotask(() => successSteps(result));
+    return;
+  }
   for (const promise of promises) {
     const promiseIndex = index;
     const fulfillmentHandler = arg => {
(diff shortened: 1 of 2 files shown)


ricea:
See my comment at https://github.com/w3ctag/promises-guide/pull/58#issuecomment-511735947 about whether or not successSteps is called synchronously. We should probably make this match whatever resolution we choose there.

domenic:
We can use https://nodejs.org/dist/latest-v12.x/docs/api/globals.html#globals_queuemicrotask_callback to implement the promises guide revision.

MattiasBuelens:
Looks like jsdom doesn't expose `queueMicrotask` on its `window` yet... I guess we could expose it ourselves in [`run-web-platform-tests.js`](https://github.com/whatwg/streams/blob/6f94580f6731d1e017c516af097d47c45aad1f56/reference-implementation/run-web-platform-tests.js#L39)?
```javascript
  const failures = await wptRunner(testsPath, {
    rootURL: 'streams/',
    setup(window) {
      window.queueMicrotask = queueMicrotask;
      window.eval(bundledJS);
    },
    // ...
  });
```

domenic:
Bah, yeah, that's probably simplest for now.

PR #1010, opened 2019-08-19
Add helpers to use initial values of Promise methods

MattiasBuelens:
Currently, the reference implementation does not use the initial value of the `Promise` constructor and its methods to implement phrases such as ["upon fulfillment of _p_ with value _v_"](https://www.w3.org/2001/tag/doc/promises-guide/#upon-fulfillment), although this is required by the promise guide. This causes the reference implementation to fail on the new tests introduced in web-platform-tests/wpt#18452.

This PR adds helper functions to call the original versions of the promise constructor and its methods. Assuming they haven't been monkey-patched before the reference implementation was loaded, these should be the initial values expected by the spec.

It also replaces all direct calls to the promise constructos and its methods with calls to the respective helper functions. This is a purely mechanical change, but it does make for quite a big diff! 😅

With this change, the reference implementation once again passes all tests. 🎉

See also:
* [Original report](https://github.com/web-platform-tests/wpt/pull/18452#issuecomment-522364285)

### DIFF
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -200,12 +200,17 @@ function uponRejection(promise, onRejected) {
   uponPromise(promise, undefined, onRejected);
 }
 
+function setPromiseIsHandledToTrue(promise) {
+  PerformPromiseThen(promise, undefined, rethrowAssertionErrorRejection);
+}
+
 exports.newPromise = newPromise;
 exports.promiseResolvedWith = promiseResolvedWith;
 exports.promiseRejectedWith = promiseRejectedWith;
 exports.uponPromise = uponPromise;
 exports.uponFulfillment = uponFulfillment;
 exports.uponRejection = uponRejection;
+exports.setPromiseIsHandledToTrue = setPromiseIsHandledToTrue;
 
 exports.PerformPromiseThen = PerformPromiseThen;
 exports.PerformPromiseCatch = PerformPromiseCatch;
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -110,7 +110,7 @@ function PromiseCall(F, V, args) {
   try {
     return PromiseResolve(Call(F, V, args));
   } catch (value) {
-    return Promise.reject(value);
+    return PromiseReject(value);
   }
 }
 
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -1,5 +1,6 @@
 'use strict';
 const assert = require('assert');
+const { rethrowAssertionErrorRejection } = require('./utils.js');
 
 const isFakeDetached = Symbol('is "detached" for our purposes');
 
@@ -184,9 +185,28 @@ function PerformPromiseCatch(promise, onRejected) {
   return PerformPromiseThen(promise, undefined, onRejected);
 }
 
+function uponPromise(promise, onFulfilled, onRejected) {
+  PerformPromiseCatch(
+    PerformPromiseThen(promise, onFulfilled, onRejected),
+    rethrowAssertionErrorRejection
+  );
+}
+
+function uponFulfillment(promise, onFulfilled) {
+  uponPromise(promise, onFulfilled);
+}
+
+function uponRejection(promise, onRejected) {
+  uponPromise(promise, undefined, onRejected);
+}
+
 exports.newPromise = newPromise;
 exports.promiseResolvedWith = promiseResolvedWith;
 exports.promiseRejectedWith = promiseRejectedWith;
+exports.uponPromise = uponPromise;
+exports.uponFulfillment = uponFulfillment;
+exports.uponRejection = uponRejection;
+
 exports.PerformPromiseThen = PerformPromiseThen;
 exports.PerformPromiseCatch = PerformPromiseCatch;
 
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -88,7 +88,7 @@ exports.CreateAlgorithmFromUnderlyingMethod = (underlyingObject, methodName, alg
       }
     }
   }
-  return () => promiseResolvedWith();
+  return () => promiseResolvedWith(undefined);
 };
 
 exports.InvokeOrNoop = (O, P, args) => {
--- a/reference-implementation/lib/helpers.js
+++ b/reference-implementation/lib/helpers.js
@@ -218,7 +218,7 @@ exports.WaitForAll = (promises, successSteps, failureSteps) => {
 exports.WaitForAllPromise = (promises, successSteps, failureSteps = undefined) => {
   let resolvePromise;
   let rejectPromise;
-  const promise = new Promise((resolve, reject) => {
+  const promise = CreatePromise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


reference-implementation/lib/helpers.js
```diff
@@ -157,11 +157,38 @@ exports.MakeSizeAlgorithmFromSizeFunction = size => {
...
+
+function CreatePromise(executor) {
+  return new originalPromise(executor);
+}
+
+function PromiseResolve(value) {
+  return originalPromiseResolve.call(originalPromise, value);
+}
+
+function PromiseReject(reason) {
```
MattiasBuelens: I wasn't sure what to name these functions, since there's no such thing as `PromiseReject` or `PerformPromiseCatch` in [the ECMAScript specification](https://tc39.es/ecma262/). I considered naming them after the phrases in the promise guide, but I found `PromiseRejectedWith(reason)` and `UponRejection(promise, onRejected)` a bit weird.

Feel free to suggest better names. 😉

reference-implementation/lib/transform-stream.js
```diff
@@ -389,18 +389,18 @@ function TransformStreamDefaultSinkCloseAlgorithm(stream) {
   TransformStreamDefaultControllerClearAlgorithms(controller);
 
   // Return a promise that is fulfilled with undefined on success.
-  return flushPromise.then(() => {
+  return PerformPromiseCatch(PerformPromiseThen(flushPromise, () => {
```
ricea: Looks like you fixed a bug here, and now the `PerformPromiseCatch` is no longer needed.

reference-implementation/lib/transform-stream.js
```diff
@@ -325,7 +325,7 @@ function TransformStreamDefaultControllerError(controller, e) {
 
 function TransformStreamDefaultControllerPerformTransform(controller, chunk) {
   const transformPromise = controller._transformAlgorithm(chunk);
-  return transformPromise.catch(r => {
+  return PerformPromiseCatch(transformPromise, r => {
```
ricea: We can use PerformPromiseThen() here and it will be closer to [the description in the promises guide](https://www.w3.org/2001/tag/doc/promises-guide/#transforming-by).

reference-implementation/lib/helpers.js
ricea: Generally where we have a function that's not an abstract operation we start it with a lowercase letter, but I'm not 100% sure whether that's the right thing to do with these.

`PerformPromiseCatch` is probably too misleading, as it makes it sound like it is a standard operation. I think we can split it into two cases: `setPromiseIsHandled()` and `rethrowAssertions()`.

I think I prefer to use the names from the promises guide, so we have `aNewPromise()`, `aPromiseRejectedWith()`, and `aPromiseResolvedWith()`. Yes, these names are ugly but they help the reference implementation match the standard text. We should probably explicitly write `aPromiseResolvedWith(undefined)` to get it to read smoothly.

Since this contradicts what you want to do, I'd like to ask @domenic to be the stylistic judge here.

domenic: I agree with sticking with the promises guide names, and using lower case. After all, this is the reference implementation; the closer to the standard text it can be, the better.

MattiasBuelens: Sounds reasonable. I'll rename them. 🙂

reference-implementation/lib/transform-stream.js
MattiasBuelens: Oh woops, that was actually a typo. The extra `)` on line 403 should have been on line 400 instead.

But you're right, the correct implementation would be to use a single `PerformPromiseThen` call with both handlers. So I *accidentally* fixed a bug. 😅

reference-implementation/lib/transform-stream.js
MattiasBuelens: You mean like this?
```javascript
return PerformPromiseThen(transformPromise, undefined, r => { /* ... */ });
```
I think it's a bit weird to not have a fulfillment handler. But yes, it would be more accurate. I'll change it.

ricea: Yes. I'm so used to doing this that it doesn't look weird to me any more :smile:.

MattiasBuelens:
This became a bit more involved than I first anticipated.

I've added the following helpers:
* `uponFulfillment`: implements [_upon fulfillment of p with value v_](https://www.w3.org/2001/tag/doc/promises-guide/#upon-fulfillment)
* `uponRejection`: implements [_upon rejection of p with reason r_](https://www.w3.org/2001/tag/doc/promises-guide/#upon-rejection)
* `uponPromise`: combines `uponFulfillment` and `uponRejection` in a single call. This keeps the code simple, since the spec often uses _upon fulfillment_ and _upon rejection_ together in subsequent steps. (I'm not 100% sure about the function name. I guess `uponFulfillmentAndRejection` would be more accurate, but I find it a bit too verbose.)
* `transformPromiseWith`: implements [_transforming p with a fulfillment and/or rejection handler_](https://www.w3.org/2001/tag/doc/promises-guide/#transforming-by) (Again, suggestions for a better name are welcome. 😛)
* `setPromiseIsHandledToTrue`: sets `[[PromiseIsHandled]]` to true

Note that the `upon*` helpers do *not* return the created promise. Only `transformPromiseWith` returns a promise, as required by the "transforming a promise" definition.

The cool bit: these helpers also take care of re-throwing assertion errors where needed. Previously, we had a few places where we accidentally used `.catch(() => {})` instead of `.catch(rethrowAssertionErrorRejection)`. Since the helpers now always attach this rejection handler to the created promise, we no longer have to worry about it in the rest of the reference implementation. 😄

reference-implementation/lib/readable-stream.js
```diff
@@ -392,12 +398,12 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
...
     // - Shutdown must stop all activity
     function pipeLoop() {
-      return new Promise((resolveLoop, rejectLoop) => {
+      return newPromise((resolveLoop, rejectLoop) => {
         function next(done) {
           if (done) {
             resolveLoop();
           } else {
-            pipeStep().then(next, rejectLoop);
+            uponPromise(pipeStep(), next, rejectLoop);
```
MattiasBuelens: I'm not sure if `uponPromise()` makes this code more readable. I feel like `PerformPromiseThen(pipeStep(), next, rejectLoop)` is more obvious, but then we have to export the low-level `then()` helper again. 😕

This also adds an extra `.catch()` handler on every pipe step, since this is now equivalent to:
```javascript
pipeStep().then(next, rejectLoop).catch(rethrowAssertionErrorRejection);
```
The `.catch()` is not necessary, since we already handle that [here](https://github.com/whatwg/streams/blob/1146228f0bcc1077f6a0c0eff2452b89332ee2ba/reference-implementation/lib/readable-stream.js#L469):
```javascript
setPromiseIsHandledToTrue(pipeLoop());
```
which is equivalent to:
```javascript
pipeLoop().catch(rethrowAssertionErrorRejection);
```
Although you wouldn't want to create this extra promise in a real implementation, I think it's fine for a reference implementation.

ricea: I think this use of `uponPromise()` is fine.

PR #1011, opened 2019-08-24
Add WritableStream.close()

MattiasBuelens:
Fixes #1007

WPT PR: web-platform-tests/wpt#18693

***
<a href="https://whatpr.org/streams/1011.html" title="Last updated on Nov 3, 2019, 10:18 PM UTC (bfbf1df)">Preview</a> | <a href="https://whatpr.org/streams/1011/d26db0e...bfbf1df.html" title="Last updated on Nov 3, 2019, 10:18 PM UTC (bfbf1df)">Diff</a>

Issue #963, opened 2018-11-07
Switching to Web IDL


domenic:
### Background

Streams as originally envisioned was meant as a ecosystem-agnostic specification at the level of ECMAScript. As such I wrote it with that specification style ([mostly](https://streams.spec.whatwg.org/#conventions)), instead of using Web IDL. See previous discussions in #45 and #732.

Since that time we've seen a shift in the ecosystem, toward using Web IDL more broadly:

- The URL and Encoding specifications (which use Web IDL), among others, are implemented in Node.js. That is, we have significant examples that Web IDL-based specifications can be implemented even in non-web environments.
- The Web Assembly specification settled on using Web IDL, instead of the ECMAScript style, despite being "engine level" and environment-agnostic. This is especially notable because most (all?) implementations of the Web Assembly API are done in the JS engine, instead of in the rendering engine with its IDL binding infrastructure. So this shows that IDL on the specification side is compatible with JS-engine-level implementations in a browser.
- The ECMAScript spec itself is [contemplating adopting an IDL](https://github.com/littledan/proposal-idl/blob/master/README.md), and at least one possible path there is to use an evolution of Web IDL.

Furthermore, not using Web IDL has caused us some pain. I think the pain breaks down in to three main areas:

1. Web platform implementers may be less familiar with this style of specification, which can be frustrating for them (e.g. #732). It can also be harder for them to implement in a way that matches the spec, using the tools available to them; they may need to use JavaScript engine APIs explicitly, or add lots of extended attributes to override Web IDL's default behavior if they use Web IDL anyway.
2. We lose some consistency with the rest of the platform. This category is about minor-ish technical things like over-generic methods (#961), non-enumerable methods, and especially argument coercion and options bag processing. We gain a little bit of consistency with ECMAScript, but ECMASCript isn't terribly internally consistent anyway, so on balance this is worse.
3. The ecosystem of tooling built up around Web IDL specs is not available to us. This involves things like automatic test coverage via idlharness.js (#962), generation of canonical TypeScript definitions (mentioned in #45), and integration with the Web Assembly host bindings proposal. It also causes us problems around how our spec integrates into the IDL ecosystem (https://github.com/whatwg/fetch/issues/780), and even our spec preprocessor (https://github.com/tabatkins/bikeshed/issues/1346). And proposals like [get-originals](https://github.com/domenic/get-originals) need to have a dedicated branch specifically for streams, instead of just treating these classes like other IDL interfaces.

[...]

PR #1021, opened 2019-11-02
Change promises-guide references to Web IDL

### LINKED ISSUE #1020: Links to "a new promise", "a promise resolved with", etc. are broken
Various terms of art like "a new promise" and "a promise resolved with" are links to a W3C document like https://www.w3.org/2001/tag/doc/promises-guide/#a-new-promise or so.  Since this text was written, however, that document has delegated its definitions to WebIDL.  Affected links need to be changed to https://heycam.github.io/webidl/#a-promise-resolved-with and similar (against whatever is the canonical version of that URL, not sure if the heycam-specific URL is that).

MattiasBuelens:
The spec text has broken links to the old [promises guide](https://www.w3.org/2001/tag/doc/promises-guide/). This is inconvenient for readers, but it also causes builds for spec PRs to fail (see [my comment on #1011](https://github.com/whatwg/streams/pull/1011#issuecomment-549029385)).

This PR fixes the links by pointing to their new Web IDL counterparts.

Todo list (non-exhaustive, feel free to suggest other changes):
* [x] Get spec to build again.
* [ ] Figure out what to do with "promise-calling".
* [x] Change usages of "react to promise" to explicitly pass one or two sets of steps (as in [this example](https://heycam.github.io/webidl/#promise-example-addDelay)), instead of passing a "fulfillment handler" and a "rejection handler".
* [ ] ~Change pairs of "upon fulfillment / upon rejection" to a single "react to promise".~ (Not that important right now.)

Fixes #1020

***
<a href="https://whatpr.org/streams/1021.html" title="Last updated on Nov 3, 2019, 9:31 PM UTC (66b04bf)">Preview</a> | <a href="https://whatpr.org/streams/1021/ae5e0cb...66b04bf.html" title="Last updated on Nov 3, 2019, 9:31 PM UTC (66b04bf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11,7 +11,7 @@ Opaque Elements: emu-alg
 </pre>
 
 <pre class=link-defaults>
-spec:promises-guide; type:dfn;
+spec:webidl; type:dfn;
     text:resolve
 </pre>
 
@@ -863,7 +863,7 @@ prototype, are created with the internal slots described in the following table:
   1. If ! IsReadableStreamAsyncIterator(*this*) is *false*, return <a>a promise rejected with</a> a *TypeError* exception.
   1. Let _reader_ be *this*.[[asyncIteratorReader]].
   1. If _reader_.[[ownerReadableStream]] is *undefined*, return <a>a promise rejected with</a> a *TypeError* exception.
-  1. Return the result of <a>transforming</a> ! ReadableStreamDefaultReaderRead(_reader_) with a fulfillment handler
+  1. Return the result of <a>reacting</a> to ! ReadableStreamDefaultReaderRead(_reader_) with a fulfillment handler
      which takes the argument _result_ and performs the following steps:
     1. Assert: Type(_result_) is Object.
     1. Let _done_ be ! Get(_result_, `"done"`).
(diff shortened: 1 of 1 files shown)


ricea:
Looks good so far.

domenic:
> Figure out what to do with "promise-calling".

I commented inline on some possible wording here.

> Change usages of "react to promise" to explicitly pass one or two sets of steps (as in this example), instead of passing a "fulfillment handler" and a "rejection handler".

No need to stress this too much; I think just changing "fulfillment handler" to "fulfillment steps" is basically good enough. In particular for the "inline" usages (like "Return the result of transforming sourceCancelPromise with a fulfillment handler that returns undefined") I don't think we need to make that into a more complicated thing with sub-bullets and sub-algorithms.

> Change pairs of "upon fulfillment / upon rejection" to a single "react to promise".

This is good but could be done as an editorial followup, as "upon fulfillment" and "upon rejection" still exist and the semantics of using them separately is the same as using "reacting" together. So if it's going to delay merging don't worry about that.

index.bs
```diff
@@ -5013,13 +5012,13 @@ nothrow>TransformStreamDefaultSinkCloseAlgorithm( <var>stream</var> )</h4>
   1. Let _controller_ be _stream_.[[transformStreamController]].
   1. Let _flushPromise_ be the result of performing _controller_.[[flushAlgorithm]].
   1. Perform ! TransformStreamDefaultControllerClearAlgorithms(_controller_).
-  1. Return the result of <a>transforming</a> _flushPromise_ with:
-    1. A fulfillment handler that performs the following steps:
+  1. Return the result of <a>reacting</a> to _flushPromise_:
+    1. If _flushPromise_ was fulfilled, then:
```
MattiasBuelens: This was already very close to the new format from [the example](https://heycam.github.io/webidl/#promise-example-addDelay), so I went ahead and refactored it. I must say: I like this new format quite a bit! 😄

index.bs
```diff
@@ -5365,7 +5364,7 @@ throws>CreateAlgorithmFromUnderlyingMethod ( <var ignore>underlyingObject</var>,
 <var>args</var> )</h4>
 
 <div class="note">
-  PromiseCall is a variant of <a>promise-calling</a> that works on methods.
+  PromiseCall is a variant of <a lt="call a user object's operation">promise-calling</a> that works on methods.
```
MattiasBuelens: > > Figure out what to do with "promise-calling".
> 
> I commented inline on some possible wording here.

@domenic What was your suggestion for promise-calling? I can't seem to find your comment. 😕

domenic: Oops, it was in an unsubmitted review.

> Maybe something like "This encapsulates the relevant promise-related parts of the Web IDL [call a user object's operation] algorithm for use while we work on [moving to Web IDL](https://github.com/whatwg/streams/issues/963)."

PR #1022, opened 2019-11-07
Set [[disturbed]] synchronously in ReadableStreamPipeTo

ricea:
Set the [[disturbed]] internal slot to true in ReadableStreamPipeTo.
This is useful for the Fetch Standard. It is not visible to user code.

See https://github.com/whatwg/fetch/pull/959 for the Fetch Standard
change that requires this.

Also fix two places where the name of the `state` internal slot was
mistakenly italicised.

***
<a href="https://whatpr.org/streams/1022.html" title="Last updated on Nov 7, 2019, 7:01 AM UTC (9e94c6b)">Preview</a> | <a href="https://whatpr.org/streams/1022/547428b...9e94c6b.html" title="Last updated on Nov 7, 2019, 7:01 AM UTC (9e94c6b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1122,6 +1122,7 @@ noticeable asymmetry between the two branches, and limits the possible <a>chunks
      discretion.
   1. Otherwise, let _reader_ be ! AcquireReadableStreamDefaultReader(_source_).
   1. Let _writer_ be ! AcquireWritableStreamDefaultWriter(_dest_).
+  1. Set _source_.[[disturbed]] to *true*.
   1. Let _shuttingDown_ be *false*.
   1. Let _promise_ be <a>a new promise</a>.
   1. If _signal_ is not *undefined*,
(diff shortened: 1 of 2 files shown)


domenic:
> It is not visible to user code.

Isn't it visible if someone does something like

```js
response.body.pipeTo(...)
console.log(response.bodyUsed);
```

?

ricea:
> Isn't it visible if someone does something like
> 
> ```js
> response.body.pipeTo(...)
> console.log(response.bodyUsed);
> ```
> 
> ?

My original thought was that this depends on the implementation of pipeTo (ie. whether or not it does the first read synchronously). But I think you can make the difference detectable in any implementation by doing

```javascript
response.body.pipeTo(new WritableStream({}, { highWaterMark: 0 });
console.log(response.bodyUsed);
```

So actually implementations will have to explicitly apply this change. Or maybe this standard is the wrong place to make this change, and it should go in Fetch instead.

yutakahirano:
Once `locked` and `disturbed`, there is no way to observe any internal state of a ReadableStream (until it gets unlocked). That is what I would expect for pipeTo, so I would like to put it in the spec.

domenic:
Yeah, I am +1 on this change, I just think we may want to accompany it with some web platform tests.

ricea:
Please see the test at https://github.com/web-platform-tests/wpt/pull/20205.

PR #1027, opened 2020-02-03
Add checks before Close and Enqueue in tee()

ricea:
ReadableStream.prototype.tee didn't check CanCloseOrEnqueue() before
calling Close() or Enqueue(), violating the interface contract. Add
checks of CanCloseOrEnqueue() to fix it.

***
<a href="https://whatpr.org/streams/1027.html" title="Last updated on Feb 3, 2020, 10:38 AM UTC (627adbf)">Preview</a> | <a href="https://whatpr.org/streams/1027/08b115d...627adbf.html" title="Last updated on Feb 3, 2020, 10:38 AM UTC (627adbf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1065,9 +1065,11 @@ noticeable asymmetry between the two branches, and limits the possible <a>chunks
       1. Let _done_ be ! Get(_result_, `"done"`).
       1. Assert: Type(_done_) is Boolean.
       1. If _done_ is *true*,
-        1. If _canceled1_ is *false*,
+        1. If _canceled1_ is *false* and !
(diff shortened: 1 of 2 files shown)


domenic:
Should we consider moving these checks into Close() and Enqueue() themselves, with an early-return if they fail? I think the original idea of separating them was that so callers that don't go through the public API could do a bit less work, but things have evolved such that that's not a great idea, so maybe we should inline them to avoid future contract violations...

ricea:
> Should we consider moving these checks into Close() and Enqueue() themselves, with an early-return if they fail? I think the original idea of separating them was that so callers that don't go through the public API could do a bit less work, but things have evolved such that that's not a great idea, so maybe we should inline them to avoid future contract violations...

Done in #1029. Closing this PR in favour of that one.

PR #1029, opened 2020-02-27
Add state checks to Close() and Enqueue()

ricea:
Exit early from abstract operations
ReadableStreamDefaultControllerClose() and
ReadableStreamDefaultControllerEnqueue() if
ReadableStreamDefaultControllerCanCloseOrEnqueue() is false. Remove the
asserts that it is true, and the requirement that callers perform the
check as part of the interface contract.

In practice the implementation of ReadableStreamTee() was failing to
properly check CanCloseOrEnqueue(). By relaxing the interface contract
we can prevent similar problems with future standards.

Also remove a now-redundant check of CanCloseOrEnqueue() in
TransformStreamDefaultControllerTerminate().

Also make similar changes to ReadableByteStreamControllerClose() and
ReadableByteStreamControllerEnqueue().

***
<a href="https://whatpr.org/streams/1029.html" title="Last updated on Feb 27, 2020, 9:40 AM UTC (3f84404)">Preview</a> | <a href="https://whatpr.org/streams/1029/08b115d...3f84404.html" title="Last updated on Feb 27, 2020, 9:40 AM UTC (3f84404)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2065,11 +2065,11 @@ export>ReadableStreamDefaultControllerClose ( <var>controller</var> )</h4>
 
 This abstract operation can be called by other specifications that wish to close a readable stream, in the same way
 a developer-created stream would be closed by its associated controller object. Specifications should <em>not</em> do
-this to streams they did not create, and must ensure they have obeyed the preconditions (listed here as asserts).
+this to streams they did not create.
 
 <emu-alg>
+  1. If ! ReadableStreamDefaultControllerCanCloseOrEnqueue(_controller_) is *false*, return.
   1. Let _stream_ be _controller_.[[controlledReadableStream]].
-  1. Assert: ! ReadableStreamDefaultControllerCanCloseOrEnqueue(_controller_) is *true*.
   1. Set _controller_.[[closeRequested]] to *true*.
   1. If _controller_.[[queue]] is empty,
     1. Perform ! ReadableStreamDefaultControllerClearAlgorithms(_controller_).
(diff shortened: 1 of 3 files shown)


ricea:
> LGTM. Tests?

It doesn't appear to be possible to reproduce the problem with `ReadableStreamTee()` in a portable fashion. The logic only fails if microtasks can be made to run synchronously.

jswalden:
As best I understand things, microtasks shouldn't be _able_ to be made to be run synchronously.  If some spec is failing to properly defer a microtask checkpoint, that's just a bug in the spec.  Or am I somehow wrong about this?

ricea:
> As best I understand things, microtasks shouldn't be _able_ to be made to be run synchronously. If some spec is failing to properly defer a microtask checkpoint, that's just a bug in the spec. Or am I somehow wrong about this?

This is how it's done in Chromium:

https://chromium.googlesource.com/chromium/src/+/12310ed05f15fea5fa6824c6a6b5d86f81532e25/third_party/blink/web_tests/http/tests/streams/chromium/concurrent-close.html

I don't know if that's a spec bug or an implementation bug.

bzbarsky:
> I don't know if that's a spec bug or an implementation bug.

Implementation bug.  Per spec, that getter runs at https://heycam.github.io/webidl/#call-a-user-objects-operation step 10.1.  But step 7 of that algorithm calls https://html.spec.whatwg.org/multipage/webappapis.html#prepare-to-run-script which pushes a JS execution context on the stack.  After that, until the matching https://html.spec.whatwg.org/multipage/webappapis.html#clean-up-after-running-script (which happens in step 15.2), there should be no microtask checkpoints.

Issue #825, opened 2017-10-04
Use [[UpperCamelCase]] in internal slots


anssiko:
The spec uses lowerCamelCase names for its internal slots while the ECMAScript spec uses UpperCamelCase. Consider adopting the UpperCamelCase notation for this spec too.

For the context, see @annevk's https://github.com/heycam/webidl/issues/453#issuecomment-334101431

PR #1035, opened 2020-04-13
Rewrite to use Web IDL, and generally modernize

### LINKED ISSUE #586: Should ReadableStreamDefaultReader be exposed?
The spec doesn't seem to cover which classes should be exposed where.

Right now `ReadableStreamDefaultReader` isn't exposed on the global. Should it be? Feels like it should be exposed somewhere for extensibility.

### LINKED ISSUE #684: Are abstract operations supposed to be in alphabetical order?
It seems like most of the abstract operations are in alphabetical order within a section, but some are not. Is there a rule here, or is it a case-by-case thing?

https://streams.spec.whatwg.org/commit-snapshots/c9d7b9de530506d42034b74f5f875a1b49b45507#contents

> 3\.9\.4 ReadableStreamDefaultControllerClose \( controller \)

### LINKED ISSUE #687: Add identifier highlighting similar to the ES spec
E.g. in https://tc39.github.io/ecma262/#sec-number-constructor-number-value, if you click _value_, _n_, or _o_, it highlights all other instances of the same value.

We can prototype this here in streams and then possibly consider expanding it to other WHATWG specs.

### LINKED ISSUE #885: Make the table of contents more useful/less dense
Streams is a bit different in structure from other specifications in that it uses a heading for every API surface and abstract operation. This leads to a much larger table of contents than other specifications, and as such, I think the TOC is not very useful in its current form.

I think we should omit:

- Everything under "Class `X`"
- Every abstract operation (but keep the categories, e.g. keep "General Readable Stream Abstract Operations")

This can be done by adding `class="no-toc"` to the appropriate headings in the `index.bs` document.

Marking this as a "good first issue" in case anyone wants to help out.

### LINKED ISSUE #907: Consider better styling for API intro notes

### LINKED ISSUE #952: Question: Should Streams have the prototype for `Symbol.toStringTag` and `toString` declared?
In certain implementations of whatwg streams, I've come across them not implementing `toString` or `Symbol.toStringTag` on the prototype meaning that if you're trying to find out if you've been given a stream (or another data type) then you have to rely on duck typing.

In the following, I'd expect to see `"[object ReadableStream]"` returned:
```js
let r = new ReadableStream()
// ReadableStream {}
r.toString()
// "[object Object]"
> Object.prototype.toString.call(r)
// "[object Object]"
```

In Safari these all return as I'd expect, in Chrome and Firefox they return `"[object Object]"`. The spec doesn't actually specify the toString behavior.

### LINKED ISSUE #963: Switching to Web IDL

### LINKED ISSUE #965: Constructor specification is unclear
Is subclassing supposed to work?

```js
class SpamStream extends ReadableStream {
    constructor() {
        let stream;
        super({
            pull(controller) {
                controller.enqueue("spam");
                stream.spamCount++;
            }
        });
        this.spamCount = 0;
        stream = this;
    }
    hasSpammed() { return this.spamCount > 0; }
}

new SpamStream().hasSpammed()
```

The standard currently says:

>  We prefix section headings with `new` to indicate they are defining constructors; when doing so, we assume that NewTarget will be checked before the algorithm starts.

What it doesn't say is how **this** gets initialized. Mozilla's current implementation ignores NewTarget and always creates an object whose [[Prototype]] is the original value of ReadableStream.prototype. So the above code would fail; `hasSpammed` would not be on that object's prototype chain.

I think we want something like

> 1. If NewTarget is **undefined**, throw a **TypeError** exception.
> 2. Let **this** be ? OrdinaryCreateFromConstructor(NewTarget, the name of the per-realm intrinsic object that is this class's original prototype, the List of this class's internal slots).

(Technically, this doesn't quite work, because OrdinaryCreateFromConstructor step 1 effectively asserts that it is being called from inside ECMA-262 and not from other standards. But I think it's an improvement anyway.)

### LINKED ISSUE #1005: Specifying QueuingStrategies in WebIDL
As part of #963, we will have to specify CountQueuingStrategy and ByteLengthQueuingStrategy in WebIDL. There are a couple of things about these classes that are "strange" from a WebIDL point-of-view:

1. size() is a function (as opposed to a method), but it is inherited
2. `highWaterMark` is added by the constructor as a property of the object, rather than coming from the prototype.

I think problem 2. can be fixed quite easily by making `highWaterMark` a normal interface attribute.

Problem 1. is harder. WebIDL doesn't seem to have functions as such at all. The closest thing is "[Callback functions](https://heycam.github.io/webidl/#idl-callback-functions)", but it doesn't appear that they can exist as attributes on an interface. There are also "static operations", which are a bit like functions, but appear as properties of the constructor, not the prototype.

### LINKED ISSUE #1017: Tracking issue for Web IDL async iterator support
A blocker for #963 is expanding Web IDL's async iterator support to be powerful enough for readable streams. The tracking issue over there is https://github.com/heycam/webidl/issues/800.

On this side I want to record a draft of the algorithms. After the following get done:

- https://github.com/heycam/webidl/pull/802
- https://github.com/heycam/webidl/issues/804
- https://github.com/heycam/webidl/issues/803
- The actual async iterator PR tracked in https://github.com/heycam/webidl/issues/800

then it will look something like this:

**Async iterator initialization steps**, given a `ReadableStream` _stream_, async iterator _iterator_, and argument _options_:

1. Let _reader_ be ? AcquireReadableStreamDefaultReader(_stream_).
1. Set _iterator_.[[reader]] to _reader_.
1. Set _iterator_.[[preventCancel]] to _options_["`preventCancel`"].

To **get the next iteration result** for `ReadableStream` given the async iterator _iterator_:

1. Let promise be a new promise.
1. Let _reader_ be _iterator_.[[reader]].
1. If _reader_.[[ownerReadableStream]] is undefined, return a promise rejected with a `TypeError` exception.
1. Return the result of reacting to ! ReadableStreamDefaultReaderRead(_reader_) with a fulfillment handler that takes an argument _result_ and performs the following steps:
    1. Assert: Type(_result_) is Object.
    1. Let done be ! Get(_result_, "`done`").
    1. Assert: Type(_done_) is Boolean.
    1. If _done_ is true,
        1. Perform ! ReadableStreamReaderGenericRelease(_reader_).
        1. Return undefined.
    1. Let _value_ be ! Get(_result_, "`value`").
    1. Return _value.

(The above will become a bit cleaner if we are able to move away from using actual ES objects for read results, which will likely happen as part of the general Web IDL conversion.)

To **close the async iterator** for `ReadableStream` given the async iterator _iterator_ and argument _value_:

1. Let _reader_ be _iterator_.[[asyncIteratorReader]].
1. If _reader_.[[ownerReadableStream]] is undefined, return a promise rejected with a `TypeError` exception.
1. If _reader_.[[readRequests]] is not empty, return a promise rejected with a `TypeError` exception.
1. If _iterator_.[[preventCancel]] is false, then:
    1. Let _result_ be ! ReadableStreamReaderGenericCancel(_reader_, _value_).
    1. Perform ! ReadableStreamReaderGenericRelease(_reader_).
    1. Return _result_.
1. Perform ! ReadableStreamReaderGenericRelease(reader).
1. Return a promise resolved with undefined.

---

I think this all works, so that's good.

domenic:
Closes #963.

Normative changes to widely-implemented features, roughly in order of most disruptive to least-disruptive:

* For the queuing strategy classes, their size and highWaterMark properties are now getters on the prototype, instead of data properties on the prototype and instance (respectively). Closes #1005. In particular this means that attempts to set either of them post-creation will throw a TypeError. Chromium already ships these semantics.

* Functions which take a dictionary no longer accept non-objects.

* For the queuing strategy classes, their highWaterMark property will no longer return a non-number from their highWaterMark properties, if one was passed to the constructor. Instead, NaN will be returned.

* All methods and accessors are now enumerable, per Web IDL defaults, instead of non-enumerable, per ECMAScript defaults.

* All classes are now exposed globally. Formerly, ReadableStreamDefaultReader, ReadableStreamBYOBReader, ReadableStreamDefaultController, ReadableByteStreamController, WritableStreamDefaultWriter, WritableStreamDefaultController, and TransformStreamDefaultController were not exposed. Closes #586.

* All classes now have `[Symbol.toStringTag]` properties. Closes #952.

* Some functions have changed their length property value.

* Some exceptions are thrown earlier, at argument-conversion time.

* Property lookup in options arguments now happens earlier, at argument-conversion time, and in alphabetical order, per dictionary rules.

Normative changes to unimplemented features:

* `ReadableStream`'s `getIterator()` method has been renamed to `values()` as part of adopting Web IDL's infrastructure for async iterators.

* The `byobRequest` property on `ReadableByteStreamController` now returns `null` when there is no BYOB request, instead of returning undefined.

* The `view` property on `ReadableStreamBYOBRequest` now returns `null` when the view cannot be written into, instead of returning undefined.

* Various byte-stream-related APIs that used to specifically prohibit detached buffers now check for zero-length views or buffers, which is a more general category.

* The async iterator's `next()` and `return()` methods now behave more like async generators, e.g. returning promises fulfilled with `{ value: undefined, done: true }` after `return()`ing the iterator, instead of returning a rejected promise.

Editorial changes:

* All APIs are specified to using Web IDL now, instead of using a modified version of the ECMAScript specification conventions. We continue using abstract operations and completion records for now, and we have to drop down to the ECMAScript level in a couple places (notably for dealing with %ObjectPrototype% vs. null-prototype iteration result objects, and transferring array buffers). But overall this removes a lot of type-checking and conversion boilerplate from the specification. Closes #963. Closes #1017. See #1036 for further followup on the iteration result objects.

* Individual abstract operations, constructors, methods, and properties no longer have their own heading. They are instead lumped together in sections. Closes #885.

* The constructors, methods, and properties are now documented in a per-class block, using the usual WHATWG "domintro" style. Closes #907.

* Abstract operations are now consistently alphabetized within their section. Closes #684.

* By using Bikeshed's `<div algorithm>` feature, we now get automatic identifier highlighting. Closes #687.

* Switched to 100-character line limits, 1-space indents, and omitting end tags, per WHATWG conventions.

* Removed usage of emu-algify in favor of using some more of Bikeshed's built-in features, plus manually annotating a few things.

* Switched to concise Bikeshed linking syntax, e.g. [=term=] and [$AbstractOp$].

* Eliminated a number of utility abstract operations, especially around calling functions, by better using Web IDL.

Other bug fixes:

* Web IDL makes constructor behavior clear, so this closes #965.

---

Note to reviewers: this isn't really a fun thing to review. I started by trying to do this incrementally, e.g. removing emu-algify first, then reformatting to WHATWG conventions/modern Bikeshed syntax, then Web IDL-ifying, to make it more reviewable. But this way turned out to be a lot easier for me, and I don't think this would have gotten done if I'd taken the harder route.

The best review strategy I can think of is just to scroll around randomly in the output preview and see if anything catches your eye.

---

TODOs:

- [x] Add back async iterators, after https://github.com/heycam/webidl/pull/808 arrives.
- [x] Update the reference implementation. Probably use [webidl2js](https://github.com/jsdom/webidl2js). (In progress)
- [x] See what tests break. Update them all. (In progress)
- [x] Add idlharness tests.
- [x] Add tests for the per-global-ness of the queuing strategy functions.
- [ ] (Optional) Consider removing the hand-curated IDs for non-exported abstract operations. They are painful to maintain and don't buy us much.
- [ ] (Optional) Consider renaming internal slot names to be less long-winded now that they are no longer used for brand checks
- [ ] (Optional) Consider renaming internal slot names to be uppercase to solve #825.

---

Here's the pull request template, which we should update as we complete the above list of TODOs. I think in particular asking for implementer interest will be more feasible after we have the test changes up, although the "normative changes" list above is reasonable by itself.

   * [Chrome](https://github.com/whatwg/streams/pull/1035#issuecomment-622316622)
   * [Firefox](https://github.com/whatwg/streams/pull/1035#issuecomment-638285307)
   * https://github.com/web-platform-tests/wpt/pull/22982
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1035.html" title="Last updated on Jun 11, 2020, 3:41 PM UTC (1da8031)">Preview</a> | <a href="https://whatpr.org/streams/1035/cfcd303...1da8031.html" title="Last updated on Jun 11, 2020, 3:41 PM UTC (1da8031)">Diff</a>


ricea:
Wow. It might take me a little time to review this.

I assume we don't have to mention minor normative changes like the order of lookup of dictionary fields changing?

domenic:
I'd love to have the list of normative changes as exhaustive as possible, at least to the extent of including general points like that. (It seems similar to my "Some exceptions are thrown earlier, at argument-conversion time".) I'll edit the OP to include that, and if you can think of any other such changes, just let me know.

ricea:
Possible breaking change: the first argument of the constructors of ReadableStream, WritableStream and TransformStream is now compulsory.

Methods which take a dictionary will be stricter about what types they accept. For example `readable.getReader(true)` will no longer work.

An additional irrelevant change for queuing strategies: currently `new CountQueuingStrategy({highWaterMark: 'sandwich'}).highWaterMark` returns `"sandwich"`. After this change, it will return `NaN`.

domenic:
> Possible breaking change: the first argument of the constructors of ReadableStream, WritableStream and TransformStream is now compulsory.

Note that this is relatively easy to reverse, and maybe we should do so to minimize unnecessary breaking changes. On the other hand, I have a hard time imagining code usefully using no-argument versions of these...

Oh wait, no-arg TransformStream is actually really important. OK, well, we should definitely allow that. I think I'll just allow it for all of them.

ricea:
Reference implementation changes look good so far.

domenic:
The reference implemenatation is mostly ported. I want to wait for the following webidl2js PRs to land and then reassess the test pass rate:

- https://github.com/jsdom/webidl2js/pull/200
- https://github.com/jsdom/webidl2js/pull/198
- https://github.com/jsdom/webidl2js/pull/194

index.bs
```diff
@@ -58,5398 +68,5539 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
+ 1. Return ! [$ReadableStreamReaderGenericCancel$]([=this=], |reason|).
 </div>
 
-<emu-alg>
-  1. If ! IsReadableStreamBYOBReader(*this*) is *false*, return <a>a promise rejected with</a> a *TypeError* exception.
-  1. If *this*.[[ownerReadableStream]] is *undefined*, return <a>a promise rejected with</a> a *TypeError* exception.
-  1. If Type(_view_) is not Object, return <a>a promise rejected with</a> a *TypeError* exception.
-  1. If _view_ does not have a [[ViewedArrayBuffer]] internal slot, return <a>a promise rejected with</a> a *TypeError*
-     exception.
-  1. If ! IsDetachedBuffer(_view_.[[ViewedArrayBuffer]]) is *true*, return <a>a promise rejected with</a> a *TypeError*
```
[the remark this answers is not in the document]
domenic: Hmm, I thought it threw an error when converting, but I see that's not accurate, at least in the spec. Uh-oh. I'll want to check implementations, but if the Web IDL spec is correct, then yeah, I made some wrong assumptions. Thanks!

index.bs
```diff
@@ -58,5398 +68,5539 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
 
-    <a href="#rbs-controller-close">close</a>()
-    <a href="#rbs-controller-enqueue">enqueue</a>(chunk)
-    <a href="#rbs-controller-error">error</a>(e)
-  }
-</code></pre>
+<xmp class="idl">
+[Exposed=(Window,Worker,Worklet)]
+interface ReadableByteStreamController {
+  readonly attribute ReadableStreamBYOBRequest byobRequest;
```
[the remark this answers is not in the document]
domenic: Good catch! It should probably switch to `null` since that's more idiomatic for Web IDL-using APIs.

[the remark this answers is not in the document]
MattiasBuelens: Yes, I always found it a bit weird that `desiredSize` was `number | null`, but `byobRequest` was `ReadableStreamBYOBRequest | undefined`.

index.bs
```diff
@@ -58,5398 +68,5539 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
+  controller any queued [=chunks=], or pull from the [=underlying source=] to get more chunks.
+</dl>
 
-  Note how we don't <code>await</code> the promise returned by {{WritableStreamDefaultWriter/write()}}; this would be
-  redundant with <code>await</code>ing the {{WritableStreamDefaultWriter/ready}} promise. Additionally, similar to <a
-  href="#example-manual-write-batch">a previous example</a>, we use the <code>.catch(() => {})</code> pattern on the
-  promises returned by {{WritableStreamDefaultWriter/write()}}; in this case we'll be notified about any failures
-  <code>await</code>ing the {{WritableStreamDefaultWriter/ready}} promise.
+(These are defined as internal methods, instead of as abstract operations, so that they can be
+called polymorphically by the {{Readab
```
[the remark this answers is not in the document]
ricea: I don't know why Github attached this comment to totally the wrong line. I was referring to line 2131.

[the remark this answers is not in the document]
ricea: This should have been attached to line 3985.

index.bs
```diff
@@ -58,5398 +68,5539 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
+ 1. Perform ! [$ReadableByteStreamControllerError$]([=this=], |e|).
 </div>
 
-<emu-alg>
-  1. If ! IsReadableByteStreamController(*this*) is *false*, throw a *TypeError* exception.
-  1. If *this*.[[closeRequested]] is *true*, throw a *TypeError* exception.
-  1. If *this*.[[controlledReadableByteStream]].[[state]] is not `"readable"`, throw a *TypeError* exception.
-  1. If Type(_chunk_) is not Object, throw a *TypeError* exception.
-  1. If _chunk_ does not have a [[ViewedArrayBuffer]] internal slot, throw a *TypeError* exception.
-  1. If ! IsDetachedBuffer(_chunk_.[[ViewedArrayBuffer]]) is *true*, throw a *TypeError* exception.
```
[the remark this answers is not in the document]
domenic: Hmm. I'm thinking we should either reject all zero-length chunks here, or accept them. We shouldn't treat detached specially.

I'll reject zero-length chunks for now, here and in respondWithNewView.

index.bs
ricea, line 3321: Should be `abort(reason)`.

ricea, line 3408:
```suggestion
 1. Let |underlyingSinkDict| be |underlyingSink|, [=converted to an IDL value=] of type
```

ricea, line 4149: Typo: manipualtion

ricea:
The preview lgtm.

I'm going to review the reference implementation changes next. After that I will do a light review of the markup.

reference-implementation/lib/ReadableStream-impl.js
```diff
@@ -0,0 +1,124 @@
...
+      this, transform.writable, options.preventClose, options.preventAbort, options.preventCancel, options.signal
+    );
+
+    setPromiseIsHandledToTrue(promise);
+
+    return transform.readable;
+  }
+
+  pipeTo(destination, options) {
+    // Conversion here is needed until https://github.com/jsdom/webidl2js/issues/81 is fixed.
```
ricea, line 86: I find the word "conversion" confusing here. Would "type checking" be acceptable instead?

reference-implementation/lib/ReadableStreamBYOBReader-impl.js
```diff
@@ -0,0 +1,53 @@
...
+  read(view) {
+    if (view.byteLength === 0) {
+      return promiseRejectedWith(new TypeError('view must have non-zero byteLength'));
+    }
+    if (view.buffer.byteLength === 0) {
+      return promiseRejectedWith(new TypeError('view\'s buffer must have non-zero byteLength'));
+    }
+
+    if (this._ownerReadableStream === undefined) {
+      return promiseRejectedWith(readerLockException('cancel'));
```
ricea, line 32:
```suggestion
      return promiseRejectedWith(readerLockException('read'));
```

reference-implementation/lib/helpers/webidl.js
```diff
@@ -0,0 +1,203 @@
...
+
+// https://heycam.github.io/webidl/#new
+// https://github.com/jsdom/webidl2js/issues/193
+exports.webidlNew = (globalObject, typeName, implModule) => {
+  if (globalObject[utils.ctorRegistrySymbol] === undefined) {
+    throw new Error('Internal error: invalid global object');
+  }
+
+  const ctor = globalObject[utils.ctorRegistrySymbol][typeName];
+  if (ctor === undefined) {
```
ricea, line 13: Maybe this should be an assert()?

reference-implementation/lib/helpers/webidl.js
```diff
@@ -0,0 +1,203 @@
...
+
+// https://heycam.github.io/webidl/#reject
+function rejectPromise(p, reason) {
+  promiseSideTable.get(p).reject(reason);
+  promiseSideTable.get(p).stateIsPending = false;
+}
+
+// https://heycam.github.io/webidl/#a-promise-resolved-with
+function promiseResolvedWith(value) {
+  // Cannot use original Promise.resolve since that will return value itself sometimes, unlike Web IDL.
```
ricea, line 79: Is this observable? Would it change the timing of resolving _startPromise_?

ricea:
Reference implementation lgtm.

reference-implementation/lib/helpers/webidl.js
domenic: It is, because then we don't end up with a weak map entry for the promise sometimes.

This implementation is actually more in line with Web IDL's "promised resolved with" than one that uses `Promise.resolve`; Web IDL always creates a new promise.

reference-implementation/lib/helpers/webidl.js
domenic: This moved into webidl2js so now it's not our problem :)

domenic:
So I'm mainly still waiting on some webidl2js updates. Although async value iterators were just merged into Web IDL, so I can work on the spec for that now. (But not yet the reference implementation.)

Any thoughts on the "(Optional)" items in the OP?

index.bs
```diff
@@ -892,7 +896,70 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  </div>
 </div>
 
-<!-- TODO async iterator stuff -->
+<h4 id="rs-asynciterator" oldids="rs-asynciterator-prototype,
+default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h4>
+
+<dl class="domintro">
+ <dt><code>for await (const <var ignore>chunk</var> of <var ignore>stream</var>) { ... }</code>
+ <dt><code>for await (const <var ignore>chunk</var> of <var ignore>stream</var>.values([{ {{ReadableStreamIteratorOptions/preventCancel}} }])) { ... }</code>
```
MattiasBuelens, line 904: I think the square brackets are a typo? You should be able to call this as `.values({ preventCancel })`, correct?
```suggestion
 <dt><code>for await (const <var ignore>chunk</var> of <var ignore>stream</var>.values({ {{ReadableStreamIteratorOptions/preventCancel}} })) { ... }</code>
```

domenic: They indicate optionality, not array-ness. I can see how it's confusing though :-/. Maybe in this case at least we could just do `{ preventCancel: true }` to show the two different styles.

MattiasBuelens: Ooooh right, I see it know. There's a similar thing going on with `reader = stream.getReader([{ mode: "byob" }])`, but for some reason I didn't interpret that as an array. I think seeing the lengthy `for await of` construct around it made me interpret the whole thing as a line of JavaScript, rather than a signature with optional arguments.

It seems like the syntax used in these non-normative sections is a bit ambiguous, since we use square brackets for both optionality and arrays:
* `stream = new ReadableStream(underlyingSource[, strategy])` (optionality)
* `[branch1, branch2] = stream.tee()` (array return value)

Then again, I can't immediately come up with a better way to present this. 🤷 
(Something like `{ preventCancel }?` would work in TypeScript, but that's not a thing in JavaScript.)

index.bs
```diff
@@ -58,5398 +68,5616 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
-<dl>
-  <dt><dfn abstract-op lt="[[CancelSteps]]">\[[CancelSteps]](<var>reason</var>)</dfn></dt>
-  <dd>The controller's steps that run in reaction to the stream being <a lt="cancel a readable stream">canceled</a>,
-  used to clean up the state stored in the controller and inform the <a>underlying source</a>.</dd>
-
-  <dt><dfn abstract-op lt="[[PullSteps]]">\[[PullSteps]]()</dfn></dt>
-  <dd>The controller's steps that run when a <a>default reader</a> is read from, used to pull from the controller any
-  queued <a>chunks</a>, or pull from the <a>underlying source</a> to get more chunks.</dd>
+  <p>By default, calling the async iterator's `return()` method will also [=cancel a readable
+  stream|cancel=] the stream. To prevent this, use the stream's `values()`method, passing true for
```
ricea, line 913:
```suggestion
  stream|cancel=] the stream. To prevent this, use the stream's `values()` method, passing true for
```

ricea:
I spot-checked the markup and it looks okay. I'm going to trust in Bikeshed that it would point out any major problems if there were any.

Async iteration changes look okay.

From the point of view of the Blink implementation, it's okay to remove the old ids for non-exported abstract operations, as long as there is some id we can link to. It will temporarily break the links in our source code, but since we have to update it anyway it's not a huge burden.

I think renaming slot names should be done in a separate change, simply because this change is already so huge it's hard to properly review it.

domenic:
Status update:

- webidl2js v16 fixed a few of the relevant bugs. The missing pieces are callback function support (landing soon) and async iterator support (not yet started). I'm unsure whether to block landing this PR on async iterator support, or whether we should investigate other approaches, e.g. hard-coding the CI to skip async iterator checks or introducing `-expected.txt` support to `wptrunner` (https://github.com/domenic/wpt-runner/issues/9).

- I did another pass through the WPTs and they now are updated so that the reference implementation matches them for all cases except callback function support and async iterator support. So probably the tests are good now, but we don't have the extra guarantee of the reference implementation to double check us.

- The OP's list of normative changes has been kept up to date. Maybe it's time to start asking implementers what they think.

ricea:
>  I'm unsure whether to block landing this PR on async iterator support, or whether we should investigate other approaches, e.g. hard-coding the CI to skip async iterator checks or introducing -expected.txt support to wptrunner (domenic/wpt-runner#9).

Obviously it's preferable if all the tests pass before we land it. If browsers start implementing async iterators based on this new spec, then they're really going to rely on the tests being correct.

However, if it's going to take more than a couple of weeks or so, then a workaround would be better. Having `-expected.txt` support is attractive to me as it lets us write tests for evil things that the reference implementation couldn't reasonably be expected to pass.

ricea:
Wearing my implementer hat, I approve these changes for Google Chrome.

domenic:
I noticed that Chrome currently does not implement a setter for highWaterMark on the queuing strategy classes, so I'll remove that. (I added it in this PR for compatibility, but it seems it's not needed.)

domenic:
OK, our reference implementation is now up to date with _all_ changes, including the async iterator changes (using the branch at https://github.com/jsdom/webidl2js/pull/224). This uncovered one slight semantic shift about what happens if you call `next()` after `return()`, which I've updated the tests for in https://github.com/web-platform-tests/wpt/pull/22982. And it also let me tighten the spec language a bit further.

Although we ideally shouldn't merge this until the reference implementation dependency (https://github.com/jsdom/webidl2js/pull/224) is merged, I think we've done all the spec and web platform test changes, and have a completely exhaustive list of everything that this effects. So, it's time to check in with implementers.

Implementers: see #963 for background, and see the OP of this thread(https://github.com/whatwg/streams/pull/1035#issue-402869237) for a complete list of changes, both normative and editorial.

In https://github.com/whatwg/streams/pull/1035#issuecomment-622316622 @ricea approved of these changes for Chrome. This will largely amount to introducing proper dictionary types, enums, etc. and moving away from `any`.

@jorendorff, do these sound reasonable for Mozilla? I imagine you'd continue to implement Streams in the JS engine, but just make the normative changes listed in the OP, guided by the test updates at https://github.com/web-platform-tests/wpt/pull/22982.

@othermaciej, how about for WebKit? At one point I know streams were already implemented in Web IDL in WebKit, but using some tricks like `[NonEnumerable]` and lots of `any`s, like Chrome's current implementation. If that's still the case, this would amount to removing those tricks.

index.bs
```diff
@@ -58,5398 +68,5601 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
...
-{{ReadableStream/locked|readableStream.locked}} or {{WritableStream/locked|writableStream.locked}} properties.
+<p class="note">The reasons for the usage of these foreign ECMAScript conventions are largely
+historical. We urge you to avoid following our example when writing your own web specifications.
 
-A reader or writer also has the capability to <dfn lt="release a lock|release a read lock|release a write lock">release
-its lock</dfn>, which makes it no longer active, and allows further readers or writers to be acquired. This is done via
-the {{ReadableStreamDefaultReader/releaseLock()|defaultReader.releaseLock()}},
-{{ReadableStreamBYOBReader/releaseLock()|byobReader.releaseLock()}}, or
-{{WritableStreamDefaultWriter/releaseLock()|writer.releaseLock()}} method, as appropriate.
+Finally, as in [[!ECMASCRIPT]], all numbers are represented as double-precision floating point
```
andreubotella, line 344: Minor nitpick, but ECMAScript has three numeric types, including the spec-only "mathematical values". I'd explicitly reference the Number type, or go with IDL's `unrestricted double`.

domenic: Thanks, I agree. I think this text originated from before the introduction of BigInt.

domenic:
Ping for thoughts from @jorendorff and @youennf :)

jorendorff:
@domenic This change is good. Thanks for doing the legwork.

PR #1039, opened 2020-06-11
Always resolve tee's cancel promise after stream closes or errors

### LINKED ISSUE #1038: Cancelling one branch of a tee'd stream never finishes if other branch reads all chunks
When teeing a stream, if one branch is cancelled while the other branch is read until the end, then the cancel promise of the first branch never resolves.

## Reproduction case
```javascript
    start(controller) {
        for (let chunk of ["a", "b", "c"]) {
            controller.enqueue(chunk);
        }
        controller.close();
    }
});
const [leftStream, rightStream] = readable.tee();

(async () => {
    const reader = leftStream.getReader();
    // Cancel the left branch immediately.
    console.log("left cancel start");
    let cancelPromise = reader.cancel();
    // The left branch never finishes cancelling.
    let cancelResult = await cancelPromise;
    console.log("left cancel done", cancelResult); // never reached
})();

(async () => {
    const reader = rightStream.getReader();
    // Read all chunks from the right branch.
    while (true) {
        const {value, done} = await reader.read();
        console.log("right read:", {value, done});
        if (done) break;
    }
    // The right branch has read all the chunks.
    let closedResult = await reader.closed;
    console.log("right closed", closedResult);
})();
```

## Expected results
The left branch should finish cancelling once all chunks have been read from the original stream (as a result of the right branch pulling them).

## Actual results
The left branch never finishes cancelling.

## Investigation
Currently, [the specification for ReadableStreamTee](https://streams.spec.whatwg.org/#readable-stream-tee) only resolves *cancelPromise* when **both** branches are cancelled. However, I believe this needs to happen when either both branches are cancelled, **or** when the last chunk has been pulled (i.e. during the "If done is true" step of *pullAlgorithm*).

MattiasBuelens:
Fixes #1038.

   * [Chromium](https://github.com/whatwg/streams/pull/1039#issuecomment-664739091)
   * In a private email, @youennf said this sounds good
   * web-platform-tests/wpt#24124
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1039.html" title="Last updated on Aug 5, 2020, 10:21 PM UTC (5c7ec3a)">Preview</a> | <a href="https://whatpr.org/streams/1039/d6ea924...5c7ec3a.html" title="Last updated on Aug 5, 2020, 10:21 PM UTC (5c7ec3a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2196,6 +2196,7 @@ create them does not matter.
        [$ReadableStreamDefaultControllerClose$](|branch1|.\[[controller]]).
     1. If |canceled2| is false, perform !
        [$ReadableStreamDefaultControllerClose$](|branch2|.\[[controller]]).
+    1. [=Resolve=] |cancelPromise| with undefined.
 
    : [=read request/error steps=]
    ::
--- a/index.bs
+++ b/index.bs
@@ -2227,6 +2227,7 @@ create them does not matter.
  1. [=Upon rejection=] of |reader|.\[[closedPromise]] with reason |r|,
   1. Perform ! [$ReadableStreamDefaultControllerError$](|branch1|.\[[controller]], |r|).
   1. Perform ! [$ReadableStreamDefaultControllerError$](|branch2|.\[[controller]], |r|).
+  1. [=Resolve=] |cancelPromise| with undefined.
  1. Return « |branch1|, |branch2| ».
 </div>
 
--- a/reference-implementation/lib/abstract-ops/readable-streams.js
+++ b/reference-implementation/lib/abstract-ops/readable-streams.js
@@ -331,10 +331,7 @@ function ReadableStreamTee(stream, cloneForBranch2) {
   let branch1;
   let branch2;
 
-  let resolveCancelPromise;
-  const cancelPromise = new Promise(resolve => {
-    resolveCancelPromise = resolve;
-  });
+  const cancelPromise = newPromise();
 
   function pullAlgorithm() {
     if (reading === true) {
@@ -376,7 +373,7 @@ function ReadableStreamTee(stream, cloneForBranch2) {
         if (canceled2 === false) {
           ReadableStreamDefaultControllerClose(branch2._controller);
         }
-        resolveCancelPromise(undefined);
+        resolvePromise(cancelPromise, undefined);
       },
       errorSteps: () => {
         reading = false;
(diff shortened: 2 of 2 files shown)


MattiasBuelens:
Keeping this as a draft for now.

I think that a similar issue could occur when the original stream **errors** while one branch has started cancelling. I suspect that would also keep the cancel promise pending indefinitely, but I first need to write a test to prove it.

index.bs
```diff
@@ -2089,6 +2090,7 @@ create them does not matter.
      |r|).
   1. Perform ! [$ReadableStreamDefaultControllerError$](|branch2|.\[[readableStreamController]],
      |r|).
+  1. [=Resolve=] |cancelPromise| with undefined.
```
MattiasBuelens, line 2093: Need some input on this: when the original stream errors while one of the two branches is canceling, should we *resolve* or *reject* the cancel promise? 🤔 

For a regular stream, when you call `controller.error(e)` while canceling, the error does not affect the promise returned by `cancel()` in any way. (Maybe it should? But that's a different discussion.) [ReadableStreamCancel](https://streams.spec.whatwg.org/#readable-stream-cancel) immediately sets the stream's state to `"closed"`, and [ReadableStreamDefaultControllerError](https://streams.spec.whatwg.org/#readable-stream-default-controller-error) does nothing if the state is not `"readable"`. The only thing that *can* affect the promise returned by `cancel()` is... well, the *sourceCancelPromise* returned by `source.cancel()`. 😛

For a tee'd stream, I'd say that `controller.error(e)` should at least let any branch finish canceling. The original stream ends up in a terminal `"errored"` state, similar to how `controller.close()` puts it in a terminal `"closed"` state. So I *think* resolving the cancel promise should be acceptable...?

domenic: > So I think resolving the cancel promise should be acceptable...?

I agree. It seems good to be symmetric with

> For a regular stream, when you call controller.error(e) while canceling, the error does not affect the promise returned by cancel() in any way.

Separately, we may indeed want to revisit that, but keeping them symmetric for now makes sense.

MattiasBuelens: All right then. 🙂

MattiasBuelens:
I think this is ready for review.

There were some other issues with WPT which needed to be fixed first. Now that those are out of the way, this PR is finally green again. 😄

ricea:
Yes, speaking for Google Chrome, we want this fix.

domenic:
@MattiasBuelens would you mind rebasing? @youennf confirmed he's good with this fix, so we can merge once that's done.

domenic:
I filed bugs for WebKit and Gecko, and updated the OP with links to them. For Chromium we'll automatically get a bug filed by the web platform tests importer, and I know @ricea has been pretty good about triaging those.

PR #1040, opened 2020-06-12
Meta: run reference-implementation tests on GitHub Actions

foolip:
This makes the .travis.yml like the other repos to make changes from
https://github.com/whatwg/spec-factory/pull/14 smaller.


foolip:
Sweet, it seems to have worked on the first try :)

foolip:
As requested by @domenic I'll roll this into a single PR moving everything to GitHub Actions.

PR #1041, opened 2020-06-12
Meta: migrate to GitHub Actions

foolip:
Part of https://github.com/whatwg/meta/issues/173.


foolip:
The first commit from this is using https://github.com/whatwg/spec-factory/pull/16.

foolip:
I'll change the branch protection rules now, then force push this to see it working, then merge and see test and deploy succeed on master.

foolip:
Success: https://streams.spec.whatwg.org/
Tests on master too: https://github.com/whatwg/streams/runs/768610769

However, I don't understand why the resulting commit is what it is:
```
commit f24f6fcd176689d0d6bfa4733771d8ecd86e3584
Author: Philip Jägenstedt <philip@foolip.org>
Date:   Sat Jun 13 20:38:09 2020 +0200

    Merge pull request #1041 from whatwg/github-actions
    
    Meta: migrate to GitHub Actions
```

It's not a merge commit (phew) but how did this happen? I thought I used the rebase option to preserve both commits. When I merged, I also briefly saw a message about master having changed and the merge failing, before it succeeds.

Seems like something a bit strange happened on the GitHub side here, and the history isn't pretty as a result.

domenic:
Hmm. Well, I'd be ok force-pushing to fix things up, unless you think that's likely to backfire.

foolip:
OK, I've done that, had to disable branch protection temporarily. I only know of two bad things that can happen:
- People having fetched in the intervening time and not knowing how to recover (not super likely on a Saturday).
- Tooling that's continuously tracking the master branch and doing something, but aren't prepared for a history rewrite. I've done this sort of thing for other repos, but I don't know why anyone would for a repo like this.

foolip:
I emailed support@github.com:

> I'd like to report a potential bug that I came across in https://github.com/whatwg/streams/pull/1041. When I merged this, I think I used "rebase and merge", but possible it was "squash and merge". It wasn't the default option (a real merge commit) because this repo disables that.
>
> As I merged it, I briefly saw an error about the merge failing because the branch had changed. I don't recall the exact wording. That went away and the PR was then merged.
>
> However, the resulting commit was something I've never seen before: https://github.com/whatwg/streams/pull/1041#issuecomment-643662026
>
> This wasn't an actual merge commit, it had just one parent. This wouldn't have resulted from either of the ways of merging that are enabled in this repo.
I hope that you can check if something strange happened in GitHub internals, and happy to provide any help debugging I can.

Issue #1036, opened 2020-05-19
Editorial: use dictionaries for { value, done } and eliminate "forAuthorCode"?


domenic:
After #1035 lands, I think it might be possible to go further.

In particular, I'm hoping to clean up the return type of `read()` to be a dictionary with `value` and `done` members.

But while looking into this, I'm noodling on ways to clean up the whole "forAuthorCode" business. Is there a way we can keep the internals of the stream agnostic to the JS object format, and only convert the chunks/end-of-stream signals into dictionaries (and then via Web IDL, JS objects) at the end?

The difficult part of this is that the internals of the streams seem bound up in the promise machinery, and promises are fundamentally resolved with JS values. Maybe we could resolve those promises with JS values for the chunks, and have a special JS value representing end-of-stream, instead of having to create prototype-pollution-prone `{ value, done }` objects and then potentially null out their prototype?

I think it's also worth figuring out what the best interface is for other specifications. If we could avoid them having to deal with raw promises for `{ value, done }` objects, that would be a plus.

ricea:
JS values for chunks would also be vulnerable to prototype pollution.

When I thought about this before, the only solution I came up with was to have something almost but not quite the same as a Promise to use internally.

domenic:
Oh, you're totally right. Dang.

Streams is in a peculiar position where it is using promises for internal flow control (throughout the spec ecosystem) as well as externally exposed to web developers. Everything else in the ecosystem only uses promises at the developer boundary, and uses "async algorithms" of some sort internally. Those async algorithms are sometimes informal, and sometimes formal (using mechanisms like "in parallel", "queue a task", etc.). https://github.com/whatwg/infra/issues/181 touches on this more.

So I think if we were to do anything here it would be best to make streams behave more like the rest of the ecosystem. In particular, I think

> the only solution I came up with was to have something almost but not quite the same as a Promise to use internally

is right, but I wonder if we'd be able to do that surgically, in only the places that handle chunks. I.e., keep using promises for closed/ready/the return values of the underlying X methods/etc. But for ReadableStreamDefaultReaderRead / [[PullSteps]] / read(), do something a bit different, passing the raw chunk around (not in promise form), until finally wrapping it up right before returning from read().

If this is surgical enough, then perhaps we could just use "spec callbacks". Callbacks are gross and messy compared to promises, but if you only go through one or two algorithms using them, it might not be so bad.

ricea:
[...]

PR #1045, opened 2020-06-22
Editorial: remove forAuthorCode and use dictionaries for read() return value

domenic:
This is a start at tackling #1036. The intention is to have no normative behavior changes. However, note that I have not introduced artificial microtasks, so there are probably be some normative timing changes. If so then we should probably fix that, unless for some reason we think this would be an improvement.

Right now I've just done the reference implementation to give people an idea of what it looks like. If we can get all the tests passing and the shape of the solution hammered out, then I can port it to the spec text.

There are two kinds of failing tests currently:

- `readable-streams/tee.any.html`: "ReadableStream teeing: errors in the source should propagate to both branches" fails.

- `readable-byte-streams/general.any.html` craps out due to a timeout in "ReadableStream with byte source: A stream must be errored if close()-d before fulfilling read(view) with Uint16Array", and then the harness doesn't continue with the rest of the tests, so there are unknown failures there.

Both of these could be due to something stupid I missed, or due to the timing changes. I'll investigate tomorrow.

In the meantime, I'd appreciate thoughts on how this looks to folks, and whether it's a promising direction. For example, does the `(chunkSteps, doneSteps, errorSteps)` tuple seem like the right way to go? For byte streams especially, I had to thread them through more places than I'd hoped; is that OK, or does it make this less worth pursuing?

***
<a href="https://whatpr.org/streams/1045.html" title="Last updated on Jun 26, 2020, 5:15 PM UTC (3761fc4)">Preview</a> | <a href="https://whatpr.org/streams/1045/cf19e8d...3761fc4.html" title="Last updated on Jun 26, 2020, 5:15 PM UTC (3761fc4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1016,16 +1016,16 @@ following table:
    [=chunks=] sooner than they are available
 </table>
 
-A <dfn>read request</dfn> is a [=struct=] containing three algorithms to perform in reaction to
-filling the [=readable stream=]'s [=internal queue=] or changing its state. It has the following
+A <dfn export>read request</dfn> is a [=struct=] containing three algorithms to perform in reaction
+to filling the [=readable stream=]'s [=internal queue=] or changing its state. It has the following
 [=struct/items=]:
 
-: <dfn for="read request">chunk steps</dfn>
+: <dfn export for="read request">chunk steps</dfn>
 :: An algorithm taking a [=chunk=], called when a chunk is available for reading
-: <dfn for="read request">close steps</dfn>
+: <dfn export for="read request">close steps</dfn>
 :: An algorithm taking no arguments, called when no [=chunks=] are available because the stream is
    closed
-: <dfn for="read request">error steps</dfn>
+: <dfn export for="read request">error steps</dfn>
 :: An algorithm taking a JavaScript value, called when no [=chunks=] are available because the
    stream is errored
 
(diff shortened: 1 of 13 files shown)


ricea:
This is better than I expected. This pretty closely reflects what we'll need to do to optimise pipeTo() anyway, so having the timing match up with the standard is useful.

I need to think a bit more about the (chunkSteps, doneSteps, errorSteps) tuple.

I think this is worth pursuing, and I'm in favour of it provided we can make it work.

domenic:
This now passes all tests. I inserted a manual microtask for teeing, which I think is reasonable. The tests were not able to detect any other instances of timing changes, and I believe that they're exhaustive. The potentially affected instances of ReadableStreamDefaultReaderRead are:

- Teeing, where the timing change matters (as seen here)
- Piping, where the timing change does not matter, since everything observable is intentionally hidden
- Public APIs, where the result gets wrapped into a promise so there are no timing changes

(ReadableStreamBYOBReaderRead only has public API callers, although in theory piping could use it too.)

I'm happy to port this over to the spec text, once folks have had a chance to make any editorial suggestions.

MattiasBuelens:
Looks good to me! 👍 

I especially like the separation between `chunkSteps` and `doneSteps`. I find [the new version](https://github.com/whatwg/streams/blob/6f6a064ec51865d0f7119169a353c2c3b3c00e1b/reference-implementation/lib/ReadableStream-impl.js#L131-L135) of `asyncIteratorNext` to be much more understandable than [the old one](https://github.com/whatwg/streams/blob/c20f174396617787e1176e11796fc5039088a5ab/reference-implementation/lib/ReadableStream-impl.js#L131-L144).

Perhaps we could use a ["struct"](https://infra.spec.whatwg.org/#structs) or a ["tuple"](https://infra.spec.whatwg.org/#tuple) in the spec text to represent all three steps, so we can thread them as a single argument through the various operations? Or would that make the reference implementation too different from the spec text?

domenic:
A struct or tuple seems very doable, although it's a tradeoff. It makes passing them through algorithms easier, but requires a bit more work to call them (e.g., it'd be something like "Perform readRequest.[[Steps]]'s done steps" instead of "Perofrm readRequest.[[doneSteps]]").

MattiasBuelens:
This might be a dumb question, but wouldn't `[[readRequests]]` and `[[readIntoRequests]]` just become a list of tuples? I don't understand why the steps would need to be nested inside an internal slot on a read request, instead of just *being* the read request. Something like "Perform *readRequest*'s done steps" should be sufficient, no?

domenic:
Oh, sure, that's true. Then we'd need to manually copy the three items over from the (chunkSteps, doneSteps, errorSteps) tuple into the read request tuple, which would slightly decrease the benefit of keeping them together, but we're already doing that anyway with them as separate parameters. So I think it would indeed be a strict improvement.

ricea:
Hmmm... having read results available synchronously is a potential footgun. If code is written naively it can run into reentrancy issues. But since AFAIK Fetch is the only spec that actually reads from a stream, the potential damage is limited.

lgtm, please write spec text when you are ready.

domenic:
Alright, this is good to go. I should probably write a Fetch patch too and hold off on merging this until that is ready.

(In general I want to work on making streams better for other specs to interface with, so that'll potentially be a bit of wasted work, but we'll see.)

PR #1047, opened 2020-06-24
Editorial: use structs instead of records

domenic:
This series of commits (meant to be squashed, but separated out for easier reviewing) moves the spec from using ES-262 [records](https://tc39.es/ecma262/#sec-list-and-record-specification-type) to Infra [structs](https://infra.spec.whatwg.org/#struct).

The main advantage here is that it lets us define the structs and their items in more detail, so that the reader has a better chance of knowing what they mean. Additionally, we can easily cross-link to the item definitions from their usage sites.

It also has the minor advantage of moving us more toward other web specifications, and reducing the problem posed by #825.

In the process I made the following editorial changes that went beyond fancy search-and-replace. (These are the ones that show up as changed in the reference implementation.)

- Instead of representing a writable stream's internal queue as containing chunks that were either the string "close" or a Record{ [[chunk]]: _jsValue_ }}, I created a "special close value" which can be put alongside the chunks directly.

- I renamed one of the items for "pending pull-into descriptors" from "ctor" to "view constructor".

---

I'd mildly prefer to merge this before continuing work on #1045, since per the latest thinking there we'll bundle the chunkSteps/doneSteps/errorSteps into a struct, and it'd be good to stay consistent.

***
<a href="https://whatpr.org/streams/1047.html" title="Last updated on Jun 25, 2020, 3:18 PM UTC (fb4e77f)">Preview</a> | <a href="https://whatpr.org/streams/1047/2b51f36...fb4e77f.html" title="Last updated on Jun 25, 2020, 3:18 PM UTC (fb4e77f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1502,7 +1502,7 @@ following table:
    calls
   <tr>
    <td>\[[pendingPullIntos]]
-   <td class="non-normative">A [=list=] of descriptors representing pending BYOB pull requests
+   <td class="non-normative">A [=list=] of [=pull-into descriptors=]
   <tr>
    <td>\[[queue]]
    <td class="non-normative">A [=list=] representing the stream's internal queue of [=chunks=]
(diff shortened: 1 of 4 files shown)


MattiasBuelens:
Slightly off-topic: it looks like the PR previews are broken in recent PRs? Previously, there'd be a `pr-preview` bot adding "Preview" and "Diff" links to the bottom of the PR description, but I don't see those anymore since #1041. 😕 

Is this intentional, or should I file a bug somewhere? 😛

index.bs
```diff
@@ -1656,9 +1682,11 @@ counterparts for default controllers, as discussed in [[#rs-abstract-ops-used-by
  1. If |autoAllocateChunkSize| is not undefined,
   1. Let |buffer| be [$Construct$]({{%ArrayBuffer%}}, « |autoAllocateChunkSize| »).
```
MattiasBuelens, line 1699: Unrelated to this PR, but it looks like we're missing the *"...and interpreting the result as a completion record"* here.

domenic: Well, generally we always treat the result as a completion record, unless it's unwrapped with ! or ?

domenic:
PR preview being broken is https://github.com/tobie/pr-preview/issues/73 :(

index.bs
MattiasBuelens: Oh woops, you're right. I missed that.

annevk:
FYI: PR Preview should be back. If you edit OP of any affected PR it'll be updated shortly with your favorite links.

index.bs
```diff
@@ -3790,6 +3860,10 @@ the following table:
    write), which writes data to the [=underlying sink=]
 </table>
 
+The <dfn>special close value</dfn> is a unique value enqueued into \[[queue]], in lieu of a
+[=chunk=], to signal that the stream is closed. It only used internally, and is never exposed to web
```
MattiasBuelens, line 3864:
```suggestion
[=chunk=], to signal that the stream is closed. It is only used internally, and is never exposed to web
```

Issue #1048, opened 2020-06-25
Editorial: move from internal slots to associated concepts?


MattiasBuelens:
In #1047, we migrated from ES-style "records" to web-style "structs". That got me thinking: should we migrate from "internal slots" to "associated concepts"? I'm not sure what the precise terminology is, but for example a `Request` class in fetch has an ["associated request"](https://fetch.spec.whatwg.org/#concept-request-request), ["associated headers"](https://fetch.spec.whatwg.org/#request-headers), etc. For streams, a `ReadableStream` would have an "associated disturbed flag", an "associated readable stream controller", etc.

This would allow cross-linking between concept definitions and their usages, and eliminate the problem from #825 entirely. We could also get rid of [this note](https://streams.spec.whatwg.org/commit-snapshots/2b51f36d9e3bf9c451dcbd1740131853faffe23f/#conventions) about our "foreign JavaScript specification conventions". 🙂

I suppose this will require a much bigger change than #1047, since we use internal slots much more than we use records.

ricea:
I find the "internal slot" style easier to read. With the "associated concept" style you end with sentences like

> If this’s do not flush is false, then set this’s decoder to a new decoder for this’s encoding, this’s stream to a new stream, and this’s BOM seen to false.

which you have to carefully parse.

I was under the impression that other WHATWG specs were going to move to the internal slot style, but I can't find any reference to it now.

domenic:
I generally preferred the associated-concept style, but I'm not sure it's worth the substantial effort it would take to move. The better cross-linking would be nice indeed, although we could also accomplish that with internal slots.

> We could also get rid of this note about our "foreign JavaScript specification conventions". 🙂

The use of completion values, with ?/!, would remain.

PR #1050, opened 2020-07-02
Editorial: add cross-links to internal slots

MattiasBuelens:
This is a quick experiment to see what would be needed to have cross-links on internal slots. For now, I've only done `ReadableStream`, but I'm pretty confident we can extend this to other classes. I also have not yet taken the line length into account, so some lines may still need wrapping.

The change is mostly mechanical: e.g. find all usages of `\[[disturbed]]` and replace them with `[=ReadableStream/[[disturbed]]=]`. For usages of `[[state]]` and `[[storedError]]`, care must be taken to check whether the usage references the slot on `ReadableStream` or on `WritableStream`.

I'm not sure if this is the most compact way to do it? A link like `[=ReadableStream/[[readableStreamController]]=]` is quite long, but I don't know if there's a way to write only `[=[[readableStreamController]]=]` and have Bikeshed figure out that it needs to use the `ReadableStream` namespace. 🤷 

See #1048.

***
<a href="https://whatpr.org/streams/1050.html" title="Last updated on Aug 13, 2020, 10:48 PM UTC (2ace378)">Preview</a> | <a href="https://whatpr.org/streams/1050/6cd5e81...2ace378.html" title="Last updated on Aug 13, 2020, 10:48 PM UTC (2ace378)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2547,7 +2547,7 @@ The following abstract operations support the implementation of the
 
  1. Let |stream| be |controller|.\[[stream]].
  1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|controller|) is false, return false.
- 1. If |controller|.\[[started]] is false, return false.
+ 1. If |controller|.[=ReadableStreamDefaultController/[[started]]=] is false, return false.
  1. If ! [$IsReadableStreamLocked$](|stream|) is true and !
     [$ReadableStreamGetNumReadRequests$](|stream|) > 0, return true.
  1. Let |desiredSize| be ! [$ReadableStreamDefaultControllerGetDesiredSize$](|controller|).
@@ -2691,7 +2691,7 @@ The following abstract operations support the implementation of the
  1. Assert: |stream|.[=ReadableStream/[[controller]]=] is undefined.
  1. Set |controller|.\[[stream]] to |stream|.
  1. Perform ! [$ResetQueue$](|controller|).
- 1. Set |controller|.\[[started]], |controller|.[=ReadableStreamDefaultController/[[closeRequested]]=], |controller|.[=ReadableStreamDefaultController/[[pullAgain]]=], and
+ 1. Set |controller|.[=ReadableStreamDefaultController/[[started]]=], |controller|.[=ReadableStreamDefaultController/[[closeRequested]]=], |controller|.[=ReadableStreamDefaultController/[[pullAgain]]=], and
     |controller|.[=ReadableStreamDefaultController/[[pulling]]=] to false.
  1. Set |controller|.\[[strategySizeAlgorithm]] to |sizeAlgorithm| and |controller|.\[[strategyHWM]]
     to |highWaterMark|.
@@ -2701,7 +2701,7 @@ The following abstract operations support the implementation of the
  1. Let |startResult| be the result of performing |startAlgorithm|. (This might throw an exception.)
  1. Let |startPromise| be [=a promise resolved with=] |startResult|.
  1. [=Upon fulfillment=]  of |startPromise|,
-  1. Set |controller|.\[[started]] to true.
+  1. Set |controller|.[=ReadableStreamDefaultController/[[started]]=] to true.
   1. Assert: |controller|.[=ReadableStreamDefaultController/[[pulling]]=] is false.
   1. Assert: |controller|.[=ReadableStreamDefaultController/[[pullAgain]]=] is false.
   1. Perform ! [$ReadableStreamDefaultControllerCallPullIfNeeded$](|controller|).
@@ -3162,7 +3162,7 @@ The following abstract operations support the implementation of the
  1. Let |stream| be |controller|.\[[stream]].
  1. If |stream|.[=ReadableStream/[[state]]=] is not "`readable`", return false.
  1. If |controller|.[=ReadableByteStreamController/[[closeRequested]]=] is true, return false.
- 1. If |controller|.\[[started]] is false, return false.
+ 1. If |controller|.[=ReadableByteStreamController/[[started]]=] is false, return false.
  1. If ! [$ReadableStreamHasDefaultReader$](|stream|) is true and !
     [$ReadableStreamGetNumReadRequests$](|stream|) > 0, return true.
  1. If ! [$ReadableStreamHasBYOBReader$](|stream|) is true and !
@@ -3187,7 +3187,7 @@ The following abstract operations support the implementation of the
  1. Set |controller|.[=ReadableByteStreamController/[[pullAgain]]=] and |controller|.[=ReadableByteStreamController/[[pulling]]=] to false.
  1. Set |controller|.\[[byobRequest]] to null.
  1. Perform ! [$ResetQueue$](|controller|).
- 1. Set |controller|.[=ReadableByteStreamController/[[closeRequested]]=] and |controller|.\[[started]] to false.
+ 1. Set |controller|.[=ReadableByteStreamController/[[closeRequested]]=] and |controller|.[=ReadableByteStreamController/[[started]]=] to false.
  1. Set |controller|.\[[strategyHWM]] to |highWaterMark|.
  1. Set |controller|.[=ReadableByteStreamController/[[pullAlgorithm]]=] to |pullAlgorithm|.
  1. Set |controller|.[=ReadableByteStreamController/[[cancelAlgorithm]]=] to |cancelAlgorithm|.
@@ -3197,7 +3197,7 @@ The following abstract operations support the implementation of the
  1. Let |startResult| be the result of performing |startAlgorithm|.
  1. Let |startPromise| be [=a promise resolved with=] |startResult|.
  1. [=Upon fulfillment=]  of |startPromise|,
-   1. Set |controller|.\[[started]] to true.
+   1. Set |controller|.[=ReadableByteStreamController/[[started]]=] to true.
    1. Assert: |controller|.[=ReadableByteStreamController/[[pulling]]=] is false.
    1. Assert: |controller|.[=ReadableByteStreamController/[[pullAgain]]=] is false.
    1. Perform ! [$ReadableByteStreamControllerCallPullIfNeeded$](|controller|).
--- a/index.bs
+++ b/index.bs
@@ -5647,14 +5647,14 @@ value given in the constructor.
  The <dfn id="cqs-constructor" constructor for="CountQueuingStrategy"
  lt="CountQueuingStrategy(init)">new CountQueuingStrategy(|init|)</dfn> constructor steps are:
 
- 1. Set [=this=].\[[highWaterMark]] to |init|["{{QueuingStrategyInit/highWaterMark}}"].
+ 1. Set [=this=].[=CountQueuingStrategy/[[highWaterMark]]=] to |init|["{{QueuingStrategyInit/highWaterMark}}"].
 </div>
 
 <div algorithm>
  The <dfn id="cqs-high-water-mark" attribute for="CountQueuingStrategy">highWaterMark</dfn>
  getter steps are:
 
- 1. Return [=this=].\[[highWaterMark]].
+ 1. Return [=this=].[=CountQueuingStrategy/[[highWaterMark]]=].
 </div>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -4480,7 +4480,7 @@ The following abstract operations support the implementation of the
  1. Set |stream|.[=WritableStream/[[controller]]=] to |controller|.
  1. Perform ! [$ResetQueue$](|controller|).
  1. Set |controller|.[=WritableStreamDefaultController/[[started]]=] to false.
- 1. Set |controller|.\[[strategySizeAlgorithm]] to |sizeAlgorithm|.
+ 1. Set |controller|.[=WritableStreamDefaultController/[[strategySizeAlgorithm]]=] to |sizeAlgorithm|.
  1. Set |controller|.[=WritableStreamDefaultController/[[strategyHWM]]=] to |highWaterMark|.
  1. Set |controller|.\[[writeAlgorithm]] to |writeAlgorithm|.
  1. Set |controller|.[=WritableStreamDefaultController/[[closeAlgorithm]]=] to |closeAlgorithm|.
@@ -4568,7 +4568,7 @@ The following abstract operations support the implementation of the
  1. Set |controller|.\[[writeAlgorithm]] to undefined.
  1. Set |controller|.[=WritableStreamDefaultController/[[closeAlgorithm]]=] to undefined.
  1. Set |controller|.[=WritableStreamDefaultController/[[abortAlgorithm]]=] to undefined.
- 1. Set |controller|.\[[strategySizeAlgorithm]] to undefined.
+ 1. Set |controller|.[=WritableStreamDefaultController/[[strategySizeAlgorithm]]=] to undefined.
 
  <p class="note">This algorithm will be performed multiple times in some edge cases. After the first
  time it will do nothing.
@@ -4617,7 +4617,7 @@ The following abstract operations support the implementation of the
  id="writable-stream-default-controller-get-chunk-size">WritableStreamDefaultControllerGetChunkSize(|controller|,
  |chunk|)</dfn> performs the following steps:
 
- 1. Let |returnValue| be the result of performing |controller|.\[[strategySizeAlgorithm]], passing
+ 1. Let |returnValue| be the result of performing |controller|.[=WritableStreamDefaultController/[[strategySizeAlgorithm]]=], passing
     in |chunk|, and interpreting the result as a [=completion record=].
  1. If |returnValue| is an abrupt completion,
   1. Perform ! [$WritableStreamDefaultControllerErrorIfNeeded$](|controller|,
--- a/index.bs
+++ b/index.bs
@@ -4793,7 +4793,7 @@ table:
   <tr>
    <td><dfn>\[[backpressureChangePromise]]</dfn>
    <td class="non-normative">A promise which is fulfilled and replaced every time the value of
-   \[[backpressure]] changes
+   [=TransformStream/[[backpressure]]=] changes
   <tr>
    <td><dfn>\[[controller]]</dfn>
    <td class="non-normative">A {{TransformStreamDefaultController}} created with the ability to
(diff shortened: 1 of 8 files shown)


index.bs
```diff
@@ -501,31 +501,31 @@ dictionary StreamPipeOptions {
...
-<table>
+<table dfn-for="ReadableStream">
  <thead>
   <tr>
    <th>Internal Slot
    <th>Description (<em>non-normative</em>)
  <tbody>
   <tr>
-   <td>\[[disturbed]]
+   <td><dfn id="rs-slot-disturbed" lt="[[disturbed]]">\[[disturbed]]</dfn>
```
domenic, line 511: We should probably omit the `id=""`s. I regret manually curating them for abstract operations, and I don't think we should set ourselves up for that maintenance burden going forward.

domenic, line 511: The fact that `lt=""` is required here seems like a Bikeshed bug (maybe it's not doing the escaping in the right order). Maybe file an issue on https://github.com/tabatkins/bikeshed?

MattiasBuelens:
> Feel free to throw [154bfb4](https://github.com/whatwg/streams/commit/154bfb4edada8f7004c621c89c6176a6c96ff075) into #1051 if you want to morph that into a general "editorial fixups" PR.

Looks like 154bfb4edada8f7004c621c89c6176a6c96ff075 has already been superseded by 63b4407f424d6ad042852b3d225b7c0ffa116b2d. 😄 I'll get rid of that commit next time I rebase this PR.

> The alternative I can think of is just assuming internal slot names are unique, and then removing the `dfn-for=""`. However, we have a few collisions these days (since we are no longer using internal slots for brand checking), and I'd like to introduce more going forward, e.g. `[[readableStreamController]]` -> `[[controller]]`. So that would only work in some cases, and we'd need `for=""` for others. I think this is reasonable.

Yeah, there's no way to make all internal slots unique, I'm afraid. Both `ReadableStreamDefaultReader` and `ReadableStreamBYOBReader` have `[[closedPromise]]` and `[[ownerReadableStream]]` slots, and although they serve the same purpose, they're *technically* different slots. Oh, and of course there's also `WritableStreamDefaultWriter` with its own `[[closedPromise]]` slot.

I'm a big fan of renaming things like `[[readableStreamController]]` to just `[[controller]]`. And I agree, we should probably do those renames *first*. 👍

index.bs
MattiasBuelens: Looks like it has already been fixed in Bikeshed, not sure when exactly. I'll remove the `lt=""` again. 🙂

MattiasBuelens:
Ran into an issue. [ReadableStreamClose](https://streams.spec.whatwg.org/commit-snapshots/d6ea924ae7e2f445b913aa1df394029bf73804b7/#readable-stream-close) has the following step:
> 6. Resolve _reader_.[[closedPromise]] with undefined.

However, `[[closedPromise]]` may refer to the internal slot of *either* `ReadableStreamDefaultReader` or `ReadableStreamBYOBReader`. So there's no single link I can put here. 😕

Do we need to introduce an "interface mixin" (like [`TextDecoderCommon` in the Encoding spec](https://encoding.spec.whatwg.org/#interface-mixin-textdecodercommon)) where we can define the internal slots shared between the two reader classes (i.e. `[[closedPromise]]` and `[[stream]]`)?

Or should we add a section "Interfacing with readers" just to define the common internal slots without any Web IDL, like we did with ["Interfacing with controllers"](https://streams.spec.whatwg.org/commit-snapshots/d6ea924ae7e2f445b913aa1df394029bf73804b7/#rs-abstract-ops-used-by-controllers)?

domenic:
Here are some options that I'd be happy with:

1. Don't link for this special case.
2. Branch the spec in a way that reads kinda strange: "If reader is a ReadableStreamDefaultReader, resolve reader.[[closedPromise]] with undefined. If reader is a ReadableStreamBYOBReader, resolve reader.[[closedPromise]] with undefined." With each sentence having a difference link.
3. Add an "Interfacing with readers" section.

Personally, I don't think we should change the IDL for this kind of internal factoring concern... although I can kind of see the appeal of formally expressing that this sub-part of the reader classes is shared, with its own identity. That is, this does feel a bit more like a shared sub-part, than like polymorphism.

Anyway, (3) is a bit weird to me. Do we move the [[stream]] slot there too? Is this all for a single [[closedPromise]] manipulation?

If such polymorphic access to [[closedPromise]] happens multiple times, then I'd feel better about (3). And if polymorphic access to [[stream]] happened, then I'd definitely say we should do (3). But if this is rare, then (1) or (2) seem more reasonable.

domenic:
Hmm, I'll also note that (2) isn't as bad as I've written it, since the algorithm already contains a specific branch for ReadableStreamDefaultReader. So, if this is the only such case, then I'd lean toward (2).

MattiasBuelens:
Unfortunately, it's not *just* this one case. All of the following abstract ops can be called on both types of readers. As such, (1) is not an option:
* `ReadableStreamReaderGenericInitialize` uses both _reader_.[[stream]] and _reader_.[[closedPromise]]
* `ReadableStreamReaderGenericRelease` uses both _reader_.[[stream]] and _reader_.[[closedPromise]]
* `ReadableStreamClose` uses _reader_.[[closedPromise]]
* `ReadableStreamError` uses _reader_.[[closedPromise]]

I don't like (2). There are too many places where we would need branching.

I think the existence of `ReadableStreamReaderGenericInitialize` and `ReadableStreamReaderGenericRelease` already suggest that there should be some concept of a "generic reader". I'm not sure whether this should be expressed in the Web IDL or not. For now, I'm leaning more towards (3).

domenic:
The fact that there are so many cases makes (3) feel pretty reasonable to me. My main concern was creating a whole new section and indirection for only 1-2 call sites, but with 6, I'm happy.

index.bs
```diff
@@ -2401,14 +2405,31 @@ the {{ReadableStream}}'s public API.
...
+ 1. Let |reader| be |stream|.[=ReadableStream/[[reader]]=].
  1. If |reader| is undefined, return false.
  1. If |reader| [=implements=] {{ReadableStreamDefaultReader}}, return true.
  1. Return false.
 </div>
 
 <h4 id="rs-reader-abstract-ops">Readers</h4>
 
+Instances of {{ReadableStreamDefaultReader}} and {{ReadableStreamBYOBReader}} are created with the
+internal slots described in the following table:
```
MattiasBuelens, line 2417: Is it weird to talk about "instances are created with the internal slots described in the following table" here, when actually they also have additional slots (i.e. `[[readRequests]]` and `[[readIntoRequests]]`) defined by the specific class as well?

index.bs
```diff
@@ -2401,14 +2405,31 @@ the {{ReadableStream}}'s public API.
...
  following steps:
 
- 1. Let |reader| be |stream|.\[[reader]].
+ 1. Let |reader| be |stream|.[=ReadableStream/[[reader]]=].
  1. If |reader| is undefined, return false.
  1. If |reader| [=implements=] {{ReadableStreamDefaultReader}}, return true.
  1. Return false.
 </div>
 
 <h4 id="rs-reader-abstract-ops">Readers</h4>
```
MattiasBuelens, line 2414: I added the shared internal slots to the "abstract operations" section for readers. Not sure if this is the best place for it. 🤷

index.bs
```diff
@@ -2401,14 +2405,31 @@ the {{ReadableStream}}'s public API.
...
+
+<table dfn-for="ReadableStreamGenericReader">
+ <thead>
+  <tr>
+   <th>Internal Slot
+   <th>Description (<em>non-normative</em>)
+ <tbody>
+  <tr>
+   <td><dfn>\[[closedPromise]]</dfn>
+   <td class="non-normative">A promise returned by the reader's <code>closed</code> getter
```
MattiasBuelens, line 2427: I can't link to the correct `closed` getter here, since there are once again two versions... 😭

I'm starting to think that adding an IDL interface that defines `.closed` and `.cancel(reason)` would make more sense. 🤔

index.bs
```diff
@@ -1039,23 +1039,25 @@ dictionary ReadableStreamDefaultReadResult {
...
    <th>Internal Slot
    <th>Description (<em>non-normative</em>)
  <tbody>
   <tr>
-   <td>\[[closedPromise]]
+   <td><dfn>\[[closedPromise]]</dfn>
    <td class="non-normative">A promise returned by the reader's
-   {{ReadableStreamDefaultReader/closed}} getter
+   {{ReadableStreamDefaultReader/closed}} getter.
+   This fulfills the [=ReadableStreamGenericReader/[[closedPromise]]=] contract.
```
MattiasBuelens, line 1052: This is probably not a legal construct. 😛

index.bs
```diff
@@ -1131,14 +1133,14 @@ to filling the [=readable stream=]'s [=internal queue=] or changing its state. I
  The <dfn id="default-reader-closed" attribute for="ReadableStreamDefaultReader">closed</dfn>
  getter steps are:
 
- 1. Return [=this=].\[[closedPromise]].
+ 1. Return [=this=].[=ReadableStreamDefaultReader/[[closedPromise]]=].
```
MattiasBuelens, line 1136: I refer to the specific class's `[[closedPromise]]` where possible, and only refer to the generic one when both types of readers could be passed in.

Not sure if this is worth the trouble? Maybe we should remove the `<dfn>\[[closedPromise]]</dfn>` on the specific readers, and only refer to the generic one? But then what should we put in the internal slots of the specific reader classes?

ricea:
> The fact that there are so many cases makes (3) feel pretty reasonable to me. My main concern was creating a whole new section and indirection for only 1-2 call sites, but with 6, I'm happy.

(3) is useful from an implementation perspective, because we need to do that anyway.

ricea:
> Honestly, I think we'd be better off with an actual `interface mixin ReadableStreamGenericReader`. That way, we could move the shared internal slots _and_ the shared `closed` getter and `cancel(reason)` method to that mixin. But I suppose such a change is no longer merely editorial, and would need a separate PR?

My non-binding opinion is that `interface mixin`s count as editorial, since you can't test for them and they don't necessarily change the implementation.

domenic:
Interesting!

Moving to an `interface mixin` is indeed editorial. I can see the appeal. I've also run into the desire to link to a single public API for things like https://whatpr.org/streams/1050.html#dom-underlyingsource-cancel; as you can see I currently use text like "`defaultReader.cancel()` or `byobReader.cancel()`".

However, I think we should game this out a bit more. If we start using mixins for something like this... will we also want to use them for controllers? A lot of internal slots _and_ public APIs are shared between the two readable stream controllers. And heck, maybe more abstract ops would be shared, if we had a common place to share them. (Not sure about that one.)

I also worry a bit about the impact on spec readability of splitting readers out so that you have to read at least 2 sections per reader, instead of 1. This is most stark for the domintro sections, aimed at web developers; would we split those up?

I'll proceed to respond to your questions and do a review based on sticking with (3), but I'm more open to discussing interface mixins now.

index.bs
domenic: There's a few ways to do this. I think it'd be best to delete the [[closedPromise]] and [[stream]] rows from this table, and then add a sentence after the table saying something like "They additionally have the internal slots described in the section XXX".

index.bs
domenic: Yeah, per my above comment, if we're going to collapse them, then I think we'd best collapse them fully.

index.bs
domenic: Seems good to me.

index.bs
domenic: You can just expand it a bit. E.g. "In addition to the ones described in their individual sections, instances..."

index.bs
domenic: If we don't go with interface mixin, I'd solve this by saying "A promise returned by the `defaultReader.closed` or `byobReader.closed` getter"

index.bs
```diff
@@ -1131,14 +1178,14 @@ to filling the [=readable stream=]'s [=internal queue=] or changing its state. I
  The <dfn id="default-reader-closed" attribute for="ReadableStreamDefaultReader">closed</dfn>
```
MattiasBuelens, line 1178: I would have liked to remove the `closed` getter and `cancel(reason)` methods on `ReadableStreamDefaultReader` and `ReadableStreamBYOBReader`, in favor for using the definitions from `ReadableStreamGenericReader`. But apparently that would break the links to MDN, so the CI doesn't let me do that here. 🤷

I could make the steps perform the steps for `ReadableStreamGenericReader.closed` and `.cancel(reason)`, but I'm not sure how to formalize that.

index.bs
```diff
@@ -1007,6 +1007,62 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
  1. Return [=a promise resolved with=] undefined.
 </div>
 
+<h3 id="generic-reader-mixin">The {{ReadableStreamGenericReader}} interface mixin</h3>
+
+The {{ReadableStreamGenericReader}} interface mixin defines common internal slots, getters and
+methods that are shared between {{ReadableStreamDefaultReader}} and {{ReadableStreamBYOBReader}}
+objects.
```
MattiasBuelens, line 1014: Shamelessly stolen from [`TextDecoderCommon`](https://encoding.spec.whatwg.org/#interface-mixin-textdecodercommon). 😁

MattiasBuelens:
> If we start using mixins for something like this... will we also want to use them for controllers? A lot of internal slots and public APIs are shared between the two readable stream controllers. And heck, maybe more abstract ops would be shared, if we had a common place to share them. (Not sure about that one.)

I suppose we could share some code between `SetUpReadableStreamDefaultController()` and `SetUpReadableByteStreamController()`. The properties and methods (`desiredSize`, `close`, `enqueue`, and `error`) all use different abstract ops, and while they do share a similar structure, they call into different ops like `ReadableStreamDefaultControllerShouldCallPull` and `ReadableByteStreamControllerShouldCallPull`. We could try using polymorphism for that, but I'm afraid it would get weird... 😕 

> I also worry a bit about the impact on spec readability of splitting readers out so that you have to read at least 2 sections per reader, instead of 1. This is most stark for the domintro sections, aimed at web developers; would we split those up?

In the Encoding spec, the domintro section about the `encoding` getter is repeated for [`TextDecoder`](https://encoding.spec.whatwg.org/#ref-for-dom-textdecoder-encoding%E2%91%A0) and [`TextDecoderStream`](https://encoding.spec.whatwg.org/#ref-for-dom-textdecoder-encoding%E2%91%A1). The links inside those sections do link back to `TextDecoderCommon.encoding`, though.

I think we could do the same: keep the domintro sections for `ReadableStreamDefaultReader` and `ReadableStreamBYOBReader`, but link to the definitions in `ReadableStreamGenericReader`.

index.bs
domenic: We need to remove them; otherwise the spec is ill-formed. But, you can use `oldids=""` to make links go to the right place.

domenic:
This is looking good to me; I'm OK with the interface mixin approach. We gotta delete the double-definitions, and it'd also be good to update https://whatpr.org/streams/1050.html#dom-underlyingsource-cancel to link to the new, consolidated `cancel()` definition. 

And then we need to update the reference implementation too. You can leave the .js files as-is and just update the .webidl as a first pass, but ideally we should match the spec and have some mixed in implementations like the spec does. See https://github.com/jsdom/jsdom/blob/88e72ef028913e78266b8105493fd6d973c68e38/lib/jsdom/living/nodes/Slotable-impl.js and references to that in jsdom for an example.

index.bs
```diff
@@ -1007,6 +1007,62 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
  1. Return [=a promise resolved with=] undefined.
 </div>
 
+<h3 id="generic-reader-mixin">The {{ReadableStreamGenericReader}} interface mixin</h3>
```
domenic, line 1010: Let's just say "mixin" instead of "interface mixin". And below, instead of "Interface definition", "Mixin definition".

index.bs
```diff
@@ -1007,6 +1007,62 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
...
+interface mixin ReadableStreamGenericReader {
+  readonly attribute Promise<void> closed;
+
+  Promise<void> cancel(optional any reason);
+};
+</xmp>
+
+<h4 id="generic-reader-internal-slots">Internal slots</h4>
+
+Instances of classes implementing the {{ReadableStreamGenericReader}} mixin are created with the
```
domenic, line 1030: implementing -> including

index.bs
MattiasBuelens: > But, you can use `oldids=""` to make links go to the right place.

Thanks, that's *exactly* what I was looking for! 😄

index.bs
```diff
@@ -2304,17 +2310,17 @@ the {{ReadableStream}}'s public API.
...
-  1. [=list/For each=] |readRequest| of |reader|.\[[readRequests]],
+  1. [=list/For each=] |readRequest| of |reader|.[=ReadableStreamDefaultReader/[[readRequests]]=],
    1. Perform |readRequest|'s [=read request/close steps=].
-  1. Set |reader|.\[[readRequests]] to an empty [=list=].
- 1. [=Resolve=] |reader|.\[[closedPromise]] with undefined.
+  1. Set |reader|.[=ReadableStreamDefaultReader/[[readRequests]]=] to an empty [=list=].
+ 1. [=Resolve=] |reader|.[=ReadableStreamGenericReader/[[closedPromise]]=] with undefined.
 
- <p class="note">The case where |stream|.\[[state]] is "`closed`", but |stream|.\[[closeRequested]]
+ <p class="note">The case where |stream|.[=ReadableStream/[[state]]=] is "`closed`", but |stream|.[=ReadableStream/[[controller]]=].\[[closeRequested]]
```
MattiasBuelens, line 2323: Drive-by fix: _stream_.\[[closeRequested]] should have been _stream_.\[[controller]].\[[closeRequested]]. (Fortunately, this note is non-normative! 😅)

MattiasBuelens, line 2323: This is the only place where we refer to `[[closeRequested]]` without a specific type of controller. I think it's a bit overkill to introduce an interface mixin *just* for this non-normative note, so I'm tempted to only link to the internal slot on the default controller, i.e.:
```suggestion
 <p class="note">The case where |stream|.[=ReadableStream/[[state]]=] is "`closed`", but |stream|.[=ReadableStream/[[controller]]=].[=ReadableStreamDefaultController/[[closeRequested]]=]
```

Thoughts?

domenic: That seems reasonable, or leaving it un-linked.

This note seems a little confusing though... the algorithm doesn't mention closeRequested, and it seems kind of redundant with https://streams.spec.whatwg.org/#readable-stream-default-controller-can-close-or-enqueue. So maybe it should just be deleted?

MattiasBuelens: Indeed, it seems redundant. I'll delete it. 🙂

MattiasBuelens:
This is now feature complete. I still need to do one big pass to fix all the line wrapping, but you can already review the changes. (If anyone has tooling suggestions for the line wrapping, let me know. 😅)

> And then we need to update the reference implementation too. You can leave the .js files as-is and just update the .webidl as a first pass, but ideally we should match the spec and have some mixed in implementations like the spec does. See https://github.com/jsdom/jsdom/blob/88e72ef028913e78266b8105493fd6d973c68e38/lib/jsdom/living/nodes/Slotable-impl.js and references to that in jsdom for an example.

Done! 😁

reference-implementation/lib/helpers/miscellaneous.js
```diff
@@ -11,3 +11,14 @@ exports.rethrowAssertionErrorRejection = e => {
     }, 0);
   }
 };
+
+exports.mixin = (target, source) => {
```
MattiasBuelens, line 15: Shamelessly stolen from [jsdom](https://github.com/jsdom/jsdom/blob/5279cfda5fe4d52f04b2eb6a801c98d81f9b55da/lib/jsdom/utils.js#L57-L66). 😁

domenic:
> (If anyone has tooling suggestions for the line wrapping, let me know. 😅)

https://domenic.github.io/rewrapper/ works on pure HTML, but I haven't tested it with the Markdown style we're doing here... maybe too much yak-shaving, but patches welcome.

index.bs
```diff
@@ -2079,24 +2112,26 @@ are even meant to be generally useful by other specifications.
     before performing any reads or writes, since they might lead to immediate shutdown.
   * <strong>Error and close states must be propagated:</strong> the following conditions must be
     applied in order.
-   1. <strong>Errors must be propagated forward:</strong> if |source|.\[[state]] is or becomes
-      "`errored`", then
+   1. <strong>Errors must be propagated forward:</strong>  if |source|.[=ReadableStream/[[state]]=]
```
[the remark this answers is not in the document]
MattiasBuelens: It's not. I think I first tried to split the line before the "if", but then changed my mind. 😛

PR #1051, opened 2020-07-03
Editorial: various fixes

MattiasBuelens:
* Fix typo in Close/Enqueue checks for readable byte streams. The internal slot is called `[[closeRequested]]` instead of `[[closeRequest]]`. Looks like we missed this in #1029. 😅
* Remove redundant `ReadableStreamDefaultControllerCanCloseOrEnqueue` check in `TransformStreamDefaultSinkCloseAlgorithm`.
* ~Add missing link to `ReadableStreamDefaultControllerEnqueue` inside `ReadableStreamTee`.~ Was already fixed in 63b4407f424d6ad042852b3d225b7c0ffa116b2d.
* Fix link IDs for `TransformStream` abstract ops, which were presumable broken in the WebIDL rewrite.

***
<a href="https://whatpr.org/streams/1051.html" title="Last updated on Jul 6, 2020, 8:48 PM UTC (2ef17ea)">Preview</a> | <a href="https://whatpr.org/streams/1051/63b4407...2ef17ea.html" title="Last updated on Jul 6, 2020, 8:48 PM UTC (2ef17ea)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5328,9 +5328,7 @@ side=] of [=transform streams=].
  1. Return the result of [=reacting=] to |flushPromise|:
   1. If |flushPromise| was fulfilled, then:
    1. If |readable|.\[[state]] is "`errored`", throw |readable|.\[[storedError]].
-   1. Let |readableController| be |readable|.\[[readableStreamController]].
-   1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|readableController|) is true,
-      perform ! [$ReadableStreamDefaultControllerClose$](|readableController|).
+   1. Perform ! [$ReadableStreamDefaultControllerClose$](|readable|.\[[readableStreamController]]).
   1. If |flushPromise| was rejected with reason |r|, then:
    1. Perform ! [$TransformStreamError$](|stream|, |r|).
    1. Throw |readable|.\[[storedError]].
--- a/index.bs
+++ b/index.bs
@@ -5215,7 +5215,7 @@ The following abstract operations support the implementaiton of the
 
 <div algorithm>
  <dfn abstract-op lt="TransformStreamDefaultControllerEnqueue"
- id="set-up-transform-stream-default-controller-enqueue"
+ id="transform-stream-default-controller-enqueue"
  export>TransformStreamDefaultControllerEnqueue(|controller|, |chunk|)</dfn> is meant to be called
  by other specifications that wish to enqueue [=chunks=] in the [=readable side=], in the same way
  a developer would enqueue chunks using the stream's associated controller object. Specifications
@@ -5241,7 +5241,7 @@ The following abstract operations support the implementaiton of the
 </div>
 <div algorithm>
  <dfn abstract-op lt="TransformStreamDefaultControllerError"
- id="set-up-transform-stream-default-controller-error"
+ id="transform-stream-default-controller-error"
  export>TransformStreamDefaultControllerError(|controller|, |e|)</dfn> is meant to be called by
  other specifications that wish to move the transform stream to an errored state, in the same way a
  developer would error the stream using the stream's associated controller object. Specifications
@@ -5254,7 +5254,7 @@ The following abstract operations support the implementaiton of the
 
 <div algorithm>
  <dfn abstract-op lt="TransformStreamDefaultControllerPerformTransform"
- id="set-up-transform-stream-default-controller-perform-transform">TransformStreamDefaultControllerPerformTransform(|controller|, |chunk|)</dfn>
+ id="transform-stream-default-controller-perform-transform">TransformStreamDefaultControllerPerformTransform(|controller|, |chunk|)</dfn>
  performs the following steps:
 
  1. Let |transformPromise| be the result of performing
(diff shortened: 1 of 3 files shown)


domenic:
Interesting. The fact that all the tests passed with this typo indicates a lack of test coverage. Would you be up for fixing that too? (Not a requirement for merging.)

index.bs
```diff
@@ -5310,9 +5310,7 @@ side=] of [=transform streams=].
  1. Return the result of [=reacting=] to |flushPromise|:
   1. If |flushPromise| was fulfilled, then:
    1. If |readable|.\[[state]] is "`errored`", throw |readable|.\[[storedError]].
-   1. Let |readableController| be |readable|.\[[readableStreamController]].
-   1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|readableController|) is true,
-      perform ! [$ReadableStreamDefaultControllerClose$](|readableController|).
+   1. Perform ! [$ReadableStreamDefaultControllerClose$](|readable|.\[[readableStreamController]]).
```
MattiasBuelens, line 5313: The `ReadableStreamDefaultControllerCanCloseOrEnqueue` check has become redundant, similar to [this change](https://github.com/whatwg/streams/commit/a3e482279576768a602e660e65ea453f9b19b239#diff-ec9cfa5f3f35ec1f84feb2e59686c34dL4997-R4994) in `TransformStreamDefaultControllerTerminate`.

MattiasBuelens:
> The fact that all the tests passed with this typo indicates a lack of test coverage. Would you be up for fixing that too? (Not a requirement for merging.)

I think the same problems arise as in [the original PR](https://github.com/whatwg/streams/pull/1029#issuecomment-593285919). These changes are inside abstract operations that are not directly accessible through the API. The API will throw a `TypeError` when these checks fail, so it's not possible to reach the abstract op with a problematic state.

You'd need to find some other abstract operation which doesn't perform these checks on its own in advance. For `ReadableStreamDefaultController`, we could look at `ReadableStreamTee` or `TransformStreamDefaultControllerEnqueue`. However, for `ReadableByteStreamController`, the abstract ops are currently **only** used through the API, which always performs these checks. So I'm afraid there's currently no way to really test the changes from this PR. (Unless we would somehow export these abstract ops only for testing purposes, which would be weird...)

MattiasBuelens:
I might be going on a bit of a tangent here, but... Now that `ReadableStreamDefaultControllerEnqueue` no longer *asserts* `CanCloseOrEnqueue()` but instead early-returns, shouldn't we do the same thing for `TransformStreamDefaultControllerEnqueue`? That means:
* Move what is currently [step 3 of `TransformStreamDefaultControllerEnqueue`](https://streams.spec.whatwg.org/commit-snapshots/c63ae9eb7de6069f0e3cf5ea6b0ef68a2c8a3971/#set-up-transform-stream-default-controller-enqueue) into [`TransformStreamDefaultController.enqueue()`](https://streams.spec.whatwg.org/commit-snapshots/c63ae9eb7de6069f0e3cf5ea6b0ef68a2c8a3971/#ts-default-controller-enqueue).
* Add a check + early-return at the start of `TransformStreamDefaultControllerEnqueue`.

(Even more of a tangent: I just noticed that the anchor links for these transform stream abstract ops are weird. For example, `TransformStreamDefaultControllerEnqueue` is linked as `#set-up-transform-stream-default-controller-enqueue`. The `set-up-` bit looks like a copy-paste error? 😅)

ricea:
> I might be going on a bit of a tangent here, but... Now that ReadableStreamDefaultControllerEnqueue no longer asserts CanCloseOrEnqueue() but instead early-returns, shouldn't we do the same thing for TransformStreamDefaultControllerEnqueue?

I think it's less clear-cut for TransformStreamDefaultControllerEnqueue because it is already handling the exceptional case, so there is no danger of state confusion. I don't feel strongly that we need to be consistent here.

domenic:
> The API will throw a TypeError when these checks fail, so it's not possible to reach the abstract op with a problematic state.

In that case, can we change them to asserts?

> shouldn't we do the same thing for TransformStreamDefaultControllerEnqueue? That means:

It looks like this might have a normative effect on SetUpTransformStreamDefaultControllerFromTransformer, where 2.2 would no longer trigger. I'm not sure though if the branch is ever reached though...

> (Even more of a tangent: I just noticed that the anchor links for these transform stream abstract ops are weird. For example, TransformStreamDefaultControllerEnqueue is linked as #set-up-transform-stream-default-controller-enqueue. The set-up- bit looks like a copy-paste error? 😅)

This was presumably broken in the Web IDL rewrite; good catch! https://streams.spec.whatwg.org/commit-snapshots/cfcd303a38158e4ec3837d1028158c88d505c35d/#transform-stream-default-controller-enqueue has the more-correct abstract operation IDs. A fix, perhaps in this PR, would be lovely.

MattiasBuelens:
> In that case, can we change them to asserts?

Uhm, I thought the whole point of #1029 was to change the asserts into early-exits? Now you want to change them back to asserts? I'm not sure I'm following... 😕

> It looks like this might have a normative effect on SetUpTransformStreamDefaultControllerFromTransformer, where 2.2 would no longer trigger. I'm not sure though if the branch is ever reached though...

Hmm, good point, I seem to have missed that usage of `TransformStreamDefaultControllerEnqueue`. That would mean we have a "default" `TransformStream` where the readable side has been cancelled, but the writable side is still trying to send a chunk.

However, even in that case, I think we won't reach `TransformStreamDefaultControllerPerformTransform`: we would have already turned the readable's cancellation into a writable error in `TransformStreamErrorWritableAndUnblockWrite`, so `TransformStreamDefaultSinkWriteAlgorithm` should not get past the *"If state is erroring, throw writable.[[storedError]]."* step.

...I *think*. I'm not confident enough about it to actually change `TransformStreamDefaultControllerEnqueue`. 😅 

> This was presumably broken in the Web IDL rewrite; good catch! https://streams.spec.whatwg.org/commit-snapshots/cfcd303a38158e4ec3837d1028158c88d505c35d/#transform-stream-default-controller-enqueue has the more-correct abstract operation IDs. A fix, perhaps in this PR, would be lovely.

Will do. 🙂

PR #1053, opened 2020-07-08
Support transferable streams (postMessage)

### LINKED ISSUE #276: can streams be transferred via postMessage()?

ricea:
Make readable, writable, and transform streams transferable via
postMessage(stream, [stream]).

The streams themselves must be transferred, but the chunks written or
read from the streams are cloned, not transferred. Support for
transferring chunks will require API changes and is expected to be added
in a future update.

There is no reference implementation of this functionality as jsdom does
not support transferable objects, and so it wouldn't be testable.

Closes #276.

- [X] At least two implementers are interested (and none opposed):
   * [Chrome](https://github.com/whatwg/streams/pull/1053#issuecomment-662114735)
   * In a private email, @youennf expressed support for shipping the current approach
- [X] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/24546
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/894838
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1659025
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=215485

***
<a href="https://whatpr.org/streams/1053.html" title="Last updated on Aug 14, 2020, 2:09 AM UTC (a5d56a9)">Preview</a> | <a href="https://whatpr.org/streams/1053/1975208...a5d56a9.html" title="Last updated on Aug 14, 2020, 2:09 AM UTC (a5d56a9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,6 +40,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: the typed array constructors table; url: #table-49
   text: typed array; url: #sec-typedarray-objects
   text: Number type; url: #sec-ecmascript-language-types-number-type
+  text: %ObjectPrototype%; url: #sec-properties-of-the-object-prototype-object
  type: abstract-op
   text: CloneArrayBuffer; url: #sec-clonearraybuffer
   text: CopyDataBlockBytes; url: #sec-copydatablockbytes
@@ -6061,7 +6062,7 @@ operations are used to implement these "cross-realm transforms".
 
  <p class="note">A JavaScript object is used for transfer to avoid having to duplicate the [=message
  port post message steps=]. The prototype of the object is set to null to avoid interference from
- the global object.</p>
+ [=%ObjectPrototype%=].</p>
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -6053,8 +6053,8 @@ operations are used to implement these "cross-realm transforms".
  the following steps:
 
  1. Let |message| be [$OrdinaryObjectCreate$](null).
- 1. Perform [$CreateDataProperty$](|message|, "`type`", |type|).
- 1. Perform [$CreateDataProperty$](|message|, `"value"`, |value|).
+ 1. Perform ! [$CreateDataProperty$](|message|, "`type`", |type|).
+ 1. Perform ! [$CreateDataProperty$](|message|, `"value"`, |value|).
  1. Let |targetPort| be the port with which |port| is entangled, if any; otherwise let it be null.
  1. Let |options| be «[ `"transfer"` → « » ]».
  1. Run the [=message port post message steps=] providing |targetPort|, |message| and |options|.
--- a/index.bs
+++ b/index.bs
@@ -6053,7 +6053,7 @@ operations are used to implement these "cross-realm transforms".
  1. Perform ! [$CreateDataProperty$](|message|, "`value`", |value|).
  1. Let |targetPort| be the port with which |port| is entangled, if any; otherwise let it be null.
  1. Let |options| be «[ "`transfer`" → « » ]».
- 1. Run the [=message port post message steps=] providing |targetPort|, |message| and |options|.
+ 1. Run the [=message port post message steps=] providing |targetPort|, |message|, and |options|.
 
  <p class="note">A JavaScript object is used for transfer to avoid having to duplicate the [=message
  port post message steps=]. The prototype of the object is set to null to avoid interference from
--- a/index.bs
+++ b/index.bs
@@ -3805,7 +3805,7 @@ as seen for example in [[#example-ws-no-backpressure]].
   <p>Sends a {{WritableStream}} to another frame, window, or worker.
 
   <p>The transferred stream can be used exactly like the original. The original will become
-  [=locked to a writer|lock=] and no longer directly usable.
+  [=locked to a writer|locked=] and no longer directly usable.
  </dd>
 </dl>
 
--- a/index.bs
+++ b/index.bs
@@ -15,6 +15,10 @@ Markup Shorthands: markdown yes
 spec:webidl; type:dfn; text:resolve
 spec:webidl; type:dfn; text:new
 spec:infra; type:dfn; text:list
+spec:html; type:dfn; text:entangle
+spec:html; type:dfn; text:create a new messageport object
+spec:html; type:dfn; text:owner
+spec:html; type:dfn; text:message port post message steps
 </pre>
 
 <pre class="anchors">
@@ -41,11 +45,14 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: CopyDataBlockBytes; url: #sec-copydatablockbytes
   text: CreateArrayFromList; url: #sec-createarrayfromlist
   text: CreateBuiltinFunction; url: #sec-createbuiltinfunction
+  text: CreateDataProperty; url: #sec-createdataproperty
   text: Construct; url: #sec-construct
   text: DetachArrayBuffer; url: #sec-detacharraybuffer
+  text: Get; url: #sec-get-o-p
   text: GetV; url: #sec-getv
   text: IsDetachedBuffer; url: #sec-isdetachedbuffer
   text: IsInteger; url: #sec-isinteger
+  text: OrdinaryObjectCreate; url: #sec-ordinaryobjectcreate
   text: SetFunctionLength; url: #sec-setfunctionlength
   text: SetFunctionName; url: #sec-setfunctionname
   text: Type; url: #sec-ecmascript-data-types-and-values
@@ -451,7 +458,7 @@ by the [=underlying source=] but not yet read by any consumer.
 The Web IDL definition for the {{ReadableStream}} class is given as follows:
 
 <xmp class="idl">
-[Exposed=(Window,Worker,Worklet)]
+[Exposed=(Window,Worker,Worklet), Transferable]
 interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
@@ -507,6 +514,9 @@ table:
    <td class="non-normative">A {{ReadableStreamDefaultController}} or
    {{ReadableByteStreamController}} created with the ability to control the state and queue of this
    stream
+  <tr>
+   <td>\[[detached]]
+   <td class="non-normative">A boolean flag set to true when the stream is transferred
   <tr>
    <td><dfn>\[[disturbed]]</dfn>
    <td class="non-normative">A boolean flag set to true when the stream has been read from or
@@ -744,6 +754,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
   <p>Note that the [=chunks=] seen in each branch will be the same object. If the chunks are not
   immutable, this could allow interference between the two branches.
 </dl>
+
+<div algorithm="ReadableStream transfer steps">
+{{ReadableStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value| and
+|dataHolder|, are:
+
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port1| be that object.
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port2| be that object.
+ 1. [=Entangle=] the |port1| and |port2| objects.
+ 1. If ! [$IsReadableStreamLocked$](|value|) is true, throw a {{TypeError}} exception.
+ 1. Let |writable| be a [=new=] {{WritableStream}}.
+ 1. Perform ! [$SetUpCrossRealmTransformWritable$](|writable|, |port1|).
+ 1. Let |promise| be ! [$ReadableStreamPipeTo$](|value|, |writable|, false, false, false, undefined).
+ 1. Set |promise|.\[[PromiseIsHandled]] to true.
+ 1. Let |messagePortDataHolder| be { \[[Type]]: `"MessagePort"` }.
+ 1. Perform {{MessagePort}}'s transfer steps <!-- TODO(ricea): Link this somehow --> with |port2|
+    and |messagePortDataHolder|.
+ 1. Set |dataHolder|.\[[port]] to |messagePortDataHolder|.
+
+</div>
+
+<div algorithm="ReadableStream transfer receiving steps">
+Their [=transfer-receiving steps=], given |dataHolder| and |value|, are:
+
+ 1. Let |port| be a new {{MessagePort}}.
+ 1. Perform {{MessagePort}}'s transfer-receiving steps with |dataHolder|.\[[port]] and |port|.
+ 1. Perform ! [$SetUpCrossRealmTransformReadable$](|value|, |dataHolder|.\[[port]])
+
+</div>
 
 <div algorithm>
  The <dfn id="rs-constructor" constructor for="ReadableStream" lt="ReadableStream(underlyingSource,
@@ -3457,7 +3497,7 @@ The {{WritableStream}} represents a [=writable stream=].
 The Web IDL definition for the {{WritableStream}} class is given as follows:
 
 <xmp class="idl">
-[Exposed=(Window,Worker,Worklet)]
+[Exposed=(Window,Worker,Worklet), Transferable]
 interface WritableStream {
   constructor(optional object underlyingSink, optional QueuingStrategy strategy = {});
 
@@ -3491,6 +3531,9 @@ table:
   <td><dfn>\[[controller]]</dfn>
   <td class="non-normative">A {{WritableStreamDefaultController}} created with the ability to
   control the state and queue of this stream
+ <tr>
+  <td>\[[detached]]
+  <td class="non-normative">A boolean flag set to true when the stream is transferred
  <tr>
   <td><dfn>\[[inFlightWriteRequest]]</dfn>
   <td class="non-normative">A slot set to the promise for the current in-flight write operation
@@ -3694,6 +3737,38 @@ as seen for example in [[#example-ws-no-backpressure]].
   can ensure nobody else can write at the same time, which would cause the resulting written data
   to be unpredictable and probably useless.
 </dl>
+
+<div algorithm="WritableStream transfer steps">
+{{WritableStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value| and
+|dataHolder|, are:
+
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port1| be that object.
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port2| be that object.
+ 1. [=Entangle=] the |port1| and |port2| objects.
+ 1. If ! [$IsWritableStreamLocked$](|value|) is true, throw a {{TypeError}} exception.
+ 1. Let |readable| be a [=new=] {{ReadableStream}}.
+ 1. Perform ! [$SetUpCrossRealmTransformReadable$](|readable|, |port1|).
+ 1. Let |promise| be ! [$ReadableStreamPipeTo$](|readable|, |value|, false, false, false,
+    undefined).
+ 1. Set |promise|.\[[PromiseIsHandled]] to true.
+ 1. Let |messagePortDataHolder| be { \[[Type]]: `"MessagePort"` }.
+ 1. Perform {{MessagePort}}'s transfer steps
+    <!-- TODO(ricea): Link this to the HTML standard somehow --> with |port2| and
+    |messagePortDataHolder|.
+ 1. Set |dataHolder|.\[[port]] to |messagePortDataHolder|.
+
+</div>
+
+<div algorithm="WritableStream transfer receiving steps">
+Their [=transfer-receiving steps=], given |dataHolder| and |value|, are:
+
+ 1. Let |port| be a new {{MessagePort}}.
+ 1. Perform {{MessagePort}}'s transfer-receiving steps with |dataHolder|.\[[port]] and |port|.
+ 1. Perform ! [$SetUpCrossRealmTransformWritable$](|value|, |port|).
+
+</div>
 
 <div algorithm>
  The <dfn id="ws-constructor" constructor for="WritableStream" lt="WritableStream(underlyingSink,
@@ -4884,7 +4959,7 @@ The {{TransformStream}} class is a concrete instance of the general [=transform 
 The Web IDL definition for the {{TransformStream}} class is given as follows:
 
 <xmp class="idl">
-[Exposed=(Window,Worker,Worklet)]
+[Exposed=(Window,Worker,Worklet), Transferable]
 interface TransformStream {
   constructor(optional object transformer,
               optional QueuingStrategy writableStrategy = {},
@@ -4918,6 +4993,9 @@ table:
    <td><dfn>\[[controller]]</dfn>
    <td class="non-normative">A {{TransformStreamDefaultController}} created with the ability to
    control [=TransformStream/[[readable]]=] and [=TransformStream/[[writable]]=]
+  <tr>
+   <td>\[[detached]]
+   <td class="non-normative">A boolean flag set to true when the stream is transferred
   <tr>
    <td><dfn>\[[readable]]</dfn>
    <td class="non-normative">The {{ReadableStream}} instance controlled by this object
@@ -5044,6 +5122,37 @@ side=], or to terminate or error the stream.
  <dd>
   <p>Returns a {{WritableStream}} representing the [=writable side=] of this transform stream.
 </dl>
+
+<div algorithm="TransformStream transfer steps">
+{{TransformStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value|
+and |dataHolder|, are:
+
+<!-- Find a way to linkify ReadableStream transfer steps and WritableStream transfer steps -->
+
+ 1. If ! [$IsReadableStreamLocked$](|value|.\[[readable]]) is true, throw a {{TypeError}} exception.
+ 1. If ! [$IsWritableStreamLocked$](|value|.\[[writable]]) is true, throw a {{TypeError}} exception.
+ 1. Let |readableDataHolder| be { \[[Type]]: `"ReadableStream"` }.
+ 1. Perform {{ReadableStream}} transfer steps with |value|.\[[readable]] and |readableDataHolder|.
+ 1. Set |dataHolder|.\[[readable]] to |readableDataHolder|.
+ 1. Let |writableDataHolder| be { \[[Type]]: `"WritableStream"` }.
+ 1. Perform {{WritableStream}} transfer steps with |value|.\[[writable]] and |writableDataHolder|.
+ 1. Set |dataHolder|.\[[writable]] to |writableDataHolder|.
+
+</div>
+
+<div algorithm="TransformStream transfer receiving steps">
+Their [=transfer-receiving steps=], given |dataHolder| and |value|, are:
+
+ 1. Perform {{ReadableStream}} transfer-receiving steps given |dataHolder|.\[[readable]] and
+    |value|.\[[readable]].
+ 1. Perform {{WritableStream}} transfer-receiving steps given |dataHolder|.\[[writable]] and
+    |value|.\[[writable]].
+
+<p class="note">These steps intentionally leave the \[[backpressure]],
+\[[backpressureChangePromise]] and \[[transformStreamController]] slots uninitialised. They are not
+used in a transferred TransformStream.</p>
+
+</div>
 
 <div algorithm>
  The <dfn id="ts-constructor" constructor for="TransformStream" lt="TransformStream(transformer,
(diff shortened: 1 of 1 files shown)


ricea:
This is ready for review. @domenic and @MattiasBuelens PTAL.

There's a minor issue that I haven't figured out how to link `{{MessagePort}}'s transfer steps`. Probably it will require an editorial change to the HTML standard, or maybe it's not worth the effort.

I need to check I don't have any gaps in my tests and port them to wpt, but I don't expect that to take long.

ricea:
Tip for easier reviewing: The changes are under
https://whatpr.org/streams/1053.html#rs-prototype
https://whatpr.org/streams/1053.html#ws-prototype
https://whatpr.org/streams/1053.html#ts-prototype
https://whatpr.org/streams/1053.html#transferrable-streams

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+  1. If ! [$Type$](|type|) is not String, return.
+  1. If |type| is `"pull"`,
+   1. If |backpressurePromise| is not undefined,
+    1. [=Resolve=] |backpressurePromise| with undefined.
+    1. Set |backpressurePromise| to undefined.
+  1. Otherwise, if |type| is `"error"`,
+   1. Perform ! [$WritableStreamDefaultControllerErrorIfNeeded$](|controller|, |value|).
+   1. If |backpressurePromise| is not undefined,
+    1. [=Resolve=] |backpressurePromise| with undefined.
+    1. Set |backpressurePromise| to undefined.
```
MattiasBuelens, line 5962:
```suggestion
    1. Set |backpressurePromise| to undefined.
   1. Disentangle |port|.
```
The readable's *cancelAlgorithm* disentangles its port when it sends the `"error"` message. Thus, the writable side should do the same when it receives that message.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Let |controller| be a [=new=] {{ReadableStreamDefaultController}}.
+ 1. Let |backpressurePromise| be [=a new promise=].
+ 1. Add a handler for |port|'s {{MessagePort/message}} event with the following steps:
+  1. Let |data| be the data of the message.
+  1. If ! [$Type$](|data|) is not Object, return.
+  1. Let |type| be ? [$Get$](|data|, `"type"`).
+  1. Let |value| be ? [$Get$](|data|, `"value"`).
+  1. If ! [$Type$](|type|) is not String, return.
+  1. If |type| is `"chunk"`,
+   1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|controller|) is true,
```
MattiasBuelens, line 5912: Is this check still necessary after #1029?

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+   1. Disentangle |port|.
+ 1. Add a handler for |port|'s {{MessagePort/messageerror}} event with the following steps:
+  1. Let |error| be a new "{{DataCloneError}}" {{DOMException}}.
+  1. Perform ! [$CrossRealmTransformSendError$](|port|, |error|).
+  1. Perform ! [$ReadableStreamDefaultControllerError$](|controller|, |error|).
+  1. Disentangle |port|.
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |pullAlgorithm| be the following steps:
+  1. Perform ! [$PackAndPostMessage$](|port|, `"pull"`, undefined).
+  1. Return |backpressurePromise|.
```
MattiasBuelens, line 5930: In [remote-web-streams](https://github.com/MattiasBuelens/remote-web-streams/), I don't maintain a backpressure promise for `pull()`. [I return nothing](https://github.com/MattiasBuelens/remote-web-streams/blob/5ca92e2aee8a0dfbc9bd60b88a524cd46e3fb130/src/readable.ts#L25) (which gets interpreted as a promise resolved with `undefined`), and instead [I rely](https://github.com/MattiasBuelens/remote-web-streams/blob/5ca92e2aee8a0dfbc9bd60b88a524cd46e3fb130/src/readable.ts#L39-L40) on the fact that `ReadableStreamDefaultControllerEnqueue` will call `pull()` again if needed.

I can see that manually maintaining a backpressure promise makes it more obvious *when* the next pull can happen. On the other hand, it feels like we're duplicating some of the internal pull handling of `ReadableStream` here. For example, in `ReadableStreamTee`, we *do* return an immediately-resolved promise from its *pullAlgorithm*, and rely on the stream to not call pull again until we enqueue a chunk.

ricea:
> From your comment on #276, I understand that we will not yet support the "double-transfer" case in this first version. Correct?

Correct. It's unfortunate, but I said there, it's very complex to support.

> That is indeed unfortunate.
> 
> For remote-web-streams, I wrote a mock implementation of MessageChannel. It doesn't actually do any (de)serialization or send stuff between realms, it's just enough so I can run my tests. But I guess that wouldn't really work for a project like jsdom... 😅

We could polyfill MessageChannel and MessagePort I suppose, but we'd need quite a high level of fidelity and at that point it might be easier just to implement it in jsdom. Anyway, I'd prefer not to block on that.

index.bs
ricea: It's my understanding that we only need to disentangle the port on one side. Although from the implementers' perspective it might be better to explicitly disentangle on both sides. Let me think about it.

index.bs
ricea: Yes, because we don't want to reset the backpressure otherwise. Although now you mention it, maybe it doesn't make any actual difference.

index.bs
ricea: I don't want to diverge too much from Blink's implementation, as until there's another implementation it's the only sanity check we have.

However, I can try this in Blink and see whether anything breaks. In principle the behaviour shouldn't change, and it would be a nice simplification to get rid of `backpressurePromise`.

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+    and |messagePortDataHolder|.
+ 1. Set |dataHolder|.\[[port]] to |messagePortDataHolder|.
+
+</div>
+
+<div algorithm="ReadableStream transfer receiving steps">
+Their [=transfer-receiving steps=], given |dataHolder| and |value|, are:
+
+ 1. Let |port| be a new {{MessagePort}}.
+ 1. Perform {{MessagePort}}'s transfer-receiving steps with |dataHolder|.\[[port]] and |port|.
```
domenic, line 783: Similarly these seems best done with a call to https://html.spec.whatwg.org/#structureddeserializewithtransfer

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+<div algorithm="ReadableStream transfer steps">
+{{ReadableStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value| and
+|dataHolder|, are:
+
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port1| be that object.
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port2| be that object.
+ 1. [=Entangle=] the |port1| and |port2| objects.
+ 1. If ! [$IsReadableStreamLocked$](|value|) is true, throw a {{TypeError}} exception.
```
domenic, line 767: I wonder if "DataCloneError" DOMException would be more idiomatic.

index.bs
```diff
@@ -4894,6 +4972,37 @@ side=], or to terminate or error the stream.
...
 
+<div algorithm="TransformStream transfer steps">
+{{TransformStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value|
+and |dataHolder|, are:
+
+<!-- Find a way to linkify ReadableStream transfer steps and WritableStream transfer steps -->
+
+ 1. If ! [$IsReadableStreamLocked$](|value|.\[[readable]]) is true, throw a {{TypeError}} exception.
+ 1. If ! [$IsWritableStreamLocked$](|value|.\[[writable]]) is true, throw a {{TypeError}} exception.
+ 1. Let |readableDataHolder| be { \[[Type]]: `"ReadableStream"` }.
```
domenic, line 4983: Is the [[Type]] tagging used anywhere? I have this question generally, although in this case it seems particularly clear that it's not used.

index.bs
```diff
@@ -4894,6 +4972,37 @@ side=], or to terminate or error the stream.
...
+Their [=transfer-receiving steps=], given |dataHolder| and |value|, are:
+
+ 1. Perform {{ReadableStream}} transfer-receiving steps given |dataHolder|.\[[readable]] and
+    |value|.\[[readable]].
+ 1. Perform {{WritableStream}} transfer-receiving steps given |dataHolder|.\[[writable]] and
+    |value|.\[[writable]].
+
+<p class="note">These steps intentionally leave the \[[backpressure]],
+\[[backpressureChangePromise]] and \[[transformStreamController]] slots uninitialized. They are not
+used in a transferred TransformStream.</p>
```
domenic, line 5002: I'm not sure that we have a well-defined concept of "uninitialized". Every other path in the spec sets all internal slots on new objects, I believe. So I think it'd be better to explicitly set them to undefined.

index.bs
```diff
@@ -15,6 +15,10 @@ Markup Shorthands: markdown yes
 spec:webidl; type:dfn; text:resolve
 spec:webidl; type:dfn; text:new
 spec:infra; type:dfn; text:list
+spec:html; type:dfn; text:entangle
```
domenic, line 18: I'm surprised these are needed; there are really other specs that define "create a new MessagePort object" and friends?

Well, hopefully we can eliminate these, even if it requires some HTML edits. We'll see.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+Transferable streams are implemented using a special kind of identity transform which has the
+[=writable side=] in one realm and the [=readable side=] in another realm. The following abstract
+operations are used to implement these "cross-realm transforms".
+
+<div algorithm>
+ <dfn abstract-op lt="PackAndPostMessage">PackAndPostMessage(|port|, |type|, |value|)</dfn> performs
+ the following steps:
+
+ 1. Let |message| be [$OrdinaryObjectCreate$](null).
+ 1. Perform [$CreateDataProperty$](|message|, `"type"`, |type|).
```
domenic, line 5863:
```suggestion
 1. Perform [$CreateDataProperty$](|message|, "`type`", |type|).
```

and elsewhere

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Perform [$CreateDataProperty$](|message|, `"value"`, |value|).
+ 1. Let |targetPort| be the port with which |port| is entangled, if any; otherwise let it be null.
+ 1. Let |options| be «[ `"transfer"` → « » ]».
```
domenic, line 5866: It'd be ideal if HTML made this linkable; TODO for me.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Run the [=message port post message steps=] providing |targetPort|, |message| and |options|.
+
+ <p class="note">A JavaScript object is used for transfer to avoid having to duplicate the [=message
```
domenic, line 5869: Hmm. Although I don't think the duplication would be that bad, I think a JS object would still have to be involved, so I think this is the best we can do, indeed.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ port post message steps=]. The prototype of the object is set to null to avoid interference from
+ the global object.</p>
```
domenic, line 5871: "interference from `Object.prototype`" is probably more accurate.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="CrossRealmTransformSendError">CrossRealmTransformSendError(|port|,
+ |error|)</dfn> performs the following steps:
+
+ 1. Perform [$PackAndPostMessage$](|port|, `"error"`, |error|), discarding the result.
+
+ <p class="note">As we are already in an errored state when this abstract operation is performed, we
+ cannot handle further errors, so we just discard them.</p>
```
domenic, line 5894: Should we set [[PromiseIsHandled]] to true?

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+
+Transferable streams are implemented using a special kind of identity transform which has the
+[=writable side=] in one realm and the [=readable side=] in another realm. The following abstract
+operations are used to implement these "cross-realm transforms".
+
+<div algorithm>
+ <dfn abstract-op lt="PackAndPostMessage">PackAndPostMessage(|port|, |type|, |value|)</dfn> performs
+ the following steps:
+
+ 1. Let |message| be [$OrdinaryObjectCreate$](null).
```
domenic, line 5862: Put `!`s before all the infallible abstract ops. Generally every abstract op call that you're not going to explicitly handle needs a ! or ?.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="SetUpCrossRealmTransformReadable">SetUpCrossRealmTransformReadable(|stream|,
+ |port|)</dfn> performs the following steps:
+
+ 1. Perform ! [$InitializeReadableStream$](|stream|).
+ 1. Let |controller| be a [=new=] {{ReadableStreamDefaultController}}.
+ 1. Let |backpressurePromise| be [=a new promise=].
+ 1. Add a handler for |port|'s {{MessagePort/message}} event with the following steps:
```
domenic, line 5905: "Add a handler" is not well-defined. I've run into this problem before; you need something like https://wicg.github.io/kv-storage/#add-a-simple-event-listener. Maybe we can upstream that into DOM.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+  1. Let |data| be the data of the message.
+  1. If ! [$Type$](|data|) is not Object, return.
```
domenic, line 5907: How could these bad cases happen?

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+  1. Let |type| be ? [$Get$](|data|, `"type"`).
```
domenic, line 5908: Should exceptions in these cases error the stream, or be thrown? If they're thrown, I guess they'll be handled like any event listener error, and go to `window.onerror`? Do these cases have tests?

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+  1. Let |value| be ? [$Get$](|data|, `"value"`).
+  1. If ! [$Type$](|type|) is not String, return.
+  1. If |type| is `"chunk"`,
+   1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|controller|) is true,
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|controller|, |value|).
+    1. [=Resolve=] |backpressurePromise| with undefined.
+    1. Set |backpressurePromise| to [=a new promise=].
+  1. Otherwise, if |type| is `"close"`,
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](|controller|).
+    1. Disentangle |port|.
```
domenic, line 5918: Wow, it's weird that this is not a defined term.

index.bs
```diff
@@ -5740,6 +5849,146 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+  1. Otherwise, if |type| is `"error"`,
+   1. Perform ! [$ReadableStreamDefaultControllerError$](|controller|, |value|)
```
domenic, line 5920:
```suggestion
   1. Perform ! [$ReadableStreamDefaultControllerError$](|controller|, |value|).
```

index.bs
ricea: It feels odd having to route the event via JavaScript in the standard when in Blink's implementation we just have [blink::NativeEventListener](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/dom/events/native_event_listener.h?q=blink::NativeEventListener). And it's not like it's some weird  edge-case either: it's used in loads of places.

index.bs
ricea: The other realm can be hostile. I thought about adding a note about the security concerns, but I'm not sure there's a concept of a "compromised agent cluster" in spec-land, so I couldn't think of a way to phrase it.

index.bs
ricea: AFAIK, there's no way for this to happen except for a compromised render process, so there's no way to write tests, and the precise behaviour isn't important as long as we don't allow ourselves to be compromised.

index.bs
ricea: Yes, I was confused when I didn't find a definition.

index.bs
domenic: In that case I would suggest ! instead of ?, since _per spec_ there should never be such an exception.

index.bs
ricea: These are just not exported. I assumed we would probably end up modifying HTML to export them.

index.bs
ricea: The duplication is not bad in spec-land, but it's a can of worms in implementation land.

index.bs
domenic: I'd suggest an assert then. You could add a general note saying something about how assertions (either "assert" or !) could fail in memory-corruption scenarios (including security exploits), although that kind of holds for all asserts in the spec, so maybe it's not necessary.

index.bs
domenic: Yeah. To be clear I don't suggest you change anything here, but I'd like to get us better primitives.

ricea:
I've uploaded the tests at https://github.com/web-platform-tests/wpt/pull/24546. I've left it as a draft PR until we are ready to land this, but please feel free to review anyway.

index.bs
ricea: I'm uncomfortable using asserts in a case that implementations absolutely must handle. There's a big difference between an error being ignored because the current render process is compromised, and an error being ignored become some other render process is compromised.

domenic: But there's no way for the situation to occur, per spec, so an assert is pretty much definitionally appropriate. We use them for cases like this all over the place (especially in the JS spec).

ricea: Okay, how about if I add a note saying something like

**Implementations are encouraged to explicitly handle failures from the asserts in this algorithm, as the input may come from an untrusted context. Failure to do so could lead to security issues.**

domenic: That sounds great to me.

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
   immutable, this could allow interference between the two branches.
 </dl>
 
+<div algorithm="ReadableStream transfer steps">
+{{ReadableStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value| and
+|dataHolder|, are:
+
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
```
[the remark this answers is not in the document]
domenic: The results are in: it turns out the owner is totally useless. https://github.com/whatwg/html/pull/5728 ; see e.g. https://whatpr.org/html/5728/web-messaging.html#dom-messagechannel for an example of the new way to create `MessagePort`s.

index.bs
ricea: I'm leaning towards "no" now, since in platonic spec-land it is redundant to disentangle from both sides, and in practice implementations need to implement MessagePort `close()` in such a way that both sides get closed after all queued messages have been delivered.

index.bs
ricea: `PackAndPostMessage()` doesn't create a promise.

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
   immutable, this could allow interference between the two branches.
 </dl>
 
+<div algorithm="ReadableStream transfer steps">
```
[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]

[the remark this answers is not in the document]

index.bs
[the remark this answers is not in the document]
ricea: Yes, done.

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port1| be that object.
+ 1. [=Create a new MessagePort object=] whose [=owner=] is the [=incumbent settings object=] and let
+    |port2| be that object.
+ 1. [=Entangle=] the |port1| and |port2| objects.
+ 1. If ! [$IsReadableStreamLocked$](|value|) is true, throw a {{TypeError}} exception.
+ 1. Let |writable| be a [=new=] {{WritableStream}}.
+ 1. Perform ! [$SetUpCrossRealmTransformWritable$](|writable|, |port1|).
+ 1. Let |promise| be ! [$ReadableStreamPipeTo$](|value|, |writable|, false, false, false, undefined).
```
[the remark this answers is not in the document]
ricea: Removed.

index.bs
```diff
@@ -745,6 +755,36 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+ 1. Set |promise|.\[[PromiseIsHandled]] to true.
+ 1. Let |messagePortDataHolder| be { \[[Type]]: `"MessagePort"` }.
+ 1. Perform {{MessagePort}}'s transfer steps <!-- TODO(ricea): Link this somehow --> with |port2|
```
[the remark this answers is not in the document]
ricea: Done. I made it slightly higher-level to avoid poking around in the internals of the serialization algorithm.

index.bs
ricea: Yes, makes sense. Done.

index.bs

index.bs
ricea: I set them to undefined but retained the note that they aren't used.

index.bs
ricea: I wasn't sure how to linkify it, so I went with `%ObjectPrototype%`, which is mostly the same thing.

index.bs
ricea: Done.

index.bs

index.bs

index.bs
ricea: I tested removing `backpressurePromise` in Blink and it works, so I've done the same thing here.

index.bs
ricea: No. Removed.

index.bs

index.bs
ricea: Now `backpressurePromise` is gone, we definitely don't need the check, so I have removed it.

index.bs
```diff
@@ -36,16 +41,20 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: the typed array constructors table; url: #table-49
   text: typed array; url: #sec-typedarray-objects
   text: Number type; url: #sec-ecmascript-language-types-number-type
+  text: %ObjectPrototype%; url: #sec-properties-of-the-object-prototype-object
```
domenic, line 44: These days they would like us to move to `%Object.prototype%`. (Same URL, just change the linking text here and below.)

index.bs
```diff
@@ -965,6 +977,44 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
  1. Return [=a promise resolved with=] undefined.
 </div>
 
+<h4 id="rs-transfer">Transfer via postMessage()</h4>
+
+<dl class="domintro">
+ <dt><code>worker.postMessage(rs, [rs]);</code>
```
domenic, line 983:
```suggestion
 <dt><code>destination.postMessage(rs, [rs]);</code>
```

index.bs
```diff
@@ -965,6 +977,44 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
...
+ <dd>
+  <p>Sends a ReadableStream to another frame, window or worker.
```
domenic, line 985:
```suggestion
  <p>Sends a {{ReadableStream}} to another frame, window, or worker.
```

index.bs
```diff
@@ -965,6 +977,44 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
  1. Return [=a promise resolved with=] undefined.
 </div>
 
+<h4 id="rs-transfer">Transfer via postMessage()</h4>
```
domenic, line 980:
```suggestion
<h4 id="rs-transfer">Transfer via `postMessage()`</h4>
```

index.bs
```diff
@@ -965,6 +977,44 @@ default-reader-asynciterator-prototype-internal-slots">Asynchronous iteration</h
...
+
+ 1. If ! [$IsReadableStreamLocked$](|value|) is true, throw a "{{DataCloneError}}" {{DOMException}}.
+ 1. Let |port1| be a [=new=] {{MessagePort}} in [=the current Realm=].
+ 1. Let |port2| be a [=new=] {{MessagePort}} in [=the current Realm=].
+ 1. [=Entangle=] |port1| and |port2|.
+ 1. Let |writable| be a [=new=] {{WritableStream}}.
+ 1. Perform ! [$SetUpCrossRealmTransformWritable$](|writable|, |port1|).
+ 1. Let |promise| be ! [$ReadableStreamPipeTo$](|value|, |writable|, false, false, false).
+ 1. Set |promise|.\[[PromiseIsHandled]] to true.
+ 1. Set |dataHolder|.\[[port]] to ! [$StructuredSerializeWithTransfer$](port2, « |port2| »).
```
domenic, line 1004:
```suggestion
 1. Set |dataHolder|.\[[port]] to ! [$StructuredSerializeWithTransfer$](|port2|, « |port2| »).
```

index.bs
```diff
@@ -3625,6 +3678,44 @@ as seen for example in [[#example-ws-no-backpressure]].
  1. Return ? [$AcquireWritableStreamDefaultWriter$]([=this=]).
 </div>
 
+<h4 id="ws-transfer">Transfer via postMessage()</h4>
```
domenic, line 3681: I'd suggest basically the same editorial tweaks for this section.

index.bs
domenic: Alternately you could make it a bit more obvious with the new overload: `destination.postMessage(rs, { transfer: [rs] })`

index.bs
```diff
@@ -4937,6 +5031,50 @@ side=], or to terminate or error the stream.
  1. Return [=this=].\[[writable]].
 </div>
 
+<h4 id="ts-transfer">Transfer via postMessage()</h4>
```
domenic, line 5034: Same editorial tweaks here

index.bs
```diff
@@ -4937,6 +5031,50 @@ side=], or to terminate or error the stream.
...
+{{TransformStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value|
+and |dataHolder|, are:
+
+<!-- Find a way to linkify ReadableStream transfer steps and WritableStream transfer steps -->
+
+ 1. If ! [$IsReadableStreamLocked$](|value|.\[[readable]]) is true, throw a "{{DataCloneError}}"
+    {{DOMException}}.
+ 1. If ! [$IsWritableStreamLocked$](|value|.\[[writable]]) is true, throw a "{{DataCloneError}}"
+    {{DOMException}}.
+ 1. Perform {{ReadableStream}} transfer steps with |value|.\[[readable]] and |readableDataHolder|.
```
domenic, line 5056: I think you can just do the same thing you did for MessagePort, passing them into StructuredSerializeWithTransfer.

index.bs
```diff
@@ -5740,6 +5878,152 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+
+Transferable streams are implemented using a special kind of identity transform which has the
+[=writable side=] in one realm and the [=readable side=] in another realm. The following abstract
+operations are used to implement these "cross-realm transforms".
+
+<div algorithm>
+ <dfn abstract-op lt="PackAndPostMessage">PackAndPostMessage(|port|, |type|, |value|)</dfn> performs
+ the following steps:
+
+ 1. Let |message| be [$OrdinaryObjectCreate$](null).
```
domenic, line 5891:
```suggestion
 1. Let |message| be ! [$OrdinaryObjectCreate$](null).
```

index.bs
```diff
@@ -5740,6 +5878,152 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Perform ! [$CreateDataProperty$](|message|, "`type`", |type|).
+ 1. Perform ! [$CreateDataProperty$](|message|, "`value`", |value|).
+ 1. Let |targetPort| be the port with which |port| is entangled, if any; otherwise let it be null.
+ 1. Let |options| be «[ "`transfer`" → « » ]».
```
domenic, line 5895: TODO for me to make these linkable

index.bs
```diff
@@ -5740,6 +5878,152 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Run the [=message port post message steps=] providing |targetPort|, |message| and |options|.
```
domenic, line 5896:
```suggestion
 1. Run the [=message port post message steps=] providing |targetPort|, |message|, and |options|.
```

index.bs
```diff
@@ -5740,6 +5878,152 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
  1. Set |container|.\[[queueTotalSize]] to 0.
 </div>
 
+<h3 id="transferrable-streams">Transferable streams</h3>
```
domenic, line 5881: Alphabetize the abstract ops in this section

index.bs
```diff
@@ -5740,6 +5878,152 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+ 1. Enable |port|'s [=port message queue=].
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |writeAlgorithm| be the following steps, taking a |chunk| argument:
+  1. If |backpressurePromise| is undefined, set |backpressurePromise| to
+     [=a promise resolved with=] undefined.
+  1. Return the result of [=reacting=] to |backpressurePromise| with the following
+     fulfillment steps:
+   1. Set |backpressurePromise| to [=a new promise=].
+   1. Let |promise| be ! [$PackAndPostMessageHandlingError$](|port|, "`chunk`", |chunk|).
+   1. If |promise| is a rejected promise, disentangle |port|.
```
domenic, line 6007: This is pretty unusual. Can we do an "upon rejection" instead?

MattiasBuelens: Perhaps it was important that the port is disentangled synchronously? Not sure.

That said, I think a better way to solve this is to make `PackAndPostMessageHandlingError` throw an error instead of returning a rejected promise. That does mean we need to change all of the call sites to turn the abrupt completion into a rejected promise, but I think it's worth it. (There's only three of them.)

index.bs
ricea: OrdinaryObjectCreate does not return a completion record. This is non-obvious from the definition https://tc39.es/ecma262/2020/#sec-ordinaryobjectcreate but clear from how it is used elsewhere: https://tc39.es/ecma262/2020/#sec-createiterresultobject.

index.bs

index.bs
ricea: Good idea. Done.

index.bs

index.bs

index.bs

index.bs
ricea: Done.

index.bs
ricea: I switched back to returning a completion record.

index.bs
domenic: Everything returns a completion record: https://tc39.es/ecma262/#sec-implicit-completion-values

index.bs
```diff
@@ -4937,6 +5027,50 @@ side=], or to terminate or error the stream.
...
+<div algorithm="TransformStream transfer steps">
+ {{TransformStream}} objects are [=transferable objects=]. Their [=transfer steps=], given |value|
+ and |dataHolder|, are:
+
+ 1. Let |readable| be |value|.\[[readable]].
+ 1. Let |writable| be |value|.\[[writable]].
+ 1. If ! [$IsReadableStreamLocked$](|readable|) is true, throw a "{{DataCloneError}}"
+    {{DOMException}}.
+ 1. If ! [$IsWritableStreamLocked$](|writable|) is true, throw a "{{DataCloneError}}"
+    {{DOMException}}.
```
domenic, line 5051: Adding these checks here, as opposed to delegating to transferring the readable and writable sides independently, avoids ending up with a "half-transferred" transform stream. That's nice :). I wonder if we should `<p class="note">` that? In either case we should be sure we have a test for it.

ricea: I've updated the tests to test this feature (which is missing from Blink's implementation 😳).

I don't feel I need to call attention to the feature with a note, as it is just a matter of polish.

index.bs
```diff
@@ -3625,6 +3676,42 @@ as seen for example in [[#example-ws-no-backpressure]].
...
 
+<h4 id="ws-transfer">Transfer via `postMessage()`</h4>
+
+<dl class="domintro">
+ <dt><code>destination.postMessage(ws, { transfer: [ws] });</code>
+ <dd>
+  <p>Sends a {{WritableStream}} to another frame, window, or worker.
+
+  <p>The transferred stream can be used exactly like the original. The original will become
+  [=locked to a writer|lock=] and no longer directly usable.
```
MattiasBuelens, line 3687:
```suggestion
  [=locked to a writer|locked=] and no longer directly usable.
```

ricea:
Can I say that Google Chrome is interested even though I'm the one who wrote the change? 😄 

@jorendorff and @youennf could I hear your views on this?

domenic:
Yes, you certainly can :)

ricea:
Google Chrome is interested!

ricea:
@MattiasBuelens Github thinks I have unresolved comments from you, but I can't tell which they are. Are you okay with landing this, or do you think it needs more work?

PR #1058, opened 2020-07-23
Update WPT

MattiasBuelens:
In web-platform-tests/wpt#24577, a window-only test was added which uses an `<iframe>` to check that `CountQueuingStrategy.prototype.size` and `ByteLengthQueuingStrategy.prototype.size` return different functions in different realms. However, we cannot emulate this easily with our JSDOM-based test runner, so the reference implementation fails this new test.

This PR changes our test script to only run `.any.js` tests, since these should work regardless of the context. I've verified that all current tests indeed are `.any.js` tests, and still run with this new check. The only test that is excluded is the new `.window.js` test, which we probably don't want (or need) to test against the reference implementation.


domenic:
Want to bump this to include the latest changes to the test suite that you've made?

MattiasBuelens:
Done!

PR #1059, opened 2020-07-23
Add examples for stopping pipes

### LINKED ISSUE #1055: Possible to undo .pipeTo()?
Originally asked on StackOverflow several days ago: https://stackoverflow.com/q/62814526/4896937

`readableStream.pipeTo(writableStream)` is a neat way of piping a `ReadableStream` to a `WritableStream`. This would appear to be more convenient than obtaining `readableStream.getReader()` and manually gluing it to `writableStream.getWriter()` or directly to underlying functionality.

I have a use case where I am provided a `ReadableStream` object which may eventually need to "go away" and be replaced by a new `ReadableStream` object. It would be nice to use a single custom `WritableStream` to wrap the functionality which these `ReadableStream` objects interface with and pipe them together for as long as needed, but have not found how to "undo" the piping process.

**Q: Is it possible to "undo"/"break"/"teardown" a pipe created with `.pipeTo()`, or are pipes "permanent" and not appropriate for this use case?**

Example using [Serial API](https://wicg.github.io/serial/), where I would like to be able to call `doConnectPort()` and then `doDisconnectPort()` repeatedly:

```javascript
var customWritable = new WritableStream(…);
var port;

async function doConnectPort() {
  port = await navigator.serial.requestPort();
  await port.open({ baudrate: …});
  await port.readable.pipeTo(customWritable);
}

async function doDisconnectPort() {
  // What to do here? Can't close the port yet—port.readable is still locked!
  await port.close();
}
```

Currently, I manually glue things together and clean up when it's time to disconnect the port:

```javascript
var port;
var portReader;

async function doConnectPort() {
  port = await navigator.serial.requestPort();
  await port.open({ baudrate: …});
  portReader = port.readable.getReader();
  while (true) {
    const { value, done } = await portReader.read();
    if (done) {
      break;
    }
    // do something with value
  }
}

async function doDisconnectPort() {
  await portReader.cancel();
  await portReader.releaseLock();
  await port.close();
}
```

domenic:
Closes #1055.

***
<a href="https://whatpr.org/streams/1059.html" title="Last updated on Jul 23, 2020, 9:43 PM UTC (76ab3b6)">Preview</a> | <a href="https://whatpr.org/streams/1059/62fe4c8...76ab3b6.html" title="Last updated on Jul 23, 2020, 9:43 PM UTC (76ab3b6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -868,6 +868,49 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
     |options|["{{StreamPipeOptions/preventClose}}"],
     |options|["{{StreamPipeOptions/preventAbort}}"],
     |options|["{{StreamPipeOptions/preventCancel}}"], |signal|).
+
+ <div class="example" id="example-pipe-abortsignal">
+  An ongoing [=pipe=] operation can be stopped using an {{AbortSignal}}, as follows:
+
+  <xmp highlight="js">
+   const controller = new AbortController();
+   readable.pipeTo(writable, { signal: controller.signal });
+
+   // ... some time later ...
+   controller.abort();
+  </xmp>
+
+  (The above omits error handling for the promise returned by {{ReadableStream/pipeTo()}}.
+  Additionally, the impact of the {{StreamPipeOptions/preventAbort}} and
+  {{StreamPipeOptions/preventCancel}} options what happens when piping is stopped are worth
+  considering.)
+ </div>
+
+ <div class="example" id="example-pipe-switch-dest">
+  The above technique can be used to switch the {{ReadableStream}} being piped, while writing into
+  the same {{WritableStream}}:
+
+  <xmp highlight="js">
+   const controller = new AbortController();
+   const pipePromise = readable1.pipeTo(writable, { preventAbort: true, signal: controller.signal });
+
+   // ... some time later ...
+   controller.abort();
+
+   // Wait for the pipe to complete before starting a new one:
+   try {
+    await pipePromise;
+   } catch (e) {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -868,6 +868,49 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+
+  <xmp highlight="js">
+   const controller = new AbortController();
+   readable.pipeTo(writable, { signal: controller.signal });
+
+   // ... some time later ...
+   controller.abort();
+  </xmp>
+
+  (The above omits error handling for the promise returned by {{ReadableStream/pipeTo()}}.
```
ricea, line 883: Having two sentences inside "()" looks a little odd to me. I think you could just omit the "()" here. If you like. I don't have really strong feelings about it.

domenic: Eh, on balance I like it, as a way of deemphasizing these sentences. I'll keep it. Thanks for weighing in though.

domenic:
FYI folks might be interested in the discussion I'm having over in https://github.com/whatwg/dom/issues/881 on making this pattern less terrible.

domenic:
Hmm, or we could do something entirely on the Streams side. For example:

```js
const controller = new AbortController();
readable1.pipeTo(writable, { preventAbort: true, signal: controller.signal });

// ... some time later ...
controller.abort();

// Start the new pipe, when ready:
readable2.pipeTo(writable, { waitForPrevious: true });
```

with the same semantics. (That would require storing the previous pipe promise while it is not settled, but that seems OK.)

I don't think we could make `waitForPrevious` the default; my intuition is that that would transform too many error cases into stalled cases. But maybe it'd work...

PR #1064, opened 2020-08-17
Editorial: Export [[detached]] slots

ricea:
Use `<dfn export>` for [[detached]] slots since they are part of the
public-facing API (consumed by the HTML standard). See
https://github.com/tabatkins/bikeshed/issues/1747.

***
<a href="https://whatpr.org/streams/1064.html" title="Last updated on Aug 18, 2020, 1:16 AM UTC (9d4ff46)">Preview</a> | <a href="https://whatpr.org/streams/1064/1fd19a0...9d4ff46.html" title="Last updated on Aug 18, 2020, 1:16 AM UTC (9d4ff46)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -516,8 +516,7 @@ table:
    {{ReadableByteStreamController}} created with the ability to control the state and queue of this
    stream
   <tr>
-   <td><!-- TODO(ricea): Style this as <dfn unused> when that is supported.
-            See https://github.com/tabatkins/bikeshed/issues/1747 --><b>\[[detached]]</b>
+   <td><dfn export>\[[detached]]</dfn>
    <td class="non-normative">A boolean flag set to true when the stream is transferred
   <tr>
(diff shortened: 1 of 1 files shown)


domenic:
Actually, shouldn't these be capitalized too?

We could also invert the relationship and link these to https://html.spec.whatwg.org/#detached, hmm. In that case maybe we'd remove them from the table, with the idea being that they're "automatically" added by virtue of the `[Transferable]` extended attribute?

ricea:
> Actually, shouldn't these be capitalized too?

Done.

> We could also invert the relationship and link these to https://html.spec.whatwg.org/#detached, hmm. In that case maybe we'd remove them from the table, with the idea being that they're "automatically" added by virtue of the `[Transferable]` extended attribute?

I imagine this could cause problems for specs that explicitly want to reference their `[[Detached]]` slots, assuming they exist.

PR #1069, opened 2020-08-19
Don't error an errored WritableStream on messageerror

ricea:
   * …
   * …
   * The change is not observable.
   * Chrome: Not needed (never had the bug)
   * Firefox: Not needed (unimplemented)
   * Safari: Not needed (unimplemented)

***
<a href="https://whatpr.org/streams/1069.html" title="Last updated on Aug 27, 2020, 9:05 PM UTC (8c8f6dc)">Preview</a> | <a href="https://whatpr.org/streams/1069/8652455...8c8f6dc.html" title="Last updated on Aug 27, 2020, 9:05 PM UTC (8c8f6dc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6146,7 +6146,7 @@ abstract operations are used to implement these "cross-realm transforms".
  1. Add a handler for |port|'s {{MessagePort/messageerror}} event with the following steps:
   1. Let |error| be a new "{{DataCloneError}}" {{DOMException}}.
   1. Perform ! [$CrossRealmTransformSendError$](|port|, |error|).
-  1. Perform ! [$WritableStreamDefaultControllerError$](|controller|, |error|).
+  1. Perform ! [$WritableStreamDefaultControllerErrorIfNeeded$](|controller|, |error|).
   1. Disentangle |port|.
  1. Enable |port|'s [=port message queue=].
  1. Let |startAlgorithm| be an algorithm that returns undefined.


domenic:
> It appears to be impossible to test in a cross-browser fashion

Can you say more? Is this change totally unobservable?

ricea:
Elaborating on the testing issue, I have been unable to find a cross-browser method to cause a messageerror event. RTCCertificate posted to a Worker should cause a messageerror by the standard because it is a Window-only type, but in practice in Chrome lets you send it to a Worker anyway. SharedArrayBuffer might be a viable option in future, but is not currently practical.

domenic:
Sending a `WebAssembly.Module` to a cross-site (note: cross-origin does not suffice) page will reliably cause a `messageerror`. See https://github.com/web-platform-tests/wpt/blob/6c2b980cd5a4bb8c64eeea5c89899051bd4f4fa0/origin-isolation/resources/helpers.mjs#L214 for some examples of how to set that up in web platform tests.

ricea:
After writing a test for this, I found it wasn't observable anyway. If an error/close occurred locally, then we won't be in a state where we can send a new chunk. If it happened remotely, then the remote side will have disentangled the message pipe already, so messageerror will never fire.

I still prefer to make this change, as the reason why it's not needed is extremely subtle.

PR #1073, opened 2020-08-26
Create a dedicated section for interfacing with other specs

### LINKED ISSUE #372: Separate out exported abstract ops from internal ones

domenic:
This incorporates much of https://fetch.spec.whatwg.org/#streams, as well as https://wicg.github.io/compression/#generic-transform-stream. It also looks at how https://wicg.github.io/native-file-system/ uses writable streams, and creates wrapper algorithms to make that easier.

Closes #372.

This is probably ready for review, but we shouldn't merge it until I have created update pull requests for at least:

- https://fetch.spec.whatwg.org/: https://github.com/whatwg/fetch/pull/1085
- https://w3c.github.io/ServiceWorker/: https://github.com/w3c/ServiceWorker/pull/1533
- https://encoding.spec.whatwg.org/: https://github.com/whatwg/encoding/pull/224
- https://wicg.github.io/compression/: https://github.com/WICG/compression/pull/35
- https://wicg.github.io/native-file-system/: https://github.com/WICG/native-file-system/pull/225

and have validated that this suffices for their purposes.

Also, I'll be opening another issue about GenericTransformStream.

***
<a href="https://whatpr.org/streams/1073.html" title="Last updated on Sep 14, 2020, 7:31 PM UTC (4c0acd1)">Preview</a> | <a href="https://whatpr.org/streams/1073/1af4aff...4c0acd1.html" title="Last updated on Sep 14, 2020, 7:31 PM UTC (4c0acd1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6194,14 +6194,14 @@ failures should be handled by the calling specification.
  <p class="note">This will throw an exception if |stream| is already [=ReadableStream/locked=].
 </div>
 
-<p algorithm>To <dfn export for="ReadableStream">read a chunk</dfn> from a
+<p algorithm>To <dfn export for="ReadableStreamReader">read a chunk</dfn> from a
 {{ReadableStreamDefaultReader}} |reader|, given a [=read request=] |readRequest|, perform !
 [$ReadableStreamDefaultReaderRead$](|reader|, |readRequest|).
 
 <div algorithm="read all bytes">
- <p>To <dfn export for="ReadableStream" lt="read all bytes|reading all bytes">read all bytes</dfn>
- from a {{ReadableStreamDefaultReader}} |reader|, perform the following steps. The result will be a
- {{Promise}} for a [=byte sequence=].
+ <p>To <dfn export for="ReadableStreamReader" lt="read all bytes|reading all bytes">read all
+ bytes</dfn> from a {{ReadableStreamDefaultReader}} |reader|, perform the following steps. The
+ result will be a {{Promise}} for a [=byte sequence=].
 
  1. Let |promise| be [=a new promise=].
  1. Let |bytes| be an empty [=byte sequence=].
@@ -6233,9 +6233,14 @@ failures should be handled by the calling specification.
  mechanism if convenient.
 </div>
 
-<p algorithm>To <dfn export for="ReadableStream">release a reader</dfn> given a
+<p algorithm>To <dfn export for="ReadableStreamReader">release a reader</dfn> given a
 {{ReadableStreamDefaultReader}} |reader|, perform !
 [$ReadableStreamReaderGenericRelease$](|reader|).
+
+<p algorithm>To <dfn export for="ReadableStreamReader">cancel</dfn> a
+{{ReadableStreamDefaultReader}} |reader| with |reason|, perform !
+[$ReadableStreamReaderGenericCancel$](|reader|, |reason|). The return value will be a promise
+that either fulfills with undefined, or rejects with a failure reason.
 
 <p algorithm>To <dfn export for="ReadableStream">cancel</dfn> a {{ReadableStream}} |stream| with
 |reason|, return ! [$ReadableStreamCancel$](|stream|, |reason|). The return value will be a promise
--- a/index.bs
+++ b/index.bs
@@ -6168,10 +6168,10 @@ JavaScript value |e|, perform !
 [$ReadableStreamDefaultControllerError$](|stream|.[=ReadableStream/[[controller]]=], |e|).
 
 <div algorithm>
- To <dfn export for="ReadableStream">create a proxy</dfn> around a {{ReadableStream}} |stream|,
- perform the following steps. The result will be a new {{ReadableStream}} object which pulls its
- data from |stream|, while |stream| itself becomes immediately [=ReadableStream/locked=] and
- [=ReadableStream/disturbed=].
+ To <dfn export for="ReadableStream" lt="create a proxy|creating a proxy">create a proxy</dfn> for a
+ {{ReadableStream}} |stream|, perform the following steps. The result will be a new
+ {{ReadableStream}} object which pulls its data from |stream|, while |stream| itself becomes
+ immediately [=ReadableStream/locked=] and [=ReadableStream/disturbed=].
 
  1. Let |identityTransform| be the result of <a>creating an identity `TransformStream`</a>.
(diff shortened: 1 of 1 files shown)


domenic:
This should also have a section on duplex streams like https://reillyeon.github.io/serial/#serialport-interface. I don't think they need any new algorithms, but stating things like "use the names `readable` and `writable`" and explaining how you can use the create algorithms is probably a good idea.

wolenetz:
Note that https://github.com/w3c/media-source/issues/14 may eventually also be resolved such that Media Source Extensions becomes another interface point for the Streams API.

ricea:
I've had an initial read-through and it looks good.

ricea:
We will also need to integrate with https://wicg.github.io/web-transport/ but that should be straightforward.

domenic:
OK, PRs sent to all of the above; I updated the OP with them.

I got all ready to send one for WebTransport, but they don't appear to be currently at a level of rigor where this makes sense; they just talk about "Set X to a ReadableStream of Uint8Arrays representing (some abstract concept), currently empty". I guess the assumption is that all the enqueuing/closing/erroring is encapsulated in that "represents", hmm.

ricea:
> I got all ready to send one for WebTransport, but they don't appear to be currently at a level of rigor where this makes sense; they just talk about "Set X to a ReadableStream of Uint8Arrays representing (some abstract concept), currently empty". I guess the assumption is that all the enqueuing/closing/erroring is encapsulated in that "represents", hmm.

No problem, we have to increase the level of rigor anyway, so we can adopt the new interfaces at the same time. Sorry for wasting your time.

index.bs
```diff
@@ -6208,6 +6080,474 @@ The following abstract operations are a grab-bag of utilities.
...
+  For the purposes of the above algorithm, to <dfn>read-loop</dfn> given |reader|, |bytes|, and
+  |promise|:
+
+  1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
+   : [=read request/chunk steps=], given |chunk|
+   ::
+    1. If |chunk| is not a {{Uint8Array}} object, [=reject=] |promise| with a {{TypeError}} and
+       abort these steps.
+    1. Append the bytes represented by |chunk| to |bytes|.
+    1. [=Read-loop=] given |reader|, |bytes|, and |promise|.
```
ricea, line 6221: This can call ReadableStreamDefaultReaderRead re-entrantly, which is a bit scary. As far as I can tell it's safe in the standard language, but implementations can fail due to stack overflow. Easiest fix would probably be to post a microtask.

index.bs
```diff
@@ -6208,6 +6080,474 @@ The following abstract operations are a grab-bag of utilities.
...
+ branch returned will have its [=chunks=] cloned (using HTML's [=serializable objects=] framework)
+ from those of the first branch. This prevents consumption of one of the branches from interfering
+ with the other.
+</div>
+
+<h4 id="other-specs-rs-introspect">Introspection</h4>
+
+The following predicates can be used on arbitrary {{ReadableStream}} objects. However, note that
+this direct introspection is not possible via the public JavaScript API, and so specifications
+should instead use the algorithms in [[#other-specs-rs-reading]]. (For
```
ricea, line 6263: Since `locked` _is_ part of the public API, this might not be the best example.

index.bs
```diff
@@ -6208,6 +6080,474 @@ The following abstract operations are a grab-bag of utilities.
...
+ returning a number; and if given, |highWaterMark| must be a non-negative, non-NaN number.
+
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |closeAlgorithmWrapper| be an algorithm that runs these steps:
+  1. If |closeAlgorithm| was given, return the result of running it.
+  1. Return [=a promise resolved with=] undefined.
+ 1. Let |abortAlgorithmWrapper| be an algorithm that runs these steps:
+  1. If |abortAlgorithm| was given, return the result of running it.
+  1. Return [=a promise resolved with=] undefined.
+ 1. If |sizeAlgorithm| is not given, then set it to an algorithm that returns 1.
```
ricea, line 6313: Why is "is not given" present tense when in the previous steps "was given" is past tense?

ricea (commented):
Apart from my concern about read-loop, this lgtm.

Sorry for the delayed review.

index.bs
MattiasBuelens: I suspect this is because the steps in *closeAlgorithmWrapper* and *abortAlgorithmWrapper* are run **after** the main steps complete.

ricea: That makes sense.

domenic: Consistency would be nicer; I'll stick with "was". It looks like this is a preexisting problem so it'll touch a few algorithms.

index.bs
domenic: Hmm. Given that there's already a note explaining that this might be implemented rather differently, I think it'd be best to make a note of the danger, and encourage implementations to mitigate it in a way that doesn't require "read all bytes" to delay for (number of chunks) microtasks.

domenic: Well, I guess such a requirement wouldn't be observable either, so scratch that part of my comment. But still.

ricea (approved):
lgtm

PR #1078, opened 2020-10-01
Editorial: Define "create a ReadableStream" instead of just "create"

jyasskin:
The current definition forces redundant text in callers:

```
... the result of [=ReadableStream/creating=] a {{ReadableStream}} given ...
```

My proposal allows just

```
... the result of [=creating a ReadableStream=] given ...
```

The general principle is that `for` works well for methods of an existing object, where readers will see the object with its type at any call site. We shouldn't use `for` in creation methods where there's no other object of the `for` type at the call site.

This change will, unfortunately, break existing calls to this algorithm. I think we could add an empty `<dfn export for=ReadableStream lt="create|creating"></dfn>` before it to keep those working if you want.

***
<a href="https://whatpr.org/streams/1078.html" title="Last updated on Jan 15, 2021, 7:26 AM UTC (bee3b46)">Preview</a> | <a href="https://whatpr.org/streams/1078/a669248...bee3b46.html" title="Last updated on Jan 15, 2021, 7:26 AM UTC (bee3b46)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6100,7 +6100,7 @@ to grow organically as needed.
 <h4 id="other-specs-rs-create">Creation and manipulation</h4>
 
 <div algorithm="create a ReadableStream">
- To <dfn export for="ReadableStream" lt="create|creating">create</dfn> a {{ReadableStream}} given an
+ To <dfn export lt="create a ReadableStream|creating a ReadableStream">create a ReadableStream</dfn> given an
  optional algorithm <dfn export for="ReadableStream/create"><var>pullAlgorithm</var></dfn>, an
  optional algorithm <dfn export for="ReadableStream/create"><var>cancelAlgorithm</var></dfn>, an
  optional number <dfn export for="ReadableStream/create"><var>highWaterMark</var></dfn> (default 1),


domenic:
I'm -1 on this change, as it messes up your typography, and also doesn't fit the mental model. This is meant to be a specialization of the general [create](https://heycam.github.io/webidl/#new) algorithm in Web IDL, which is specifically for=ReadableStream.

jyasskin:
I wonder if @tabatkins has a good idea for constructors. I don't think it makes sense to think of this as a "specialization" of WebIDL's [new](https://heycam.github.io/webidl/#new), since [=new=] already [works](https://streams.spec.whatwg.org/#create-readable-stream) on `ReadableStream`s without specialization, [=new=] isn't written to expect specializations, and this thing takes different arguments from WebIDL's thing ... but it does make sense to think of this as a constructor, which deserves appropriate markup and rendering.

domenic:
Well, it's a specialization in the sense that if you do Web IDL's new on a ReadableStream, you will get an incompletely-constructed one. So you should use this version instead, and never use Web IDL's new.

PR #1080, opened 2020-10-27
Fix readable byte stream example

nidhijaju:
This change fixes a few bugs in the example for readable byte streams. 
One of the bugs in the existing example is that there seems to be infinite 
recursion for streams that have a total length of less than 1024. 
Furthermore, another bug is the return value of reader.read() is not 
destructured.

In this change, iteration is used instead of recursion, as well as async/
await.

***

### :boom: Error: Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 28, 2020, 1:47 AM UTC)_.

<details>
<summary>More</summary>

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/streams%231080.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -426,19 +426,25 @@ particularly important for the data structure described in [[#queue-with-sizes]]
  const reader = readableStream.getReader({ mode: "byob" });
 
  let startingAB = new ArrayBuffer(1024);
- readInto(startingAB)
-   .then(buffer => console.log("The first 1024 bytes:", buffer))
-   .catch(e => console.error("Something went wrong!", e));
+ const buffer = await readInto(startingAB);
+ console.log("The first 1024 bytes: ", buffer);
 
- function readInto(buffer, offset = 0) {
-   if (offset === buffer.byteLength) {
-     return Promise.resolve(buffer);
-   }
+ async function readInto(buffer) {
+   let offset = 0;
+   let view;
+   let done;
 
-   const view = new Uint8Array(buffer, offset, buffer.byteLength - offset);
-   return reader.read(view).then(newView => {
-     return readInto(newView.buffer, offset + newView.byteLength);
-   });
+   do {
+     ({value: view, done} =
+      await reader.read(new Uint8Array(buffer, offset, buffer.byteLength - offset)));
+     buffer = view.buffer;
+     if (done) {
+       break;
+     }
+     offset += view.byteLength;
+   } while (offset < buffer.byteLength);
+
+   return buffer;
  }
  </xmp>
 
--- a/index.bs
+++ b/index.bs
@@ -431,18 +431,16 @@ particularly important for the data structure described in [[#queue-with-sizes]]
 
  async function readInto(buffer) {
    let offset = 0;
-   let view;
-   let done;
 
-   do {
-     ({value: view, done} =
-      await reader.read(new Uint8Array(buffer, offset, buffer.byteLength - offset)));
+   while (offset < buffer.byteLength) {
+     const {value: view, done} =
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -426,19 +426,25 @@ particularly important for the data structure described in [[#queue-with-sizes]]
...
+     buffer = view.buffer;
+     if (done) {
+       break;
+     }
+     offset += view.byteLength;
+   } while (offset < buffer.byteLength);
 
-   const view = new Uint8Array(buffer, offset, buffer.byteLength - offset);
-   return reader.read(view).then(newView => {
-     return readInto(newView.buffer, offset + newView.byteLength);
```
ricea, line 440: Since the new version no longer has a `newView` variable, we should change the explanation below the example to refer to `view` instead.

ricea:
Please wait for @domenic to also take a look at it in case I missed any bugs.

> Furthermore, another bug is the return value of reader.read() is not
destructed, and the buffer is used after detaching.

"destructured", not "destructed" 😄 

The version in the standard never had the bug that the buffer was used after detaching, so you can remove that comment.

index.bs
```diff
@@ -426,19 +426,25 @@ particularly important for the data structure described in [[#queue-with-sizes]]
...
+ const buffer = await readInto(startingAB);
+ console.log("The first 1024 bytes: ", buffer);
+
+ async function readInto(buffer) {
+   let offset = 0;
+   let view;
+   let done;
+
+   do {
+     ({value: view, done} =
```
MattiasBuelens, line 438: I think this would read better as:
```javascript
const {value: view, done} = await reader.read(new Uint8Array(buffer, offset, buffer.byteLength - offset));
```
Then we don't need the extra parentheses or the `let view` and `let done` declarations at the top.

index.bs
```diff
@@ -426,19 +426,25 @@ particularly important for the data structure described in [[#queue-with-sizes]]
...
-   }
+ const buffer = await readInto(startingAB);
+ console.log("The first 1024 bytes: ", buffer);
+
+ async function readInto(buffer) {
+   let offset = 0;
+   let view;
+   let done;
+
+   do {
```
MattiasBuelens, line 437: I *think* this could be a regular `while` loop? If `buffer.byteLength === 0`, then we don't even need to bother calling `read()`.
```suggestion
   while (offset < buffer.byteLength) {
```

MattiasBuelens (commented):
Oh god, that example was very broken indeed. How did we not even check the `done` and `value` properties from the result of `read()`? 😛

The new version already looks much better! 👍 I think I can actually understand what it's doing now. 😁

index.bs
ricea: Good! I hadn't noticed that we were no longer using `view` and `done` outside the loop.

index.bs
ricea: Agree. A regular `while` loop is easier to understand.

index.bs
nidhijaju: Sorry, yes I've changed the explanation to reflect that now. Thank you!

index.bs
nidhijaju: That's a good point! I've just changed it to that in the last commit. Thank you!

index.bs
nidhijaju: That makes sense, I've just added your suggestion in my last commit. Thank you for taking a look and the suggestions as well! 😃

domenic (approved):
LGTM, thank you!

Issue #1018, opened 2019-10-15
ReadableStream.from(X)


ricea:
It would be good to have a convenience method to create a ReadableStream in a DWIM-ish fashion from something else.

Logic being something like
1. If X is an async iterable, then create a stream and enqueue each value as a chunk, applying flow control, closing when the async iterable is done.
2. Else if X is an iterable, then create a stream and enqueue each value as a chunk, closing when the iterable is done.
3. Else if X is a Blob, return the original value of Blob.prototype.stream called on X.
4. Else create a stream, enqueue X, and close it.

domenic:
I wonder if we want to omit (4) for future-proofing, in case there are other idiomatic types we want to add to this.

(3) is also somewhat dubious as it's not much value over `blob.stream()`.

What are code examples where this would be most helpful?

ricea:
The main motivating example is this from the [CompressionStream explainer](https://github.com/ricea/compressstream-explainer/blob/master/README.md#example-code):

```javascript
async function compressArrayBuffer(in) {
  const cs = new CompressionStream('deflate');
  const writer = cs.writable.getWriter();
  writer.write(in);
  writer.close();
  const out = [];
  const reader = cs.readable.getReader();
  let totalSize = 0;
  while (true) {
    const { value, done } = await reader.read();
    if (done)
      break;
    out.push(value);
    totalSize += value.byteLength;
  }
  const concatenated = new Uint8Array(totalSize);
  let offset = 0;
  for (const array of out) {
    concatenated.set(array, offset);
    offset += array.byteLength;
  }
  return concatenated;
}
```
where it took 3 lines just to create a stream containing a single ArrayBuffer chunk, something that feels like it should be a one-liner. This is also a possible solution to https://github.com/whatwg/fetch/issues/809.

ricea:
I just realised that
```javascript
ReadableStream.from([in]).pipeTo(cs.writable);
```
would also work, is only two more characters, and requires less DWIM.

domenic:
Awesome. Yeah just doing (1) and (2) seems totally uncontroversial (although it'll need a bit of Web IDL work, which I'm happy to do).

jimmywarting:
like this proposal. this would make the conversion between node/whatwg stream easy if both have @@asyncIterator

jimmywarting:
> I just realised that
> ```js
> ReadableStream.from([in]).pipeTo(cs.writable);
> ```

That's not that far off from 
```js
[...]

PR #1083, opened 2020-11-10
ReadableStream.from(asyncIterable)

MattiasBuelens:
I want to give #1018 a try. 😁

This adds a static `ReadableStream.from()` method, which takes an async iterable and returns a `ReadableStream` pulling chunks from that async iterable. Sync iterables (including arrays and generators) are also supported, since [`GetIterator()`](https://tc39.es/ecma262/#sec-getiterator) already has all the necessary handling to adapt a sync iterator into an async iterator. So that's nice. 😄

I think it should be fairly easy to add `ReadableStream.of(...chunks)` as well, which would simply call through to `ReadableStream.from(chunks)`. I'll look into it.

This is a very early draft, so there are no tests yet. I first want to get most of the semantics done, then we can worry about all the nasty error handling and edge cases.

   * Chrome (`@ricea`, https://github.com/whatwg/streams/pull/1083#issuecomment-820205616)
   * Firefox (`@mgaudet`, https://github.com/whatwg/streams/pull/1083#issuecomment-1145991409)
   * web-platform-tests/wpt#27009
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1442600
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1772772
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=255086

***
<a href="https://whatpr.org/streams/1083.html" title="Last updated on Jun 7, 2023, 4:55 PM UTC (bf5f360)">Preview</a> | <a href="https://whatpr.org/streams/1083/058f290...bf5f360.html" title="Last updated on Jun 7, 2023, 4:55 PM UTC (bf5f360)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -823,7 +823,7 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  1. Let |stream| be undefined.
  1. Let |iteratorRecord| be undefined.
  1. Let |startAlgorithm| be the following steps:
-  1. Set |iteratorRecord| to ? [$GetIterator$](|asyncIterable|, async).
+  1. Set |iteratorRecord| to ? [$GetIterator$](|asyncIterable|, `async`).
  1. Let |pullAlgorithm| be the following steps:
   1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
   1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
@@ -840,7 +840,7 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
     1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
        |value|).
  1. Let |cancelAlgorithm| be the following steps:
-  1. Let |returnMethod| be ? [$GetMethod$](|iteratorRecord|.\[[Iterator]], "return").
+  1. Let |returnMethod| be ? [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
   1. If |returnMethod| is undefined, return [=a promise resolved with=] undefined.
   1. Let |returnResult| be ? [$Call$](|returnMethod|, |iteratorRecord|.\[[Iterator]]).
   1. Return [=a promise resolved with=] |returnResult|.
--- a/index.bs
+++ b/index.bs
@@ -821,9 +821,8 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  are:
 
  1. Let |stream| be undefined.
- 1. Let |iteratorRecord| be undefined.
- 1. Let |startAlgorithm| be the following steps:
-  1. Set |iteratorRecord| to ? [$GetIterator$](|asyncIterable|, `async`).
+ 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, `async`).
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
  1. Let |pullAlgorithm| be the following steps:
   1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
   1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
--- a/index.bs
+++ b/index.bs
@@ -829,7 +829,7 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
      |nextResult|.\[[Value]].
   1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
   1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
-     given the argument |iterResult|:
+     given |iterResult|:
    1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
    1. Let |done| be ? [$IteratorComplete$](|iterResult|).
    1. If |done| is true:
@@ -838,12 +838,13 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
     1. Let |value| be ? [$IteratorValue$](|iterResult|).
     1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
        |value|).
- 1. Let |cancelAlgorithm| be the following steps:
+ 1. Let |cancelAlgorithm| be the following steps, given |reason|:
   1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
   1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
      |returnMethod|.\[[Value]].
   1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
-  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iteratorRecord|.\[[Iterator]]).
+  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iteratorRecord|.\[[Iterator]],
+     « |reason| »).
   1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
      |returnResult|.\[[Value]].
   1. Return [=a promise resolved with=] |returnResult|.\[[Value]].
--- a/index.bs
+++ b/index.bs
@@ -2110,6 +2110,49 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
  1. Return true.
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
+
+ 1. Let |stream| be undefined.
+ 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, async).
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |pullAlgorithm| be the following steps:
+  1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
+  1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
+     |nextResult|.\[[Value]].
+  1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
+  1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
+     given |iterResult|:
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
+       |value|).
+       <!-- TODO (future): If we allow changing the queuing strategy, this Enqueue might throw.
+             We'll then need to catch the error and close the async iterator. -->
+ 1. Let |cancelAlgorithm| be the following steps, given |reason|:
+  1. Let |iterator| be |iteratorRecord|.\[[Iterator]].
+  1. Let |returnMethod| be [$GetMethod$](|iterator|, "`return`").
+  1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
+     |returnMethod|.\[[Value]].
+  1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
+  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iterator|, « |reason| »).
+  1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
+     |returnResult|.\[[Value]].
+  1. Let |returnPromise| be [=a promise resolved with=] |returnResult|.\[[Value]].
+  1. Return the result of [=reacting=] to |returnPromise| with the following fulfillment steps,
+     given |iterResult|:
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Return undefined.
+ 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|,
+    0).
+ 1. Return |stream|.
+</div>
+
 <div algorithm="ReadableStreamPipeTo">
  <dfn abstract-op lt="ReadableStreamPipeTo"
  id="readable-stream-pipe-to">ReadableStreamPipeTo(|source|, |dest|, |preventClose|, |preventAbort|,
@@ -2547,49 +2590,6 @@ create them does not matter.
     |cancel2Algorithm|).
  1. Perform |forwardReaderError|, given |reader|.
  1. Return « |branch1|, |branch2| ».
-</div>
-
-<div algorithm>
- <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
- ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
-
- 1. Let |stream| be undefined.
- 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, async).
- 1. Let |startAlgorithm| be an algorithm that returns undefined.
- 1. Let |pullAlgorithm| be the following steps:
-  1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
-  1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
-     |nextResult|.\[[Value]].
-  1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
-  1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
-     given |iterResult|:
-   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
-   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
-   1. If |done| is true:
-    1. Perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
-   1. Otherwise:
-    1. Let |value| be ? [$IteratorValue$](|iterResult|).
-    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
-       |value|).
-       <!-- TODO (future): If we allow changing the queuing strategy, this Enqueue might throw.
-             We'll then need to catch the error and close the async iterator. -->
- 1. Let |cancelAlgorithm| be the following steps, given |reason|:
-  1. Let |iterator| be |iteratorRecord|.\[[Iterator]].
-  1. Let |returnMethod| be [$GetMethod$](|iterator|, "`return`").
-  1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
-     |returnMethod|.\[[Value]].
-  1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
-  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iterator|, « |reason| »).
-  1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
-     |returnResult|.\[[Value]].
-  1. Let |returnPromise| be [=a promise resolved with=] |returnResult|.\[[Value]].
-  1. Return the result of [=reacting=] to |returnPromise| with the following fulfillment steps,
-     given |iterResult|:
-   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
-   1. Return undefined.
- 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|,
-    0).
- 1. Return |stream|.
 </div>
 
 <h4 id="rs-abstract-ops-used-by-controllers">Interfacing with controllers</h4>
--- a/index.bs
+++ b/index.bs
@@ -2561,7 +2561,7 @@ create them does not matter.
  ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
 
  1. Let |stream| be undefined.
- 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, `async`).
+ 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, async).
  1. Let |startAlgorithm| be an algorithm that returns undefined.
  1. Let |pullAlgorithm| be the following steps:
   1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
--- a/index.bs
+++ b/index.bs
@@ -2574,15 +2574,19 @@ create them does not matter.
        <!-- TODO (future): If we allow changing the queuing strategy, this Enqueue might throw.
              We'll then need to catch the error and close the async iterator. -->
  1. Let |cancelAlgorithm| be the following steps, given |reason|:
-  1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
+  1. Let |iterator| be |iteratorRecord|.\[[Iterator]].
+  1. Let |returnMethod| be [$GetMethod$](|iterator|, "`return`").
   1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
      |returnMethod|.\[[Value]].
   1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
-  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iteratorRecord|.\[[Iterator]],
-     « |reason| »).
+  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iterator|, « |reason| »).
   1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
      |returnResult|.\[[Value]].
-  1. Return [=a promise resolved with=] |returnResult|.\[[Value]].
+  1. Let |returnPromise| be [=a promise resolved with=] |returnResult|.\[[Value]].
+  1. Return the result of [=reacting=] to |returnPromise| with the following fulfillment steps,
+     given |iterResult|:
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Return undefined.
  1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|,
     0).
  1. Return |stream|.
--- a/index.bs
+++ b/index.bs
@@ -2573,10 +2573,10 @@ create them does not matter.
    1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
    1. Let |done| be ? [$IteratorComplete$](|iterResult|).
    1. If |done| is true:
-    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
    1. Otherwise:
     1. Let |value| be ? [$IteratorValue$](|iterResult|).
-    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
        |value|).
  1. Let |cancelAlgorithm| be the following steps, given |reason|:
   1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
--- a/index.bs
+++ b/index.bs
@@ -485,7 +485,6 @@ interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
   static ReadableStream from(any asyncIterable);
-  static ReadableStream of(any... chunks);
 
   readonly attribute boolean locked;
 
@@ -822,12 +821,6 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  are:
 
  1. Return ? [$ReadableStreamFromIterable$](|asyncIterable|).
-</div>
-
-<div algorithm>
- The static <dfn id="rs-of" method for="ReadableStream">of(...|chunks|)</dfn> method steps are:
-
- 1. Return ? [$ReadableStreamFromIterable$](|chunks|).
 </div>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -485,6 +485,7 @@ interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
   static ReadableStream from(any asyncIterable);
+  static ReadableStream of(any... chunks);
 
   readonly attribute boolean locked;
 
@@ -821,6 +822,12 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  are:
 
  1. Return ? [$ReadableStreamFromIterable$](|asyncIterable|).
+</div>
+
+<div algorithm>
+ The static <dfn id="rs-of" method for="ReadableStream">of(...|chunks|)</dfn> method steps are:
+
+ 1. Return ? [$ReadableStreamFromIterable$](|chunks|).
 </div>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -2578,6 +2578,8 @@ create them does not matter.
     1. Let |value| be ? [$IteratorValue$](|iterResult|).
     1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
        |value|).
+       <!-- TODO (future): If we allow changing the queuing strategy, this Enqueue might throw.
+             We'll then need to catch the error and close the async iterator. -->
  1. Let |cancelAlgorithm| be the following steps, given |reason|:
   1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
   1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
--- a/index.bs
+++ b/index.bs
@@ -820,36 +820,7 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
  The static <dfn id="rs-from" method for="ReadableStream">from(|asyncIterable|)</dfn> method steps
  are:
 
- 1. Let |stream| be undefined.
- 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, `async`).
- 1. Let |startAlgorithm| be an algorithm that returns undefined.
- 1. Let |pullAlgorithm| be the following steps:
-  1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
-  1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
-     |nextResult|.\[[Value]].
-  1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
-  1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
-     given |iterResult|:
-   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
-   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
-   1. If |done| is true:
-    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
-   1. Otherwise:
-    1. Let |value| be ? [$IteratorValue$](|iterResult|).
-    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
-       |value|).
- 1. Let |cancelAlgorithm| be the following steps, given |reason|:
-  1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
-  1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
-     |returnMethod|.\[[Value]].
-  1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
-  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iteratorRecord|.\[[Iterator]],
-     « |reason| »).
-  1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
-     |returnResult|.\[[Value]].
-  1. Return [=a promise resolved with=] |returnResult|.\[[Value]].
- 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|).
- 1. Return |stream|.
+ 1. Return ? [$ReadableStreamFromIterable$](|asyncIterable|).
 </div>
 
 <div algorithm>
@@ -2576,6 +2547,42 @@ create them does not matter.
     |cancel2Algorithm|).
  1. Perform |forwardReaderError|, given |reader|.
  1. Return « |branch1|, |branch2| ».
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
+
+ 1. Let |stream| be undefined.
+ 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, `async`).
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |pullAlgorithm| be the following steps:
+  1. Let |nextResult| be [$IteratorNext$](|iteratorRecord|).
+  1. If |nextResult| is an abrupt completion, return [=a promise rejected with=]
+     |nextResult|.\[[Value]].
+  1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
+  1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
+     given |iterResult|:
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+       |value|).
+ 1. Let |cancelAlgorithm| be the following steps, given |reason|:
+  1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
+  1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
+     |returnMethod|.\[[Value]].
+  1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
+  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iteratorRecord|.\[[Iterator]],
+     « |reason| »).
+  1. If |returnResult| is an abrupt completion, return [=a promise rejected with=]
+     |returnResult|.\[[Value]].
+  1. Return [=a promise resolved with=] |returnResult|.\[[Value]].
+ 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|).
+ 1. Return |stream|.
 </div>
 
 <h4 id="rs-abstract-ops-used-by-controllers">Interfacing with controllers</h4>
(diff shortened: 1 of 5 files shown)


index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+       |value|).
+ 1. Let |cancelAlgorithm| be the following steps:
+  1. Let |returnMethod| be ? [$GetMethod$](|iteratorRecord|.\[[Iterator]], "return").
+  1. If |returnMethod| is undefined, return [=a promise resolved with=] undefined.
+  1. Let |returnResult| be ? [$Call$](|returnMethod|, |iteratorRecord|.\[[Iterator]]).
+  1. Return [=a promise resolved with=] |returnResult|.
+ 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|).
```
domenic, line 826: This needs to be ?, since startAlgorithm could throw.

index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
 </div>
 
+<div algorithm>
+ The static <dfn id="rs-from" method for="ReadableStream">from(|asyncIterable|)</dfn> method steps
+ are:
+
+ 1. Let |stream| be undefined.
+ 1. Let |iteratorRecord| be undefined.
+ 1. Let |startAlgorithm| be the following steps:
+  1. Set |iteratorRecord| to ? [$GetIterator$](|asyncIterable|, async).
```
domenic, line 805: Putting this inside startAlgorithm, instead of just doing it straightaway, means that from() also works if [Symbol.asyncIterator]() returns a promise for an async iterator. That's probably not intended.

index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+       |value|).
+ 1. Let |cancelAlgorithm| be the following steps:
+  1. Let |returnMethod| be ? [$GetMethod$](|iteratorRecord|.\[[Iterator]], "return").
```
domenic, line 822:
```suggestion
  1. Let |returnMethod| be ? [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
```

index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+  1. If |returnMethod| is undefined, return [=a promise resolved with=] undefined.
+  1. Let |returnResult| be ? [$Call$](|returnMethod|, |iteratorRecord|.\[[Iterator]]).
```
domenic, line 824: Probably we should call it with the cancel reason?

index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+  1. Return [=a promise resolved with=] |returnResult|.
+ 1. Set |stream| to ! [$CreateReadableStream$](|startAlgorithm|, |pullAlgorithm|, |cancelAlgorithm|).
+ 1. Return |stream|.
```
domenic, line 827: Can we combine steps 1, step 6, and step 7 into just "Return ? CreateReadableStream(...)"?

domenic (commented):
Very nice!! I originally was thinking we should add an async iterable type (e.g. `async sequence<>`) to Web IDL, but probably it's best to stick with `any` plus manual plumbing for now until there's a second consumer on the web platform that would use it.

index.bs
MattiasBuelens: I need *stream* to call `ReadableStreamDefaultControllerEnqueue` and `ReadableStreamDefaultControllerClose` in *pullAlgorithm*... 😕

index.bs
MattiasBuelens: Hmm, good point.

Would you prefer moving the current *startAlgorithm* steps to the "main steps", and keep the `!` here? The result should the same, I think.

MattiasBuelens: Ah, right, that would also fix the edge case where `[Symbol.asyncIterator]()` returns a promise. I'll move those steps out. 🙂

index.bs
```diff
@@ -787,6 +795,41 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+       |value|).
+ 1. Let |cancelAlgorithm| be the following steps:
+  1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
```
MattiasBuelens, line 821: It would be nicer to use [`AsyncIteratorClose()`](https://tc39.es/ecma262/#sec-asynciteratorclose) instead, but I don't think that's possible since it has an [`Await()`](https://tc39.es/ecma262/#await) in the middle. 😕

index.bs
MattiasBuelens: Good point, that would make `ReadableStream.from(new ReadableStream()).cancel("reason")` behave nicer. 👍

index.bs
```diff
@@ -787,6 +795,38 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+  1. Let |nextPromise| be [=a promise resolved with=] |nextResult|.\[[Value]].
+  1. Return the result of [=reacting=] to |nextPromise| with the following fulfillment steps,
+     given the argument |iterResult|:
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
```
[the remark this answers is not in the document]
MattiasBuelens: Ah, that reminds me: there's currently no way to change the queuing strategy. 😅

Do we want `ReadableStream.from(asyncIterable, queuingStrategy = {})`?

That might conflict with a [possible future extension](https://github.com/whatwg/streams/issues/1018#issuecomment-573450716) to pass `{ type: 'bytes' }` to create a readable byte stream. However, we could also introduce `ReadableStream.fromBytes(asyncIterable, queuingStrategy)` for that, [as suggested by Jake](https://github.com/whatwg/streams/issues/1018#issuecomment-587350179).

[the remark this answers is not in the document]
ricea: I suggest we leave that to a future extension. I think there should be no backwards-compatibility issues if we add strategy and/or byte support later?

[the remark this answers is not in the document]
MattiasBuelens: No, there shouldn't be. But when we decide to add such extension in the future, we mustn't forget to turn this `!` into a `?` and handle the error (by closing the iterator).

reference-implementation/lib/abstract-ops/ecmascript.js
```diff
@@ -34,3 +34,67 @@ exports.TransferArrayBuffer = O => {
...
+
+exports.GetIterator = (obj, hint = 'sync', method) => {
+  assert(hint === 'sync' || hint === 'async');
+  if (method === undefined) {
+    if (hint === 'async') {
+      method = exports.GetMethod(obj, Symbol.asyncIterator);
+      if (method === undefined) {
+        const syncMethod = exports.GetMethod(obj, Symbol.iterator);
+        const syncIterator = exports.GetIterator(obj, 'sync', syncMethod);
+        return syncIterator; // TODO sync to async iterator
```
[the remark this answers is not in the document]
MattiasBuelens: I made an implementation for [`CreateAsyncFromSyncIterator`](https://tc39.es/ecma262/#sec-createasyncfromsynciterator)... but I don't know how I feel about it. 😅
```javascript
function CreateAsyncFromSyncIterator(iteratorRecord) {
  const syncIterator = {
    [Symbol.iterator]: () => iteratorRecord.iterator
  };
  const asyncIterator = (async function* () {
    return yield* syncIterator;
  }());
  const nextMethod = asyncIterator.next;
  return { iterator: asyncIterator, nextMethod, done: false };
}
```
Basically, I use an async generator function that delegates to the given sync iterator (with `yield*`), and use that as the async iterator.

It *works*, but maybe it's too confusing or distracting for the *reference* implementation? 🤷 We could also leave it as it is now, with a comment explaining that the reference implementation doesn't really care if `GetIterator()` returns a sync or an async iterator.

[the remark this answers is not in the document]
ricea: That's amazing. I feel like I can just barely understand it, but it's a stretch.

I have a slight preference for leaving it as is, on the grounds that a comment saying "we don't need to do this bit" is easier to understand than the actual implementation. But on the other hand, we do try to keep as close to the standard text as possible, so it could go either way. I will let @domenic cast the deciding vote on this one.

index.bs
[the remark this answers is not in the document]
domenic: A future extension seems reasonable to me. Feel free to add a `<!-- comment -->` to the source file to remind us.

That said, it's hard enough to motivate implementers to add convenience features like async iteration in the first place; I worry that doing this in two pieces would reduce the chance of the second piece ever happening. Hmm.

index.bs
```diff
@@ -787,6 +795,42 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
      |underlyingSource|, |underlyingSourceDict|, |highWaterMark|, |sizeAlgorithm|).
 </div>
 
+<div algorithm>
+ The static <dfn id="rs-from" method for="ReadableStream">from(|asyncIterable|)</dfn> method steps
+ are:
+
+ 1. Let |stream| be undefined.
+ 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, `async`).
```
domenic, line 803:
```suggestion
 1. Let |iteratorRecord| be ? [$GetIterator$](|asyncIterable|, async).
```

Where the ES spec has special formatting or fonts, web specs ignore them all, generally. (Plus imposing our own formatting for strings.)

index.bs
domenic: Yeah, those seem largely designed for `for`-`await`-`of` consumers.

index.bs
```diff
@@ -787,6 +795,42 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
+   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
+   1. If |done| is true:
+    1. Perform ! [$ReadableStreamDefaultControllerClose$](stream.[=ReadableStream/[[controller]]=]).
+   1. Otherwise:
+    1. Let |value| be ? [$IteratorValue$](|iterResult|).
+    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](stream.[=ReadableStream/[[controller]]=],
+       |value|).
+ 1. Let |cancelAlgorithm| be the following steps, given |reason|:
+  1. Let |returnMethod| be [$GetMethod$](|iteratorRecord|.\[[Iterator]], "`return`").
```
domenic, line 821: Hmm, should we grab `return` before cancel time, or delay it as this PR does? The ES spec grabs the `next` method ahead of time... but I guess it usually delays the `"return"` method, so, yeah, this is correct as-is.

index.bs
domenic: The reason I didn't catch this is because those references don't use `|var|` syntax, so they didn't get highlighted when I clicked step 1's "stream" variable.

reference-implementation/lib/ReadableStream-impl.js
```diff
@@ -158,6 +160,57 @@ exports.implementation = class ReadableStreamImpl {
...
+    function pullAlgorithm() {
+      let nextResult;
+      try {
+        nextResult = IteratorNext(iteratorRecord);
+      } catch (e) {
+        return promiseRejectedWith(e);
+      }
+      const nextPromise = promiseResolvedWith(nextResult);
+      return transformPromiseWith(nextPromise, iterResult => {
+        if (typeof iterResult !== 'object') {
```
domenic, line 179: `null` should also throw. There might be a `typeIsObject` predicate in one of the helper files? Or maybe I killed it when we moved stuff to IDL and it became unnecessary.

(Tests would be ideal to catch this case.)

reference-implementation/lib/abstract-ops/ecmascript.js
```diff
@@ -34,3 +34,67 @@ exports.TransferArrayBuffer = O => {
 exports.IsDetachedBuffer = O => {
   return isFakeDetached in O;
 };
+
+exports.Call = (F, V, args) => {
+  if (typeof F !== 'function') {
+    throw new TypeError('Argument is not a function');
+  }
+
+  return Function.prototype.apply.call(F, V, args);
```
domenic, line 43: Slightly nicer: `Reflect.apply(F, V, args)`

reference-implementation/lib/abstract-ops/ecmascript.js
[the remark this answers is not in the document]
domenic: > The ReadableStream.from() implementation already wraps all results from next() and return() in new promises, so we don't really need the sync-to-async adapter for that.

Hmm, interesting. Maybe it shouldn't? But, [`for-await-of`](https://tc39.es/ecma262/#sec-runtime-semantics-forin-div-ofbodyevaluation-lhs-stmt-iterator-lhskind-labelset) does, indirectly, by using Await().

But what is the point of async-from-sync iterator then? I know I introduced it for a reason back when I wrote that spec...

I'll ask around.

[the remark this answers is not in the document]
domenic: It looks like AsyncFromSyncIterator is kind of pointless: https://github.com/tc39/ecma262/issues/1765

So yeah, maybe stating that, with a link to that issue thread, is best.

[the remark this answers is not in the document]
domenic: No, wait. The issue is what happens if you have a sync iterator of promises. With AsyncFromSyncIterator, you get the promises unwrapped. Without it, you get the promises.

So I think omitting this step would break test cases like `ReadableStream.from([promiseFor1, promiseFor2])`. That should be equivalent to `ReadableStream.from([1, 2])` but without the AsyncFromSyncIterator, it would give you a stream where `(await reader.read()).value` was a promise.

index.bs
MattiasBuelens: Woops, my bad. I'll add the `|var|`s.

index.bs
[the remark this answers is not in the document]
MattiasBuelens: I've added a todo for now. I'll leave this discussion open, to discuss whether we already want to add a strategy argument.

index.bs
```diff
@@ -787,6 +796,19 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
...
+ The static <dfn id="rs-from" method for="ReadableStream">from(|asyncIterable|)</dfn> method steps
+ are:
+
+ 1. Return ? [$ReadableStreamFromIterable$](|asyncIterable|).
+</div>
+
+<div algorithm>
+ The static <dfn id="rs-of" method for="ReadableStream">of(...|chunks|)</dfn> method steps are:
+
+ 1. Return ? [$ReadableStreamFromIterable$](|chunks|).
```
domenic, line 809: This doesn't really work, as |chunks| is an Infra list of `any` values, not a JS array. A separate abstract op is probably best.

But personally, I'm not sure it's worth adding `of()` vs. just `from([])`. I rarely see people use `Array.of()`.

MattiasBuelens:
I've moved the implementation to a new abstract op `ReadableStreamFromIterable`. Unfortunately, this marks most of the review comments as outdated. Sorry about that. 😞

I also added the suggested `ReadableStream.of(...chunks)` method, since it's almost trivial now.

I think I'll start working on some tests next. 🙂

index.bs
MattiasBuelens: (Oh wow, that was a fast review. 😅)

Hmm, fair point. I believe the argument behind `of()` was because people might accidentally use `ReadableStream.from(uint8Array)` and get a stream of individual bytes instead of a stream with one byte chunk ([comment](https://github.com/whatwg/streams/issues/1018#issuecomment-587350179)). But then again, that's easily fixed with `ReadableStream.from([uint8Array])`.

I'll remove `of()` for now.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1456,6 +1456,6 @@ function ReadableStreamFromIterable(asyncIterable) {
     return promiseResolvedWith(returnResult);
   }
 
-  stream = CreateReadableStream(startAlgorithm, pullAlgorithm, cancelAlgorithm);
+  stream = CreateReadableStream(startAlgorithm, pullAlgorithm, cancelAlgorithm, 0);
```
MattiasBuelens, line 1459: Setting the high-water mark to zero ensures `ReadableStream.from(otherReadableStream)` doesn't immediately start pulling from `otherReadableStream`, and matches the default HWM for `TransformStream.readable` (step 5 in [the `TransformStream` constructor](https://streams.spec.whatwg.org/commit-snapshots/660d259fa263e3c0646415a41b6e7b9a13ac4d04/#ts-constructor)).

In comparison, `ReadableStream.tee()` *does* immediately start pulling, even if no `read()` was called on any of the returned branches. In step 16 and 17 of [the `ReadableStreamTee` abstract op](https://streams.spec.whatwg.org/commit-snapshots/660d259fa263e3c0646415a41b6e7b9a13ac4d04/#readable-stream-tee), no value is passed for *highWaterMark* and thus it defaults to 1. This seems to be intentional, as there are tests that rely on this. (In particular, we rely on it to forward closing of the parent stream to the two branches.)

I *think* we want to follow the example of `TransformStream.readable` here, and use zero as HWM? Or am I missing something important?

MattiasBuelens (commented):
Found some time to write a few tests. More to come. 🙂

Don't worry too much about reviewing this right now, it's the holiday season after all. Enjoy! 😄

reference-implementation/lib/abstract-ops/ecmascript.js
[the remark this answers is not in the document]
MattiasBuelens: I added a test for the general case of [a sync iterable of promises](https://github.com/web-platform-tests/wpt/blob/9ce2d37efeb10b9b6b8df22ddd0ca33b0af936a7/streams/readable-streams/from.any.js#L52-L64), and the specific case of [an array of promises](https://github.com/web-platform-tests/wpt/blob/9ce2d37efeb10b9b6b8df22ddd0ca33b0af936a7/streams/readable-streams/from.any.js#L9-L14). These tests require a correct implementation of `CreateAsyncFromSyncIterator`, so I added the above snippet (with extra comments) to the reference implementation. 🙂

index.bs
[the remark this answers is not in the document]
MattiasBuelens: One possible use case would be if you had an (slow) async iterable, and you wanted to quickly add buffering around it:
```javascript
const bufferedReadable = ReadableStream
  .from(createSlowIterable(), new CountQueuingStrategy({ highWaterMark: 10 }));
```
That said, you can also achieve this with an intermediate `TransformStream`, which is already explained in [this example](https://streams.spec.whatwg.org/#example-transform-identity):
```javascript
const slowReadable = ReadableStream.from(createSlowIterable());
const bufferedReadable = slowReadable
  .pipeThrough(new TransformStream(undefined, new CountQueuingStrategy({ highWaterMark: 10 })));
```
(Note that this assumes that `ReadableStream.from` creates a stream with HWM = 0, [see other thread](https://github.com/whatwg/streams/pull/1083#discussion_r549040797).)

I don't know how compelling this use case is though... 🤷

reference-implementation/lib/abstract-ops/readable-streams.js
domenic: Hmm. I have a slight bias toward keeping our default, since it must have been picked for a reason :). In particular, I believe our historical intention has been that by default readable streams should try to keep something in their internal queue, whether that something is derived from a `controller.enqueue()` call in `start()`, or one in `pull()`. I think that same logic applies to deriving it from an async iterator.

Transform streams are a bit different, as we want introducing them to be more of a no-op and not cause a bunch of chunks to sit in intermediate transform-stream queues, unless explicitly requested.

On the other hand, you could view `ReadableStream.from()` as a sort of transform, such that the same no-op logic should apply. Hmm. @ricea, any thoughts?

MattiasBuelens: Oh right, I didn't update the spec text yet to pass a zero high-water mark. But I'll first wait until we decide what that high-water mark should be. 😉

> Transform streams are a bit different, as we want introducing them to be more of a no-op and not cause a bunch of chunks to sit in intermediate transform-stream queues, unless explicitly requested.

Hmm... it doesn't look like that's working as intended? 😕 

By default, a `TransformStream` is indeed created with a readable HWM of 0, but it still has a **writable HWM** of 1. So piping a stream into it *will* pull at least one chunk, to fill the writable end's queue.
```javascript
var rs1 = new ReadableStream({
  pull(c) {
    console.log('pull');
    c.enqueue('a');
  }
}, { highWaterMark: 0 });
// no "pull" is logged yet
var rs2 = rs1.pipeThrough(new TransformStream());
// logs "pull" once
```

There's also no good way to fix this. You can't set both the readable and writable HWM to 0, since then the pipe will stall:
```javascript
var rs1 = new ReadableStream({
  pull(c) {
    console.log('pull');
    c.enqueue('a');
  }
}, { highWaterMark: 0 });
var rs2 = rs.pipeThrough(new TransformStream({}, { highWaterMark: 0 }));
var r = rs.getReader();
await r.read(); // never resolves
```

MattiasBuelens: It feels like there's a connection missing somewhere... There's no way for the writable end of a transform stream to know that the readable end is being read from. 🤔 
```javascript
let { readable, writable } = new TransformStream({}, { highWaterMark: 0 });
let r = readable.getReader();
let w = writable.getWriter();

// we start reading from the readable end...
let read1 = r.read();

// ...but the writable end is still applying backpressure. :-/
w.desiredSize; // still 0
w.ready; // still pending
```

ricea: > By default, a `TransformStream` is indeed created with a readable HWM of 0, but it still has a **writable HWM** of 1. So piping a stream into it _will_ pull at least one chunk, to fill the writable end's queue.

Yes. As you observed, a writable stream with a HWM of 0 will always have backpressure. So adding an identity TransformStream to a pipe can't be a complete no-op: it always increases the total queue size by 1.

> Hmm. I have a slight bias toward keeping our default, since it must have been picked for a reason :).

In implementation practice we seem to be setting readable HWM to 0 whenever we create a platform stream, because it permits maximum control over backpressure. So I'm not sure the default of 1 is correct.

MattiasBuelens:
I can't think of any more interesting tests to add, so I'm putting this up for review. 🙂

Quick recap of the discussions so far:
* An optional `queuingStrategy` argument is currently left as a future extension. However, [Domenic is concerned](https://github.com/whatwg/streams/pull/1083#discussion_r527221755) that implementers might not be motivated to add a convenience method if they already know that they might have to revisit it in the future.
* The high water mark of the stream is currently set to 0 (instead of the default value of 1). [The question remains](https://github.com/whatwg/streams/pull/1083#discussion_r549040797) whether we should keep it at 0 (to align with `TransformStream.readable`) or revert it to 1 (to restore the default).

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Another reason for setting HWM to 0: you can treat `ReadableStream.from()` as the "inverse" of `ReadableStream.prototype.values()`.
```javascript
const it1 = ['a', 'b', 'c'];
// This async iterator should behave the same as the original
const it2 = ReadableStream.from(it1).values();
for await (const chunk of it2) { /* ... */ }
```
```javascript
const rs1 = new ReadableStream({
  start(c) {
    c.enqueue('a');
    c.enqueue('b');
    c.enqueue('c');
    c.close();
  }
});
// This stream should behave the same as the original
const rs2 = ReadableStream.from(rs1.values());
const reader = rs2.getReader();
for (let result = await reader.read(); !result.done; result = await reader.read()) { /* ... */ }
```

ricea (commented):
lgtm.

Sorry for the delay.

ricea:
Chrome is interested.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1411,3 +1413,54 @@ function SetUpReadableByteStreamControllerFromUnderlyingSource(
...
+    let nextResult;
+    try {
+      nextResult = IteratorNext(iteratorRecord);
+    } catch (e) {
+      return promiseRejectedWith(e);
+    }
+    const nextPromise = promiseResolvedWith(nextResult);
+    return transformPromiseWith(nextPromise, iterResult => {
+      if (!typeIsObject(iterResult)) {
+        throw new TypeError();
```
ricea, line 1433: Please give the exceptions a message to help us grep for them when debugging.

ricea (approved):
lgtm

ricea:
I realised that for synchronous iterators we could get optimum performance if we just enqueued everything in startAlgorithm. It's probably not worth the extra complexity, and I guess an iterator could return an infinite stream, which would be bad. What do you think?

MattiasBuelens:
I think we should only do that if we are *absolutely sure* that the user-visible behavior is exactly the same. So we shouldn't do this with arbitrary synchronous iterators, since they can be infinite (as you correctly noted) or they might be modifying some external state inside `next()` (such as variables or synchronous I/O).

That doesn't mean we can't allow some optimizations. For example, if the input is an `Array` with its *original* `[Symbol.iterator]` method, then the implementation *can* indeed use a "fast path" for such a case. The same logic could apply to other built-in iterables like `Set` or `Map` or `Uint8Array` (although I don't know if these are worth optimizing).

ricea:
Thanks for the feedback. I have a suspicion that it's detectable, but I don't know how.

I'm ready to land this if you are.

MattiasBuelens:
Good to go. 🙂

ricea:
Ah, wait, we need interest from another implementer. @youennf, @evilpie, any thoughts?

MattiasBuelens:
Uhoh... It looks like optimizing for arrays might not even work. 😕

You can `push()` new elements to an array *while you're iterating it*:
```javascript
let array = ['a', 'b'];
const it = array[Symbol.iterator]();

it.next(); // { value: 'a', done: false }
it.next(); // { value: 'b', done: false }

array.push('c');

it.next(); // { value: 'c', done: false }
it.next(); // { value: undefined, done: true }
```

So by that logic, this should also work:
```javascript
let array = ['a', 'b'];
const rs = ReadableStream.from(array);
const reader = rs.getReader();

await reader.read(); // { value: 'a', done: false }
await reader.read(); // { value: 'b', done: false }

array.push('c');

await reader.read(); // { value: 'c', done: false }
await reader.read(); // { value: undefined, done: true }
```

With the current implementation, this does indeed work. `ReadableStream.from()` will call `it.next()` only when a new chunk is needed (since HWM = 0), so it *will* still see the newly added element.

However, if we add a special case for arrays that *immediately* iterates over the array and enqueues all its elements, then the above snippet would no longer work.

How do we feel about this? Should we allow this, and make this optimization impossible? (If so, I'll add a test for the above.) Or not?

ricea:
That's really interesting.

I don't like either option, but I'm leaning towards special-casing arrays.

My reasoning is that I want people to use the `ReadableStream.from([chunk]).pipeTo(writable)` pattern, but if it performs poorly then they either won't use it or they'll have a bad time.

domenic:
I'm doubtful that special-casing arrays would give serious performance benefits, and so I'd prefer we keep things generic (with uniform observable behavior) unless we have some performance data that it'd be problematic. I think we're talking about avoiding a couple microtasks, and some `{ value, done }` wrapping/unwrapping, which doesn't seem like it'd be significant.

ricea:
It depends on your definition of "serious performance benefits".

I went away and implemented it, and in a microbenchmark
```
ReadableStream.from([chunk]).pipeTo(writable);
```
takes 83μs with the array optimisation and 139μs without it.

Obviously with a larger array the performance difference would be much more striking, but I mostly care about the performance of the single-chunk case.

Unfortunately the presence of thenables in the array complicates things. I didn't go to the trouble of matching the behaviour that the ES standard gives to them. It would require falling back to an async algorithm the first time we see a thenable. The performance impact would not be significant in the normal case, but the implementation would be messy. I don't actually like the behaviour of following the thenable, so I'd prefer to ignore the problem, but I don't think we can.

My draft CL is at https://chromium-review.googlesource.com/c/chromium/src/+/3023854.

By the way @MattiasBuelens this is some amazing specification work. I didn't find a single issue while implementing.

domenic:
I think the optimization is even trickier :(. Checking using V8's `IsArray()` doesn't guarantee that the optimization is nonobservable, because someone could have messed with `Array.prototype` to insert a getter for `0`, or override the getter for `[Symbol.iterator]`, or something like that. I suspect V8 has ways of detecting such problems so that they can deopt arrays generally, but I doubt they're part of the public API, and they're probably intertwined with the VM pipeline instead of something that could be easily exposed.

Having actual data is great. I'm unsure how to interpret the results... I mean, nanoseconds are very small. But a 1.67x overhead feels large. I'd be interested in @syg's perspective (or that of his colleagues) as to whether this sort of thing feels like something we could optimize if needed in the future, or whether it's fundamentally going to be about 1.67x more work to do the whole async-to-sync-iterator dance vs. just iterating an array.

At a conceptual level, one angle is to wonder what `ReadableStream.from()` is supposed to mean. Some of the discussions in #1018 indicated it would be a general-purpose "feed me anything" API. Notably, the original idea had some behavior for blobs, and we discussed special-casing `ReadableStream.from(uint8Array)` or `ReadableStream.from(string)`---even if just to throw to avoid people getting a bunch of single-byte/code point chunks. But the current spec draft here leans hard into "treat the argument always as an async iterable". That has some elegance, but perhaps bad consequences for the string and Uint8Array case.

The relevance of the above paragraph is: if we decide to special-case more types, then special-casing arrays seems natural.

Also, I'm very excited that you created a draft CL!!

evilpie:
Special casing arrays vs other iterables seems wrong to me. Is there no room for just optimizing array iterables more? I know SpiderMonkey has a specific [optimization for iterating over arrays](https://searchfox.org/mozilla-central/source/js/src/vm/ForOfIterator.cpp#101). (Please note that I don't want to make any performance claims here, this might well still be slower than Chrome anyway)

MattiasBuelens:
> Is there no room for just optimizing array iterables more? I know SpiderMonkey has a specific [optimization for iterating over arrays](https://searchfox.org/mozilla-central/source/js/src/vm/ForOfIterator.cpp#101).

It looks like [`ForOfIterator`](https://searchfox.org/mozilla-central/rev/77f0b36028b2368e342c982ea47609040b399d89/js/public/ForOfIterator.h#29-49) is only meant to imitate the behavior of a `for..of` loop? For `ReadableStream.from()`, we need the behavior of a `for await..of` loop instead, so you'd still have to wrap it with `CreateAsyncFromSyncIterator`.

---

I don't think it's a valid optimization to *exhaustively* iterate an array. For such optimization to be unobservable, you'd have to prove that the array can never be mutated after constructing the stream, and that all of its elements are non-thenables (so they won't trigger a delay when using `it.next()`). This seems like too complex of a condition to check at runtime.

If I understand the code from the CL correctly, you already have special handling for synchronous iterators in `ScriptAsyncIterator`. Perhaps it can also optimize for the common case where `it.next().value` is a non-thenable, and thus `AsyncFromSyncIteratorContinuation` doesn't need to attach an `UnwrapFunction`? Maybe that can save a microtask? Or is the overhead not coming from using too many microtasks?

ricea:
To be clear, I was only trying to evaluate the performance of treating Array as a special case, I was not trying to make the behaviour match the async iterator case.

Having thought about it a bit more, making Array a special case is not really justifiable since as @domenic noted it gets weird if someone has tampered with the object or the prototype.

So I've experimented with making sync iterators a special case instead. As with the Array case, I enqueue everything synchronously when the stream is created. And as with the Array case, this means it will fill up your memory and crash if you put an infinite iterator in there.

The performance is actually weirdly good. I'm getting a median of 82μs, which is within the margin of error of the Array special case. I can't explain why it's so fast, but I'm not complaining.

Here's what I'm proposing:

1. Check if _asyncIterable_ has an `@@iterator` method. If it does then call GetIterator(_asyncIterable_, sync) and synchronously enqueue all the values in the startAlgorithm.
2. Otherwise, call GetIterator(_asyncIterable_, async) and lazily pull items as before.
3. Tell people that if they want to use an infinite iterator with `from()` they have to make it async.

ricea:
For reference, here's the benchmark I'm using: https://gist.github.com/ricea/ca766d90ba985dfb7e1cd54bba241565

If you want to run it, don't open the console until the benchmark is over or it will mess up the results. Also, you need to divide by 10000 in your head to get the per-iteration time.

MattiasBuelens:
> 3\. Tell people that if they want to use an infinite iterator with `from()` they have to make it async.

If all they have is an infinite synchronous iterable, they might find it difficult to turn it into an asynchronous iterable. Effectively, they'd have to resort to things [like this](https://github.com/whatwg/streams/blob/2d40fe365180b231dc9cd4509cd013f7b690abca/reference-implementation/lib/abstract-ops/ecmascript.js#L73-L75):
```javascript
ReadableStream.from(async function* () {
  return yield* syncIterable;
}());
```
I don't know if we want to encourage the use of this sort of pattern. But then again, I also don't know how often people will actually want to use an infinite synchronous iterable. 🤷

domenic:
Hmm, I haven't thought about the tradeoffs in depth yet, but disallowing infinite sync iterables would preclude this from working:

```js
ReadableStream.from(Number.range(0, Infinity));
```

using https://github.com/tc39/proposal-Number.range

We could have separate from() and fromSync() methods? It's a bit more justifiable since the behavior is so different.

MattiasBuelens:
It would be nice to know *where* we are losing time. Is it because we have to go through more microtasks? Would it be worth trying to optimize the `ReadableStream` machinery itself for cases where `startAlgorithm` and `pullAlgorithm` run *synchronously*? That is:
* In `SetUpReadableStreamDefaultController`, don't wrap `startResult` in a promise, but instead set `[[started]]` to true immediately.
* In `ReadableStreamDefaultControllerCallPullIfNeeded`, allow `[[pullAlgorithm]]` to return a non-promise (or just ignore its result altogether) and set `[[pulling]]` to false immediately after calling it.

Or is it not about the number of microtasks, and are we just running more code within each microtask? 🤔

ricea:
I added support for backpressure for sync iterators (calling `next()` inside pullAlgorithm) to see how much difference it would make. I also redid my measurements with a million iterations.

Behaviour | Results (μs)
----|----
No special cases | 127
Special case sync iterables, respecting backpressure | 117
Special case sync iterables, ignoring backpressure | 103
Special case arrays, ignoring backpressure | 92

I think there's a couple of factors at play:
1. The ES standard sync-to-async conversion is just really complicated and expensive
2. Doing things asynchronously is slower than doing them synchronously, not just because of task dispatch, but because you need to keep more state on the heap

Although in my new results "Special case sync iterables, ignoring backpressure" is no longer as fast as "Special case arrays, ignoring backpressure", it could definitely be made as fast by exposing the iterations algorithm from the JS engine. Whereas the ES standard sync-to-async behaviour is not going to be that fast with any amount of heroics.

index.bs
```diff
@@ -475,6 +481,8 @@ The Web IDL definition for the {{ReadableStream}} class is given as follows:
 interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
+  static ReadableStream from(any asyncIterable);
```
jasnell, line 484: It would be excellent to see this move forward. With regards to the question about the special case for byte sources, the signature here can echo the signature for the `new ReadableStream()` constructor...

```js
const readable1 = ReadableStream.from(iter, new CountQueueingStrategy({ highWaterMark: 10 });

const readable2 = ReadableStream.from({ source: iter, type: 'bytes', autoAllocateChunkSize: 1024 });
```

That is, make the first argument either a dictionary or an iterator...

```
dictionary FromInit {
  (AsyncIterator | Iterator) source;
  ReadableStreamType type;
  unsigned long long autoAllocateChunkSize;
}
```

MattiasBuelens:
It's been a while, but shall I add [the proposed test from my previous comment](https://github.com/whatwg/streams/pull/1083#issuecomment-876767187) to the WPT PR? That is: make it a normative requirement that `ReadableStream.from(array)` iterates lazily, such that `array.push()` still works?

domenic:
@evilpie @annevk ping on https://github.com/whatwg/streams/pull/1083#issuecomment-942780668 :

> > We haven't fully evaluated this proposal, but we aren't opposed. To me personally this looks good to me without any special casing.
> 
> Thanks. Would you like us to wait for a full evaluation before merging it into the spec? Or should we go ahead?

Also always interested to hear from WebKit, e.g. @youennf.

reference-implementation/lib/ReadableStream.webidl
```diff
@@ -2,6 +2,8 @@
 interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
+  static ReadableStream from(any asyncIterable);
```
saschanaz, line 5: Should this be fromAsync to match TC39's [Array.fromAsync(asyncIterable) proposal](https://github.com/tc39/proposal-array-from-async/), and to explicitly show that it supports async iterable unlike Array.from? (But that one returns a promise while this one does not 🤔)

MattiasBuelens: It makes sense for `Array.from(syncIterable)` and `Array.fromAsync(asyncIterable)` to be two different methods, since they also return different results (`Array` versus `Promise<Array>`).

I don't think such distinction is necessary from `ReadableStream`. We're never going to need *both* `ReadableStream.from()` and `ReadableStream.fromAsync()`. A `ReadableStream` is *inherently* asynchronous, so we never need to return a `Promise<ReadableStream>`.

Node.js also has [`Readable.from()`](https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#streamreadablefromiterable-options), which also accepts both sync and async iterables.

I think `ReadableStream.from()` is still be the better option. It makes it clear that this is *the* primary method to convert anything into a `ReadableStream`. A `from()` method usually also implies "idempotency", that is: passing an array to `Array.from()` returns a new array. On the other hand, `fromAsync()` feels more "foreign".

saschanaz: Makes sense, LGTM then.

mgaudet:
Just to weigh in here: I think we (Mozilla) are generally supportive of the proposal; it's a great user feature. 

I did a scan of the specification today, and while others have pointed out lots of details, at a high level it seems to nicely accomplish the goals :)

evilpie:
I started working on a draft implementation in https://bugzilla.mozilla.org/show_bug.cgi?id=1772772.

My feedback so far is that the _cancelAlgorithm_ doesn't seem to follow what seems to be the usual pattern of JS behavior. I think the reason is that you want to use the value returned by the `return` function. Maybe we should just use `IteratorClose` and resolve with undefined?

mgaudet:
(Further to the above, it also doesn't seem to follow the pattern of the other _cancelAlgorithm_s in the spec, all of those I spotted which either resolve with undefined, or throw a result)

MattiasBuelens:
Thanks for the feedback!

> My feedback so far is that the _cancelAlgorithm_ doesn't seem to follow what seems to be the usual pattern of JS behavior. I think the reason is that you want to use the value returned by the `return` function.

> Further to the above, it also doesn't seem to follow the pattern of the other _cancelAlgorithm_s in the spec, all of those I spotted which either resolve with undefined, or throw a result)

You're right, the resolution value of _cancelAlgorithm_ doesn't matter, because [`ReadableStreamCancel`](https://streams.spec.whatwg.org/commit-snapshots/e9355ce79925947e8eb496563d599c329769d315/#readable-stream-cancel) replaces it with `undefined` anyway.

I looked again at `IteratorClose`. If I read this correctly, then:
1. If there's a `return()` method, try calling it with no arguments.
2. If the original completion was a `throw`, return it.
3. If the completion from calling `return()` (in step 1) was a `throw`, return it (thus *replacing the original completion*).
4. If the return value from `return()` is not an object (i.e. not an iterator result), throw a `TypeError`.
5. Otherwise, return the original completion.

I noticed a couple of things:
* _cancelAlgorithm_ calls `return()` with the given cancel _reason_, whereas `IteratorClose` calls it with no arguments. It turns out that `return()` is only ever called *with* an argument in the context of [`yield *` inside (async) generator functions](https://tc39.es/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation)... 🤔 So I suppose it makes more sense to leave it out, and align closer to `for..await.of` semantics?
* _cancelAlgorithm_ is missing the check in step 4.

So indeed, as currently specified, _cancelAlgorithm_  is a bit sloppy... 😬 

> Maybe we should just use `IteratorClose` and resolve with undefined?

Ideally, we would use `AsyncIteratorClose`. But as discussed in https://github.com/whatwg/streams/pull/1083#discussion_r520908762, there's an `Await()` in the middle which (we think) wouldn't be allowed in the Streams specification.

We also cannot use `IteratorClose`, because then we would finish closing the iterator too early (since we don't await the promise from `return()`. Also, in step 4 we would be checking if the *promise for the return value* is an object (which it always is), whereas really we need to be checking the *resolution value*.

---

To fix this, we should match the behavior of `AsyncIteratorClose` *exactly*. If we cannot use the abstract operation directly, then I suppose we have to make a copy of it that uses [Web IDL's "react to promise"](https://webidl.spec.whatwg.org/#dfn-perform-steps-once-promise-is-settled) instead of `Await()`? Or could we somehow make `Await()` work? 😕

MattiasBuelens:
> * _cancelAlgorithm_ calls `return()` with the given cancel _reason_, whereas `IteratorClose` calls it with no arguments. It turns out that `return()` is only ever called _with_ an argument in the context of [`yield *` inside (async) generator functions](https://tc39.es/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation)... 🤔 So I suppose it makes more sense to leave it out, and align closer to `for..await.of` semantics?

Hmmm... but if we removed the argument for `return()`, then we'd no longer have `ReadableStream.from` as the "reverse" of `ReadableStream.prototype.values` (see https://github.com/whatwg/streams/pull/1083#discussion_r598005739), because the cancel reason would no longer be propagated. 😕 
```javascript
let rs1 = new ReadableStream({
  cancel(reason) { console.log(reason) }
});
let rs2 = ReadableStream.from(rs1);
rs2.cancel("boom!");
// -> rs1 gets cancelled with `undefined` (should have been "boom!")

**Comment by evilpie** [comment]
Thanks for your example. I think this is a good enough justification. `AsyncGenerator.prototype.return` accepts an value argument so we might as well just use it. I don't know who was heavily involved in the design of AsyncGenerator at TC39, but it might still be nice to ask what they think.

[My implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1772772) should now be feature complete, but I haven't ran your tests yet.

**Comment by jasnell** [comment]
What's the status on getting this merged? :-)

**Comment by andreubotella** [comment]
Hi. In the WinterCG effort we're interested in this PR and willing to help push it towards the finish line.

My understanding is that there is only one remaining issue to clarify about the semantics/behavior of `ReadableStream.from`, which is whether canceling the stream should call `return` with the cancel reason. And per https://github.com/whatwg/streams/pull/1083#issuecomment-1147591568 and https://github.com/whatwg/streams/pull/1083#issuecomment-1172950193 it seems like there are good reasons to pass the cancel reason.

Is there anything else that's not yet settled?

**Comment by andreubotella** [comment]
@MattiasBuelens We're still interested in pushing this PR as part of [WinterCG](https://wintercg.org/), as well as contribute to getting it implemented in browsers. There doesn't seem to be any remaining spec issues – especially since per https://github.com/whatwg/streams/pull/1083#issuecomment-1147591568 it would be better to not use `AsyncIteratorClose`, and therefore the TC39 `Await()` consideration wouldn't apply. If so, is anything else blocking this PR?

**Comment by MattiasBuelens** [comment]
Sorry for the delay, I didn't have much free time to do spec work lately.

I've added this extra check from `IteratorClose`, which came up in https://github.com/whatwg/streams/pull/1083#issuecomment-1147557078:
> If the return value from return() is not an object (i.e. not an iterator result), throw a TypeError.

I've also added a test for this case. For me, this resolves the last outstanding spec issue.

---

It looks like the tests are failing because `wpt-runner` is a bit behind the latest changes in WPT? I'll see what I can do about that.

File: index.bs [hunk 18]
```diff
@@ -2534,6 +2549,49 @@ create them does not matter.
  1. Return « |branch1|, |branch2| ».
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
```
domenic, line 2554: Editorial nit: this has only one call site, so it might be best to inline it?

Otherwise, it'd be good to alphabetize it within the abstract ops section.

MattiasBuelens: I prefer it as an abstract op, in case we ever want to export it for other specifications. Besides, all of the other `ReadableStream` methods also have corresponding abstract ops.

I'll fix the alphabetical order though.

MattiasBuelens:
@domenic or @ricea: Tests have landed and implementation bugs have been filed. Okay if I merge? 🙂

MattiasBuelens:
I've also opened implementation bugs for [Deno](https://github.com/denoland/deno/issues/19417) and [Node](https://github.com/nodejs/node/issues/48389).

I just realized that I forgot to write the "For web developers" documentation for the new `from()` method... 😓 Except a follow-up PR soon.

PR #1087, opened 2020-11-17
Editorial: Remove redundant steps from AcquireDefaultReader & BYOBReader

nidhijaju:
The second step in both AcquireReader() functions call their respective SetUp() functions, which is redundant as SetUp() is already called inside the constructor in the first step. Hence, this change removes these redundant steps from the standard.

***
<a href="https://whatpr.org/streams/1087.html" title="Last updated on Nov 17, 2020, 4:10 AM UTC (d92e3a5)">Preview</a> | <a href="https://whatpr.org/streams/1087/670bb6a...d92e3a5.html" title="Last updated on Nov 17, 2020, 4:10 AM UTC (d92e3a5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1998,7 +1998,6 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
  the following steps:
 
  1. Let |reader| be a [=new=] {{ReadableStreamBYOBReader}}.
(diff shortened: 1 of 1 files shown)


nidhijaju:
Actually, this is invalid because `new` in the Standard doesn't actually call the SetUp() according the WebIDL description.

PR #1088, opened 2020-11-24
Update Byte Streams Explainer

nidhijaju:
Based on a response on https://github.com/w3ctag/design-reviews/issues/567, this pull request adds some relevant sections (e.g. the goals, non-goals, alternatives, and exact API proposed) to the explainer to provide more clarity on the Readable Byte Streams API.

### DIFF
--- a/byte-streams-explainer.md
+++ b/byte-streams-explainer.md
@@ -16,9 +16,9 @@ readers can reduce the number of times we run garbage collection, because we can
 ## API Proposed
 
 *   Constructing a Readable Byte Stream
-    *   `new ReadableStream({ type: "bytes" })`
+    *   <code>new [ReadableStream](https://streams.spec.whatwg.org/#rs-constructor)({ type: "bytes" })</code>
 *   Getting a BYOB reader
-    *   `getReader({ mode: "byob" })`
+    *   <code>[getReader](https://streams.spec.whatwg.org/#rs-get-reader)({ mode: "byob" })</code>
 *   As part of the implementation, there are 3 main classes that will be added to the Streams API:
     *   The [ReadableStreamBYOBReader](https://streams.spec.whatwg.org/#byob-reader-class) class
         *   This class represents a BYOB reader designed to be vended by a `ReadableStream` instance.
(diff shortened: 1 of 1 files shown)


byte-streams-explainer.md
```diff
@@ -86,3 +102,24 @@ function makeReadableByteStream() {
 With this in hand, we can create and use BYOB readers for the returned `ReadableStream`. The adaptation between the
 low-level byte tracking of the underlying byte source shown here, and the higher-level chunk-based consumption of
 a default reader, is all taken care of automatically by the streams implementation.
+
+
+## Goals
+
+*   Support for {type: “bytes”} in the [ReadableStream constructor](https://streams.spec.whatwg.org/#rs-constructor).
```
yutakahirano, line 109: I think these are means rather than goals. How about the following?

 - Provide a way to represent a ReadableStream for bytes efficiently.
 - Avoid races caused by multiple access for the same buffer.

nidhijaju: Thank you for the suggestion! I've changed the goals section to have those points instead.

byte-streams-explainer.md
```diff
@@ -86,3 +102,23 @@ function makeReadableByteStream() {
 With this in hand, we can create and use BYOB readers for the returned `ReadableStream`. The adaptation between the
 low-level byte tracking of the underlying byte source shown here, and the higher-level chunk-based consumption of
 a default reader, is all taken care of automatically by the streams implementation.
+
+
+## Goals
+
+*   Provide a way to represent a [ReadableStream](https://streams.spec.whatwg.org/#rs-class) for bytes efficiently.
+*   Avoid races caused by multiple access for the same buffer.
```
ricea, line 110: Add `* Permit buffer re-use to reduce GC churn`.

byte-streams-explainer.md
```diff
@@ -86,3 +102,23 @@ function makeReadableByteStream() {
...
+
+## Non-Goals
+
+*   Ports for non-binary types will not be supported; only Bytes, Int16, etc. are supported by the default readable
```
ricea, line 114: I find this a little confusing, since the default readable stream supports all types. How about something like

> Non-binary chunk types will not be supported. They can still use the default type of readable stream.

?

byte-streams-explainer.md
```diff
@@ -86,3 +102,23 @@ function makeReadableByteStream() {
...
+stream and readable byte stream implementations.
+*   Shared array buffers will still not be supported. Currently, we always detach buffers, but shared array buffers
```
ricea, line 116: I don't think we need the word "still" here.

ricea (commented):
Great! Just a few tweaks needed.

byte-streams-explainer.md
nidhijaju: I agree, that makes more sense. Thank you!

ricea (approved):
lgtm. Let's see if @domenic has any comments before committing.

yutakahirano (approved):

PR #1092, opened 2020-12-01
Editorial: refactor autoAllocateChunkSize checking

domenic:
This throws the exception only from
SetUpReadableByteStreamControllerFromUnderlyingSource, i.e. from parts
of the spec reached by arbitrary web-developer provided values, instead
of from SetUpReadableByteStreamController, which is deeper in.

Then, we can assert that by the time the value reaches
SetUpReadableByteStreamController, it's positive. This makes any other
specs doing the wrong thing into assertion failures, reserving thrown
exceptions for web-developer provided zeros.

/cc @nidhijaju

***
<a href="https://whatpr.org/streams/1092.html" title="Last updated on Dec 1, 2020, 4:53 PM UTC (ac10db6)">Preview</a> | <a href="https://whatpr.org/streams/1092/946fac9...ac10db6.html" title="Last updated on Dec 1, 2020, 4:53 PM UTC (ac10db6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3247,8 +3247,7 @@ The following abstract operations support the implementation of the
  1. Assert: |stream|.[=ReadableStream/[[controller]]=] is undefined.
  1. If |autoAllocateChunkSize| is not undefined,
   1. Assert: ! [$IsInteger$](|autoAllocateChunkSize|) is true.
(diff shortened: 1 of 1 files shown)


ricea (approved):
lgtm

Issue #1074, opened 2020-08-26
GenericTransformStream mixin vs. subclassing


domenic:
https://github.com/whatwg/streams/pull/1073 works to formalize and formally recommend the pattern used by https://encoding.spec.whatwg.org/ and https://wicg.github.io/compression/, of wrapping up an internal `TransformStream` and re-exposing it through `readable` and `writable` properties.

However, I've lost track of why we do this, instead of subclassing `TransformStream`.

NativeFS shows [how to subclass `WritableStream`](https://wicg.github.io/native-file-system/#api-filesystemwritablefilestream), and #1073 makes it much easier by allowing you to reduce much of the boilerplate there to `[=WritableStream/Set up=] [=this=] with |writeAlgorithm| and |closeAlgorithm|`. It seems we could similarly make subclassing `TransformStream` easy.

This isn't very consequential. In general I expect the most noticable effect to be on `instanceof`. I don't expect us to add any significant public APIs to `TransformStream` that we'd want subclasses to inherit.

However, I am a bit concerned about recommending a totally different pattern for `TransformStream` than we do for `WritableStream` and `ReadableStream`. So I thought it'd be worth raising, and seeing if perhaps we want to change course.

ricea:
See https://github.com/whatwg/encoding/issues/72#issuecomment-396882904 and so on for the historical background.

I'd summarise it as: subclassing from TransformStream places constraints on implementations that don't benefit users. In the case of ReadableStream and WritableStream there's a lot of stuff you get from subclassing, but TransformStream is just a {readable, writable} pair.

Since this debate happened, there have been a few relevant developments:
- Blink has switched to WebIDL for TransformStream, which makes subclassing it technically easy for Blink, not necessarily other browsers.
- Transferable streams have added a kind of "fake" TransformStream which just wraps an existing {readable, writable} pair while bypassing the internal machinery.
- All specs so far using GenericTransformStream delegate to TransformStream internally anyway.

I still prefer to have the GenericTransformStream mixin. Maybe we could make TransformStream include the mixin as well?

domenic:
Right, OK, thank you for digging that up.

My main concern is that, looking at the section I wrote up in #1073, it looks kind of pointless. The wrapping is basically isomorphic to subclassing; instead of setting up `this` and operating on it, you create `this's transform` and operate on `this's transform`. And, this wrapping is inconsistent with how we handle specializations of `WritableStream` and `ReadableStream`.

[...]

PR #1110, opened 2021-03-10
Remove "create" algorithms for other specifications

### LINKED ISSUE #1084: Using streams from other specs: you can't really use "create"
The algorithm for other specs to [create](https://streams.spec.whatwg.org/#readablestream-create) a readable stream isn't really usable, at least for the case of a readable stream with a pull underlying source.

In particular, you need to pass a _pullAlgorithm_. But that _pullAlgorithm_ will need to [enqueue](https://streams.spec.whatwg.org/#readablestream-enqueue), close, etc. the stream. But the stream hasn't been created yet!

I'm not sure what the right fix is here. People can work around it using Web IDL's "new" + our [set up](https://streams.spec.whatwg.org/#readablestream-set-up), but that kind of sucks.

I haven't checked if this same problem applies to other cases (e.g. push underlying sources, writable streams, transform streams) but I suspect it might.

domenic:
Closes #1084 by encouraging use of "set up" instead. Also partially addresses #1074 by softening the language about subclassing vs. mixins.

I am willing to send PRs to fix specs that use "create". The ones I am aware of are:

- Fetch (/cc @annevk), which sidesteps the problem discussed in #1084 by having parts of the fetch algorithm, separate from the pullAlgorithm, enqueue into the stream.
- Compression streams, which suffer from the problem in #1084: they reference an undeclared variable _ds_.

https://w3c.github.io/webtransport/ also uses "create" but the spec text there is full of TODOs and doesn't seem to create in a well-defined way, so I'm not sure what to do there. /cc @yutakahirano.

***
<a href="https://whatpr.org/streams/1110.html" title="Last updated on Mar 10, 2021, 8:58 PM UTC (410b42f)">Preview</a> | <a href="https://whatpr.org/streams/1110/4b5b3cd...410b42f.html" title="Last updated on Mar 10, 2021, 8:58 PM UTC (410b42f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -60,6 +60,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: Type; url: #sec-ecmascript-data-types-and-values
  text: TypeError; url: #sec-native-error-types-used-in-this-standard-typeerror; type: exception
  text: map; url: #sec-array.prototype.map; type: method; for: Array.prototype
+url: https://wicg.github.io/compression/#compressionstream; spec: COMPRESSION; type: interface; text: CompressionStream
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


MattiasBuelens (approved):

yutakahirano:
After https://github.com/w3c/webtransport/pull/220, I'll specify creation of the ReadableStream for incoming datagrams in more detail. So I will be able to fix the WebTransport part altogether, I think.

annevk:
Would passing the stream to the pull algorithm be too weird? Anyway, seems fine. It's probably not worth having both create and set up.

domenic:
> Would passing the stream to the pull algorithm be too weird? Anyway, seems fine. It's probably not worth having both create and set up.

That was indeed my first thought, and I started writing it up even, but then I looked at the usage sites and a lot of them were already using "set up" and so I thought it'd be simpler to just have one algorithm.

domenic:
@ricea can you review?

index.bs
```diff
@@ -6152,14 +6132,21 @@ to grow organically as needed.
  1. Perform ! [$SetUpReadableStreamDefaultController$](|stream|, |controller|, |startAlgorithm|,
     |pullAlgorithmWrapper|, |cancelAlgorithmWrapper|, |highWaterMark|, |sizeAlgorithm|).
 
- <p class="note">This algorithm is mainly intended for use in the [=constructor operations=] of any
- {{ReadableStream}} subclasses.
+ <div class="example" id="example-set-up-rs">
+  Creating a {{ReadableStream}} from other specifications is thus a two-step process, like so:
+
+  1. Let |readableStream| be a [=new=] {{ReadableStream}}.
+  1. [=ReadableStream/Set up=] |readableStream| given….
```
ricea, line 6139: This needs to say something about how to use _readableStream_ in the creation of the algorithms.

domenic:
> CompressionStreams are not broken: the ds variable comes from the value of this when transformAlgorithm is created. TextEncoderStream and TextDecoderStream do the same thing.

I guess you're saying, they're a bit broken, but they'd be fixed if you replaced _ds_ with "this"? Right now [the DecompressionStream constructor](https://wicg.github.io/compression/#dom-decompressionstream-decompressionstream) references an undefined _ds_ variable.

I agree that seems to work, but it's only because the somewhat-unusual choice we've made (#1074) of having different types of transform streams be wrappers instead of subclasses. That means there's an easy way to indirectly capture the stream by passing "this" and then later referencing "this's transform" (but only after "this's transform" has been set up).

Taking a step back, I guess we have the following sorts of consumers:

- Subclassers (of ReadableStream and WritableStream)
  - They need to use "set up" for sure
- Wrappers (of TransformStream)
  - They can use create + the this's transform trick
- Wrappers (of ReadableStream and WritableStream), such as fetch `Response` or Web RTC Encoded Transform (insertable streams)
  - They can use create in a natural way *only if* they enqueue from outside the pullAlgorithm, like fetch does.
  - Or they could use a custom level of indirection, such as: "Let stream be null. Let pullAlgorithm be the following steps: ... enqueue chunk into stream. Set stream to the result of creating a ReadableStream given pullAlgorithm."
     - This is even trickier if you want to define your pullAlgorithm to call a separate set of steps. Then you need a mutable structure, e.g. "Let holder be an empty list. Let pullAlgorithm be the following steps: perform the pull algorithm given holder. Set holder[0] to the result of creating a ReableStream given pullAlgorithm. The pull algorithm, given a list holder, is: ... enqueue chunk in holder[0]."
  - Otherwise they have to use Web IDL new + set up.

I don't like this inconsistent landscape, which is my motivation for moving everything to "set up". WDYT?

ricea:
Thanks for the clear summary of the situation.

> I don't like this inconsistent landscape, which is my motivation for moving everything to "set up". WDYT?

I'm confused by the behaviour of WebIDL's `new`, but that's a pre-existing issue. Other than that, this is fine. Let's do it.

domenic:
Great, thanks! But:

> Merging can be performed automatically with 1 approving review.

:)

ricea (approved):
lgtm

Issue #1111, opened 2021-03-19
Teeing a readable byte stream


MattiasBuelens:
Currently, teeing a readable byte stream always returns two "default" readable streams. This prevents developers from using a BYOB reader on a tee'd byte stream.

I think we should change `ReadableStreamTee` to return two *readable byte streams* when given a readable byte stream as its input. This should also help with e.g. whatwg/fetch#267: when we make `Response.body` a readable byte stream, we need to make sure that `Response.clone()` creates two new *byte streams* as well.

An easy solution would be to start from [the current `ReadableStreamTee` abstract op](https://streams.spec.whatwg.org/commit-snapshots/4b5b3cdb0f22a791fc1a8e734296a32a94ad753a/#readable-stream-tee), and adapt it a bit:
* Use `CreateReadableByteStream` instead of `CreateReadableStream` for creating the two branches. (Note that this abstract op was removed in #1035 because it was unused, but it shouldn't be difficult to bring it back.)
* Use the `ReadableByteStreamController` and its abstract ops instead of a `ReadableStreamDefaultController`.
* Clone every read chunk, and enqueue the original chunk and the cloned chunk to the two branches. (If one of the two branches is already cancelled, we don't need to clone the chunk.)

Note that such an implementation would still use a `ReadableStreamDefaultReader` to read the byte chunks. This means any BYOB request from a tee'd stream cannot reach the original stream, so the original stream must always allocate a new buffer for every chunk and cannot make an informed decision on how big that buffer should be. It may be worth investigating if we can switch to a `ReadableStreamBYOBReader` inside the teeing logic whenever a tee'd branch has a pending BYOB request.

domenic:
Right, I think the blocking question here has always been whether we want to do the easy solution or come up with a harder solution.

PR #1114, opened 2021-03-20
Support teeing readable byte streams

MattiasBuelens:
With this PR, teeing a readable byte stream will now create two *readable byte streams* (instead of two "default" readable streams). See #1111 for context.

This implements the "hard" solution: when a branch is being read using a BYOB reader, the view of that read request is forwarded to the original stream, so it can read directly into that view. After the view is filled, an identical copy is created and enqueued on the other branch, so they both see the same data.

To do:
* [x] Write tests
* [x] Update specification text to match new reference implementation

---

   * …
   * …
   * web-platform-tests/wpt#29190
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1114.html" title="Last updated on Jul 8, 2021, 8:45 PM UTC (75a80a8)">Preview</a> | <a href="https://whatpr.org/streams/1114/c46742f...75a80a8.html" title="Last updated on Jul 8, 2021, 8:45 PM UTC (75a80a8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2328,8 +2328,7 @@ create them does not matter.
  1. Assert: |stream| [=implements=] {{ReadableStream}}.
  1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
     {{ReadableByteStreamController}}.
- 1. Let |reader| be undefined.
- 1. Let |readerType| be undefined.
+ 1. Let |reader| be ? [$AcquireReadableStreamDefaultReader$](|stream|).
  1. Let |reading| be false.
  1. Let |canceled1| be false.
  1. Let |canceled2| be false.
@@ -2338,34 +2337,21 @@ create them does not matter.
  1. Let |branch1| be undefined.
  1. Let |branch2| be undefined.
  1. Let |cancelPromise| be [=a new promise=].
- 1. Let |createReader| be the following steps, taking a |newReaderType| argument:
-  1. Let |newReader| be undefined.
-  1. If |newReaderType| is "`default`", set |newReader| to ?
-     [$AcquireReadableStreamDefaultReader$](|stream|).
-  1. Otherwise,
-   1. Assert: |newReaderType| is "`byob`".
-   1. Set |newReader| to ? [$AcquireReadableStreamBYOBReader$](|stream|).
-  1. Set |reader| to |newReader|.
-  1. Set |readerType| to |newReaderType|.
-  1. [=Upon rejection=] of |newReader|.[=ReadableStreamGenericReader/[[closedPromise]]=] with reason
+ 1. Let |forwardReaderError| be the following steps, taking a |thisReader| argument:
+  1. [=Upon rejection=] of |thisReader|.[=ReadableStreamGenericReader/[[closedPromise]]=] with reason
      |r|,
-   1. If |reader| is not |newReader|, return.
+   1. If |reader| is not |thisReader|, return.
    1. Perform ! [$ReadableByteStreamControllerError$](|branch1|.[=ReadableStream/[[controller]]=],
       |r|).
    1. Perform ! [$ReadableByteStreamControllerError$](|branch2|.[=ReadableStream/[[controller]]=],
       |r|).
    1. If |canceled1| is false or |canceled2| is false, [=resolve=] |cancelPromise| with undefined.
- 1. Let |releaseReader| be the following steps:
-  1. If |readerType| is "`default`",
-   1. Assert: |reader|.[=ReadableStreamDefaultReader/[[readRequests]]=] is [=list/is empty|empty=].
-  1. Otherwise,
-   1. Assert: |readerType| is "`byob`".
-   1. Assert: |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=] is [=list/is empty|empty=].
-  1. Perform ! [$ReadableStreamReaderGenericRelease$](|reader|).
  1. Let |pullWithDefaultReader| be the following steps:
-  1. If |readerType| is not "`default`",
-    1. Perform |releaseReader|.
-    1. Perform |createReader|, given "`default`".
+  1. If |reader| [=implements=] {{ReadableStreamBYOBReader}},
+   1. Assert: |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=] is [=list/is empty|empty=].
+   1. Perform ! [$ReadableStreamReaderGenericRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|stream|).
+   1. Perform |forwardReaderError|, given |reader|.
   1. Let |readRequest| be a [=read request=] with the following [=struct/items=]:
    : [=read request/chunk steps=], given |chunk|
    ::
@@ -2407,9 +2393,11 @@ create them does not matter.
     1. Set |reading| to false.
   1. Perform ! [$ReadableStreamDefaultReaderRead$](|reader|, |readRequest|).
  1. Let |pullWithBYOBReader| be the following steps, given |view| and |forBranch2|:
-  1. If |readerType| is not "`byob`",
-    1. Perform |releaseReader|.
-    1. Perform |createReader|, given "`byob`".
+  1. If |reader| [=implements=] {{ReadableStreamDefaultReader}},
+   1. Assert: |reader|.[=ReadableStreamDefaultReader/[[readRequests]]=] is [=list/is empty|empty=].
+   1. Perform ! [$ReadableStreamReaderGenericRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamBYOBReader$](|stream|).
+   1. Perform |forwardReaderError|, given |reader|.
   1. Let |readIntoRequest| be a [=read-into request=] with the following [=struct/items=]:
    : [=read-into request/chunk steps=], given |chunk|
    ::
@@ -2518,12 +2506,11 @@ create them does not matter.
    1. [=Resolve=] |cancelPromise| with |cancelResult|.
   1. Return |cancelPromise|.
  1. Let |startAlgorithm| be an algorithm that returns undefined.
- 1. Perform |createReader|, given "`default`".
-    <!-- TODO This can throw if the stream is locked, is that allowed? -->
  1. Set |branch1| to ! [$CreateReadableByteStream$](|startAlgorithm|, |pull1Algorithm|,
     |cancel1Algorithm|).
  1. Set |branch2| to ! [$CreateReadableByteStream$](|startAlgorithm|, |pull2Algorithm|,
     |cancel2Algorithm|).
+ 1. Perform |forwardReaderError|, given |reader|.
  1. Return « |branch1|, |branch2| ».
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -2329,7 +2329,7 @@ create them does not matter.
  1. Let |forwardReaderError| be the following steps, taking a |thisReader| argument:
   1. [=Upon rejection=] of |thisReader|.[=ReadableStreamGenericReader/[[closedPromise]]=] with reason
      |r|,
-   1. If |reader| is not |thisReader|, return.
+   1. If |thisReader| is not |reader|, return.
    1. Perform ! [$ReadableByteStreamControllerError$](|branch1|.[=ReadableStream/[[controller]]=],
       |r|).
    1. Perform ! [$ReadableByteStreamControllerError$](|branch2|.[=ReadableStream/[[controller]]=],
--- a/index.bs
+++ b/index.bs
@@ -2407,20 +2407,16 @@ create them does not matter.
        1. Let |clonedChunk| be ? [$CloneAsUint8Array$](|chunk|).
        1. Perform ! [$ReadableByteStreamControllerEnqueue$](|branch1|.[=ReadableStream/[[controller]]=],
           |clonedChunk|).
-      1. If |canceled2| is true,
-       1. Set |chunk| to ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
-          chunk.\[[ByteOffset]], 0 »).
-      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
+      1. If |canceled2| is false,
+       1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
          |chunk|).
      1. Otherwise,
       1. If |canceled2| is false,
        1. Let |clonedChunk| be ? [$CloneAsUint8Array$](|chunk|).
        1. Perform ! [$ReadableByteStreamControllerEnqueue$](|branch2|.[=ReadableStream/[[controller]]=],
           |clonedChunk|).
-      1. If |canceled1| is true,
-       1. Set |chunk| to ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
-          chunk.\[[ByteOffset]], 0 »).
-      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
+      1. If |canceled1| is false,
+       1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
          |chunk|).
 
     <p class="note">The microtask delay here is necessary because it takes at least a microtask to
--- a/index.bs
+++ b/index.bs
@@ -2376,9 +2376,9 @@ create them does not matter.
    ::
     1. Set |reading| to false.
     1. If |canceled1| is false, perform !
-       [$ReadableStreamDefaultControllerClose$](|branch1|.[=ReadableStream/[[controller]]=]).
+       [$ReadableByteStreamControllerClose$](|branch1|.[=ReadableStream/[[controller]]=]).
     1. If |canceled2| is false, perform !
-       [$ReadableStreamDefaultControllerClose$](|branch2|.[=ReadableStream/[[controller]]=]).
+       [$ReadableByteStreamControllerClose$](|branch2|.[=ReadableStream/[[controller]]=]).
     1. If |branch1|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
        is not [=list/is empty|empty=], perform !
        [$ReadableByteStreamControllerRespond$](|branch1|.[=ReadableStream/[[controller]]=], 0).
@@ -2433,9 +2433,9 @@ create them does not matter.
     1. Assert: |chunk|.\[[ByteLength]] is 0.
     1. Set |reading| to false.
     1. If |canceled1| is false, perform !
-       [$ReadableStreamDefaultControllerClose$](|branch1|.[=ReadableStream/[[controller]]=]).
+       [$ReadableByteStreamControllerClose$](|branch1|.[=ReadableStream/[[controller]]=]).
     1. If |canceled2| is false, perform !
-       [$ReadableStreamDefaultControllerClose$](|branch2|.[=ReadableStream/[[controller]]=]).
+       [$ReadableByteStreamControllerClose$](|branch2|.[=ReadableStream/[[controller]]=]).
     1. If |forBranch2| is true,
      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
         |chunk|).
--- a/index.bs
+++ b/index.bs
@@ -2409,13 +2409,10 @@ create them does not matter.
        1. Perform ! [$ReadableByteStreamControllerEnqueue$](|branch1|.[=ReadableStream/[[controller]]=],
           |clonedChunk|).
       1. If |canceled2| is true,
-       1. Let |emptyChunk| be ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
+       1. Set |chunk| to ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
           chunk.\[[ByteOffset]], 0 »).
-       1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
-          |emptyChunk|).
-      1. Otherwise,
-       1. Perform ! [$ReadableByteStreamControllerRespond$](|branch2|.[=ReadableStream/[[controller]]=],
-          |chunk|.\[[ByteLength]]).
+      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
+         |chunk|).
      1. Otherwise,
       1. If |canceled2| is false,
        1. Let |clonedChunk| be ! [$StructuredDeserialize$](? [$StructuredSerialize$](|chunk|),
@@ -2423,13 +2420,10 @@ create them does not matter.
        1. Perform ! [$ReadableByteStreamControllerEnqueue$](|branch2|.[=ReadableStream/[[controller]]=],
           |clonedChunk|).
       1. If |canceled1| is true,
-       1. Let |emptyChunk| be ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
+       1. Set |chunk| to ! [$Construct$]({{%Uint8Array%}}, « |chunk|.\[[ViewedArrayBuffer]],
           chunk.\[[ByteOffset]], 0 »).
-       1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
-          |emptyChunk|).
-      1. Otherwise,
-       1. Perform ! [$ReadableByteStreamControllerRespond$](|branch1|.[=ReadableStream/[[controller]]=],
-          |chunk|.\[[ByteLength]]).
+      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
+         |chunk|).
 
     <p class="note">The microtask delay here is necessary because it takes at least a microtask to
     detect errors, when we use |reader|.[=ReadableStreamGenericReader/[[closedPromise]]=] below.
--- a/index.bs
+++ b/index.bs
@@ -2043,10 +2043,9 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="CreateReadableByteStream"
- id="create-readable-byte-stream">CreateReadableByteStream(|startAlgorithm|, |pullAlgorithm|,
- |cancelAlgorithm|[, |highWaterMark|, [, |autoAllocateChunkSize|]])</dfn> performs the following
- steps:
+ <dfn abstract-op lt="CreateReadableByteStream">CreateReadableByteStream(|startAlgorithm|,
+ |pullAlgorithm|, |cancelAlgorithm|[, |highWaterMark|, [, |autoAllocateChunkSize|]])</dfn>
+ performs the following steps:
 
  1. If |highWaterMark| was not passed, set it to 0.
  1. If |autoAllocateChunkSize| was not passed, set it to undefined.
@@ -2237,9 +2236,8 @@ create them does not matter.
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableStreamDefaultTee"
- id="readable-stream-default-tee">ReadableStreamDefaultTee(|stream|, |cloneForBranch2|)</dfn>
- performs the following steps:
+ <dfn abstract-op lt="ReadableStreamDefaultTee">ReadableStreamDefaultTee(|stream|,
+ |cloneForBranch2|)</dfn> performs the following steps:
 
  1. Assert: |stream| [=implements=] {{ReadableStream}}.
  1. Assert: |cloneForBranch2| is a boolean.
@@ -2321,8 +2319,7 @@ create them does not matter.
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableByteStreamTee"
- id="readable-byte-stream-tee">ReadableByteStreamTee(|stream|)</dfn>
+ <dfn abstract-op lt="ReadableByteStreamTee">ReadableByteStreamTee(|stream|)</dfn>
  performs the following steps:
 
  1. Assert: |stream| [=implements=] {{ReadableStream}}.
@@ -3234,8 +3231,7 @@ The following abstract operations support the implementation of the
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableByteStreamControllerGetBYOBRequest"
- id="readable-byte-stream-controller-get-byob-request">ReadableByteStreamControllerGetBYOBRequest(|controller|)</dfn>
+ <dfn abstract-op lt="ReadableByteStreamControllerGetBYOBRequest">ReadableByteStreamControllerGetBYOBRequest(|controller|)</dfn>
  performs the following steps:
 
  1. If |controller|.[=ReadableByteStreamController/[[byobRequest]]=] is null and
@@ -6413,8 +6409,7 @@ The following abstract operations are a grab-bag of utilities.
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="CloneAsUint8Array"
- id="clone-as-uint8-array">CloneAsUint8Array(|O|)</dfn> performs the following steps:
+ <dfn abstract-op lt="CloneAsUint8Array">CloneAsUint8Array(|O|)</dfn> performs the following steps:
 
  1. Assert: [$Type$](|O|) is Object.
  1. Assert: |O| has an \[[ViewedArrayBuffer]] internal slot.
--- a/index.bs
+++ b/index.bs
@@ -2426,24 +2426,25 @@ create them does not matter.
 
    : [=read-into request/close steps=], given |chunk|
    ::
-    1. Assert: |chunk|.\[[ByteLength]] is 0.
     1. Set |reading| to false.
     1. If |canceled1| is false, perform !
        [$ReadableByteStreamControllerClose$](|branch1|.[=ReadableStream/[[controller]]=]).
     1. If |canceled2| is false, perform !
        [$ReadableByteStreamControllerClose$](|branch2|.[=ReadableStream/[[controller]]=]).
-    1. If |forBranch2| is true,
-     1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
-        |chunk|).
-     1. If |branch1|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
-        is not [=list/is empty|empty=], perform !
-        [$ReadableByteStreamControllerRespond$](|branch1|.[=ReadableStream/[[controller]]=], 0).
-    1. Otherwise,
-     1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
-        |chunk|).
-     1. If |branch2|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
-        is not [=list/is empty|empty=], perform !
-        [$ReadableByteStreamControllerRespond$](|branch2|.[=ReadableStream/[[controller]]=], 0).
+    1. If |chunk| is not undefined,
+     1. Assert: |chunk|.\[[ByteLength]] is 0.
+     1. If |forBranch2| is true,
+      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch2|.[=ReadableStream/[[controller]]=],
+         |chunk|).
+      1. If |branch1|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
+         is not [=list/is empty|empty=], perform !
+         [$ReadableByteStreamControllerRespond$](|branch1|.[=ReadableStream/[[controller]]=], 0).
+     1. Otherwise,
+      1. Perform ! [$ReadableByteStreamControllerRespondWithNewView$](|branch1|.[=ReadableStream/[[controller]]=],
+         |chunk|).
+      1. If |branch2|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
+         is not [=list/is empty|empty=], perform !
+         [$ReadableByteStreamControllerRespond$](|branch2|.[=ReadableStream/[[controller]]=], 0).
     1. If |canceled1| is false or |canceled2| is false, [=resolve=] |cancelPromise| with undefined.
 
    : [=read-into request/error steps=]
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -3205,7 +3442,7 @@ The following abstract operations support the implementation of the
  1. Let |firstDescriptor| be |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
  1. If |firstDescriptor|'s [=pull-into descriptor/byte offset=] + |firstDescriptor|' [=pull-into
     descriptor/bytes filled=] is not |view|.\[[ByteOffset]], throw a {{RangeError}} exception.
- 1. If |firstDescriptor|'s [=pull-into descriptor/byte length=] is not |view|.\[[ByteLength]], throw
+ 1. If |firstDescriptor|'s [=pull-into descriptor/byte length=] < |view|.\[[ByteLength]], throw
```
MattiasBuelens, line 3445: I believe this was a missing feature of `byobRequest.respondWithNewView(view)`. You could already call `byobRequest.respondWith(bytesWritten)` with `0 <= bytesWritten <= byobRequest.view.byteLength`, but for some reason `.respondWithNewView(view)` **insisted** that you fill the entire view. Now, you can pass any view whose size is less than or equal to the original view's size. 🙂

Or would you prefer fixing this in a separate PR?

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -422,6 +451,234 @@ function ReadableStreamTee(stream, cloneForBranch2) {
...
+        // This needs to be delayed a microtask because it takes at least a microtask to detect errors (using
+        // reader._closedPromise below), and we want errors in stream to error both branches immediately. We cannot let
+        // successful synchronously-available reads get ahead of asynchronously-available errors.
+        queueMicrotask(() => {
+          reading = false;
+
+          const chunk1 = chunk;
+          let chunk2 = chunk;
+          if (canceled1 === false && canceled2 === false) {
+            chunk2 = new chunk.constructor(chunk);
```
MattiasBuelens: Ugh, right, I'm trying to be too clever. This constructor works for typed arrays, but not for `DataView`s. 🤦‍♂️

I guess I'll have to write a helper function. 😛

index.bs
```diff
@@ -2277,6 +2302,199 @@ create them does not matter.
...
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|stream|).
+   1. Perform |forwardReaderError|, given |reader|.
+  1. Let |readRequest| be a [=read request=] with the following [=struct/items=]:
+   : [=read request/chunk steps=], given |chunk|
+   ::
+    1. [=Queue a microtask=] to perform the following steps:
+     1. Set |reading| to false.
+     1. Let |chunk1| and |chunk2| be |chunk|.
+     1. If |canceled1| is false and |canceled2| is false, set |chunk2| to !
+        [$StructuredDeserialize$](? [$StructuredSerialize$](|chunk|), [=the current Realm=]).
```
MattiasBuelens, line 2344: Perhaps this is a bit overkill? This copies the entire backing `ArrayBuffer`, but we only really need the bytes within the `chunk`'s view. 🤔

I suppose we could make a dedicated `CloneArrayBufferView` helper, rather than using structured serialize/deserialize.

MattiasBuelens: Okay, this is a difficult one...

If we want to guarantee that [the tee wrapper algorithm structurally clones every chunk](https://streams.spec.whatwg.org/commit-snapshots/dd76a17a3738d78708a8dfd8f0508e717d6a1988/#readablestream-tee), then we *have* to copy the entire `ArrayBuffer`.

But this buffer comes from the BYOB request of one of the branches. So if we copy it in its entirety, we might leak data from one branch to the other branch!
```javascript
  type: 'bytes',
  pull(c) {
    c.byobRequest.view[0] = 1;
    c.byobRequest.respond(1);
  }
});

// The first reader reads into its own buffer with a BYOB reader
const reader1 = branch1.getReader({ mode: 'byob' });
const buffer1 = new Uint8Array([42, 42, 42]);
const result1 = await reader1.read(buffer);
result1.value; // [1]
new Uint8Array(result1.value.buffer); // [1, 42, 42]

// The second reader reads the same data, but using a default reader
const reader2 = branch2.getReader();
const result2 = await reader2.read();
// The read chunk has the expected data
result2.value; // [1]
// ...but its backing buffer was copied from the first reader's buffer!
new Uint8Array(result2.value.buffer); // [1, 42, 42]
```
This can be bad if the first reader also puts other, potentially sensitive information inside `buffer1`. 😱 

Thoughts? Personally, I think it'd be more efficient and safer if we only copied the bytes within `chunk`'s view. So in the previous example, the last line would log `[1]` instead.

ricea: I agree with your analysis. I think we should only copy the bytes within `chunk`'s view. To me it's more intuitive, even if it isn't a real structural clone.

index.bs
```diff
@@ -2277,6 +2302,193 @@ create them does not matter.
  1. Return « |branch1|, |branch2| ».
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableByteStreamTee"
+ id="readable-byte-stream-tee">ReadableByteStreamTee(|stream|)</dfn>
```
domenic, line 2307: The fact that cloneForBranch2 is ignored (and treated as always-true) for byte streams is interesting and worth noting carefully, probably in the web-developer facing domintro for tee().

Or, would it be worthwhile and possible to respect it? Probably not given all the transfers...

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -422,6 +451,230 @@ function ReadableStreamTee(stream, cloneForBranch2) {
...
+        reading = false;
+      }
+    };
+    ReadableStreamDefaultReaderRead(reader, readRequest);
+  }
+
+  function pullWithBYOBReader(view, forBranch2) {
+    if (ReadableStreamDefaultReader.isImpl(reader)) {
+      assert(reader._readRequests.length === 0);
+      ReadableStreamReaderGenericRelease(reader);
```
domenic, line 539: Wow, so this switches off readers according to which each branch is doing? Very cool!

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1287,8 +1556,8 @@ function ReadableByteStreamControllerRespondWithNewView(controller, view) {
   if (firstDescriptor.byteOffset + firstDescriptor.bytesFilled !== view.byteOffset) {
     throw new RangeError('The region specified by view does not match byobRequest');
   }
-  if (firstDescriptor.byteLength !== view.byteLength) {
-    throw new RangeError('The buffer of view has different capacity than byobRequest');
+  if (firstDescriptor.byteLength < view.byteLength) {
```
domenic, line 1559: Seems like this needs a corresponding spec change (and probably independent tests)

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -422,6 +451,230 @@ function ReadableStreamTee(stream, cloneForBranch2) {
...
+
+    const readIntoRequest = {
+      chunkSteps: chunk => {
+        // This needs to be delayed a microtask because it takes at least a microtask to detect errors (using
+        // reader._closedPromise below), and we want errors in stream to error both branches immediately. We cannot let
+        // successful synchronously-available reads get ahead of asynchronously-available errors.
+        queueMicrotask(() => {
+          reading = false;
+
+          if (forBranch2 === true) {
```
domenic, line 553: What are your thoughts on creating separate read into requests for each branch vs. having this forBranch2 thing?

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -422,6 +451,230 @@ function ReadableStreamTee(stream, cloneForBranch2) {
...
+              chunk = new Uint8Array(chunk.buffer, chunk.byteOffset, 0);
+            }
+            ReadableByteStreamControllerRespondWithNewView(branch2._controller, chunk);
+          } else {
+            if (canceled2 === false) {
+              const clonedChunk = CloneArrayBufferView(chunk);
+              ReadableByteStreamControllerEnqueue(branch2._controller, clonedChunk);
+            }
+            if (canceled1 === true) {
+              chunk = new Uint8Array(chunk.buffer, chunk.byteOffset, 0);
```
domenic, line 568: I've gotten lost trying to figure out the purpose of this wrapper... I kind of thought the stream machinery handled the case of returning the buffers when closed for us?

index.bs
MattiasBuelens: Correct, both `ReadableByteStreamController.enqueue()` and `ReadableStreamBYOBRequest.respond()` transfer the buffer. So it's impossible to pass the same chunk to both branches, we *must* copy it first.

I'll try to write something for the domintro. 🙂

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Yup! I've done something similar before in [web-streams-adapter](https://github.com/MattiasBuelens/web-streams-adapter/blob/v0.1.0-alpha.5/src/readable-wrapper.ts#L235-L257), so that's where I got the idea. 😁

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: I agree, this deserves its own spec change with tests. ([See also my other comment.](https://github.com/whatwg/streams/pull/1114#discussion_r600081235))

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: That would also work, but I'm a bit worried there would be a lot of duplication between the *pullWithBYOBFromBranch1* and *pullWithBYOBFromBranch2* algorithms. But if you think it'd be more readable that way, I can try that.

domenic: I'm not sure; I was genuinely asking :). I guess I was hoping the duplication would be nicer because once you read a single one of them, you'd be able to glaze over when reading the second. But who knows...

index.bs
domenic: Sorry for missing this. Doing this in the same PR makes sense, but I'm wondering if maybe we had a reason for this in the first place...

In particular, consider a consumer which wants to allocate and use, say, 1 MiB of memory. It allocates a 1 MiB buffer, and passes it through the stream, and gets it back, round and round.

My understanding is that in the current spec, no matter what, they will always get back a 1 MiB buffer. Even if the underlying byte source calls `byobRequest.respond(5)`, the returned buffer will still be 1 MiB (even if the _view_ is only 5 bytes). So when they cycle that buffer into another `reader.read(newViewOntoSameBuffer)` call, they can get up to 1 MiB that next time.

Whereas, if we allow smaller results to be sent back from `respondWithNewView()`, then the consumer would lose their 1 MiB buffer forever, and only get back a 5 byte buffer in return.

Now, that said, having written that up, it seems like we could still allow smaller views onto larger buffers. I.e. we could change the check from checking the view size to checking the buffer size. But I think we probably don't want to allow respondWithNewView() to take arbitrarily-small buffers?

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: This is a tricky one! This is how we get to this point:
1. _branch1_ pulls with a BYOB request. We call _pullWithBYOBReader_ with `forBranch2 = false`, and start a BYOB read on the original stream using the buffer from _branch1_'s BYOB request. Note that this **transfers control of the buffer** from _branch1_ to the original stream.
2. _branch1_ gets canceled. We set `canceled1 = true` in _cancel1Algorithm_. However, we cannot yet "finish" _branch1_'s BYOB request at this point, because we cannot yet return control of the BYOB request's buffer back to _branch1_. Any attempt to call `byobRequest.respond()` here would throw an "already transferred" error (at step 1 in [ReadableByteStreamControllerRespondInClosedState](https://streams.spec.whatwg.org/commit-snapshots/dd76a17a3738d78708a8dfd8f0508e717d6a1988/#readable-byte-stream-controller-respond-in-closed-state)).
3. The BYOB read on the original stream completes. We get passed _chunk_, filled with new data. But more importantly: we get control back over _chunk_'s underlying buffer! So **now** we can finally respond to _branch1_'s BYOB request. But we can't use `byobRequest.respond()`, since that will still try to use the old (transferred) buffer. We **have** to use `.respondWithNewView()` instead, so it'll use the new buffer.
4. Since `canceled1 === true`, `branch1.[[state]] === "closed"`. This means we **must** respond to the BYOB request with exactly 0 bytes written, as per step 3.1 in [ReadableByteStreamControllerRespondInternal](https://streams.spec.whatwg.org/commit-snapshots/dd76a17a3738d78708a8dfd8f0508e717d6a1988/#readable-byte-stream-controller-respond-internal). But the chunk *was* filled with some data (which we can still copy and pass to _branch2_), so we first have to turn the filled view into an empty one before responding.

index.bs
```diff
@@ -2277,6 +2302,193 @@ create them does not matter.
...
+   1. Perform |pullWithDefaultReader|.
+  1. Otherwise,
+   1. Perform |pullWithBYOBReader|, given |byobRequest|.[=ReadableStreamBYOBRequest/[[view]]=] and true.
+  1. Return [=a promise resolved with=] undefined.
+ 1. Let |cancel1Algorithm| be the following steps, taking a |reason| argument:
+  1. Set |canceled1| to true.
+  1. Set |reason1| to |reason|.
+  1. If |reading| is false and
+     |branch1|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
+     is not [=list/is empty|empty=],
```
MattiasBuelens, line 2464: This condition isn't correct yet. I'm still working on it, but I think it will end up being something along these lines:
> If _branch1_.\[[controller]].\[[pendingPullIntos]] is not empty and we're currently not pulling with a BYOB reader from _branch1_, ...

The idea is that if a branch is canceled and still has a pending BYOB request, but we're not already using that BYOB request to BYOB read from the original stream (and thus we haven't transferred its buffer), then we don't need to wait with responding to this branch's BYOB request.

However, if we *are* using this branch's BYOB request for our currently pending BYOB read, then we have transferred control of that request's buffer to the original stream and we must wait for that BYOB read to complete before we can transfer it back to the branch. Hence why we need the second part of the condition.

...Have I already mentioned that this is quite tricky to get right? 😅

index.bs
MattiasBuelens: I agree. `byobRequest.respondWithNewView(newView)` should not allow passing a view whose buffer is *smaller* than the original buffer of the original BYOB request.

I would go even further: I'd say `byobRequest.respondWithNewView(newView)` must only be called with a view on the **same buffer** that was originally passed as `byobRequest.view.buffer`, after it has been **transferred** zero or more times. All of the data that was originally inside `byobRequest.view.buffer` should also be in `newView.buffer` (except for the part spanned by `newView`'s start and end, since that will contain the read data).

Ideally, I would make `.respondWithNewView()` enforce this: if `newView.buffer` is not the "transferred version" of `byobRequest.view.buffer`, then it should throw an error. Unfortunately, I don't think the ECMAScript specification provides any way to check this... 😞

We could approximate it by storing the original `byobRequest.view.buffer.byteLength` as part of the pull-into descriptor, and then checking whether `newView.buffer.byteLength` equals that original length. However, that doesn't prevent someone from creating a new `ArrayBuffer` with the same size and using that for `newView`... 😕

---

Either way, you **must** be able to pass a smaller view onto that buffer. Otherwise, it's impossible to satisfy the assert in step 3.1 of [ReadableByteStreamControllerRespondInternal](https://streams.spec.whatwg.org/commit-snapshots/dd76a17a3738d78708a8dfd8f0508e717d6a1988/#readable-byte-stream-controller-respond-internal) when called from `.respondWithNewView()`. And [we definitely need this](https://github.com/whatwg/streams/pull/1114#discussion_r601890120)! 😁

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: > I kind of thought the stream machinery handled the case of returning the buffers when closed for us?

No, it doesn't. If you look at [ReadableStreamClose](https://streams.spec.whatwg.org/commit-snapshots/dd76a17a3738d78708a8dfd8f0508e717d6a1988/#readable-stream-close), it only runs the close steps of read requests from a *default reader*. It doesn't run any close steps of read-into requests from a *BYOB reader*.

And that's because *it cannot*. In order to run those close steps, it must construct a _chunk_ with a transferred version of the original read-into request's buffer. But it doesn't *have* that buffer, since the underlying byte source might have transferred it from `byobRequest.view` to some other API to do the reading. (For example, by passing it to another `byobReader.read(view)` call like in `ReadableByteStreamTee`.)

We can only call these close steps when we have received control back of that buffer, which must happen through either `byobRequest.respond()` (if the byte source didn't transfer it and instead wrote directly into the buffer) or `.respondWithNewView()` (if it did transfer it). 

This also means that *every underlying byte source implementation* must do something like this in its `cancel` algorithm:
```javascript
new ReadableStream({
  type: 'bytes',
  start(controller) {
    this._controller = controller;
  },
  pull(controller) {
    // call controller.enqueue() or controller.byobRequest.respond() somehow...
  },
  cancel() {
    // finish any pending BYOB request (even if we never used it ourselves)
    this._controller.byobRequest?.respond(0);
  }
});
```
Otherwise, the last BYOB read can never resolve to `{ value: emptyChunk, done: true }` after canceling the stream.

index.bs
MattiasBuelens: Note that [this example in the specification](https://streams.spec.whatwg.org/#example-manual-read-bytes) *relies* on `read()` returning a view on the same (transferred) buffer. Otherwise the `offset < buffer.byteLength` check wouldn't work, and you might read more or fewer bytes than initially requested. It also assumes that data from previous reads is still available in `buffer`, so the underlying byte source cannot change any of the bytes outside of `byobRequest.view` either.

To summarize, the caller of `byobRequest.respondWithNewView(newView)` (i.e. the underlying byte source) must guarantee the following:
* `newView.byteOffset` must equal `byobRequest.view.byteOffset`, and `newView.byteLength` must be `<= byobRequest.view.byteLength`.
* `newView.buffer` has the exact same `byteLength` as `byobRequest.view.buffer`. This can be done by either transferring the original buffer, or by copying it (although that would defeat the purpose of using a BYOB request).
* Within `newView.buffer`, the byte ranges from `0` to `byobRequest.view.byteOffset` and from `byobRequest.view.byteOffset + byobRequest.view.byteLength` to `byobRequest.view.buffer.byteLength` must be unmodified. In other words, only bytes within the range of `byobRequest.view` can be changed. (This also applies to `byobRequest.respond(bytesWritten)`.)

Currently, we check the first condition. We could *feasibly* check the second one, by remembering the original `byteLength` as part of the pull-into descriptor. However, it's not possible to check the third condition without copying the original buffer upfront, which would defeat the purpose of BYOB. So we have to *assume* that the developer doesn't do anything weird in their underlying byte source... 😅

domenic: > Currently, we check the first condition.

Well, _currently_ we kind of check the second one (by checking the view length, which must be >= the buffer length).

I'm thinking now we should pull out this change (to check both 1 + 2 cleanly), including the domintro change you suggest (which is a great idea), into a separate PR. Ideally we'd also add some usage examples for `respondWithNewView()`; I was surprised to not be able to find any in the spec.

reference-implementation/lib/abstract-ops/readable-streams.js
domenic: Thanks so much for the analysis and walking me through it!

> This also means that every underlying byte source implementation must do something like this in its cancel algorithm:

This seems like a significant footgun, especially given that none of the [example byte sources in the spec](https://streams.spec.whatwg.org/#example-rbs-push) do so.

Can we make this easier somehow? E.g.:

- Automatically introspect `byobRequest.view` and if it's transferred, call (the equivalent of) `respondWithNewView(constructedView)`, and if it's not transferred, call `respond(0)`?

- Relax the requirement that the last BYOB read should resolve with `value: emptyChunk`?

MattiasBuelens: > This seems like a significant footgun, especially given that none of the [example byte sources in the spec](https://streams.spec.whatwg.org/#example-rbs-push) do so.

Yes, it really is. We'll need to update those to handle it properly.

> Can we make this easier somehow?
> * Automatically introspect `byobRequest.view` and if it's transferred, call (the equivalent of) `respondWithNewView(constructedView)` ...

I don't think that's possible? If a buffer is transferred, we have no way to know *where it was transferred to*. We definitely cannot re-construct an `ArrayBuffer` around something that has been transferred. For all we know, that buffer is now being used by a different thread in a Web Worker.

Until the underlying byte source actually calls `.respond()` or `.respondWithNewView()` itself, we have no way to "reclaim" the transferred buffer.

> * ...and if it's not transferred, call `respond(0)`?

That *might* work, but it can cause unexpected behavior:
```javascript
  type: "bytes",
  pull(controller) {
    const byobRequest = controller.byobRequest;
    if (byobRequest) {
      // Note: we're not returning this promise, so the stream will not wait for it before canceling.
      fetchSomeBytes(byobRequest.view.byteLength).then((data) => {
        // If the stream is canceled while fetching the data, then the stream would have already "auto-responded"
        // to the BYOB request. This invalidates the request: its buffer is transferred back to the stream,
        // and its view is set to null.
        byobRequest.view.set(data, 0); // 💥
        byobRequest.respond(data.byteLength);
      });
    }
  }
}
```

> * Relax the requirement that the last BYOB read should resolve with `value: emptyChunk`?

This would break [the `readInto()` example](https://streams.spec.whatwg.org/#example-manual-read-bytes). The BYOB reader *needs* to get its buffer back.

domenic: > That might work, but it can cause unexpected behavior:

Eh, seems OK. Your fault for not returning the promise to signal you were still planning to use `byobRequest`.

> This would break the readInto() example. The BYOB reader needs to get its buffer back.

Well, the readInto() example doesn't cancel the stream. I guess maybe that's my suggestion (and perhaps even the current intent of the spec): if you cancel the stream with a BYOB request pending, you don't expect to get your buffer back. You should have waited to cancel until the outstanding BYOB request completed. How bad would that be?

MattiasBuelens: > Eh, seems OK. Your fault for not returning the promise to signal you were still planning to use `byobRequest`.

Huh, hang on. We don't wait until `source.pull()` resolves/rejects before we call `source.cancel()`. So even *if* you returned a promise from `pull()`, you'd still have to provide your own `cancel()` method which ensures that the previous `pull()` promise stops early and doesn't use `byobRequest` anymore *after* `cancel()` settles.

I guess that makes sense. We have the same thing for `ReadableByteStreamController.enqueue()`: if the stream's state is `"closed"`, it throws a `TypeError`. So if the `pull()` method still uses it after the stream has been cancelled, it'll (hopefully) propagate the error to the returned promise and the stream will ignore that promise's result (because the state is no longer `"readable"`).

> Well, the readInto() example doesn't cancel the stream. I guess maybe that's my suggestion (and perhaps even the current intent of the spec): if you cancel the stream with a BYOB request pending, you don't expect to get your buffer back. You should have waited to cancel until the outstanding BYOB request completed. How bad would that be?

Hmm, that's also true. That would change `ReadableStreamBYOBReadResult` slightly, since `value` can now be `undefined` if the read resolves as a result of canceling the stream.

But really, what *would* a developer still want to do with the buffer from the `read()` result after they've cancelled the stream? Is there ever a case where you start a BYOB read, then change your mind and cancel it *but still want your buffer back*? 🤔

I'll try it out on the reference implementation, see where it leads us. But I'm *definitely* going to move it to a separate PR now! 😁

index.bs
MattiasBuelens: I've added a `CloneAsUint8Array` helper to copy the bytes of `chunk` to a new `Uint8Array`. 🙂

index.bs
MattiasBuelens: Solved in #1129.

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Solved in #1129. Since we no longer need to respond after the branch was cancelled, I've removed this bit. 🙂

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Solved in #1123.

index.bs
MattiasBuelens: Solved in #1123.

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: I've tried [something different](https://github.com/whatwg/streams/blob/78b76ee7dd6720c8bffb82cc142c1bcc05bb10ca/index.bs#L2407-L2410) to reduce duplication. Have a look! 🙂

index.bs
```diff
@@ -2055,6 +2044,29 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
  |startAlgorithm| throws.
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="CreateReadableByteStream"
+ id="create-readable-byte-stream">CreateReadableByteStream(|startAlgorithm|, |pullAlgorithm|,
+ |cancelAlgorithm|[, |highWaterMark|, [, |autoAllocateChunkSize|]])</dfn> performs the following
```
domenic, line 2050: Since nobody ever passes the other args, maybe omit them?

index.bs
```diff
@@ -2055,6 +2044,29 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
  |startAlgorithm| throws.
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="CreateReadableByteStream"
+ id="create-readable-byte-stream">CreateReadableByteStream(|startAlgorithm|, |pullAlgorithm|,
```
domenic, line 2049: I think we landed on no manual ID curation for new abstract ops?

index.bs
```diff
@@ -2218,6 +2230,19 @@ create them does not matter.
 
  It performs the following steps:
 
+ 1. Assert: |stream| [=implements=] {{ReadableStream}}.
+ 1. Assert: |cloneForBranch2| is a boolean.
+ 1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
+  1. Return ? [$ReadableByteStreamTee$](|stream|).
```
domenic, line 2236: Nit: although we're not consistent, recently I've preferred folding single-line blocks into their parents (i.e. "If X, y" instead of two separate list items). Unless that would break symmetry between If/Otherwise, I guess.

index.bs
```diff
@@ -2297,6 +2322,172 @@ create them does not matter.
...
+ 1. Let |canceled2| be false.
+ 1. Let |reason1| be undefined.
+ 1. Let |reason2| be undefined.
+ 1. Let |branch1| be undefined.
+ 1. Let |branch2| be undefined.
+ 1. Let |cancelPromise| be [=a new promise=].
+ 1. Let |forwardReaderError| be the following steps, taking a |thisReader| argument:
+  1. [=Upon rejection=] of |thisReader|.[=ReadableStreamGenericReader/[[closedPromise]]=] with reason
+     |r|,
+   1. If |reader| is not |thisReader|, return.
```
domenic, line 2345: Makes more sense to me flipped, as if thisReader is not reader?

index.bs
```diff
@@ -2297,6 +2322,172 @@ create them does not matter.
...
+      |r|).
+   1. If |canceled1| is false or |canceled2| is false, [=resolve=] |cancelPromise| with undefined.
+ 1. Let |pullWithDefaultReader| be the following steps:
+  1. If |reader| [=implements=] {{ReadableStreamBYOBReader}},
+   1. Assert: |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=] is [=list/is empty|empty=].
+   1. Perform ! [$ReadableStreamReaderGenericRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|stream|).
+   1. Perform |forwardReaderError|, given |reader|.
+  1. Let |readRequest| be a [=read request=] with the following [=struct/items=]:
+   : [=read request/chunk steps=], given |chunk|
```
domenic, line 2358: We should use either _chunk_ or _value_ as variable names uniformly between the two algorithms. I'm OK with either, although I suspect _value_ is more prevalent in other parts of the spec?

index.bs
```diff
@@ -2297,6 +2322,172 @@ create them does not matter.
...
+   ::
+    1. [=Queue a microtask=] to perform the following steps:
+     1. Set |reading| to false.
+     1. Let |byobCanceled| be |canceled2| if |forBranch2| is true, and |canceled1| otherwise.
+     1. Let |otherCanceled| be |canceled2| if |forBranch2| is false, and |canceled1| otherwise.
+     1. If |otherCanceled| is false,
+      1. Let |clonedChunk| be ? [$CloneAsUint8Array$](|chunk|).
+      1. Perform ! [$ReadableByteStreamControllerEnqueue$](|otherBranch|.[=ReadableStream/[[controller]]=],
+         |clonedChunk|).
+     1. If |byobCanceled| is false,
```
domenic, line 2415: Nit: flip the order here, so byob is always before "other"?

domenic (approved):
LGTM with nits. Exciting...

index.bs
MattiasBuelens: Sure, I can remove them for now. That said, we might need to bring them back if/when we introduce "transform byte streams".

index.bs
MattiasBuelens: Seems like _chunk_ is more common. Currently [on `main`](https://github.com/whatwg/streams/blob/9beefc2ac041a1a848088962d56eb03e89a8bf07/index.bs):
* `[=read request/chunk steps=], given |chunk|`: 5 matches
* `[=read request/chunk steps=], given |value|`: 1 match
* `[=read request/chunk steps=], given |view|`: 1 match
* `[=read-into request/chunk steps=], given |chunk|`: 2 matches
* `[=read-into request/chunk steps=], given |filledView|`: 1 match

I think I'll change `ReadableStreamTee` to use _chunk_ instead of _value_. 🙂

index.bs
MattiasBuelens: We still need to clone the chunk first, since `ReadableByteStreamControllerRespondWithNewView` will transfer it. So we'd need do:

> 1. If _otherCanceled_ is false, set _clonedChunk_ to CloneAsUint8Array(_chunk_)
> 1. If _byobCanceled_ is false, perform ReadableByteStreamControllerRespondWithNewView(..., _chunk_)
> 1. If _otherCanceled_ is false, perform ReadableByteStreamControllerEnqueue(..., _clonedChunk_)

Might be a bit weird? Or we'd have repeat the `ReadableByteStreamControllerRespondWithNewView` step:

> 1. If _otherCanceled_ is false,
>    1. Let _clonedChunk_ be CloneAsUint8Array(_chunk_)
>    1. If _byobCanceled_ is false, perform ReadableByteStreamControllerRespondWithNewView(..., _chunk_)
>    1. Perform ReadableByteStreamControllerEnqueue(..., _clonedChunk_)
> 1. Otherwise, if _byobCanceled_ is false, perform ReadableByteStreamControllerRespondWithNewView(..., _chunk_)

MattiasBuelens: I went for the second option.

ricea (commented):
I've checked the algorithms and cross-referenced against the reference implementation and it all looks good.

I haven't analysed it for re-entrancy risk. I'm a bit worried about that.

index.bs
```diff
@@ -2297,6 +2312,174 @@ create them does not matter.
...
+   : [=read-into request/error steps=]
+   ::
+    1. Set |reading| to false.
+  1. Perform ! [$ReadableStreamBYOBReaderRead$](|reader|, |view|, |readIntoRequest|).
+ 1. Let |pull1Algorithm| be the following steps:
+  1. If |reading| is true, return [=a promise resolved with=] undefined.
+  1. Set |reading| to true.
+  1. Let |byobRequest| be ! [$ReadableByteStreamControllerGetBYOBRequest$](|branch1|.[=ReadableStream/[[controller]]=]).
+  1. If |byobRequest| is null,
+   1. Perform |pullWithDefaultReader|.
```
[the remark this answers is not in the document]
MattiasBuelens: Right, [he did](https://github.com/whatwg/streams/pull/1114#discussion_r645566297), but I only fixed one instance. I'll go through the whole thing again.

index.bs
```diff
@@ -2297,6 +2312,174 @@ create them does not matter.
...
+   1. Perform ! [$ReadableStreamReaderGenericRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|stream|).
+   1. Perform |forwardReaderError|, given |reader|.
+  1. Let |readRequest| be a [=read request=] with the following [=struct/items=]:
+   : [=read request/chunk steps=], given |chunk|
+   ::
+    1. [=Queue a microtask=] to perform the following steps:
+     1. Set |reading| to false.
+     1. Let |chunk1| and |chunk2| be |chunk|.
+     1. If |canceled1| is false and |canceled2| is false, set |chunk2| to ?
```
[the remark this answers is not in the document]
MattiasBuelens: Hmm, good point. In that case, *think* we're going to need to cancel the stream and error both branches?

It looks like this problem already existed in [the current ReadableStreamTee](https://streams.spec.whatwg.org/commit-snapshots/9beefc2ac041a1a848088962d56eb03e89a8bf07/#readable-stream-tee)... We should fix that as well. 🤔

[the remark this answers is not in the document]
domenic: Great catch. I think this changed in https://github.com/whatwg/streams/commit/63b4407f424d6ad042852b3d225b7c0ffa116b2d . [Before](https://streams.spec.whatwg.org/commit-snapshots/c63ae9eb7de6069f0e3cf5ea6b0ef68a2c8a3971/#readable-stream-tee), the failure was in a promise handler for a promise that was explicitly marked as [[PromiseIsHandled]]. So I guess such failures were ignored.

These days they go to HTML's "report an exception" for all unhandled exceptions.

Also related: I think the `? ReadableStreamDefaultControllerEnqueue` steps inside the chunk steps should be `!`. (Maybe all call sites for it should be??)

Anyway, what is the correct behavior? I think another option would be to only error branch 2. WDYT? @ricea, any ideas on what makes the most sense from the perspective of how implementations use tee-with-cloneForBranch2-set?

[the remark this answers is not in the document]
MattiasBuelens: > Also related: I think the `? ReadableStreamDefaultControllerEnqueue` steps inside the chunk steps should be `!`. (Maybe all call sites for it should be??)

Indeed, it will only throw if `strategy.size()` throws, but that can never happen since we're using the default strategy. This is the same reasoning as we had for `ReadableStream.from()` in [#1083](https://github.com/whatwg/streams/pull/1083#discussion_r520886364).

> Anyway, what is the correct behavior? I think another option would be to only error branch 2. WDYT?

Well, we shouldn't ignore the failure like we did previously, since that means both branches get "stuck" (never receive a new chunk, or become closed or errored).

Only erroring branch 2 seems too risky. One place where we'll want to tee a readable byte stream is when Fetch clones the request to pass to the service worker (step 5.1 of [HTTP fetch](https://fetch.spec.whatwg.org/commit-snapshots/3babdcb8d1e564040b30bc824420ade870709e4f/#concept-http-fetch)). If the service worker reads the request body but does *not* provide a response (i.e. its `fetch` event handler does not call `event.respondWith()`), then we'll skip step 5.5 and we'll continue the fetch with the original request. However, if we've *failed to clone* a chunk, then the service worker would have successfully read the chunk (because its branch was providing the BYOB request), but our original request's body would have become errored, breaking the fetch.

To me, it seems better to error both branches if we can't clone the chunk.

index.bs
```diff
@@ -2233,19 +2248,25 @@ create them does not matter.
...
-     1. If |canceled2| is false and |cloneForBranch2| is true, set |value2| to ?
-        [$StructuredDeserialize$](? [$StructuredSerialize$](|value2|), [=the current Realm=]).
-     1. If |canceled1| is false, perform ?
+     1. Let |chunk1| and |chunk2| be |chunk|.
+     1. If |canceled2| is false and |cloneForBranch2| is true,
+      1. Let |cloneResult| be [$StructuredClone$](|chunk2|).
+      1. If |cloneResult| is an abrupt completion,
+       1. Perform ! [$ReadableStreamDefaultControllerError$](|branch1|.[=ReadableStream/[[controller]]=], |cloneResult|.\[[Value]]).
+       1. Perform ! [$ReadableStreamDefaultControllerError$](|branch2|.[=ReadableStream/[[controller]]=], |cloneResult|.\[[Value]]).
+       1. [=Resolve=] |cancelPromise| with ! [$ReadableStreamCancel$](|stream|, |cloneResult|.\[[Value]]).
```
MattiasBuelens, line 2261: There's another re-entrancy risk here. `ReadableStreamDefaultControllerError` will call `readRequest.errorSteps(e)`, which *could* call both `branch1.cancel()` and `branch2.cancel()`. Thus, our `cancel1Algorithm` (or `cancel2Algorithm` depending on which branch is cancelled last) will call `ReadableStreamCancel(stream, e)` twice. This is "okay" since it'll immediately return an already-resolved promise, but we'll also try to *resolve* our `cancelPromise` twice, which we want to avoid.

Possible solutions:
* Swap the order: cancel first, error branches second. However, this doesn't really solve the problem, since the stream's cancel steps might *also* cancel the branches and make us cancel the stream *again*. I guess we could add a separate `canceled` flag to remember if we already called `ReadableStreamCancel` before, but that seems messy... 😕 
* Add an extra _"if `canceled1` is false or `canceled2` is false"_ check before cancelling.

I think I'll go with the second option, with a note about why this check is necessary. Not sure if we can trigger this edge case from WPT, because we always have at least one micro-task delay from the user-facing `reader.read()` promises.

index.bs
```diff
@@ -2233,19 +2248,25 @@ create them does not matter.
...
      1. Set |reading| to false.
-     1. Let |value1| and |value2| be |value|.
-     1. If |canceled2| is false and |cloneForBranch2| is true, set |value2| to ?
-        [$StructuredDeserialize$](? [$StructuredSerialize$](|value2|), [=the current Realm=]).
-     1. If |canceled1| is false, perform ?
+     1. Let |chunk1| and |chunk2| be |chunk|.
+     1. If |canceled2| is false and |cloneForBranch2| is true,
+      1. Let |cloneResult| be [$StructuredClone$](|chunk2|).
+      1. If |cloneResult| is an abrupt completion,
+       1. Perform ! [$ReadableStreamDefaultControllerError$](|branch1|.[=ReadableStream/[[controller]]=], |cloneResult|.\[[Value]]).
```
MattiasBuelens, line 2259: Should we use the original error as-is (i.e. a `"DataCloneError"` `DOMException`), or should we do "let `cloneError` be a new `TypeError`" instead? 🤔

MattiasBuelens (commented):
I've added handling for errors from cloning chunks.

Still gotta find some time to write tests for all these edge cases. 😛

index.bs
MattiasBuelens: Added a note, have a look. 😉

index.bs
MattiasBuelens: Hmmm... actually, this part is wrong:
> `ReadableStreamDefaultControllerError` will call `readRequest.errorSteps(e)`, which could call both `branch1.cancel()` and `branch2.cancel()`.

After a branch is errored, you can no longer cancel it. Once we call `ReadableStreamDefaultControllerError` for `branch2`, we know that `cancel2Algorithm` can never be called and thus `canceled2` can never become true. So there's no re-entrancy risk and we don't need any extra checks. 😁

index.bs
domenic: Original error as-is seems good. It might contain useful information!

domenic (approved):
Spec looks good to me at this point! I guess the new error cases, including for non-byte tee() are testable (although probably not in the reference implementation).

domenic (approved):
LGTM. Not sure why the bots are failing; try rebasing?

@ricea feel free to take a look

MattiasBuelens:
Yay! 🥳

I assume this means Chrome is interested in implementing? I'll go ahead and file an implementation bug then. 🙂

ricea:
Yes, Chrome is interested.

PR #1118, opened 2021-03-27
Fix cancelPromise in ReadableStreamTee being resolved twice (again)

MattiasBuelens:
I liked #1112 so much that I wanted to fix it twice! 😛

The following test case throws an `AssertionError`:
```javascript
const rs = new ReadableStream({});

const [branch1, branch2] = rs.tee();

const cancel1 = branch1.cancel();
await flushAsyncEvents();
const cancel2 = branch2.cancel();

await Promise.all([cancel1, cancel2]);
```

This time, the error comes from [this line](https://github.com/whatwg/streams/blob/dd76a17a3738d78708a8dfd8f0508e717d6a1988/reference-implementation/lib/abstract-ops/readable-streams.js#L376):
```
promise_test: Unhandled rejection with value: object "AssertionError: false == true"
Error
    at AssertionError.get_stack (http://127.0.0.1:51551/resources/testharness.js:3553:21)
    at new AssertionError (http://127.0.0.1:51551/resources/testharness.js:3546:27)
    at assert (http://127.0.0.1:51551/resources/testharness.js:3539:19)
    at Test.<anonymous> (http://127.0.0.1:51551/resources/testharness.js:618:29)
    at Test.step (http://127.0.0.1:51551/resources/testharness.js:1977:25)
    at http://127.0.0.1:51551/resources/testharness.js:2002:35
```

- [x] At least two implementers are interested (and none opposed):
   * Chome (@ricea)
   * Firefox (@evilpie)
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * web-platform-tests/wpt#28266
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1194266
   * Firefox: …
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=223962

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/streams/1118.html" title="Last updated on Mar 30, 2021, 11:21 PM UTC (9a5bcc7)">Preview</a> | <a href="https://whatpr.org/streams/1118/03bd01a...9a5bcc7.html" title="Last updated on Mar 30, 2021, 11:21 PM UTC (9a5bcc7)">Diff</a>

PR #1119, opened 2021-03-30
Fix dfn refs for arguments of "TransformStream/set up"

MattiasBuelens:
The references for the _transformAlgorithm_ and _flushAlgorithm_ arguments of "set up TransformStream" were incorrect in #1110. This fixes it.

Not sure how/why [`make deploy` succeeded](https://github.com/whatwg/streams/runs/2230655379) for 8c215873d9c720b76878638e1c0ee235dc08164a, but it failed on my PR #1118. 🤷

***
<a href="https://whatpr.org/streams/1119.html" title="Last updated on Mar 30, 2021, 10:47 PM UTC (a6b7f4f)">Preview</a> | <a href="https://whatpr.org/streams/1119/8c21587...a6b7f4f.html" title="Last updated on Mar 30, 2021, 10:47 PM UTC (a6b7f4f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6407,7 +6407,7 @@ reason.
 
  1. Let |transformStream| be a [=new=] {{TransformStream}}.
  1. [=TransformStream/Set up=] |transformStream| with <var
-    ignore>[=TransformStream/create/transformAlgorithm=]</var> set to an algorithm which, given
+    ignore>[=TransformStream/set up/transformAlgorithm=]</var> set to an algorithm which, given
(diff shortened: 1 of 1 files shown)


domenic:
Thanks for catching this. I believe it probably referenced the "live spec" at `https://streams.spec.whatwg.org/`, which at the time of deploy, exported these. There's no way to tell Bikeshed "this is an internal reference, so if it doesn't exist in the current document, error!"

domenic (approved):

Issue #1001, opened 2019-05-10
DetachArrayBuffer can throw


domenic:
Discovered and discussed in https://github.com/whatwg/html/issues/4601.

This impacts https://streams.spec.whatwg.org/#transfer-array-buffer, and all of its transitive callers. We need to find the appropriate point at which we'd catch the errors and transform them into rejected promises or errored streams.

This only impacts byte streams, which nobody implements yet, so the urgency is not super-high here, but we should still fix it.

Issue #495, opened 2016-08-04
Brainstorming a buffer-reusing version of writable streams


domenic:
I'd like to use this meta-issue to track any ideas for a version of writable streams that has some sort of buffer-reuse semantics, like our BYOB readable streams.

#465 has some previous discussion.

One thing I noticed while writing an example is that it's possible for an underlying sink to always transfer away a buffer. This is probably what most high-performance binary data underlying sinks will do, actually. (It's either transfer, copy, or expose raciness to the web.) But this has two drawbacks:

- The underlying sink doesn't see the buffer until the other buffers in the queue ahead of it have been consumed. So it doesn't get detached immediately. So there is a bit of an async race (not a multithreaded one) between any code that manipulates the buffer, and the underlying sink.
- The underlying sink has no opportunity to give the buffer back to the producer if they want to reuse it.

Here is a real example of this. Assume `writableStream` is a stream whose underlying sink transfers any chunks given to it (and errors if not given Uint8Arrays). Then consider this code:

```js
  async function writeRandomBytesForever(writableStream) {
    const writer = writableStream.getWriter();

    while (true) {
      if (Math.random() > 0.5) {
        await writer.ready;
      }

      const bytes = new Uint8Array(1024);
      window.crypto.getRandomValues(bytes);

      const promise = writer.write(bytes);
      doStuffWith(bytes);
      await promise;
    }
  }
```

Here the two problems are illustrated:

- `doStuffWith(bytes)` can maybe modify bytes, or maybe not. (If we did `await writer.ready`, the queue will possibly be empty, in which case you cannot modify, since the underlying sink already transferred it. But if we did not, then you can modify.)
- We have to reallocate new typed arrays every time through the loop; we cannot reuse the same one.

ricea:
How it works for non-blocking TCP is that userspace asks to write N bytes, and the kernel responds synchronously that it managed to write M bytes, where M <= N. Because it's synchronous, there's no risk that the buffer will be modified while the write happens (unless threads are involved, and then all bets are off). And once the write call returns, userspace is free to reuse the buffer any way it likes.

I'm hesitant to introduce synchronous APIs to Javascript, but I think it can work if a synchronous copy is done when the sink is unable to do synchronous writes.

When specifying standards involving writable streams, it would have to be unspecified whether the underlying sink was sync-capable or not, because different browser architectures would have different constraints. For example, I can imagine that Chrome might have streams that were sync-capable in the main thread but not in workers.

[...]

PR #1123, opened 2021-04-16
Various fixes for readable byte streams

MattiasBuelens:
This started as a few fixes that we encountered while working on #1114, but it ended up being... a lot of fixes. 😅

* `.respondWithNewView(newView)` can now be called with a `newView` that is *smaller than* the BYOB request's view. This aligns it with `.respond(bytesWritten)`, which also allows `bytesWritten <= byobRequest.view.byteLength`.
* `.respondWithNewView()` must now be called with an empty view when the stream is closed. This aligns it with `.respond(bytesWritten)`, which requires `bytesWritten` to be 0 when closed.
* `.respondWithNewView(newView)` must now be called with a view whose `view.buffer.byteLength` matches that of the BYOB request. Ideally, we would like to assert that the new view's buffer is the "transferred version" of the BYOB request's buffer, but that's not possible yet with the tools currently provided by the ECMAScript specification.
* `.enqueue(chunk)` and `.respondWithNewView(newView)` now check that the given view's buffer is actually *transferable*. Previously, we only checked whether the buffer is *not yet detached*, but this is insufficient: a `WebAssembly.Memory`'s buffer is *never* transferable. We also make sure to not transfer the given buffer until *after* we've checked all other preconditions, so the buffer is still intact if these methods were to throw an error.
* `.enqueue()` and `.respond()` now check that the BYOB request's view has *not* been transferred, since otherwise it's not possible to copy bytes into its buffer and/or transfer the buffer when committing. This could crash Chrome, but has already been fixed in https://crbug.com/1200302. The reproduction case (see below) is now covered by a WPT test.
* `.enqueue()`, `.respond()` and `.respondWithNewView()` immediately invalidate the BYOB request. Previously, we only did this if we actually filled the first pull-into descriptor, which doesn't *always* happen. (For example: if the pull-into descriptor's element size is 4, but we only have filled 1 or 2 bytes.)
* We now always transfer the pull-into descriptor's buffer when committing it (to fulfill a read request or read-into request). This is mainly a sanity check: the stream should never use this buffer after it has been committed.

<details>
<summary>Original description</summary>

While working on #1114, we found that `byobRequest.respondWithNewView(newView)` didn't support all the same use cases as a regular `.respond(bytesWritten)` call. In particular:
* `newView` was required to be the same length as the BYOB request's view. This would have required the underlying byte source to always fill *the entire view*. On the other hand, `.respond()` allows responding with fewer bytes *up to the view's length*, which is the desired behavior.
* When the stream is closed, `.respond()` can only be called with `bytesWritten` set to 0. To match this, `.respondWithNewView()` should only be called with an empty view in the closed state.

We now also require the view passed to `.respondWithNewView(newView)` to have the same capacity as the BYOB request's original buffer. Ideally, we would like to assert that the new view's buffer is the "transferred version" of the BYOB request's buffer, but that's not possible yet with the tools currently provided by the ECMAScript specification.

I also found cases where the stream would try to transfer an already transferred buffer. For example, this code crashes the tab in Chrome 92 (error code: `STATUS_ACCESS_VIOLATION`):
```javascript
let rs = new ReadableStream({
    type: 'bytes',
    pull(c) {
        let buffer = c.byobRequest.view.buffer;
        // Detach the buffer.
        postMessage(buffer, '*', [buffer]);
        // Try to enqueue with a new buffer.
        // The stream will try to copy to the pull-into request's view, but it can't because it has lost the buffer.
        c.enqueue(new Uint8Array([42]));
    }
});
let reader = rs.getReader({ mode: 'byob' });
await reader.read(new Uint8Array(1));
```
I fixed this by requiring the BYOB request's buffer to not be detached when calling `byteStreamController.enqueue()` or `.close()`.

</details>

   * …
   * …
   * web-platform-tests/wpt#28557
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1123.html" title="Last updated on May 26, 2021, 4:23 PM UTC (b22d61f)">Preview</a> | <a href="https://whatpr.org/streams/1123/cc29ed3...b22d61f.html" title="Last updated on May 26, 2021, 4:23 PM UTC (b22d61f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1992,7 +1992,6 @@ following table:
  The <dfn id="rs-byob-request-respond-with-new-view" method
  for="ReadableStreamBYOBRequest">respondWithNewView(|view|)</dfn> method steps are:
 
- 1. If |view|.\[[ByteLength]] is 0, throw a {{TypeError}} exception.
  1. If |view|.\[[ViewedArrayBuffer]].\[[ArrayBufferByteLength]] is 0, throw a {{TypeError}}
     exception.
  1. If [=this=].[=ReadableStreamBYOBRequest/[[controller]]=] is undefined, throw a {{TypeError}}
@@ -3203,6 +3202,7 @@ The following abstract operations support the implementation of the
      |firstDescriptor|).
  1. Otherwise,
   1. Assert: |state| is "`readable`".
+  1. If |bytesWritten| is 0, throw a {{TypeError}} exception.
   1. Perform ? [$ReadableByteStreamControllerRespondInReadableState$](|controller|, |bytesWritten|,
      |firstDescriptor|).
  1. Perform ! [$ReadableByteStreamControllerCallPullIfNeeded$](|controller|).
--- a/index.bs
+++ b/index.bs
@@ -2922,6 +2922,8 @@ The following abstract operations support the implementation of the
       |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
    1. If ! [$CanTransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
       is false, throw a {{TypeError}} exception.
+   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
+      [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
   1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|,
      |transferredBuffer|, |byteOffset|, |byteLength|).
   1. Perform ! [$ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue$](|controller|).
@@ -3150,6 +3152,8 @@ The following abstract operations support the implementation of the
   1. If |bytesWritten| is 0, throw a {{TypeError}} exception.
   1. If |firstDescriptor|'s [=pull-into descriptor/bytes filled=] + |bytesWritten| >
      |firstDescriptor|'s [=pull-into descriptor/byte length=], throw a {{RangeError}} exception.
+ 1. Set |firstDescriptor|'s [=pull-into descriptor/buffer=] to !
+    [$TransferArrayBuffer$](|firstDescriptor|'s [=pull-into descriptor/buffer=]).
  1. Perform ? [$ReadableByteStreamControllerRespondInternal$](|controller|, |bytesWritten|).
 </div>
 
@@ -3158,8 +3162,6 @@ The following abstract operations support the implementation of the
  id="readable-byte-stream-controller-respond-in-closed-state">ReadableByteStreamControllerRespondInClosedState(|controller|,
  |firstDescriptor|)</dfn> performs the following steps:
 
- 1. Set |firstDescriptor|'s [=pull-into descriptor/buffer=] to !
-    [$TransferArrayBuffer$](|firstDescriptor|'s [=pull-into descriptor/buffer=]).
  1. Assert: |firstDescriptor|'s [=pull-into descriptor/bytes filled=] is 0.
  1. Let |stream| be |controller|.[=ReadableByteStreamController/[[stream]]=].
  1. If ! [$ReadableStreamHasBYOBReader$](|stream|) is true,
@@ -3242,7 +3244,8 @@ The following abstract operations support the implementation of the
     |view|.\[[ViewedArrayBuffer]].\[[ByteLength]], throw a {{RangeError}} exception.
  1. If |firstDescriptor|'s [=pull-into descriptor/bytes filled=] + |view|.\[[ByteLength]] >
     |firstDescriptor|'s [=pull-into descriptor/byte length=], throw a {{RangeError}} exception.
- 1. Set |firstDescriptor|'s [=pull-into descriptor/buffer=] to |view|.\[[ViewedArrayBuffer]].
+ 1. Set |firstDescriptor|'s [=pull-into descriptor/buffer=] to !
+    [$TransferArrayBuffer$](|view|.\[[ViewedArrayBuffer]]).
  1. Perform ? [$ReadableByteStreamControllerRespondInternal$](|controller|, |view|.\[[ByteLength]]).
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -3220,7 +3220,7 @@ The following abstract operations support the implementation of the
     descriptor/bytes filled=] is not |view|.\[[ByteOffset]], throw a {{RangeError}} exception.
  1. If |firstDescriptor|'s [=pull-into descriptor/buffer byte length=] is not
     |view|.\[[ViewedArrayBuffer]].\[[ByteLength]], throw a {{RangeError}} exception.
- 1. If |firstDescriptor|'s [=pull-into descriptor/byte length=] is not |view|.\[[ByteLength]], throw
+ 1. If |firstDescriptor|'s [=pull-into descriptor/byte length=] < |view|.\[[ByteLength]], throw
     a {{RangeError}} exception.
  1. Set |firstDescriptor|'s [=pull-into descriptor/buffer=] to |view|.\[[ViewedArrayBuffer]].
  1. Perform ? [$ReadableByteStreamControllerRespondInternal$](|controller|, |view|.\[[ByteLength]]).
--- a/index.bs
+++ b/index.bs
@@ -2914,8 +2914,8 @@ The following abstract operations support the implementation of the
     [=list/is empty|empty=],
   1. Let |firstPendingPullInto| be
      |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
-  1. If ! [$CanTransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
-     is false, throw a {{TypeError}} exception.
+  1. If ! [$IsDetachedBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
+     is true, throw a {{TypeError}} exception.
   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
      [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
--- a/index.bs
+++ b/index.bs
@@ -2918,10 +2918,7 @@ The following abstract operations support the implementation of the
    1. Let |firstPendingPullInto| be
       |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
    1. If ! [$CanTransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
-      is false,
-    1. Let |e| be a new {{TypeError}} exception.
-    1. Perform ! [$ReadableByteStreamControllerError$](|controller|, |e|).
-    1. Throw |e|.
+      is false, throw a {{TypeError}} exception.
   1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|,
      |transferredBuffer|, |byteOffset|, |byteLength|).
   1. Perform ! [$ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue$](|controller|).
--- a/index.bs
+++ b/index.bs
@@ -1986,6 +1986,8 @@ following table:
 
  1. If [=this=].[=ReadableStreamBYOBRequest/[[controller]]=] is undefined, throw a {{TypeError}}
     exception.
+ 1. If ! [$CanTransferArrayBuffer$](|view|.\[[ViewedArrayBuffer]]) is false,
+    throw a {{TypeError}} exception.
  1. Return ?
     [$ReadableByteStreamControllerRespondWithNewView$]([=this=].[=ReadableStreamBYOBRequest/[[controller]]=],
     |view|).
@@ -3215,6 +3217,7 @@ The following abstract operations support the implementation of the
 
  1. Assert: |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not [=list/is
     empty|empty=].
+ 1. Assert: ! [$CanTransferArrayBuffer$](|view|.\[[ViewedArrayBuffer]]) is true.
  1. Let |firstDescriptor| be |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
  1. If |firstDescriptor|'s [=pull-into descriptor/byte offset=] + |firstDescriptor|' [=pull-into
     descriptor/bytes filled=] is not |view|.\[[ByteOffset]], throw a {{RangeError}} exception.
--- a/index.bs
+++ b/index.bs
@@ -2902,6 +2902,14 @@ The following abstract operations support the implementation of the
  1. Let |stream| be |controller|.[=ReadableByteStreamController/[[stream]]=].
  1. If |controller|.[=ReadableByteStreamController/[[closeRequested]]=] is true or
     |stream|.[=ReadableStream/[[state]]=] is not "`readable`", return.
+ 1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
+    [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be
+     |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. If ! [$CanTransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
+     is false, throw a {{TypeError}} exception.
+  1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
+     [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
  1. Let |buffer| be |chunk|.\[[ViewedArrayBuffer]].
  1. Let |byteOffset| be |chunk|.\[[ByteOffset]].
  1. Let |byteLength| be |chunk|.\[[ByteLength]].
@@ -2916,14 +2924,6 @@ The following abstract operations support the implementation of the
       |byteOffset|, |byteLength| »).
    1. Perform ! [$ReadableStreamFulfillReadRequest$](|stream|, |transferredView|, false).
  1. Otherwise, if ! [$ReadableStreamHasBYOBReader$](|stream|) is true,
-  1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
-     [=list/is empty|empty=],
-   1. Let |firstPendingPullInto| be
-      |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
-   1. If ! [$CanTransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
-      is false, throw a {{TypeError}} exception.
-   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
-      [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
   1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|,
      |transferredBuffer|, |byteOffset|, |byteLength|).
   1. Perform ! [$ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue$](|controller|).
--- a/index.bs
+++ b/index.bs
@@ -1817,6 +1817,8 @@ for="ReadableByteStreamController">enqueue(|chunk|)</dfn> method steps are:
  1. If |chunk|.\[[ByteLength]] is 0, throw a {{TypeError}} exception.
  1. If |chunk|.\[[ViewedArrayBuffer]].\[[ArrayBufferByteLength]] is 0, throw a {{TypeError}}
     exception.
+ 1. If ! [$CanTransferArrayBuffer$](|chunk|.\[[ViewedArrayBuffer]]) is false,
+    throw a {{TypeError}} exception.
  1. If [=this=].[=ReadableByteStreamController/[[closeRequested]]=] is true, throw a {{TypeError}}
     exception.
  1. If [=this=].[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[state]]=] is not
@@ -2894,6 +2896,7 @@ The following abstract operations support the implementation of the
  id="readable-byte-stream-controller-enqueue">ReadableByteStreamControllerEnqueue(|controller|,
  |chunk|)</dfn> performs the following steps:
 
+ 1. Assert: ! [$CanTransferArrayBuffer$](|chunk|.\[[ViewedArrayBuffer]]) is true.
  1. Let |stream| be |controller|.[=ReadableByteStreamController/[[stream]]=].
  1. If |controller|.[=ReadableByteStreamController/[[closeRequested]]=] is true or
     |stream|.[=ReadableStream/[[state]]=] is not "`readable`", return.
--- a/index.bs
+++ b/index.bs
@@ -1824,8 +1824,6 @@ for="ReadableByteStreamController">enqueue(|chunk|)</dfn> method steps are:
  1. If |chunk|.\[[ByteLength]] is 0, throw a {{TypeError}} exception.
  1. If |chunk|.\[[ViewedArrayBuffer]].\[[ArrayBufferByteLength]] is 0, throw a {{TypeError}}
     exception.
- 1. If ! [$CanTransferArrayBuffer$](|chunk|.\[[ViewedArrayBuffer]]) is false,
-    throw a {{TypeError}} exception.
  1. If [=this=].[=ReadableByteStreamController/[[closeRequested]]=] is true, throw a {{TypeError}}
     exception.
  1. If [=this=].[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[state]]=] is not
@@ -2907,6 +2905,11 @@ The following abstract operations support the implementation of the
  1. Let |stream| be |controller|.[=ReadableByteStreamController/[[stream]]=].
  1. If |controller|.[=ReadableByteStreamController/[[closeRequested]]=] is true or
     |stream|.[=ReadableStream/[[state]]=] is not "`readable`", return.
+ 1. Let |buffer| be |chunk|.\[[ViewedArrayBuffer]].
+ 1. Let |byteOffset| be |chunk|.\[[ByteOffset]].
+ 1. Let |byteLength| be |chunk|.\[[ByteLength]].
+ 1. If ! [$IsDetachedBuffer$](|buffer|) is true, throw a {{TypeError}} exception.
+ 1. Let |transferredBuffer| be ? [$TransferArrayBuffer$](|buffer|).
  1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
     [=list/is empty|empty=],
   1. Let |firstPendingPullInto| be
@@ -2916,10 +2919,6 @@ The following abstract operations support the implementation of the
   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
      [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
- 1. Let |buffer| be |chunk|.\[[ViewedArrayBuffer]].
- 1. Let |byteOffset| be |chunk|.\[[ByteOffset]].
- 1. Let |byteLength| be |chunk|.\[[ByteLength]].
- 1. Let |transferredBuffer| be ! [$TransferArrayBuffer$](|buffer|).
  1. If ! [$ReadableStreamHasDefaultReader$](|stream|) is true
   1. If ! [$ReadableStreamGetNumReadRequests$](|stream|) is 0,
    1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|,
--- a/index.bs
+++ b/index.bs
@@ -61,6 +61,11 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: Type; url: #sec-ecmascript-data-types-and-values
  text: TypeError; url: #sec-native-error-types-used-in-this-standard-typeerror; type: exception
  text: map; url: #sec-array.prototype.map; type: method; for: Array.prototype
+urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASM-JS-API-1
+ type: interface
+  text: WebAssembly.Memory; url: #memory
+ type: attribute
+  text: buffer; for: WebAssembly.Memory; url: #dom-memory-buffer
 url: https://wicg.github.io/compression/#compressionstream; spec: COMPRESSION; type: interface; text: CompressionStream
 </pre>
 
@@ -6136,6 +6141,9 @@ The following abstract operations are a grab-bag of utilities.
  1. Let |arrayBufferData| be |O|.\[[ArrayBufferData]].
  1. Let |arrayBufferByteLength| be |O|.\[[ArrayBufferByteLength]].
  1. Perform ? [$DetachArrayBuffer$](|O|).
+    <p class="note">This will throw an exception if |O| has an \[[ArrayBufferDetachKey]]
+    that is not undefined, such as a {{WebAssembly.Memory}}'s {{WebAssembly.Memory/buffer}}.
+    [[WASM-JS-API-1]]</p>
  1. Return a new {{ArrayBuffer}} object, created in [=the current Realm=], whose
     \[[ArrayBufferData]] internal slot value is |arrayBufferData| and whose
     \[[ArrayBufferByteLength]] internal slot value is |arrayBufferByteLength|.
--- a/index.bs
+++ b/index.bs
@@ -1721,6 +1721,8 @@ has the following [=struct/items=]:
 
 : <dfn for="pull-into descriptor">buffer</dfn>
 :: An {{ArrayBuffer}}
+: <dfn for="pull-into descriptor">buffer byte length</dfn>
+:: A positive integer representing the initial byte length of [=pull-into descriptor/buffer=]
 : <dfn for="pull-into descriptor">byte offset</dfn>
 :: A nonnegative integer byte offset into the [=pull-into descriptor/buffer=] where the
    [=underlying byte source=] will start writing
@@ -1887,8 +1889,9 @@ counterparts for default controllers, as discussed in [[#rs-abstract-ops-used-by
    1. Perform |readRequest|'s [=read request/error steps=], given |buffer|.\[[Value]].
    1. Return.
   1. Let |pullIntoDescriptor| be a new [=pull-into descriptor=] with [=pull-into descriptor/buffer=]
-     |buffer|.\[[Value]], [=pull-into descriptor/byte offset=] 0, [=pull-into descriptor/byte
-     length=] |autoAllocateChunkSize|, [=pull-into descriptor/bytes filled=] 0, [=pull-into
+     |buffer|.\[[Value]], [=pull-into descriptor/buffer byte length=] |autoAllocateChunkSize|,
+     [=pull-into descriptor/byte offset=] 0, [=pull-into descriptor/byte length=]
+     |autoAllocateChunkSize|, [=pull-into descriptor/bytes filled=] 0, [=pull-into
      descriptor/element size=] 1, [=pull-into descriptor/view constructor=] {{%Uint8Array%}}, and
      [=pull-into descriptor/reader type=] "`default`".
   1. [=list/Append=] |pullIntoDescriptor| to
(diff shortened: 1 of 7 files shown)


MattiasBuelens:
I also noticed that [ReadableByteStreamControllerRespondWithNewView](https://streams.spec.whatwg.org/#readable-byte-stream-controller-respond-with-new-view) can still throw an error *after* we've already updated the first pull-into descriptor's buffer to the new view's buffer. Specifically, [ReadableByteStreamControllerRespondInternal](https://streams.spec.whatwg.org/#readable-byte-stream-controller-respond-internal) may still throw a `TypeError` if *bytesWritten* is or is not zero (depending on whether we're in the readable or closed state).

I think we will need to turn these checks into asserts inside `RespondInternal()`, and throw errors in `Respond()` and `RespondWithNewView()`?

index.bs
```diff
@@ -3206,8 +3217,10 @@ The following abstract operations support the implementation of the
  1. Let |firstDescriptor| be |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
  1. If |firstDescriptor|'s [=pull-into descriptor/byte offset=] + |firstDescriptor|' [=pull-into
     descriptor/bytes filled=] is not |view|.\[[ByteOffset]], throw a {{RangeError}} exception.
- 1. If |firstDescriptor|'s [=pull-into descriptor/byte length=] is not |view|.\[[ByteLength]], throw
-    a {{RangeError}} exception.
+ 1. If |firstDescriptor|'s [=pull-into descriptor/buffer byte length=] is not
+    |view|.\[[ViewedArrayBuffer]].\[[ByteLength]], throw a {{RangeError}} exception.
+ 1. If |firstDescriptor|'s [=pull-into descriptor/bytes filled=] + |view|.\[[ByteLength]] >
+    |firstDescriptor|'s [=pull-into descriptor/byte length=], throw a {{RangeError}} exception.
```
MattiasBuelens, line 3223: This repeats the check in the first step of [ReadableByteStreamControllerRespondInReadableState](https://streams.spec.whatwg.org/#readable-byte-stream-controller-respond-in-readable-state). But [as I mentioned](https://github.com/whatwg/streams/pull/1123#issuecomment-821727018), we throw this error too late... 😕

index.bs
```diff
@@ -1981,7 +1991,6 @@ following table:
  The <dfn id="rs-byob-request-respond-with-new-view" method
  for="ReadableStreamBYOBRequest">respondWithNewView(|view|)</dfn> method steps are:
 
- 1. If |view|.\[[ByteLength]] is 0, throw a {{TypeError}} exception.
  1. If |view|.\[[ViewedArrayBuffer]].\[[ArrayBufferByteLength]] is 0, throw a {{TypeError}}
     exception.
```
MattiasBuelens, line 1995: Hmm, I suppose this check is now made redundant by this check in `ReadableByteStreamControllerRespondWithNewView`:
> 4. If firstDescriptor’s buffer byte length is not view.[[ViewedArrayBuffer]].[[ByteLength]], throw a RangeError exception.

...given that a pull-into's buffer can never be zero length.

I guess we can delete this?

MattiasBuelens:
Hmm, we should probably also tackle #1001 while we're at it. It's often not enough to check if `IsDetachedBuffer` returns false, we should actually try to transfer any user-provided `ArrayBuffer` as soon as possible and propagate any errors *before* we're adding/changing pull-into descriptors deep inside our abstract ops. 🤔

reference-implementation/lib/ReadableByteStreamController-impl.js
```diff
@@ -38,6 +39,13 @@ exports.implementation = class ReadableByteStreamControllerImpl {
       throw new TypeError(`The stream (in ${state} state) is not in the readable state and cannot be closed`);
     }
 
+    if (this._pendingPullIntos.length > 0) {
```
domenic, line 42: Why locate these checks here, instead of e.g. in ReadableByteStreamControllerClose step 4?

index.bs
domenic: Yeah, I guess so, although we'd want to update the tests to change the exception type expected.

domenic (commented):
> I think we will need to turn these checks into asserts inside RespondInternal(), and throw errors in Respond() and RespondWithNewView()?

That does sound nicer. It's possible our original reasoning was that in error states like these, there are no guarantees, but we might as well make it better.

reference-implementation/lib/ReadableByteStreamController-impl.js
MattiasBuelens: Good point, I'll move things around a bit. 👍

Currently, this also prevents valid usages like:
```javascript
const { done, value } = await transferAndFillViewSomehow(c.byobRequest.view);
if (done) {
  c.close(); // would throw when implemented as currently written
  c.byobRequest.respondWithNewView(value);
}
```
so the check *definitely* needs to move. 🙂

ricea:
I'm working on a fix for the Chrome crash that Mattias found, which I will land before the rest of these changes.

reference-implementation/lib/ReadableByteStreamController-impl.js
MattiasBuelens: Done. I removed this check, and added tests to ensure we can call `close()` and afterwards respond with a *transferred* view (see web-platform-tests/wpt@9b7f4358098f0aea35eb8de8333ec5b6b6f46446).

index.bs
```diff
@@ -6081,6 +6101,17 @@ abstract operations are used to implement these "cross-realm transforms".
...
 The following abstract operations are a grab-bag of utilities.
 
+<div algorithm>
+ <dfn abstract-op lt="CanTransferArrayBuffer"
+ id="can-transfer-array-buffer">CanTransferArrayBuffer(|O|)</dfn> performs the following steps:
+
+ 1. Assert: [$Type$](|O|) is Object.
+ 1. Assert: |O| has an \[[ArrayBufferData]] internal slot.
+ 1. If ! [$IsDetachedBuffer$](|O|) is true, return false.
+ 1. If [$SameValue$](|O|.\[[ArrayBufferDetachKey]], undefined) is false, return false.
```
MattiasBuelens, line 6111: This is the important bit. `IsDetachedBuffer(O) === false` does not guarantee that `DetachArrayBuffer(O)` will not throw an error, since the buffer may be "non-detachable" (e.g. if it's a `WebAssembly.Memory`'s buffer). We have to perform the exact same steps as `DetachArrayBuffer(O)`, except for actually detaching the buffer.

index.bs
```diff
@@ -3214,7 +3226,15 @@ The following abstract operations support the implementation of the
...
     empty|empty=].
+ 1. Assert: ! [$CanTransferArrayBuffer$](|view|.\[[ViewedArrayBuffer]]) is true.
  1. Let |firstDescriptor| be |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+ 1. Let |state| be
+    |controller|.[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[state]]=].
+ 1. If |state| is "`closed`",
+  1. If |view|.\[[ByteLength]] is not 0, throw a {{TypeError}} exception.
+ 1. Otherwise,
+  1. Assert: |state| is "`readable`".
+  1. If |view|.\[[ByteLength]] is 0, throw a {{TypeError}} exception.
```
MattiasBuelens, line 3237: These steps were moved from `RespondInternal()` to both `Respond()` and `RespondWithNewView()`. This way, we can throw errors *before* reaching this step in `RespondWithNewView()`:
> 10. Set *firstDescriptor*’s buffer to *view*.[[ViewedArrayBuffer]].

In `RespondInternal()`, these checks were turned into asserts.

index.bs
```diff
@@ -1820,15 +1817,13 @@ for="ReadableByteStreamController">enqueue(|chunk|)</dfn> method steps are:
...
  1. If [=this=].[=ReadableByteStreamController/[[closeRequested]]=] is true, throw a {{TypeError}}
     exception.
  1. If [=this=].[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[state]]=] is not
     "`readable`", throw a {{TypeError}} exception.
- 1. If [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] is not [=list/is empty|empty=],
-  1. Let |firstDescriptor| be [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
-  1. If [$IsDetachedBuffer$](|firstDescriptor|'s [=pull-into descriptor/buffer=]) is true,
-     throw a {{TypeError}} exception.
- 1. Return ! [$ReadableByteStreamControllerEnqueue$]([=this=], |chunk|).
+ 1. Return ? [$ReadableByteStreamControllerEnqueue$]([=this=], |chunk|).
```
MattiasBuelens: The transferable (previously: "not detached") check is moved to `ReadableByteStreamControllerEnqueue()`, so it can now throw.

Or would you prefer to keep the check in `ReadableByteStreamController.enqueue()`, and have `ReadableByteStreamControllerEnqueue()` assert that the pull-into descriptor's buffer is indeed still transferable?

index.bs
```diff
@@ -1806,10 +1807,6 @@ has the following [=struct/items=]:
     exception.
  1. If [=this=].[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[state]]=] is not
     "`readable`", throw a {{TypeError}} exception.
- 1. If [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] is not [=list/is empty|empty=],
-  1. Let |firstDescriptor| be [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
-  1. If [$IsDetachedBuffer$](|firstDescriptor|'s [=pull-into descriptor/buffer=]) is true,
-     throw a {{TypeError}} exception.
```
MattiasBuelens: [As mentioned here](https://github.com/whatwg/streams/pull/1123#discussion_r616209252), we need to ensure that we can call `close()` and still respond with a *transferred* view. Tests are in web-platform-tests/wpt@9b7f4358098f0aea35eb8de8333ec5b6b6f46446.

ricea:
I included a test in the fix, which is here: https://github.com/web-platform-tests/wpt/pull/28586

MattiasBuelens:
Thanks! I'll see if I can integrate your test in my new tests somewhere.

That's a clever idea to call `controller.error()` after the failed `controller.enqueue()`, to make sure `reader.read()` still settles. I think I'll use that in my new tests too. 👍

MattiasBuelens:
I noticed [this TODO comment](https://github.com/whatwg/streams/blob/cc29ed3921eceaae700fa9a36cb74c9d8ede0236/reference-implementation/lib/abstract-ops/readable-streams.js#L1244) in `ReadableByteStreamControllerRespondInReadableState`:
```javascript
  if (pullIntoDescriptor.bytesFilled < pullIntoDescriptor.elementSize) {
    // TODO: Figure out whether we should detach the buffer or not here.
    return;
  }
```
This seems relevant, so let's discuss it.

I think that after you successfully called `controller.byobRequest.respond()` or `.respondWithNewView()`, the BYOB request should be invalidated and its buffer should be transferred. You should no longer be able to modify the old value of `controller.byobRequest.view`:
```javascript
reader.read(new Uint32Array(1));

const byobRequest1 = controller.byobRequest;
const view1 = byobRequest1.view; // 4-byte view on a 4-byte buffer
view1[0] = 0x01;
byobRequest1.respond(1);

// At this point, `byobRequest1` should invalidated and `view1.buffer` should be detached.
view1.buffer.byteLength === 0;
view1[0] = 42; // should not work
```

Instead, you should always request a new `controller.byobRequest` after calling `respond()` and fill the new view on the remaining part of the pull-into descriptor:
```javascript
const byobRequest2 = controller.byobRequest;
const view2 = byobRequest2.view; // 3-byte view on a 4-byte buffer
view2[0] = 0x02;
view2[1] = 0x03;
view2[2] = 0x04;
byobRequest2.respond(3);

// read() resolves with 0x04030201
```
Thoughts?

MattiasBuelens:
Hmm, `enqueue()` also invalidates the BYOB request but doesn't always detach the previous request's buffer either.
```javascript
reader.read(new Uint32Array(1));

const byobRequest1 = controller.byobRequest;
const view1 = byobRequest1.view;
controller.enqueue(new Uint8Array([0x01]));
view1[0] === 0x01;
```
Maybe this was intentional? Should the underlying byte stream be able to continue using the old view as long as we haven't committed the pull-into descriptor? 🤔

---

One thing that we definitely *should* fix, is actually transferring the request's buffer when we commit the pull-into descriptor. Right now, we aren't actually doing that when using `enqueue()`... oops! 😅
```javascript
let byobView;
  type: 'bytes',
  pull(controller) {
    byobView = controller.byobRequest.view;
    controller.enqueue(new Uint8Array([0x01]));
  }
})

const reader = rs.getReader({ mode: 'byob' });
const { done, value } = await reader.read(new Uint8Array(1));
value[0] === 0x01;
byobView.buffer === value.buffer; // true... oops, forgot to transfer!
```

domenic:
> I think that after you successfully called controller.byobRequest.respond() or .respondWithNewView(), the BYOB request should be invalidated and its buffer should be transferred. You should no longer be able to modify the old value of controller.byobRequest.view

This makes sense to me.

> Hmm, enqueue() also invalidates the BYOB request but doesn't always detach the previous request's buffer either. Maybe this was intentional?

I don't think it was intentional... unless we wrote tests exercising it. Probably tightening it up is safer until/unless someone discovers a use case.

---

This PR is getting rather large, but all good fixes. I have no problem doing them all together, but would you be able to update the OP with a running summary of the changes, in the form of what an eventual commit message would look like?

MattiasBuelens:
PR description updated.

As discussed, I made `.enqueue()`, `.respond()` and `.respondWithNewView()` immediately invalidate the BYOB request. I still need to write a test for this (i.e. check that the old request's `.view` becomes `null`).

domenic:
I apologize for us leaving this PR alone for so long. I had a few weeks in a row of vaccine recovery, vacation, and then working on other stuff. (Although, I see it's still in draft state; is there more to do?)

Of the goals listed in the OP, I'm slightly concerned about

> .respondWithNewView() must now be called with an empty view when the stream is closed. This aligns it with .respond(bytesWritten), which requires bytesWritten to be 0 when closed.

and

> .respondWithNewView(newView) must now be called with a view whose view.buffer.byteLength matches that of the BYOB request. Ideally, we would like to assert that the new view's buffer is the "transferred version" of the BYOB request's buffer, but that's not possible yet with the tools currently provided by the ECMAScript specification.

since they have the potential to break code. I'm not as worried about compat, but slightly worried about potentially shutting down valid use cases. I admit this is largely because I lost all the context of our previous discussions; I recall being somewhat convinced these were good changes.

I'd also appreciate @ricea's review from an implementation point of view!

MattiasBuelens:
> I apologize for us leaving this PR alone for so long. I had a few weeks in a row of vaccine recovery, vacation, and then working on other stuff. (Although, I see it's still in draft state; is there more to do?)

No worries! Glad to hear you got your vaccinations. The rollout has been a bit slower here in Europe, hopefully I can get one in June or July. 🤞

The draft state is because I wasn't sure if more tests are needed. I'll have another look to see if I can think of other interesting cases that need to be covered.

> Of the goals listed in the OP, I'm slightly concerned about [...] and [...] since they have the potential to break code. I'm not as worried about compat, but slightly worried about potentially shutting down valid use cases.

The way I see it, `.respondWithNewView()` in its current state is *preventing* a lot of valid use cases. These changes are not so much about imposing more restrictions, but actually making it work as advertised.

For example, right now it's *impossible* to use `.respondWithNewView()` when the stream is in the closed state. If you try to call it with a non-empty view:
```javascript
  type: 'bytes',
  async pull(c) {
    const view = c.byobRequest.view;
    c.close();
    try {
      c.byobRequest.respondWithNewView(view);
    } catch (e) {
      // pull() silences any rejections after calling close(), so we log this separately:
      console.error("respondWithNewView errored with:", e);
    }
  }
});
const reader = rs.getReader({ mode: 'byob' });
await reader.read(new Uint8Array(1));
```
it throws the following error:
> TypeError: Failed to execute 'respondWithNewView' on 'ReadableStreamBYOBRequest': bytes written is not 0

If you change it to `respondWithNewView(view.subarray(0, 0))` to pass a zero-length view (to satisfy the "bytes written must be 0" requirement), then you get this error instead:
> TypeError: Failed to execute 'respondWithNewView' on 'ReadableStreamBYOBRequest': ArrayBufferView is empty

If you're unable to call `respondWithNewView`, then you cannot return ownership of the BYOB request's buffer back to the stream, and so the stream cannot fulfill the pending read-into request. Thus, the stream is stuck.

That's why this change is necessary:
> .respondWithNewView() must now be called with an empty view when the stream is closed. This aligns it with .respond(bytesWritten), which requires bytesWritten to be 0 when closed.

With this, the second case (passing `view.subarray(0, 0)`) would be accepted when the stream is closed. Ownership of the buffer is returned, and the read-into request can be fulfilled.

As for the second change:
> .respondWithNewView(newView) must now be called with a view whose view.buffer.byteLength matches that of the BYOB request. Ideally, we would like to assert that the new view's buffer is the "transferred version" of the BYOB request's buffer, but that's not possible yet with the tools currently provided by the ECMAScript specification.

This is to fulfill the contract of `read(view)`. From the [domintro](https://streams.spec.whatwg.org/#byob-reader-prototype):
> If the chunk does become available, the promise will be fulfilled with an object of the form `{ value: theChunk, done: false }`. In this case, view will be detached and no longer usable, but `theChunk` will be a new view (of the same type) onto the same backing memory region, with the chunk’s data written into it. 

`read(view)` transfers the view's buffer to the stream, which then constructs a BYOB request around it. If we are ever to resolve the `read(view)` promise with a new view *on the same backing memory region*, then the underlying byte source *must give us back the buffer from the BYOB request*.

If it doesn't transfer the BYOB request's view, then it can simply call `.respond()`. The stream will assert that the BYOB request's buffer has not yet been transferred, so it can safely give it back to the result of `read(view)`.

If it *does* transfer the BYOB request's view (e.g. pass it to a worker to fill it), then it must ensure that it (eventually) receives that transferred view back (e.g. the worker passing it back). The stream has no way of knowing where the original BYOB request's view has been transferred to, so the underlying byte source *must* call `.respondWithNewView()`. It's at this point where we would have liked to assert that this new view is indeed backed by the same memory that was in the original BYOB request, so that we can ensure that `read(view)` will get its original buffer back. But alas, there's no way to check this with the current ECMAScript specification, so the closest we can get is to assert that the backing buffer's length is the same as it was when we created the BYOB request (and hope that the underlying byte source didn't do something fishy with it).

In case you want to toy around with transferring views inside `pull()`, here's an example:
```javascript
function transferView(view) {
  // this could also be done with postMessage()
    type: 'bytes',
    pull(c) {
      c.byobRequest.respond(c.byobRequest.view.byteLength);
    }
  });
  return rs.getReader({ mode: 'byob' }).read(view).then(result => result.value);
}

  type: 'bytes',
  async pull(c) {
    var view = await transferView(c.byobRequest.view);
    c.byobRequest.respondWithNewView(view);
  }
});
const reader = rs.getReader({ mode: 'byob' });
await reader.read(new Uint8Array([42])); // should return a result with a transferred but otherwise unmodified view
```

I hope that helps. If it's still unclear, just ask. 😉

domenic:
> With this, the second case (passing view.subarray(0, 0)) would be accepted when the stream is closed. Ownership of the buffer is returned, and the read-into request can be fulfilled.

Right, OK, this is coming back to me now. So I guess the question is whether we want to support this use case. In particular, it seems like an underlying source that closes a stream without first returning the pending BYOB request is doing the wrong thing. Allowing it to do the wrong thing but then return the buffer by doing `view.subarray(0, 0)` doesn't seem that useful?

From the other direction, if a stream consumer wants to cancel the stream (thus closing it), do they really expect to get their buffer back? It seems like annoying boilerplate to require that underlying source authors intercept cancels and send back the buffer. I'm inclined to just say that if you cancel the stream, you won't get your buffer back.

> As for the second change: [...] This is to fulfill the contract of read(view). From the domintro:

So my main question is whether the contract, as explained in the domintro, is meant to be 100% accurate, or is instead meant to cover the happy path.

I wish we had better historical notes on what the purpose of respondWithNewView() was. (The only reference I can find is https://github.com/whatwg/streams/issues/495#issuecomment-256945073 which suports your interpretation.) But it seems possible that one envisioned use case was when it's wasteful for the stream to write into the same backing memory region, for whatever reason. For example, the stream was handed some data from a non-BYOB API, or had some queued up because backpressure hadn't fully propagated, or whatever. In that case, respondWithNewView() can be used to say "I know you handed me a buffer, but instead of copying into that and wasting everyone's time, let me just give you a new one".

Probably this is not accurate. E.g. the spec's existing check on view length implies that such freedom was not intended. So I am comfortable with this change. I just thought I'd throw out the above paragraph in case you had thoughts :).

MattiasBuelens:
> Right, OK, this is coming back to me now. So I guess the question is whether we want to support this use case. In particular, it seems like an underlying source that closes a stream without first returning the pending BYOB request is doing the wrong thing. Allowing it to do the wrong thing but then return the buffer by doing `view.subarray(0, 0)` doesn't seem that useful?

The problem with that is we currently also *disallow* responding with 0 bytes written when still in the *readable* state. So it's not possible to do `respond(0); close();`, you *have* to do `close(); respond(0);`. If you also transferred the buffer, the equivalent of that would be to use `close(); respondWithNewView(transferredView.subarray(0, 0));`

One alternative would be to have a separate `closeWithNewView(transferredView);` method to do both at the same time. But then again, `close(); respondWithNewView(emptyTransferredView);` already does the trick. So unless there is a considerable benefit to this (some sort of optimization that is impossible with close+respond), I don't think it's worth adding a new API for this.

> From the other direction, if a stream consumer wants to cancel the stream (thus closing it), do they really expect to get their buffer back? It seems like annoying boilerplate to require that underlying source authors intercept cancels and send back the buffer. I'm inclined to just say that if you cancel the stream, you won't get your buffer back.

Right, we started the discussion in [#1114 (comment)](https://github.com/whatwg/streams/pull/1114#discussion_r605259271). The idea was that if you cancel the stream, then we *immediately* fulfill the pending read-into requests with `{ done: true, value: undefined }` and we don't bother waiting for the underlying byte source to give us those buffers back.

I'm not opposed to the idea. We'd have to refactor the specification some more though, so I'd probably want to do it in a separate PR (since this one is already quite big).

> For example, the stream was handed some data from a non-BYOB API, or had some queued up because backpressure hadn't fully propagated, or whatever. In that case, respondWithNewView() can be used to say "I know you handed me a buffer, but instead of copying into that and wasting everyone's time, let me just give you a new one".

We already have `ReadableByteStreamController.enqueue()` for this exact case. That will put the new data into the queue, and then try to fill up any pull-into requests by copying data from the queue. 😉

The way I understand it, `respondWithNewView()`'s sole purpose is to allow the underlying byte source to respond with a transferred version of the BYOB request's buffer. For all other use cases, there's `respond()` (if you didn't transfer) or `enqueue()` (if you didn't use the BYOB request at all).

domenic:
Haha, how do I keep forgetting about enqueue(). I'm pretty convinced; there might be a few things to discuss on the first point, but by now I feel comfortable reviewing the PR for editorial issues and the tests for coverage come Monday. I'll be back then!

ricea:
Given the limited use case of `respondWithNewView()` and large number of ways to get it wrong, could we just eliminate it?

MattiasBuelens:
> Given the limited use case of `respondWithNewView()` and large number of ways to get it wrong, could we just eliminate it?

Not really. Although the use cases are rare, they *are* important:
* To tee a readable byte stream (#1114), we want to use the BYOB request from a branch to perform a BYOB read on the original stream. That `read(view)` call *always* transfers the given buffer, so the branch's BYOB request becomes detached. The only way to get the result from `read(view)` back into the branch is [by calling `respondWithNewView()`](https://github.com/whatwg/streams/blob/8aa13163a1cfaf1d8be0070e505ef15aa418a806/reference-implementation/lib/abstract-ops/readable-streams.js#L565). (We *could* [implement teeing without using a BYOB reader](https://github.com/whatwg/streams/blob/8aa13163a1cfaf1d8be0070e505ef15aa418a806/reference-implementation/lib/abstract-ops/readable-streams.js#L484), but that kind of defeats the point of using a readable byte stream over a "regular" readable stream with `Uint8Array` chunks...)
* If we ever want to make readable byte streams transferable *without copying chunks around*, we will *need* the ability to transfer BYOB requests between realms. This won't be for any time soon though: transferable streams are currently implemented as pipes between cross-realm readable/writable streams, and we don't have "writable byte streams" yet (#495).

domenic (approved):
Overall LGTM, although @ricea's review would be good. One potential note:

> .enqueue(chunk) and .respondWithNewView(newView) now check that the given view's buffer is actually transferable. Previously, we only checked whether the buffer is not yet detached, but this is insufficient: a WebAssembly.Memory's buffer is never transferable. We also make sure to not transfer the given buffer until after we've checked all other preconditions, so the buffer is still intact if these methods were to throw an error.

I'm wondering if we should restructure things to naturally let TransferArrayBuffer throw (and perhaps that thrown exception would get turned into a promise rejection in some cases). This only works if we can transfer after all precondition checks, but before any other state mutations. But at a glance that appears to be how you've factored things?

I guess that's kind of a subtle hidden requirement on us as spec maintainers though, which is not great...

The reason I say this is because I suspect in implementations it's much easier to just transfer than it is to check for transferability. So they might want to implement things that way anyway.

MattiasBuelens:
> I'm wondering if we should restructure things to naturally let TransferArrayBuffer throw (and perhaps that thrown exception would get turned into a promise rejection in some cases). This only works if we can transfer after all precondition checks, but before any other state mutations. But at a glance that appears to be how you've factored things?

I think I started with that once, but then ran into some roadblock which required me to check transferability in advance. But that's a long time ago, perhaps the more recent changes made it possible. I'll have another try. 🙂

> I guess that's kind of a subtle hidden requirement on us as spec maintainers though, which is not great...

Indeed, we must be careful to use `?` instead of `!` in places where we *might* be trying to transfer a user-provided (and thus potentially WebAssembly-backed) `ArrayBuffer`. Fortunately, I don't think there are *that* many places where this can happen, most of our `TransferArrayBuffer` calls happen internally on an already-transferred (and thus definitely transferable) `ArrayBuffer`.

reference-implementation/lib/ReadableByteStreamController-impl.js
```diff
@@ -48,6 +49,9 @@ exports.implementation = class ReadableByteStreamControllerImpl {
     if (chunk.buffer.byteLength === 0) {
       throw new TypeError('chunk\'s buffer must have non-zero byteLength');
     }
+    if (CanTransferArrayBuffer(chunk.buffer) === false) {
```
MattiasBuelens, line 52: This check can easily be removed. We can move [these lines](https://github.com/whatwg/streams/blob/5da8d5c2bfdde4e3ce34ec3e63f4bd3f86922390/reference-implementation/lib/abstract-ops/readable-streams.js#L1010-L1013) in `ReadableByteStreamControllerEnqueue` to the start of that function, so that `TransferArrayBuffer` happens earlier (before we invalidate the BYOB request).

reference-implementation/lib/ReadableStreamBYOBRequest-impl.js
```diff
@@ -14,7 +14,7 @@ exports.implementation = class ReadableStreamBYOBRequestImpl {
       throw new TypeError('This BYOB request has been invalidated');
     }
 
-    if (IsDetachedBuffer(this._view.buffer) === true) {
+    if (CanTransferArrayBuffer(this._view.buffer) === false) {
```
MattiasBuelens, line 17: We can turn this back into a regular `IsDetachedBuffer` check without problems. `this._view` is always an `Uint8Array` that we constructed ourselves around the buffer of the first pull-into descriptor, which we have been checked earlier (i.e. received from `BYOBReader.read(view)` or created ourselves through `autoAllocateChunkSize`).

reference-implementation/lib/ReadableStreamBYOBReader-impl.js
```diff
@@ -17,6 +18,9 @@ class ReadableStreamBYOBReaderImpl {
     if (view.buffer.byteLength === 0) {
       return promiseRejectedWith(new TypeError('view\'s buffer must have non-zero byteLength'));
     }
+    if (CanTransferArrayBuffer(view.buffer) === false) {
```
MattiasBuelens, line 21: This one can also be changed to `IsDetachedBuffer`. We need to change `ReadableByteStreamControllerPullInto` to do a `try..catch` around [this line](https://github.com/whatwg/streams/blob/5da8d5c2bfdde4e3ce34ec3e63f4bd3f86922390/reference-implementation/lib/abstract-ops/readable-streams.js#L1175), and then immediately call `readIntoRequest.errorSteps()`.

reference-implementation/lib/ReadableStreamBYOBRequest-impl.js
```diff
@@ -25,17 +25,14 @@ exports.implementation = class ReadableStreamBYOBRequestImpl {
...
-    }
-    if (view.buffer.byteLength === 0) {
-      throw new TypeError('chunk\'s buffer must have non-zero byteLength');
-    }
-
     if (this._controller === undefined) {
       throw new TypeError('This BYOB request has been invalidated');
     }
 
+    if (CanTransferArrayBuffer(view.buffer) === false) {
```
MattiasBuelens, line 32: I think it's fine to change this back to `IsDetachedBuffer` too? We'll throw an error on [this line](https://github.com/whatwg/streams/blob/5da8d5c2bfdde4e3ce34ec3e63f4bd3f86922390/reference-implementation/lib/abstract-ops/readable-streams.js#L1339) instead, which is still *before* any state is mutated.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -994,6 +995,18 @@ function ReadableByteStreamControllerEnqueue(controller, chunk) {
     return;
   }
 
+  if (controller._pendingPullIntos.length > 0) {
+    const firstPendingPullInto = controller._pendingPullIntos[0];
+    if (CanTransferArrayBuffer(firstPendingPullInto.buffer) === false) {
```
MattiasBuelens, line 1000: We already know that the buffer of a pull-into descriptor has been transferred before when we constructed it, so this can be changed to `IsDetachedBuffer` too.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1263,18 +1289,17 @@ function ReadableByteStreamControllerRespondInReadableState(controller, bytesWri
 
 function ReadableByteStreamControllerRespondInternal(controller, bytesWritten) {
   const firstDescriptor = controller._pendingPullIntos[0];
+  assert(CanTransferArrayBuffer(firstDescriptor.buffer) === true);
```
MattiasBuelens, line 1292: `IsDetachedBuffer` is sufficient here. Both `ReadableByteStreamControllerRespond` and `ReadableByteStreamControllerRespondWithNewView` will have called `TransferArrayBuffer` already.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1283,24 +1308,42 @@ function ReadableByteStreamControllerRespondInternal(controller, bytesWritten) {
 
 function ReadableByteStreamControllerRespondWithNewView(controller, view) {
   assert(controller._pendingPullIntos.length > 0);
+  assert(CanTransferArrayBuffer(view.buffer) === true);
```
MattiasBuelens, line 1311: We'll change this back to `IsDetachedBuffer`. The `TransferArrayBuffer` call in this function happens before any state mutations, so it's fine.

MattiasBuelens (commented):
Okay, I think it *will* work to remove `CanTransferArrayBuffer`! 😁

reference-implementation/lib/abstract-ops/readable-streams.js
domenic: I'd be happy leaving CanTransferArrayBuffer in the asserts, for spec clarity.

index.bs
```diff
@@ -55,11 +55,17 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
...
   text: SetFunctionLength; url: #sec-setfunctionlength
   text: SetFunctionName; url: #sec-setfunctionname
   text: Type; url: #sec-ecmascript-data-types-and-values
  text: TypeError; url: #sec-native-error-types-used-in-this-standard-typeerror; type: exception
  text: map; url: #sec-array.prototype.map; type: method; for: Array.prototype
+urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASM-JS-API-1
+ type: interface
+  text: WebAssembly.Memory; url: #memory
+ type: attribute
+  text: buffer; for: WebAssembly.Memory; url: #dom-memory-buffer
```
MattiasBuelens, line 68: No idea if I'm doing this right. It *seems* to work fine? 😬

I'm also not sure why I even need to specify this manually. Am I missing some auto-linking magic that would allow me to just write `{{WebAssembly.Memory}}` and have it automatically find the correct spec (like `{{Response}}` [here](https://github.com/whatwg/streams/blob/0ff6d4508b58acce28386696b2854b2a1570d205/index.bs#L87))?

index.bs
```diff
@@ -6082,12 +6140,13 @@ The following abstract operations are a grab-bag of utilities.
...
- 1. Assert: [$Type$](|O|) is Object.
- 1. Assert: |O| has an \[[ArrayBufferData]] internal slot.
  1. Assert: ! [$IsDetachedBuffer$](|O|) is false.
  1. Let |arrayBufferData| be |O|.\[[ArrayBufferData]].
  1. Let |arrayBufferByteLength| be |O|.\[[ArrayBufferByteLength]].
- 1. Perform ! [$DetachArrayBuffer$](|O|).
+ 1. Perform ? [$DetachArrayBuffer$](|O|).
+    <p class="note">This will throw an exception if |O| has an \[[ArrayBufferDetachKey]]
+    that is not undefined, such as a {{WebAssembly.Memory}}'s {{WebAssembly.Memory/buffer}}.
+    [[WASM-JS-API-1]]</p>
```
MattiasBuelens, line 6149: Maybe double-check this bit too? 😇

MattiasBuelens (commented):
There we go. We now use `? TransferArrayBuffer(buffer)` where needed. 😁

I also added a note about `TransferArrayBuffer` and `WebAssembly.Memory`, but I'm not 100% sure if I got the Bikeshed syntax correct.

index.bs
domenic: I think the interface name should be `Memory`, so you'd do something like `{{Memory|WebAssembly.Memory}}`. However it appears not to work. I suspect that's because the Bikeshed database does not yet contain the spec: Maybe only when the AppHistory flag is enabled, I dunn

I'd suggest updating this to use `Memory` as the interface name, here and at the usage sites, so that in the future we can delete this link block.

index.bs
```diff
@@ -3151,8 +3186,8 @@ The following abstract operations support the implementation of the
  id="readable-byte-stream-controller-respond-in-readable-state">ReadableByteStreamControllerRespondInReadableState(|controller|,
  |bytesWritten|, |pullIntoDescriptor|)</dfn> performs the following steps:
 
- 1. If |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] + |bytesWritten| >
-    |pullIntoDescriptor|'s [=pull-into descriptor/byte length=], throw a {{RangeError}} exception.
+ 1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] + |bytesWritten| <=
```
domenic, line 3189:
```suggestion
 1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] + |bytesWritten| ≤
```

domenic (commented):
LGTM with nits

ricea (commented):
lgtm except for one thing.

index.bs
```diff
@@ -2898,7 +2908,17 @@ The following abstract operations support the implementation of the
...
+ 1. If ! [$IsDetachedBuffer$](|buffer|) is true, throw a {{TypeError}} exception.
+ 1. Let |transferredBuffer| be ? [$TransferArrayBuffer$](|buffer|).
+ 1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
+    [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be
+     |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. If ! [$IsDetachedBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
+     is true, throw a {{TypeError}} exception.
+  1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
+     [$TransferArrayBuffer$]([=pull-into descriptor/buffer=]).
```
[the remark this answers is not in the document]
MattiasBuelens: Thanks! 😄

ricea:
I am interested in this on behalf of Chrome.

Do you want me to file an issue for Chrome?

MattiasBuelens:
Sure, go ahead. 🙂

Who do we need to ping for interest from other browsers?

Do we need to file bugs for other browsers? AFAIK readable byte streams are only implemented in Chrome at the moment, other browsers haven't implemented it yet.

domenic:
IMO it's OK to treat bugfixes in readable byte streams as falling under the general precedent that other browsers have already expressed interest in readable byte streams; we don't need to block on them reviewing the fixes. (Although they're welcome to comment!) It's kind of a judgment call what's a bugfix vs. a normative change that would rise to the level of needing confirmation, but IMO these are all fine.

index.bs
```diff
@@ -1950,6 +1960,13 @@ following table:
   {{ReadableStreamBYOBRequest/view}}, the [=underlying byte source=] is providing a new
   {{ArrayBufferView}}, which will be given to the [=consumer=] of the [=readable byte stream=].
 
+  <p>The new |view| has to be a view onto the same backing memory region as
+  {{ReadableStreamBYOBRequest/view}}, i.e. its buffer has to equal (or be a
+  <a href="#transfer-array-buffer">transferred</a> version of) {{ReadableStreamBYOBRequest/view}}'s
+  buffer. Its <code>byteOffset</code> has to equal {{ReadableStreamBYOBRequest/view}}'s
+  <code>byteOffset</code>, and its <code>byteLength</code> (representing the number of bytes written)
+  has to be less than or equal to that of {{ReadableStreamBYOBRequest/view}}.
+
```
MattiasBuelens, line 1969: Almost forgot: [we wanted to document](https://github.com/whatwg/streams/pull/1114#discussion_r603504851) how to correctly use `respondWithNewView`. So here's that.

ricea: lgtm

MattiasBuelens:
FYI since I couldn't find a Chromium issue for this spec change yet, I've filed one myself: https://crbug.com/1223565

ricea:
> FYI since I couldn't find a Chromium issue for this spec change yet, I've filed one myself: https://crbug.com/1223565

Thanks! I dropped the ball there.

PR #1125, opened 2021-05-10
Remove unused parameter

Dragiyski:
``WritableStreamFinishErroring`` is only invoked with one parameter, but it has signature with two parameters, the second one seems unused. It seems ``reason`` was added by mistake due to similarly named ``WritableStreamStartErroring``, while it actually should exists only as an argument to ``Upon promise rejection`` step.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4294,7 +4294,7 @@ the {{WritableStream}}'s public API.
 
 <div algorithm>
  <dfn abstract-op lt="WritableStreamFinishErroring"
- id="writable-stream-finish-erroring">WritableStreamFinishErroring(|stream|, |reason|)</dfn>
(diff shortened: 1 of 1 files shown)


ricea:
lgtm

It's weird that this bug is only in the standard, and not in the reference implementation.

Could you sign the participation agreement at https://participate.whatwg.org/agreement-status?user=Dragiyski&repo=streams&pull=1125 ?

Dragiyski:
> lgtm
> 
> It's weird that this bug is only in the standard, and not in the reference implementation.

It is possible this is already caught by a linter tool, although I do not know what the reference implementation uses.

domenic (approved):
Thanks for finding this!

PR #1129, opened 2021-05-26
Resolve BYOB reads immediately on cancel

MattiasBuelens:
Right now, if you cancel the stream while a BYOB read is pending, the stream has to wait for the underlying byte source to call `respond(0)` before it can return the BYOB request's buffer to the caller. Effectively, every single underlying byte source would need to look like this:
```javascript
  type: 'bytes',
  autoAllocateChunkSize: 1024,
  start(controller) {
    this._controller = controller;
  },
  async pull(controller) {
    const bytesWritten = await readBytesIntoViewSomehow(controller.byobRequest.view);
    controller.byobRequest.respond(bytesWritten);
  },
  cancel(reason) {
    this._controller.byobRequest?.respond(0); // must not forget this!
  }
});
```
This is a footgun: we *have* to rely on the underlying byte source to do this correctly. We are unable to do this automatically, since we don't know if or how the source might be using the BYOB request (for example, it might have transferred the buffer). If the source fails to call `respond(0)`, then the BYOB read never resolves and the reader gets stuck. If you forget to implement `cancel()` altogether, then the default implementation will not help either, since it is specified to do nothing.

In [#1114 (comment)](https://github.com/whatwg/streams/pull/1114#discussion_r605259271) and [#1123 (comment)](https://github.com/whatwg/streams/pull/1123#issuecomment-841413902), @domenic suggested that if you cancel the stream while a BYOB read is pending, the stream should simply not give you back your buffer. This means that `cancel()` can immediately resolve all pending BYOB reads with the classic `{ done: true, value: undefined }`, without having to wait for the underlying byte source. This solves the problem, and would make it easier to implement an underlying byte source.

This PR implements this suggestion. To make it work, I also had to change one other thing: when the stream is cancelled, any pending BYOB request is now *immediately invalidated*, so the underlying byte source doesn't erroneously think that it still needs to provide a response. (This aligns with the behavior of `controller.enqueue()`, which throws if the stream is already closed.)

   * …
   * …
   * web-platform-tests/wpt#29128
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1129.html" title="Last updated on Jun 1, 2021, 10:11 PM UTC (1b0a482)">Preview</a> | <a href="https://whatpr.org/streams/1129/033c6d9...1b0a482.html" title="Last updated on Jun 1, 2021, 10:11 PM UTC (1b0a482)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1322,8 +1322,8 @@ the following [=struct/items=]:
 : <dfn export for="read-into request">chunk steps</dfn>
 :: An algorithm taking a [=chunk=], called when a chunk is available for reading
 : <dfn export for="read-into request">close steps</dfn>
-:: An algorithm taking a [=chunk=], called when no chunks are available because the stream is
-   closed
+:: An algorithm taking a [=chunk=] or undefined, called when no chunks are available because
+   the stream is closed
 : <dfn export for="read-into request">error steps</dfn>
 :: An algorithm taking a JavaScript value, called when no [=chunks=] are available because the
    stream is errored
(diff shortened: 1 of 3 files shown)


MattiasBuelens:
[I initially thought](https://github.com/whatwg/streams/pull/1123#issuecomment-841511098) that we'd need to refactor large parts of the specification to make this work, but surprisingly there aren't that many changes needed. So that's good... assuming I didn't miss anything important. 😅

ricea (approved):
lgtm

PR #1130, opened 2021-06-02
Allow other specifications to create readable byte streams

### LINKED ISSUE #1121: Allow other specifications to create readable byte streams
I'd like to solve https://github.com/WICG/serial/issues/127 and https://github.com/whatwg/fetch/issues/267. /cc @reillyeon @yutakahirano. This issue is trying to figure out how to do that.

Note that Web Serial has a proper pull algorithm and so is very similar to https://streams.spec.whatwg.org/#example-rbs-pull . Fetch instead is more like https://streams.spec.whatwg.org/#example-rbs-push, except it has backpressure support; it can send a suspend and resume signal.

---

We'd have an algorithm "Set up a readable byte stream" similar to https://streams.spec.whatwg.org/#readablestream-set-up. It takes a mandatory autoAllocateChunkSize. (Although maybe consumer specs would just set that to an implementation-defined value anyway, so perhaps we should have streams do that as a default instead of making it mandatory?).

Then we have two cases:

- Specs like Web Serial which only want to enqueue from the pull algorithm. They will always get a BYOB request they can fill exactly.

- Specs like Fetch which want to enqueue whenever the network has data. They will only sometimes have a BYOB request available.

I think the basic primitives they would use are:

- "enqueue bytes": takes an Infra byte sequence and wraps it into a Uint8Array and enqueues it, ignoring any BYOB requests. Specs "should" avoid doing this whenever there's a current BYOB request.
- "current BYOB request": returns... just the number of bytes requested? I think the actual identity of the ArrayBuffer, as well as the offset, are not interesting. So maybe this is "current bytes requested". It needs to also be able to return null if there is no outstanding BYOB request, in which case the spec should use "enqueue". It will never be null inside a pullAlgorithm.
- "respond to current BYOB request": takes an Infra byte sequence (of max length = current bytes requested) and puts it into the appropriate offset in the BYOB request's view. Then calls the spec-equivalent of `byobRequest.respond(byteSequenceLength)`.

  The way this is written makes it seem like there are necessarily copies (from the byte sequence into the ArrayBufferView), so we'd have to note that more efficient implementations are possible, where the OS directly writes into the ArrayBufferView's backing memory, and the byte sequences are just for spec convenience.

Also, we'd need to update https://streams.spec.whatwg.org/#readablestream-need-more-data and its following definitions to work on byte controllers.

---

In practice I think this would look like the following:

- [Web Serial](https://wicg.github.io/serial/#readable-attribute):
  - The pullAlgorithm consults the "current bytes requested" instead of the desired size.
  - The pullAlgorithm uses "respond to BYOB request" with the bytes it gets, instead of wrapping into a Uint8Array and enqueueing.
- [Fetch](https://fetch.spec.whatwg.org/#concept-http-network-fetch):
  - Says something vague about attempting to use the "current bytes requested" to ask the network for the appropriate amount??
  - If "current bytes requested" is non-null, then:
    - If the (post-content-coding) bytes fit into current bytes requested, then "respond to BYOB request" with them.
    - Otherwise... just enqueue them? Or split up the bytes so that you respond to the BYOB request with the first N and then enqueue the rest?
  - If "current bytes requested" is null, then:
    - Enqueue the bytes
    - Ask the user agent to suspend the ongoing fetch

---

Any thoughts? How does this sound to spec writers who might use it?

domenic:
Closes #1121.

Users:

- https://github.com/whatwg/fetch/pull/1246
- https://github.com/WICG/serial/pull/137
- https://github.com/w3c/webtransport/pull/297

***
<a href="https://whatpr.org/streams/1130.html" title="Last updated on Jul 14, 2021, 10:35 PM UTC (b6e8404)">Preview</a> | <a href="https://whatpr.org/streams/1130/cada812...b6e8404.html" title="Last updated on Jul 14, 2021, 10:35 PM UTC (b6e8404)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6556,7 +6556,7 @@ mark=] is greater than zero.
  1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
     then perform !
     [$ReadableByteStreamControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
- 1. Perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
+ 1. Otherwise, perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
 </div>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -6576,24 +6576,39 @@ mark=] is greater than zero.
     |e|).
 </div>
 
-<hr>
-
-The following algorithm must only be used on {{ReadableStream}} instances initialized via the above
-[=ReadableStream/set up=] algorithm:
-
 <div algorithm>
  To <dfn export for="ReadableStream">enqueue</dfn> the JavaScript value |chunk| into a
  {{ReadableStream}} |stream|:
 
- 1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
-    {{ReadableStreamDefaultController}}.
- 1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
-    |chunk|).
+ 1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=]
+    {{ReadableStreamDefaultController}},
+  1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
+     |chunk|).
+ 1. Otherwise,
+  1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
+     {{ReadableByteStreamController}}.
+  1. Assert: |chunk| is an {{ArrayBufferView}}.
+  1. Let |byobRequest| be !
+     [$ReadableByteStreamControllerGetBYOBRequest$](|stream|.[=ReadableStream/[[controller]]=]).
+  1. If |byobRequest| is non-null, and |chunk|.\[[ViewedArrayBuffer]].\[[ArrayBufferData]] is the
+     same [=Data Block=] as |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].\[[ArrayBufferData]],
+     then:
+   1. Assert: |chunk|.\[[ByteOffset]] is |byobRequest|.\[[view]].\[[ByteOffset]].
+   1. Assert: |chunk|.\[[ByteLength]] ≤ |byobRequest|.\[[view]].\[[ByteLength]].
+      <p class="note">These asserts ensure that the caller does not write outside the requested
+      range in the [=ReadableStream/current BYOB request view=].
+   1. Perform ?
+      [$ReadableByteStreamControllerRespondWithNewView$](|stream|.[=ReadableStream/[[controller]]=],
+      |chunk|).
+      <p class="note">This works fine even in the case where |chunk|.\[[ViewedArrayBuffer]] equals
+      |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].
+  1. Otherwise, perform ?
+     [$ReadableByteStreamControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |chunk|).
 </div>
 
 <hr>
 
-The following algorithms must only be used on {{ReadableStream}} instances initialized via the above
+The following algorithm must only be used on {{ReadableStream}} instances initialized via the above
 [=ReadableStream/set up with byte reading support=] algorithm:
 
 <div algorithm>
@@ -6609,35 +6624,13 @@ The following algorithms must only be used on {{ReadableStream}} instances initi
  1. Return |byobRequest|.[=ReadableStreamBYOBRequest/[[view]]=].
 </div>
 
-<div algorithm>
- To <dfn export for="ReadableStream">enqueue a view</dfn> into a {{ReadableStream}} |stream| given
- an {{ArrayBufferView}} |view|:
-
- 1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
-    {{ReadableByteStreamController}}.
- 1. Let |byobRequest| be !
-    [$ReadableByteStreamControllerGetBYOBRequest$](|stream|.[=ReadableStream/[[controller]]=]).
- 1. If |byobRequest| is non-null, and |view|.\[[ViewedArrayBuffer]].\[[ArrayBufferData]] is the same
-    [=Data Block=] as |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].\[[ArrayBufferData]], then:
-  1. Assert: |view|.\[[ByteOffset]] is |byobRequest|.\[[view]].\[[ByteOffset]].
-  1. Assert: |view|.\[[ByteLength]] ≤ |byobRequest|.\[[view]].\[[ByteLength]].
-     <p class="note">These asserts ensure that the caller does not write outside the requested
-     range in the [=ReadableStream/current BYOB request view=].
-  1. Perform ?
-     [$ReadableByteStreamControllerRespondWithNewView$](|stream|.[=ReadableStream/[[controller]]=],
-     |view|).
-     <p class="note">This works fine even in the case where |view|.\[[ViewedArrayBuffer]] equals
-     |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].
- 1. Otherwise, perform ?
-    [$ReadableByteStreamControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |view|).
-</div>
-
 Specifications should, when possible, [=ArrayBufferView/write=] into the [=ReadableStream/current
-BYOB request view=] when it is non-null, and then call [=enqueue a view=] with that view. They should
-only [=ArrayBufferView/create=] a new {{ArrayBufferView}} to pass to [=enqueue a view=] when the
-[=ReadableStream/current BYOB request view=] is null, or when they have more bytes on hand than the
-[=ReadableStream/current BYOB request view=]'s [=BufferSource/byte length=]. This avoids unnecessary
-copies and better respects the wishes of the stream's [=consumer=].
+BYOB request view=] when it is non-null, and then call [=ReadableStream/enqueue=] with that view.
+They should only [=ArrayBufferView/create=] a new {{ArrayBufferView}} to pass to
+[=ReadableStream/enqueue=] when the [=ReadableStream/current BYOB request view=] is null, or when
+they have more bytes on hand than the [=ReadableStream/current BYOB request view=]'s
+[=BufferSource/byte length=]. This avoids unnecessary copies and better respects the wishes of the
+stream's [=consumer=].
 
 
 <h4 id="other-specs-rs-reading">Reading</h4>
--- a/index.bs
+++ b/index.bs
@@ -6566,7 +6566,7 @@ mark=] is greater than zero.
  1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
     then perform ! [$ReadableByteStreamControllerError$](|stream|.[=ReadableStream/[[controller]]=],
     |e|).
- 1. Perform ! [$ReadableStreamDefaultControllerError$](|stream|.[=ReadableStream/[[controller]]=],
+ 1. Otherwise, perform ! [$ReadableStreamDefaultControllerError$](|stream|.[=ReadableStream/[[controller]]=],
     |e|).
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -6575,9 +6575,15 @@ mark=] is greater than zero.
 The following algorithm must only be used on {{ReadableStream}} instances initialized via the above
 [=ReadableStream/set up=] algorithm:
 
-<p algorithm>To <dfn export for="ReadableStream">enqueue</dfn> the JavaScript value |chunk| into a
-{{ReadableStream}} |stream|, perform !
-[$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |chunk|).
+<div algorithm>
+ To <dfn export for="ReadableStream">enqueue</dfn> the JavaScript value |chunk| into a
+ {{ReadableStream}} |stream|:
+
+ 1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
+    {{ReadableStreamDefaultController}}.
+ 1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
+    |chunk|).
+</div>
 
 <hr>
 
--- a/index.bs
+++ b/index.bs
@@ -42,6 +42,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: the typed array constructors table; url: #table-49
   text: typed array; url: #sec-typedarray-objects
   text: Number type; url: #sec-ecmascript-language-types-number-type
+  text: Data Block; url: #sec-data-blocks
  type: abstract-op
   text: CloneArrayBuffer; url: #sec-clonearraybuffer
   text: CopyDataBlockBytes; url: #sec-copydatablockbytes
@@ -165,9 +166,9 @@ readable stream tee">branches</dfn>, which can be consumed independently.
 For streams representing bytes, an extended version of the [=readable stream=] is provided to handle
 bytes efficiently, in particular by minimizing copies. The [=underlying source=] for such a readable
 stream is called an <dfn>underlying byte source</dfn>. A readable stream whose underlying source is
-an underlying byte source is sometimes called a <dfn>readable byte stream</dfn>. Consumers of a
-readable byte stream can acquire a [=BYOB reader=] using the stream's {{ReadableStream/getReader()}}
-method.
+an underlying byte source is sometimes called a <dfn export>readable byte stream</dfn>. Consumers of
+a readable byte stream can acquire a [=BYOB reader=] using the stream's
+{{ReadableStream/getReader()}} method.
 
 <h3 id="ws-model">Writable streams</h3>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6203,42 +6211,149 @@ to grow organically as needed.
...
+ for="ReadableStream/set up with byte reading support"><var>highWaterMark</var></dfn> (default 0),
+ perform the following steps. If given, |highWaterMark| must be a non-negative, non-NaN number.
 
-  1. Let |readableStream| be a [=new=] {{ReadableStream}}.
-  1. [=ReadableStream/Set up=] |readableStream| given….
- </div>
+ 1. Let |startAlgorithm| be an algorithm that returns undefined.
+ 1. Let |pullAlgorithmWrapper| be an algorithm that runs these steps:
+  1. If |pullAlgorithm| was given, run it.
+  1. Return [=a promise resolved with=] undefined.
```
ricea, line 6232: TODO: support returning a promise as with https://github.com/whatwg/streams/pull/1131

index.bs
```diff
@@ -6203,42 +6211,149 @@ to grow organically as needed.
...
+ 1. Let |byobRequest| be !
+    [$ReadableByteStreamControllerGetBYOBRequest$](|stream|.[=ReadableStream/[[controller]]=]).
+ 1. If |byobRequest| is null, or
+    |byobRequest|.[=ReadableStreamBYOBRequest/[[view]]=].\[[ByteLength]] &lt; |bytes|'s [=byte
+    sequence/length=], then:
+  1. Let |uint8Array| be a new {{Uint8Array}} wrapping an {{ArrayBuffer}} backed by |bytes|.
+  1. Perform ! [$ReadableByteStreamControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
+     |uint8Array|).
+ 1. Otherwise,
+  1. Write |bytes| into
```
ricea, line 6339: Do we need an assertion that it will fit?

ricea (approved):
lgtm

index.bs
domenic: Once we start using https://github.com/heycam/webidl/pull/987 there will be such an assertion transitively, but it can't hurt to add it here too.

domenic:
Marking as "do not merge yet" pending https://github.com/whatwg/streams/issues/1121 and updating to use https://github.com/heycam/webidl/pull/987.

domenic:
This is ready to merge but we're getting some good feedback in https://github.com/w3c/webtransport/pull/297#discussion_r659855231 so I'll hold off a bit...

index.bs
```diff
@@ -6247,42 +6257,150 @@ to grow organically as needed.
...
+ {{ArrayBufferView}} |view|:
+
+ 1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
+    {{ReadableByteStreamController}}.
+ 1. Let |byobRequest| be !
+    [$ReadableByteStreamControllerGetBYOBRequest$](|stream|.[=ReadableStream/[[controller]]=]).
+ 1. If |byobRequest| is non-null, and |view|.\[[ViewedArrayBuffer]].\[[ArrayBufferData]] is the same
+    [=Data Block=] as |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].\[[ArrayBufferData]], then:
+  1. Assert: |view|.\[[ByteOffset]] is |byobRequest|.\[[view]].\[[ByteOffset]].
+  1. Assert: |view|.\[[ByteLength]] ≤ |byobRequest|.\[[view]].\[[ByteLength]].
```
domenic, line 6384: I made these asserts instead of if tests (compared to https://github.com/w3c/webtransport/pull/297#discussion_r661130619) since it seems like it would be bad if the caller used the view but went past the requested bounds.

index.bs
```diff
@@ -6247,42 +6257,150 @@ to grow organically as needed.
...
+     <p class="note">These asserts ensure that the caller does not write outside the requested
+     range in the [=ReadableStream/current BYOB request view=].
+  1. If |view|.\[[ViewedArrayBuffer]] equals |byobRequest|.\[[view]].\[[ViewedArrayBuffer]], then
+     perform ? [$ReadableByteStreamControllerRespond$](|stream|.[=ReadableStream/[[controller]]=],
+     |view|.\[[ByteLength]]).
+  1. Otherwise, perform ?
+     [$ReadableByteStreamControllerRespondWithNewView$](|stream|.[=ReadableStream/[[controller]]=],
+     |view|).
```
domenic, line 6392: I handled the case where you transfer the backing Data Block here, which goes a bit beyond https://github.com/w3c/webtransport/pull/297#discussion_r661130619 but seems like a good idea.

domenic (commented):
@yutakahirano, can you take a look at the latest commit? I should also update the Fetch and Web Serial PRs to make sure this works well for them too.

index.bs
```diff
@@ -6356,51 +6358,48 @@ The following algorithms must only be used on {{ReadableStream}} instances initi
...
-number of copies involved and will better respect the wishes of stream [=consumers=].
+ 1. If |byobRequest| is non-null, and |view|.\[[ViewedArrayBuffer]].\[[ArrayBufferData]] is the same
+    [=Data Block=] as |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].\[[ArrayBufferData]], then:
+  1. Assert: |view|.\[[ByteOffset]] is |byobRequest|.\[[view]].\[[ByteOffset]].
+  1. Assert: |view|.\[[ByteLength]] ≤ |byobRequest|.\[[view]].\[[ByteLength]].
+     <p class="note">These asserts ensure that the caller does not write outside the requested
+     range in the [=ReadableStream/current BYOB request view=].
+  1. If |view|.\[[ViewedArrayBuffer]] equals |byobRequest|.\[[view]].\[[ViewedArrayBuffer]], then
+     perform ? [$ReadableByteStreamControllerRespond$](|stream|.[=ReadableStream/[[controller]]=],
+     |view|.\[[ByteLength]]).
```
MattiasBuelens, line 6389: Is this needed? Can't we use do `ReadableByteStreamControllerRespondWithNewView` unconditionally? 🤔

index.bs
```diff
@@ -1791,18 +1793,7 @@ has the following [=struct/items=]:
...
-  1. Let |view| be ! [$Construct$]({{%Uint8Array%}}, « |firstDescriptor|'s [=pull-into
-     descriptor/buffer=], |firstDescriptor|'s [=pull-into descriptor/byte offset=] +
-     |firstDescriptor|'s [=pull-into descriptor/bytes filled=], |firstDescriptor|'s [=pull-into
-     descriptor/byte length=] − |firstDescriptor|'s [=pull-into descriptor/bytes filled=] »).
-  1. Let |byobRequest| be a [=new=] {{ReadableStreamBYOBRequest}}.
-  1. Set |byobRequest|.[=ReadableStreamBYOBRequest/[[controller]]=] to [=this=].
-  1. Set |byobRequest|.[=ReadableStreamBYOBRequest/[[view]]=] to |view|.
-  1. Set [=this=].[=ReadableByteStreamController/[[byobRequest]]=] to |byobRequest|.
- 1. Return [=this=].[=ReadableByteStreamController/[[byobRequest]]=].
+ 1. Return ! [$ReadableByteStreamControllerGetBYOBRequest$]([=this=]).
```
MattiasBuelens, line 1796: Don't forget to update the reference implementation as well. Or, to make your life easier, just steal [my commit](https://github.com/whatwg/streams/commit/7a8d7e72549f3cc8e5ac52fc083e580faf2e8410) from #1114. 😛

yutakahirano (approved):
Thank you, LGTM!

index.bs
domenic: Yeah, I'm not sure it's actually needed. It felt weird responding with a new view when we didn't have to though... maybe that's just a naming quirk. I.e. we named it `respondWithNewView` instead of `respondWithPossiblyNewView` for brevity.

domenic: Switched to always use ReadableByteStreamControllerRespondWithNewView

domenic:
Almost done. Still TODO: update Fetch and Web Serial PRs for the new WebTransport-inspired model, to make sure this still works.

domenic:
Seems to work. There's a bit of boilerplate if you, like the Web Serial and Fetch specs, already have a byte sequence on hand. Then you need to branch as to whether there's a BYOB request (and, in the case of Fetch, whether the byte sequence fits). I considered adding a helper for this, which would put us back to "enqueue bytes given a byte sequence", but I think it's OK to be more explicit. In particular it's more obviously optimizable by reading the spec in question, instead of having to go to Streams.

Still, to leave open the door for such a helper in the future, I'll rename the operation from "enqueue bytes" to "enqueue view".

Any final reviews, of either this, https://github.com/WICG/serial/pull/137, or https://github.com/whatwg/fetch/pull/1246 would be appreciated. If there aren't any, I'll merge this tomorrow-ish.

index.bs
```diff
@@ -6481,42 +6484,149 @@ to grow organically as needed.
...
+     |view|).
+     <p class="note">This works fine even in the case where |view|.\[[ViewedArrayBuffer]] equals
+     |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].
+ 1. Otherwise, perform ?
+    [$ReadableByteStreamControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |view|).
+</div>
+
+Specifications should, when possible, [=ArrayBufferView/write=] into the [=ReadableStream/current
+BYOB request view=] when it is non-null, and then call [=enqueue bytes=] with that view. They should
+only [=ArrayBufferView/create=] a new {{ArrayBufferView}} to pass to [=enqueue bytes=] when the
```
MattiasBuelens, line 6625:
```suggestion
BYOB request view=] when it is non-null, and then call [=enqueue a view=] with that view. They should
only [=ArrayBufferView/create=] a new {{ArrayBufferView}} to pass to [=enqueue a view=] when the
```

index.bs
```diff
@@ -6481,42 +6484,149 @@ to grow organically as needed.
...
+</div>
+
+<hr>
+
+The following algorithm must only be used on {{ReadableStream}} instances initialized via the above
+[=ReadableStream/set up=] algorithm:
 
 <p algorithm>To <dfn export for="ReadableStream">enqueue</dfn> the JavaScript value |chunk| into a
 {{ReadableStream}} |stream|, perform !
 [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |chunk|).
```
MattiasBuelens, line 6580: For consistency (and our own sanity), we should make sure we're dealing with a non-byte stream:
```
 1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
    {{ReadableStreamDefaultController}}.
```

MattiasBuelens (changes requested):

index.bs
domenic: Added, although now I'm wondering if we should combine the algorithms into "enqueue", which branches on the controller type. (And asserts that |chunk| is an ArrayBufferView in the byte streams case.) Hmm...

MattiasBuelens: That seems doable.

MattiasBuelens:
Okay, there's still an important problem. After a specification calls "close a readable stream" on a readable byte stream, *the current BYOB request is still pending*. We must somehow perform the equivalent of `controller.byobRequest.respond(0)` (like we do in [our tests](https://github.com/web-platform-tests/wpt/blob/87a4c80598aee5178c385628174f1832f5a28ad6/streams/readable-byte-streams/general.any.js#L1752-L1753)) to return ownership of the BYOB buffer and fulfill any pending `byobReader.read(view)` calls.

I suppose we have to extend "close a readable stream" to handle this? But then:
* Is it safe to assume that, if the other specification calls "close a readable stream", then it is done using the current BYOB request view?
* What if the other specification has transferred the current BYOB request view to a different realm? 😬

domenic:
I think we can mandate that other specifications not transfer the current BYOB request view. Even if implementations do so (and then do the correspondingly-different ending dance), it wouldn't be observable, so we can write the specifications to assume they don't.

Similarly I think we can require that other specifications not write into the current BYOB request view after closing the readable stream.

index.bs
```diff
@@ -6481,42 +6484,155 @@ to grow organically as needed.
...
+<p algorithm>A {{ReadableStream}} <dfn export for="ReadableStream" lt="need more data|needs
+more data">needs more data</dfn> if its [=ReadableStream/desired size to fill up to the high water
+mark=] is greater than zero.
+
+<div algorithm>
+ To <dfn export for="ReadableStream">close</dfn> a {{ReadableStream}} |stream|:
+
+ 1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
+    then perform !
+    [$ReadableByteStreamControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
```
MattiasBuelens, line 6558:
```suggestion
 1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
  1. Perform !
     [$ReadableByteStreamControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
  1. If |stream|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
     is not [=list/is empty|empty=], perform !
     [$ReadableByteStreamControllerRespond$](|stream|.[=ReadableStream/[[controller]]=], 0).
     <p class="note">The caller has to make sure not to write into the
     [=ReadableStream/current BYOB request view=] or transfer it to a different realm before
     closing the stream.
```

MattiasBuelens (changes requested):
That seems reasonable. In that case, we need to do something similar as in the read request close steps of [`ReadableByteStreamTee`](https://streams.spec.whatwg.org/commit-snapshots/cada8129edcc4803b2878a7a3f5e1d8325dc0c23/#abstract-opdef-readablebytestreamtee). (See suggestion below.)

domenic:
Alright, this is ready to merge, but since previous waiting periods have turned up great finds, I'll give it another day before doing so.

MattiasBuelens (approved):

index.bs
```diff
@@ -6481,42 +6484,154 @@ to grow organically as needed.
...
+  1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
+     |chunk|).
+ 1. Otherwise,
+  1. Assert: |stream|.[=ReadableStream/[[controller]]=] [=implements=]
+     {{ReadableByteStreamController}}.
+  1. Assert: |chunk| is an {{ArrayBufferView}}.
+  1. Let |byobRequest| be !
+     [$ReadableByteStreamControllerGetBYOBRequest$](|stream|.[=ReadableStream/[[controller]]=]).
+  1. If |byobRequest| is non-null, and |chunk|.\[[ViewedArrayBuffer]].\[[ArrayBufferData]] is the
+     same [=Data Block=] as |byobRequest|.\[[view]].\[[ViewedArrayBuffer]].\[[ArrayBufferData]],
```
ricea, line 6594: There may be a problem here. If the calling specification needs to transfer `|byobRequest|.[[view]].[[ViewedArrayBuffer]]` for some reason, then it will be detached and not match.

domenic: Ah, but [[ArrayBufferData]] will still match. That's the hidden internal slot that the ECMAScript spec uses to abstract the concept of "memory regions", so it will survive such transfers.

index.bs
```diff
@@ -6481,42 +6484,154 @@ to grow organically as needed.
...
+<div algorithm>
+ To <dfn export for="ReadableStream">close</dfn> a {{ReadableStream}} |stream|:
+
+ 1. If |stream|.[=ReadableStream/[[controller]]=] [=implements=] {{ReadableByteStreamController}},
+  1. Perform !
+     [$ReadableByteStreamControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
+  1. If |stream|.[=ReadableStream/[[controller]]=].[=ReadableByteStreamController/[[pendingPullIntos]]=]
+     is not [=list/is empty|empty=], perform !
+     [$ReadableByteStreamControllerRespond$](|stream|.[=ReadableStream/[[controller]]=], 0).
+     <p class="note">The caller has to make sure not to write into the
```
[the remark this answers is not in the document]
domenic: Yeah, I think this does not match my understanding of https://github.com/whatwg/streams/pull/1130#issuecomment-878611929 . Reviewing that, I think instead we want to say something like:

- The calling specification must never transfer the current BYOB request view. Note: implementations could do something equivalent to a transfer, but in that case they'd need to modify the steps in close to use RespondWithNewView instead of Respond(,0) to keep the same observable behavior as not transfering.
- The calling specification must not write into the current BYOB request view after closing the stream.

@MattiasBuelens, WDYT?

index.bs
ricea: But if it's transferred it will be detached, right? And if it's detached then [[ArrayBufferData]] is set to **null**: https://tc39.es/ecma262/#sec-detacharraybuffer

domenic: Oh, dang, you're right. In that case we should probably solve it by, per the above discussion, prohibiting transferring at a spec level (but allowing it an implementation level as long as the implementation does appropriate behind-the-scenes contortions).

ricea: I think that's fine. If someone finds they need to transfer the buffer we can revisit.

index.bs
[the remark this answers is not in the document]
MattiasBuelens: > The calling specification must never transfer the current BYOB request view. Note: implementations could do something equivalent to a transfer, but in that case they'd need to modify the steps in close to use RespondWithNewView instead of Respond(,0) to keep the same observable behavior as not transfering.

That's correct.

Do we want to provide an algorithm for closing a readable byte stream after the calling specification has already transferred the current BYOB request view? Or maybe add an optional `newByobBuffer` argument and switch on that to use either `Respond(0)` or `RespondWithNewView(emptyViewOnNewByobBuffer)`?

> The calling specification must not write into the current BYOB request view after closing the stream.

Makes sense. If they write into the BYOB request, they should enqueue it first, and *then* close the stream.

index.bs
MattiasBuelens: D'oh, how could I forget? I even [explained this in #1123](https://github.com/whatwg/streams/pull/1123#issuecomment-840861165). 😛 

We *could* make two separate algorithms: one for enqueuing a new chunk, and another for responding with a (transferred) BYOB request view. We still can't check if the given buffer is a transferred version of the BYOB request's buffer, so we'd have to rely on the calling specification to not mess this up.

Or yeah, we could just forbid transferring for now. 😁

ricea (approved):
lgtm

PR #1131, opened 2021-06-02
Be uniformly lenient in what other-spec algorithms can do

domenic:
Previously:

* Setting up a readable stream would ignore the return value (or thrown exception) from pullAlgorithm and sizeAlgorithm, instead always returning an immediately-fulfilled promise.
* Setting up a writable stream would require closeAlgorithm and abortAlgorithm, if given, to return promises.
* Setting up a transform stream would require transformAlgorithm and flushAlgorithm to act synchronously; it would catch any exceptions, but otherwise always return an immediately-fulfilled promise.

Now, all of these algorithms are maximally flexible, to make it easier to write other specs:

* If not given, the wrapper algorithms return immediately-fulfilled promises.
* If they return promises, that promise will be returned, allowing async action. (Web Serial does this currently, even though per the above the Streams spec did not account for it.)
* If they synchronously throw exceptions, those exceptions are converted into rejected promises (and never ignored).
* If they return nothing, i.e. act synchronously, then the wrapper algorithms take care of returning immediately-fulfilled promises.

***
<a href="https://whatpr.org/streams/1131.html" title="Last updated on Jun 4, 2021, 12:41 PM UTC (04b1156)">Preview</a> | <a href="https://whatpr.org/streams/1131/8a7d92b...04b1156.html" title="Last updated on Jun 4, 2021, 12:41 PM UTC (04b1156)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6191,16 +6191,21 @@ to grow organically as needed.
  up"><var>pullAlgorithm</var></dfn>, an optional algorithm <dfn export for="ReadableStream/set
  up"><var>cancelAlgorithm</var></dfn>, an optional number <dfn export for="ReadableStream/set
  up"><var>highWaterMark</var></dfn> (default 1), an optional algorithm <dfn export
- for="ReadableStream/set up"><var>sizeAlgorithm</var></dfn>, perform the following steps. If given,
- |sizeAlgorithm| must be an algorithm accepting [=chunk=] objects and returning a number; and if
- given, |highWaterMark| must be a non-negative, non-NaN number.
+ for="ReadableStream/set up"><var>sizeAlgorithm</var></dfn>, perform the following steps. If
+ given, |pullAlgorithm| and |cancelAlgorithm| may return a promise. If given, |sizeAlgorithm| must
+ be an algorithm accepting [=chunk=] objects and returning a number; and if given, |highWaterMark|
+ must be a non-negative, non-NaN number.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6365,16 +6370,21 @@ for="ReadableStream">locked</dfn> if ! [$IsReadableStreamLocked$](|stream|) retu
...
+ given, |closeAlgorithm| and |abortAlgorithm| may return a promise. If given, |sizeAlgorithm| must
  be an algorithm accepting [=chunk=] objects and returning a number; and if given, |highWaterMark|
  must be a non-negative, non-NaN number.
 
  1. Let |startAlgorithm| be an algorithm that returns undefined.
  1. Let |closeAlgorithmWrapper| be an algorithm that runs these steps:
-  1. If |closeAlgorithm| was given, return the result of running it.
+  1. Let |result| be the result of running |closeAlgorithm|, if |closeAlgorithm| was given, or
+     null otherwise. If this throws an exception |e|, return [=a promise rejected with=] |e|.
+  1. Let |result| be the result of running |closeAlgorithm|, if |closeAlgorithm| was given, or null otherwise.
```
ricea, line 6381: Are we running _closeAlgorithm_ twice?

ricea (commented):
I like it.

index.bs
MattiasBuelens: Yes, looks like a copy-paste error. This second step should be removed.

ricea (approved):
lgtm

PR #1132, opened 2021-06-03
Define WritableStreamDefaultController.signal

### LINKED ISSUE #1015: Make underlying sink writes abortable
We should pass an `AbortSignal` object to underling sink `write()` calls and signal abort if the stream is aborted. `write()` implementations could either ignore it (giving the existing behaviour), or reject the promise they returned, making the abort operation complete more quickly than it would otherwise.

This would also work with `pipeTo()` abort (as long as _preventAbort_ was not set).

yutakahirano:
Define the property to signal abort to UnderlyingSink even
when it has a pending write or close operation.

Fixes #1015.

   * Chrome
   * Firefox
   * https://github.com/web-platform-tests/wpt/pull/29310
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1215992
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1714341
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=226575

***
<a href="https://whatpr.org/streams/1132.html" title="Last updated on Jun 24, 2021, 5:57 PM UTC (dde2430)">Preview</a> | <a href="https://whatpr.org/streams/1132/8a7d92b...dde2430.html" title="Last updated on Jun 24, 2021, 5:57 PM UTC (dde2430)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4105,7 +4105,7 @@ closed. It is only used internally, and is never exposed to web developers.
 <h4 id="ws-default-controller-prototype">Methods and properties</h4>
 
 <dl class="domintro non-normative">
- <dt><code><var ignore>controller</var>.signal</code>
+ <dt><code><var ignore>controller</var>.{{WritableStreamDefaultController/signal}}</code>
  <dd>
   <p>An AbortSignal that can be used to abort the pending write or close operation when the stream is
   [=abort a writable stream|aborted=].
--- a/index.bs
+++ b/index.bs
@@ -4274,6 +4274,10 @@ The following abstract operations operate on {{WritableStream}} instances at a h
     |stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[signal]]=].
  1. Let |state| be |stream|.[=WritableStream/[[state]]=].
  1. If |state| is "`closed`" or "`errored`", return [=a promise resolved with=] undefined.
+
+ <p class="note">We re-check the state because [=signaling abort=] runs author code and that might
+ have changed the state.
+
  1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined, return
     |stream|.[=WritableStream/[[pendingAbortRequest]]=]'s [=pending abort request/promise=].
  1. Assert: |state| is "`writable`" or "`erroring`".
@@ -4441,8 +4445,7 @@ the {{WritableStream}}'s public API.
  1. [=Reject=] |stream|.[=WritableStream/[[inFlightCloseRequest]]=] with |error|.
  1. Set |stream|.[=WritableStream/[[inFlightCloseRequest]]=] to undefined.
  1. Assert: |stream|.[=WritableStream/[[state]]=] is "`writable`" or "`erroring`".
- 1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined and |error| is not an
-    "{{AbortError}}" {{DOMException}},
+ 1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined,
   1. [=Reject=] |stream|.[=WritableStream/[[pendingAbortRequest]]=]'s [=pending abort
      request/promise=] with |error|.
   1. Set |stream|.[=WritableStream/[[pendingAbortRequest]]=] to undefined.
--- a/index.bs
+++ b/index.bs
@@ -4442,7 +4442,7 @@ the {{WritableStream}}'s public API.
  1. Set |stream|.[=WritableStream/[[inFlightCloseRequest]]=] to undefined.
  1. Assert: |stream|.[=WritableStream/[[state]]=] is "`writable`" or "`erroring`".
  1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined and |error| is not an
-    {{AbortError}},
+    "{{AbortError}}" {{DOMException}},
   1. [=Reject=] |stream|.[=WritableStream/[[pendingAbortRequest]]=]'s [=pending abort
      request/promise=] with |error|.
   1. Set |stream|.[=WritableStream/[[pendingAbortRequest]]=] to undefined.
--- a/index.bs
+++ b/index.bs
@@ -4716,6 +4716,7 @@ The following abstract operations support the implementation of the
  1. Set |controller|.[=WritableStreamDefaultController/[[stream]]=] to |stream|.
  1. Set |stream|.[=WritableStream/[[controller]]=] to |controller|.
  1. Perform ! [$ResetQueue$](|controller|).
+ 1. Set |controller|.[=WritableStreamDefaultController/[[abortReason]]=] to undefined.
  1. Set |controller|.[=WritableStreamDefaultController/[[signal]]=] to a new {{AbortSignal}}.
  1. Set |controller|.[=WritableStreamDefaultController/[[started]]=] to false.
  1. Set |controller|.[=WritableStreamDefaultController/[[strategySizeAlgorithm]]=] to
--- a/index.bs
+++ b/index.bs
@@ -4041,6 +4041,7 @@ The Web IDL definition for the {{WritableStreamDefaultController}} class is give
 <xmp class="idl">
 [Exposed=(Window,Worker,Worklet)]
 interface WritableStreamDefaultController {
+  readonly attribute any abortReason;
   readonly attribute AbortSignal signal;
   undefined error(optional any e);
 };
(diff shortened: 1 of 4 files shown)


ricea (commented):
We're going to need some tests.

index.bs
```diff
@@ -4252,6 +4268,10 @@ The following abstract operations operate on {{WritableStream}} instances at a h
  id="writable-stream-abort">WritableStreamAbort(|stream|, |reason|)</dfn> performs the following
  steps:
 
+ 1. If |stream|.[=WritableStream/[[state]]=] is "`closed`" or "`errored`", return
+    [=a promise resolved with=] undefined.
+ 1. [=Signal abort=] on
```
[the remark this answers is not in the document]
domenic: I guess technically it could throw if some spec added a throwing algorithm to the "abort algorithms" list, but that seems unlikely. I filed https://github.com/whatwg/dom/issues/984 on making it impossible.

index.bs
```diff
@@ -4097,9 +4102,13 @@ The <dfn>close sentinel</dfn> is a unique value enqueued into
 [=WritableStreamDefaultController/[[queue]]=], in lieu of a [=chunk=], to signal that the stream is
 closed. It is only used internally, and is never exposed to web developers.
 
-<h4 id="ws-default-controller-prototype">Methods</h4>
+<h4 id="ws-default-controller-prototype">Methods and properties</h4>
 
 <dl class="domintro non-normative">
+ <dt><code><var ignore>controller</var>.signal</code>
```
domenic, line 4108:
```suggestion
 <dt><code><var ignore>controller</var>.{{WritableStreamDefaultController/signal}}</code>
```

index.bs
```diff
@@ -4252,6 +4268,10 @@ The following abstract operations operate on {{WritableStream}} instances at a h
  id="writable-stream-abort">WritableStreamAbort(|stream|, |reason|)</dfn> performs the following
  steps:
 
+ 1. If |stream|.[=WritableStream/[[state]]=] is "`closed`" or "`errored`", return
+    [=a promise resolved with=] undefined.
+ 1. [=Signal abort=] on
+    |stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[signal]]=].
  1. Let |state| be |stream|.[=WritableStream/[[state]]=].
  1. If |state| is "`closed`" or "`errored`", return [=a promise resolved with=] undefined.
```
domenic, line 4276: Maybe add a note explaining that we need to re-check after signaling abort because author code might have closed or errored the stream.

index.bs
```diff
@@ -4421,7 +4441,8 @@ the {{WritableStream}}'s public API.
  1. [=Reject=] |stream|.[=WritableStream/[[inFlightCloseRequest]]=] with |error|.
  1. Set |stream|.[=WritableStream/[[inFlightCloseRequest]]=] to undefined.
  1. Assert: |stream|.[=WritableStream/[[state]]=] is "`writable`" or "`erroring`".
- 1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined,
+ 1. If |stream|.[=WritableStream/[[pendingAbortRequest]]=] is not undefined and |error| is not an
```
domenic, line 4444: This kind of switching on error type is a bit unusual. What is the intention here? What kind of test would pass with this change that would fail without it?

yutakahirano:
```
let reject;
close_promise = new Promise((res, rej) => reject = rej);
  start(c) {
    c.signal.addEventListener('abort', () => {
      const abort_error = new DOMException('foobar', 'AbortError');
      reject(abort_error);
    });
  }, close() {
    return close_promise;
  }
});

writer.close();
const reason = Exception('hoge');
writer.abort(reason);
```

In this example, I think it's natural that `writer.closed` will be rejected by `reason`, but without this clause it'll be rejected with `abort_error`.

Another solution would be to expose `pendingAbortRequest.reason` in the controller so that the underlying sink can reject the result of `close()` with that.

domenic: > In this example, I think it's natural that writer.closed will be rejected by reason, but without this clause it'll be rejected with abort_error.

Hmm, I'm not sure I agree. The close() underlying sink hook explicitly said it failed to close (i.e. it returned a promise rejected with an "AbortError" DOMException, indicating that it tried to close but closing got aborted). I think that failure should be propagated to the `closed` promise...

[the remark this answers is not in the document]
MattiasBuelens: Agreed. We currently don't have any special cases for specific rejection values, so it would feel weird to start doing that. This could also potentially break existing code which *happens* to throw an `AbortError` DOMException in `close()`.

yutakahirano: But in this case, the operation fails because of the abort call coming from the producer, so I think it should be rejected with the abort reason. With the other solution, that code would look like:

```
let reject;
close_promise = new Promise((res, rej) => reject = rej);
  start(c) {
    c.signal.addEventListener('abort', () => {
      reject(c.abortReason);
    });
  }, close() {
    return close_promise;
  }
});

writer.close();
const reason = Exception('hoge');
writer.abort(reason);
```

MattiasBuelens: I think I prefer `reject(c.abortReason)`.

domenic: > But in this case, the operation fails because of the abort call coming from the producer, so I think it should be rejected with the abort reason

Well, but it only "failed" because the underlying sink said it failed. The underlying sink should just not return a rejection from the `close()` method at all, IMO. Then the usual logic [where calling `abort(r)` during close causes `closed` to reject with `r`](https://github.com/web-platform-tests/wpt/blob/master/streams/writable-streams/aborting.any.js#L231-L245) will take over.

MattiasBuelens: Hmm. So you'd do `resolve()` instead? Or something like `return close_promise.catch(() => {})`? 🤔

domenic: I think I need a non-artificial example. Why did you reject the closed promise in the first place? I would think that the signal would have no impact on the promise returned from close, but instead would be used while opening a file or writing to it.

E.g.

```js
let fd;
  start(c) {
    fd = await fs.open(filePath, { signal: c.signal });
  },
  write(chunk, c) {
    await fd.write(chunk, { signal: c.signal });
  }
  close() {
    await fd.close(); // no signal; closing isn't abortable
  }
});
```

If you changed the line to `await fd.close({ signal: c.signal })`, i.e. you let someone abort in the middle of closing---then yes, I'd expect `ws.closed` to reject with an "AbortError" DOMException, because we failed to close the stream due to the closing process being aborted!

yutakahirano: Let me quote @vasilvv's example for WebTransport.

>  Imagine I have a web app where I can do drawing online.  Every 100ms or so, I send the entirety of the drawing on a dedicated WebTransport stream.  Note that if a network connection becomes spotty for a short period of time, this will cause a lot of streams to be open at once, so I want to keep only the three most recent streams open, and reset all streams that are older than the last three.

Here we want to map [the reset operation](https://w3c.github.io/webtransport/#stream-reset) to `abort()`. We are planning to resolve the promise returned by `close()` when the the last chunk is ack-ed, and until then the close operation is always cancellable - by just sending a RESET_STREAM frame.

domenic: OK, so in that case, it seems like you can avoid the problem by not rejecting the promise returned by close when the reset operation is done. Returning a rejected promise from close indicates the closing failed, which it doesn't sound like is accurate in that case.

yutakahirano: Reverted the edit. I'm not sure if we don't need to access `abortReason`.

index.bs
yutakahirano: Done.

index.bs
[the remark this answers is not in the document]
yutakahirano: Reverted the edit.

yutakahirano:
Tests are available at https://github.com/web-platform-tests/wpt/pull/29310.

domenic (approved):

yutakahirano:
Thank you! @domenic, do you have contact of streams implementers for Gecko/WebKit?

domenic:
I suspect you might have better contacts than me at this point, given your work on WebTransport with Gecko and WebKit? But usually I ping @youennf and @jorendorff :)

yutakahirano:
Thank you, I'll talk about this at the next WebTransport WG meeting.

ricea:
Do we need an operation to **add an algorithm to signal** for use by other standards?

yutakahirano:
I was planning to write something like: [=AbortSignal/add=] |algorithm| to |stream|.[[controller]].[[signal]].

Isn't that enough?

ricea:
> Isn't that enough?

It's certainly enough for now. We might want to change it in future so that other standards don't have to make references to WritableStreamDefaultController's internal slots.

yutakahirano:
I realized that we need `abortReason`, as it may contain [the code of RESET_STREAM](https://datatracker.ietf.org/doc/html/rfc9000#section-19.4). Hence I added WritableStreamDefaultController.abortReason in the last commit. @domenic, can you take a look again? I'll update the tests accordingly.

yutakahirano:
@jan-ivar Could you express an implementer interest (Mozilla)? This is needed for https://github.com/w3c/webtransport/issues/259 and https://github.com/w3c/webtransport/issues/260.

domenic:
LGTM

jan-ivar:
Yes we have implementer interest in WebTransport, and therefore this issue.

domenic:
Great!

The reference implementation is failing the tests at https://github.com/web-platform-tests/wpt/pull/29310 indicating a problem either with the reference implementation or the tests. Let's get that straightened out, and then we have to do the usual merge-tests-update-submodule-merge spec dance.

yutakahirano:
Fixed. Could you double-check?

PR #1135, opened 2021-06-04
Fix coverage script for reference implementation

MattiasBuelens:
As noted in [my comment on web-platform-tests/wpt#29190](https://github.com/web-platform-tests/wpt/pull/29190#issuecomment-855036997), it looks like `npm run coverage` has been broken for a while on the reference implementation. 😬

[Previously](https://github.com/whatwg/streams/blob/99ffba7a632c261566c12da60bb35800a42e3b3f/reference-implementation/run-web-platform-tests.js), we would use `require()` to load the reference implementation, which Istanbul can instrument. [Nowadays](https://github.com/whatwg/streams/blob/9beefc2ac041a1a848088962d56eb03e89a8bf07/reference-implementation/run-web-platform-tests.js) however, we first bundle the reference implementation with Browserify and then `eval()` that inside the test's `window`. As a result, Istanbul is unable to instrument our code and cannot report any coverage.

This PR fixes this by instrumenting the bundle with `babel-plugin-istanbul`. (There's also a `browserify-istanbul` plugin, but it uses an old version of Istanbul which doesn't support all the latest JavaScript syntax. The Babel plugin seems to be the recommended approach.) We also need to make sure that coverage results are kept between tests, so we inject the `__coverage__` variable from Node's global scope into each test's `window`.

To do:
* [ ] Only enable Istanbul transform if we're running the tests with coverage reporting enabled? However I don't know how we could detect that... (`global.__coverage__` is initially still `undefined` even with coverage reporting enabled.)

### DIFF
--- a/reference-implementation/package.json
+++ b/reference-implementation/package.json
@@ -12,8 +12,10 @@
   "author": "Domenic Denicola <d@domenic.me> (https://domenic.me/)",
   "license": "(CC0-1.0 OR MIT)",
   "devDependencies": {
+    "@babel/core": "^7.14.3",
+    "babel-plugin-istanbul": "^6.0.0",
+    "babelify": "^10.0.0",
     "browserify": "^16.5.1",
-    "browserify-istanbul": "^3.0.1",
     "debug": "^4.1.1",
     "eslint": "^6.8.0",
     "minimatch": "^3.0.4",
--- a/reference-implementation/package.json
+++ b/reference-implementation/package.json
@@ -13,6 +13,7 @@
   "license": "(CC0-1.0 OR MIT)",
   "devDependencies": {
     "browserify": "^16.5.1",
+    "browserify-istanbul": "^3.0.1",
     "debug": "^4.1.1",
     "eslint": "^6.8.0",
     "minimatch": "^3.0.4",
--- a/reference-implementation/run-web-platform-tests.js
+++ b/reference-implementation/run-web-platform-tests.js
@@ -6,7 +6,7 @@ const path = require('path');
 const fs = require('fs');
 const { promisify } = require('util');
 const browserify = require('browserify');
-const istanbul = require('browserify-istanbul');
+const babelify = require('babelify');
 const wptRunner = require('wpt-runner');
 const minimatch = require('minimatch');
 const readFileAsync = promisify(fs.readFile);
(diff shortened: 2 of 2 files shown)


ricea (commented):
lgtm.

Impressive. I had assumed that it couldn't be fixed.

I will leave it to @domenic to make a decision about whether we want to take on the extra dependencies.

domenic:
Maybe we could use [this command](https://istanbul.js.org/docs/advanced/instrument/) to instrument before bundling, without any additional dependencies?

domenic:
Stepping back a bit: I don't really care about more dependencies for the reference implementation. Everything about the reference implementation is designed to make it easier for us to write the spec and tests, so if this helps, I'm happy with that.

I do worry a bit about the overhead of adding coverage and transpilation to every run of the tests, as that could slow our velocity as we try to iterate locally. Thus my above suggestions for how to maybe help with your TODO. But if you have some before/after measurements showing that these extra steps don't add significant overhead, that'd resolve my concern there.

MattiasBuelens:
On my machine, the `bundle` function takes ~500 ms on `main` and ~3.9 s on this branch. So there is definitely some overhead. We should probably either pass an extra flag to `run-web-platform-tests.js` to enable/disable the instrumentation, or create the instrumented bundle ahead of time.

Out of interest, I looked into whether we could use [c8](https://github.com/bcoe/c8) instead. Because c8 leverages Node's built-in coverage reporting, we wouldn't need to instrument the code ourselves and it *should* be almost as fast as a regular test run. I'm not sure if it'll work though: I was thinking we could make Browserify add a source map to the end of the bundle (by passing `{ debug: true }`), but I can't seem to figure out if/how Node's code coverage will recognize that source map when we run `window.eval(bundledJS)`. 😕

MattiasBuelens:
I managed to get c8 working! 😁

Turns out that Node's `vm.Script` (which jsdom uses for `eval()`) is quite finicky about source maps, so I need to massage it a bit by adding a `// #sourceURL` comment. That URL also has to point to an *actual file*, so I'm now running Browserify separately as part of `pretest` and have it write to `bundle.js`. (I wanted to make it write to `generated/bundle.js`, but then the file names in the coverage report are wrong... 🙄)

If it's too farfetched, I can revert back to Istanbul if you want. But I thought this approach would be nicer, since it doesn't negatively affect test runs without coverage. 🙂

domenic (approved):
I love it!

PR #1138, opened 2021-06-25
Fix reference implementation's ReadableStream{Default,BYOB}ReadResult.webidl

ninevra:
The [spec](https://streams.spec.whatwg.org/#default-reader-class-definition) says this should be `{value: any, done: boolean}`, not `{chunk: any, done: boolean}`.

The code generated by this webidl file is never run, because `ReadableStreamDefaultReadResult`s are only ever used wrapped in `Promise`s, which `webidl2js` currently doesn't unwrap.

### DIFF
--- a/reference-implementation/lib/ReadableStreamBYOBReadResult.webidl
+++ b/reference-implementation/lib/ReadableStreamBYOBReadResult.webidl
@@ -1,4 +1,4 @@
 dictionary ReadableStreamBYOBReadResult {
-  ArrayBufferView chunk;
+  ArrayBufferView value;
   boolean done;
 };
--- a/reference-implementation/lib/ReadableStreamDefaultReadResult.webidl
+++ b/reference-implementation/lib/ReadableStreamDefaultReadResult.webidl
@@ -1,4 +1,4 @@
 dictionary ReadableStreamDefaultReadResult {
-  any chunk;
+  any value;
   boolean done;
 };


MattiasBuelens:
Good catch! Looks like [`ReadableStreamBYOBReadResult.webidl`](https://github.com/whatwg/streams/blob/d0bd912c2c1547aabee61c950ff624e0937d8779/reference-implementation/lib/ReadableStreamBYOBReadResult.webidl) is also wrong. Could you fix that one as well? 🙂

MattiasBuelens (approved):

domenic (approved):

ninevra:
I expected I'd be able to sign the participation agreement in short order, but it's a lot of text (given that it pulls in the entire list of WHATWG policies), and it does contain some things I'm not comfortable with, such as the physical location field, the explicit national security exception in the privacy policy, and the ongoing, dubiously terminable requirement to inform WHATWG of any change in employment. It'd take me at least a few days to read everything carefully and evaluate whether I want to sign, which is kind of silly for such a tiny PR.

Meanwhile if someone else wants to submit their own PR for this issue and close this one they're welcome to do that.

TimothyGu:
@ninevra Regarding the change in employment clause, generally it's enough to make a PR yourself to update your affiliation in the [participant-data](https://github.com/whatwg/participant-data) repo, which I have done myself [numerous times in the past](https://github.com/whatwg/participant-data/pulls?q=is%3Apr+author%3ATimothyGu+is%3Aclosed). Additionally, my belief is that generally updating the participant-data the next time you contribute to one of the WHATWG specs will be enough; there's no need to inform WHATWG if you don't "participate" in the future.

For the "national security exception", FWIW note that most of the data you submit on the participation form is already public on participant-data. So your email and physical location would be the only information subject to that exception.

ninevra:
Participation agreement is signed.

ninevra:
> For the "national security exception", FWIW note that most of the data you submit on the participation form is already public on participant-data. So your email and physical location would be the only information subject to that exception.


I'm not actually afraid that the WHATWG might decide to release my personal data for reasons of national security. I just think it reflects poorly on WHATWG (as stewards in part of a piece of global information infrastructure that necessarily interacts with, is influenced by, and comes into conflict with all world governments) that WHATWG prioritizes the "national security" of one particular nation & treats the actual identity of that nation as obvious/implicit, and affirmatively announces its intention to support (a particular) state & law enforcement interests.

domenic:
Hi @ninevra,

It looks like you signed the participant agreement with the name "ninevra", which I believe is a pseudonym. Per https://github.com/whatwg/sg/issues/93 we need the legal name that you use to sign contracts. Are you able to fix that by sending a pull request to update https://github.com/whatwg/participant-data/blob/main/individuals.json ?

ninevra:
It's not a pseudonym, actually.

domenic:
Ah OK, so that's the name you sign contracts with in the real world?

ninevra:
I mean, no that's not my full name. (Most people don't sign most contracts in their full name afaik, omitting some name-parts is culturally accepted at least in the US including in legal contexts. Policies around which name-parts to require are highly variable and often serve to prioritize names from a particular cultural context while mangling others.)
I do have more names, which I add when required.
But by preference, yes, that is the name I use irl and the name I sign contracts with.

The rest of my names are mostly googleable, though, so I've no particular privacy objection to adding them here, if it's a sticking point? I just wasn't going to default to that unless required. Which name-parts does WHATWG require? I see that most entries in `participant-data` list two names, some list more, some list one, some list one + one or more initials...

domenic:
We need enough per https://github.com/whatwg/sg/issues/93#issuecomment-557643376 for the purposes of

> confirming the real identity of the individual, [without which] we cannot confirm the validity of the contract with WHATWG, and can also not make a determination about whether the participation agreement can be signed as an individual, or needs to be signed by the employer

This is all pretty vague and lawyer-driven, so it's hard to give a clear answer as to what's required. But my job as an editor is to do basic due diligence before verifying. It would be easier for me to do so if you included given name and family name. Otherwise I'd feel a bit uncomfortable verifying straightaway, and would need to consult with the SG to ensure they have no concerns before proceeding.

It sucks that this is required for small contributions. Unfortunately we don't have a process for distinguishing between small contributions (which are unlikely impact patent and copyright issues) and bigger ones (for which all this machinery serves a real purpose). That's roughly https://github.com/whatwg/sg/issues/63 .

ninevra:
PR submitted: https://github.com/whatwg/participant-data/pull/38.

PR #1140, opened 2021-06-29
Editorial: Update usage of CreateBuiltinFunction

TimothyGu:
ECMAScript updated the abstract operation to take length and name as arguments.

***
<a href="https://whatpr.org/streams/1140.html" title="Last updated on Jun 29, 2021, 10:42 PM UTC (5393aa1)">Preview</a> | <a href="https://whatpr.org/streams/1140/d0bd912...5393aa1.html" title="Last updated on Jun 29, 2021, 10:42 PM UTC (5393aa1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -56,8 +56,6 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
   text: IsInteger; url: #sec-isinteger
   text: OrdinaryObjectCreate; url: #sec-ordinaryobjectcreate
   text: SameValue; url: #sec-samevalue
-  text: SetFunctionLength; url: #sec-setfunctionlength
-  text: SetFunctionName; url: #sec-setfunctionname
   text: Type; url: #sec-ecmascript-data-types-and-values
  text: TypeError; url: #sec-native-error-types-used-in-this-standard-typeerror; type: exception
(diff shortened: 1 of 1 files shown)


ricea:
Thanks!

ricea (approved):
lgtm

index.bs
```diff
@@ -5747,9 +5745,8 @@ in the constructor.
 
  1. Let |steps| be the following steps, given |chunk|:
   1. Return ? [$GetV$](|chunk|, "`byteLength`").
- 1. Let |F| be ! [$CreateBuiltinFunction$](|steps|, « », |globalObject|'s [=relevant Realm=]).
- 1. Perform ! [$SetFunctionName$](|F|, "`size`").
- 1. Perform ! [$SetFunctionLength$](|F|, 1).
+ 1. Let |F| be ! [$CreateBuiltinFunction$](|steps|, 1, "`size`", « », |globalObject|'s [=relevant
+    Realm=]).
```
TimothyGu, line 5749: Actually sorry, this is technically a normative change as it changes the enumeration order of `name` vs `length`, but Chrome and Firefox both already conform to this change (`length` comes before `name).

ricea: Thanks for the followup, that's interesting. I think we don't need to go through the normative process for this.

PR #1142, opened 2021-06-30
Reference impl: Use arrow functions for `size` functions

ninevra:
The reference implementation uses `function` expressions to create the `size()` functions for `CountQueuingStrategy` and `ByteLengthQueuingStrategy`, which results in them being constructible and having `prototype` properties. This PR changes to using arrow functions to better conform to the spec; see original discussion below:

> Printing
> 
> ```js
> new ByteLengthQueuingStrategy({ highWaterMark: 100 }).size.prototype
> ```
>
> gives undefined in Chrome and Firefox. The Streams spec calls CreateBuiltinFunction without calling MakeConstructor, which is the abstract op that adds the `prototype` property. So it would appear that we should in fact use arrow functions after all
>
> _Originally posted by @TimothyGu in https://github.com/jsdom/jsdom/pull/3200#discussion_r661008480_

### DIFF
--- a/reference-implementation/lib/ByteLengthQueuingStrategy-impl.js
+++ b/reference-implementation/lib/ByteLengthQueuingStrategy-impl.js
@@ -19,8 +19,7 @@ function initializeSizeFunction(globalObject) {
   }
 
   // We need to set the 'name' property:
-  // eslint-disable-next-line prefer-arrow-callback
-  sizeFunctionWeakMap.set(globalObject, function size(chunk) {
-    return chunk.byteLength;
-  });
+  // The size function must not have a prototype property nor be a constructor
+  const size = chunk => chunk.byteLength;
+  sizeFunctionWeakMap.set(globalObject, size);
 }
--- a/reference-implementation/lib/CountQueuingStrategy-impl.js
+++ b/reference-implementation/lib/CountQueuingStrategy-impl.js
@@ -19,8 +19,7 @@ function initializeSizeFunction(globalObject) {
   }
 
   // We need to set the 'name' property:
-  // eslint-disable-next-line prefer-arrow-callback
-  sizeFunctionWeakMap.set(globalObject, function size() {
-    return 1;
-  });
+  // The size function must not have a prototype property nor be a constructor
+  const size = () => 1;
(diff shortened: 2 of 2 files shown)


ricea:
This doesn't set the name property. I'm not sure if it's sufficient just to assign it after creating the function. @domenic what's the best thing to do here?

MattiasBuelens:
Chrome and Firefox do seem to infer a `name` property from the variable assignment. But I don't think ECMAScript requires that, so it's probably not a good idea to rely on it for the reference implementation.
```javascript
const size = () => 1;
size.name // -> "size"
```

TimothyGu:
ECMAScript does indeed require the `name` property to be set. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#inferred_function_names.

ninevra:
I do find this to very odd and unexpected behavior, but it is required by ecmascript.

Relevant part of the spec, from [14.3.1.2 Runtime Semantics: Evaluation](https://tc39.es/ecma262/#sec-let-and-const-declarations-runtime-semantics-evaluation) of `let` or `const` bindings:

    3. If IsAnonymousFunctionDefinition(Initializer) is true, then
        a. Let value be NamedEvaluation of Initializer with argument bindingId.

capturing `bindingId` and passing it into the evaluation of the anonymous function expression, where it's eventually passed through to [15.3.4 Runtime Semantics: InstantiateArrowFunctionExpression](https://tc39.es/ecma262/#sec-runtime-semantics-instantiatearrowfunctionexpression), and then to [SetFunctionName](https://tc39.es/ecma262/#sec-setfunctionname).

ricea (approved):
Thanks!

ricea:
@ninevra Sorry for the trouble, but could you sign the CLA? https://participate.whatwg.org/agreement-status?user=ninevra&repo=streams&pull=1142

ricea:
Optionally, if you could add a test to https://github.com/web-platform-tests/wpt/blob/master/streams/queuing-strategies.any.js that fails without this change it would be a big help.

ninevra:
@ricea I think I've already signed the CLA; it says I need to be verified?

I'll try to get a test submitted today.

Issue #1143, opened 2021-07-12
ReadableStreamBYOBReader.prototype.readFully(view)


MattiasBuelens:
As suggested in [#1019 (comment)](https://github.com/whatwg/streams/issues/1019#issuecomment-565499158), developers may want to read into a BYOB buffer until it is *completely filled* (instead of only *partially* filling it with however many bytes the underlying source provides). We already have [an example for this](https://streams.spec.whatwg.org/commit-snapshots/cada8129edcc4803b2878a7a3f5e1d8325dc0c23/#example-manual-read-bytes) with a `readInto()` helper function, but it may be worth integrating this into the standard itself for better ergonomics and/or optimization opportunities.

Concretely, I suggest we extend `ReadableStreamBYOBReader`:
```
interface ReadableStreamBYOBReader {
  Promise<ReadableStreamBYOBReadResult> readFully(ArrayBufferView view);
};
```
This method performs a BYOB read into `view`:
* If the underlying source responds with `n` bytes where `n < view.byteLength`, then the stream will immediately start pulling again with the "remainder" of `view` (i.e. `view.subarray(n)`, assuming `view` is a `Uint8Array`).
* If the underlying source responds with `n` bytes where `n === view.byteLength`, then the method fulfills with `{ done: false, value }` where `value` is the completely filled view.
* If the underlying source closes before the view is completely filled, then the method fulfills with `{ done: true, value }` where `value` is the (partially) filled view.
* If the underlying source cancels, then the method returns with `{ done: true, value: undefined }`, like with regular BYOB reads.

Some remarks:
* We can't return a `Promise<ArrayBufferView>`, because then you wouldn't know if the stream is closed. Also, this wouldn't work if the stream is cancelled... unless we're okay with returning `Promise<undefined>` in that case (which I personally don't find very pretty).
* We have to be very careful with how we create a pull-into descriptor for the remainder. There may already be *other* pull-into descriptors into the queue, for example a user could do this:
  ```javascript
  let pullCount = 0;
  const rs = new ReadableStream({
    type: 'bytes',
    pull(c) {
      ++pullCount;
      c.byobRequest.view[0] = pullCount;
      c.byobRequest.respond(1);
    }
  });
  const reader = rs.getReader({ mode: 'byob' });
  const read1 = reader.readFully(new Uint8Array(3));
  const read2 = reader.read(new Uint8Array(1));
  await read1; // should be [1, 2, 3], NOT [1, 3, 4]
[...]

Issue #1019, opened 2019-10-15
ReadableStream.prototype.arrayBuffer()


ricea:
It would be good to have a convenience method to concatenate the contents of a ReadableStream into a (promise for a) Uint8Array.

It is unclear what should happen if the stream contains chunks that are not ArrayBuffer or ArrayBufferViews.

Other similar methods we might consider are `.blob()` and `.text()`.

domenic:
A more generic method might be `stream.toArray()` which returns a promise for an array of chunks. (If https://github.com/tc39/proposal-iterator-helpers lands this would basically become an alias for `stream.values().toArray()`, although potentially easier to optimize since it's more direct.)

If we had that, then we'd just need a way of turning an array of Uint8Arrays into a single Uint8Array. Surprisingly I can't find any easy way to do this, especially not performantly. So, maybe that is not the right path.

annevk:
Probably also want to consider an equivalent API where the buffer is already allocated (similar to `encodeInto()`, BYOB).

MattiasBuelens:
> Probably also want to consider an equivalent API where the buffer is already allocated (similar to `encodeInto()`, BYOB).

I like the idea of a method that fills up an entire `ArrayBuffer` (or `ArrayBufferView`). I think it would work better as a reader method, for example `ReadableBYOBReader.readFully(view)` (taken from Java's [`DataInput.readFully()`](https://docs.oracle.com/javase/8/docs/api/java/io/DataInput.html#readFully-byte:A-)). Optionally, we could also add a shorthand method on `ReadableStream`, similar to how `ReadableStream.cancel()` is a shorthand for `ReadableStreamDefaultReader.cancel()`. The `readInto()` function in [this example](https://streams.spec.whatwg.org/#example-manual-read-bytes) seems like a good starting point.

It would be different from `ReadableStream.arrayBuffer()` and/or `ReadableStream.toArray()` though. Those two methods would always read the entire contents of the stream, whereas a `readFully()` method may leave some bytes on the stream if the given buffer is not large enough to fit the entire contents.

jakearchibald:
How about:

* `array()` - Array of chunks
* `blob(BlobPropertyBag)` - Blob. Rejects and cancels stream if any chunks are not BlobPart.
* `arrayBuffer({ into })` - Rejects and cancels stream if any chunks are not buffer sources, or if `into` is provided and isn't big enough to hold the data. Resolves with an arraybuffer, which will be a slice of `into` if it's provided.

[...]

PR #1145, opened 2021-07-15
Add ReadableStreamBYOBReader.prototype.read(view, { min })

MattiasBuelens:
Implements #1143 and #1175.

In my first attempt, I did most of the heavy lifting in a `ReadableStreamBYOBReaderReadFully()` helper that calls `read(view)`. If the first read didn't fill the entire view yet, it immediately calls `read(view)` again but with a special "addToFront" flag so that the new read-into request goes to the *front* of the queue rather than the back. This works fine, but it's not very pretty. It also means we're possibly calling `read(view)` *re-entrantly* from within `ReadIntoRequest.chunkSteps()`, which is somewhat pushing the boundaries of what you can reasonably expect from the spec. 😬 (I've kept this attempt in the commit history, so you can judge for yourselves.)

I then realized that we already have logic that keeps a read-into request in the queue after a `respond()`. When the user performs a BYOB read with a multi-byte typed array, the stream only fulfills the request when at least `elementSize` have been written into the pull-into descriptor. If not, the request stays pending and the pull-into descriptor remains at the head of the queue. Therefore, in my second attempt, I add a special type of pull-into descriptor which must stay in the queue as long as `bytesFilled < byteLength`, rather than the usual `bytesFilled < elementSize`. This also works very well, but requires changes across multiple parts of the code.

In #1175, we realized that it would be more powerful to allow reading *at least* a given number of bytes, rather than reading an *exact* number of bytes. Thus, in my third attempt, every pull-into descriptor now has a "minimum fill" slot that tracks how many bytes must be filled before this descriptor can fulfill its associated `read(view)` request. As a result, the `bytesFilled < elementSize` and `bytesFilled < byteLength` conditions from the previous solution are now replaced with a single `bytesFilled < minimumFill` condition. This greatly simplifies the logic.

To do:
* [x] Decide upon the final API
  * Current proposal: `byobReader.read(view, { min })`
  * ~~Do we still want `byobReader.fill(view)` too? This would basically be a shorthand for `read()` where `min = view.length`~~
* [x] Write (many more) tests to prove that this actually handles all possible edge cases
* [x] Write the specification text once we've decided on the approach

---

   * Chrome (`@ricea`, [per comment](https://github.com/whatwg/streams/pull/1145#issuecomment-902285322))
   * Safari (`@annevk`, [per comment](https://github.com/whatwg/streams/pull/1145#issuecomment-1582499354))
   * (Non-browser implementers: [Deno](https://github.com/whatwg/streams/pull/1145#issuecomment-1023491792), [Node.js](https://github.com/whatwg/streams/issues/1143#issuecomment-956895141))
   * web-platform-tests/wpt#29723
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1145.html" title="Last updated on Nov 13, 2023, 10:55 AM UTC (b938532)">Preview</a> | <a href="https://whatpr.org/streams/1145/007d729...b938532.html" title="Last updated on Nov 13, 2023, 10:55 AM UTC (b938532)">Diff</a>


reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1236,7 +1251,10 @@ function ReadableByteStreamControllerCommitPullIntoDescriptor(stream, pullIntoDe
 
   let done = false;
   if (stream._state === 'closed') {
-    assert(pullIntoDescriptor.bytesFilled === 0);
+    assert(pullIntoDescriptor.bytesFilled % pullIntoDescriptor.elementSize === 0);
+    if (!pullIntoDescriptor.readFully) {
```
domenic, line 1255: Maybe make this if (readFully) { assert1 } else { assert2 }.

Even though your first assert makes sense in both cases, it feels a bit redundant and less clear.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1453,7 +1477,7 @@ function ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue(contro
   }
 }
 
-function ReadableByteStreamControllerPullInto(controller, view, readIntoRequest) {
+function ReadableByteStreamControllerPullInto(controller, view, readIntoRequest, readFully = false) {
```
domenic, line 1480: I don't think the defaulting is needed? It only has one (now two) call sites.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1550,7 +1575,7 @@ function ReadableByteStreamControllerRespond(controller, bytesWritten) {
 }
 
 function ReadableByteStreamControllerRespondInClosedState(controller, firstDescriptor) {
-  assert(firstDescriptor.bytesFilled === 0);
+  assert(firstDescriptor.bytesFilled % firstDescriptor.elementSize === 0);
```
domenic, line 1578: Should this be a dual assert like the above?

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1570,6 +1595,10 @@ function ReadableByteStreamControllerRespondInReadableState(controller, bytesWri
     return;
   }
 
+  if (pullIntoDescriptor.readFully && pullIntoDescriptor.bytesFilled < pullIntoDescriptor.byteLength) {
+    return;
```
domenic, line 1599: Maybe add a comment here (eventually a spec note) explaining that, IIUC, this is where the magic happens. If we're trying to read fully and haven't filled things up yet, just return, leaving the modified pull-into descriptor in the queue. The next time around we'll pick up right back here, until we're full.

domenic (commented):
This seems way too easy! I expected something more like the manual loop you would have to do if writing this in userspace.

If it works, I definitely like this approach. I imagine tests will let us know if it does...

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: There are two places where the magic happens:
* Here, in `ReadableByteStreamControllerRespondInReadableState`, we make sure not to shift the first pull-into descriptor if a call to `respond()` or `respondWithNewView()` did not fill the entire view.
* [Over in `ReadableByteStreamControllerFillPullIntoDescriptorFromQueue`](https://github.com/whatwg/streams/commit/383c9e543a63938a2c2497a627b0a216d89e7250#diff-db46b8b33e26be52c078fa1540a3516fca7f75af6effb18e9b152efbd114b4bfL1350-R1370), we make sure to indicate that the pull-into descriptor is "not ready" if the total queue size is smaller than the view's length. This is used by the callers to decide if the pull-into descriptor should be shifted out or not. This handles the case where there were already some chunks queued up before `readFully()` was called, or if the underlying source uses `enqueue()` to (partially) fill up the pull-into descriptor.

I'll try to explain all of this in a code comment. 🙂

MattiasBuelens:
I looked at a few other languages to see what their API looks like:
* [Java `java.io.DataInput.readFully`](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/io/DataInput.html#readFully(byte%5B%5D))
  * If end of file is detected before filling up to `b.length`, then this throws an `EOFException`. In this case, it may be that some but not all bytes of `b` have been updated with data from the input stream.
* [Rust `std::io::Read::read_exact`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact)
  * If this function encounters an “end of file” before completely filling the buffer, it returns an error of the kind `ErrorKind::UnexpectedEof`. The contents of `buf` are unspecified in this case.
* [.NET `System.IO.Stream.ReadAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-5.0#System_IO_Stream_ReadAsync_System_Memory_System_Byte__System_Threading_CancellationToken_)
  * Behaves like `ReadableStreamBYOBReader.read(view)`, in that it doesn't necessarily fill the entire buffer. I couldn't find a direct equivalent that does fill up the entire buffer. Even [`BinaryReader.ReadBytes()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binaryreader.readbytes?view=net-5.0#System_IO_BinaryReader_ReadBytes_System_Int32_) may read less than the requested number of bytes.
  * [`BinaryReader.ReadUint32()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binaryreader.readuint32?view=net-5.0#System_IO_BinaryReader_ReadUInt32) throws an `EndOfStreamException` if it detects EOF before reading 4 bytes.

It looks like most other languages throw an error if they encounter EOF before completely filling the buffer. That said, the Streams standard already takes a different approach for "regular" reads too: instead of returning the number of bytes read into the given buffer, we transfer the given buffer and return a `{ done, value }` pair with a view on that buffer. So it seems appropriate to align closer with our existing `read(view)` and use the `done` flag to indicate if we encountered EOF earlier. We can also guarantee that the returned view contains all data up to EOF, rather than leaving the buffer contents unspecified.

If we want to bikeshed a bit on the name, here are some ideas:
* `readFully(view)`
* `readExact(view)`

index.bs
```diff
@@ -1756,6 +1813,8 @@ has the following [=struct/items=]:
 : <dfn for="pull-into descriptor">reader type</dfn>
 :: Either "`default`" or "`byob`", indicating what type of [=readable stream reader=] initiated this
    request
+: <dfn for="pull-into descriptor">read fully</dfn>
+:: A boolean flag indicating whether this is a {{ReadableStreamBYOBReader/readFully()}} request.
```
MattiasBuelens, line 1817: Nit-pick: do we prefer enums over boolean flags? Something like "request type" which can be "default read", "byob read" or "byob read fully"?

Or perhaps we can combine this with "reader type"?

MattiasBuelens (commented):
I suppose we should update (or remove) [the `readInto()` example](https://streams.spec.whatwg.org/commit-snapshots/109f1d4f9fba4eed42238f8de5b9201b8769cd8c/#example-manual-read-bytes) too? Since that can be now replaced with a single `readFully()` call.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1341,13 +1360,22 @@ function ReadableByteStreamControllerFillPullIntoDescriptorFromQueue(controller,
...
-  if (maxAlignedBytes > currentAlignedBytes) {
-    totalBytesToCopyRemaining = maxAlignedBytes - pullIntoDescriptor.bytesFilled;
-    ready = true;
+  if (pullIntoDescriptor.readFully) {
+    // A descriptor for a readFully() request that is not yet completely filled will stay at the head of the queue,
+    // so the underlying source can keep filling it.
+    if (maxBytesFilled >= pullIntoDescriptor.byteLength) {
+      totalBytesToCopyRemaining = pullIntoDescriptor.byteLength - pullIntoDescriptor.bytesFilled;
+      ready = true;
+    }
```
MattiasBuelens, line 1372: At the end of this function, we assert that if `ready === false`, then `controller._queueTotalSize === 0`. In other words, if the pull-into descriptor is not ready yet, then we must have drained the entire queue into the descriptor. With [the previous implementation](https://github.com/whatwg/streams/blob/383c9e543a63938a2c2497a627b0a216d89e7250/reference-implementation/lib/abstract-ops/readable-streams.js#L1366-L1371), we would incorrectly leave some unaligned bytes in the queue if they would not fill up the entire descriptor. With the new implementation, we always copy those bytes.

I added [a new test](https://github.com/web-platform-tests/wpt/blob/984cdd21400ba73419bad9fe66884c5d2436bbf6/streams/readable-byte-streams/read-fully.any.js#L476-L491) for this specific edge case.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1341,13 +1360,22 @@ function ReadableByteStreamControllerFillPullIntoDescriptorFromQueue(controller,
...
+  } else {
+    const maxAlignedBytes = maxBytesFilled - maxBytesFilled % elementSize;
+    if (maxAlignedBytes > currentAlignedBytes) {
```
MattiasBuelens, line 1375: Can't this be simplified to `maxAlignedBytes > 0`? 🤔 If the pull-into descriptor was still in the queue, then it must have `bytesFilled < elementSize`. So `currentAlignedBytes = bytesFilled - bytesFilled % elementSize` must also be `0`. (I feel like I never understood why `currentAlignedBytes` was necessary.)

All tests pass fine when I add an `assert(currentAlignedBytes === 0)` here, so I think this is valid? Unless I'm missing an edge case (that isn't covered by any WPT test)?

MattiasBuelens: I'm pretty sure this is right, so I changed it. I added some asserts for good measure. 😁

MattiasBuelens:
I've been thinking about whether we want to expose `pullIntoDescriptor.[[readFully]]` to the underlying byte source, e.g. something like `ReadableStreamBYOBRequest.isReadFullyRequest`.
* Pro: the underlying byte source might be able to optimize for large reads, instead of receiving repeated `pull()` calls with continuations of the same BYOB request. For example, for `ReadableByteStreamTee`, we could forward a `readFully()` request from the branch all the way to the original stream.
* Con: we might end up with *too much* buffering. Again using the `ReadableByteStreamTee` example: if the first branch does `readFully(new Uint8Array(1024))` and then the second branch does `read(new Uint8Array(1))`, the second branch wouldn't receive any data until the underlying source of the original stream has filled up the entire view of the forwarded `readFully()` request from the first branch. (This case probably deserves its own WPT test... 🤔)

So yeah, it might be an optimization opportunity, but it might also be a footgun. At least for now, I'll leave `ReadableStreamBYOBRequest` untouched. 🙂

index.bs
```diff
@@ -1382,6 +1392,34 @@ value: newViewOnSameMemory, done: true }</code> for closed streams. If the strea
...
 
+ <dt><code>{ <var ignore>value</var>, <var ignore>done</var> } = await <var ignore>reader</var>.{{ReadableStreamBYOBReader/readFully()|readFully}}(<var ignore>view</var>)</code>
+ <dd>
+  <p>Attempts to read bytes into |view| until it is completely full, and returns a promise resolved
+  with the result:
+
+  <ul>
+   <li>If enough bytes become available to fill |view|, the promise will be fulfilled with an
+   object  of the form <code highlight="js">{ value: theView, done: false }</code>. In this case,
+   |view| will be [=ArrayBuffer/detached=] and no longer usable, but <code>theView</code> will be a
```
domenic, line 1403: "view" vs. "theView" is confusing. Above we use "view" vs. "theChunk" which is OK. We could also change both to "view" vs. "newView".

index.bs
```diff
@@ -1382,6 +1392,34 @@ value: newViewOnSameMemory, done: true }</code> for closed streams. If the strea
...
+   new view (of the same type) onto the same backing memory region, with the received bytes written
+   into it.
+
+   <li>If the stream becomes closed before |view| is full, the promise will be fulfilled with an
+   object of the form <code highlight="js">{ value: theView, done: true }</code>. In this case,
+   |view| will be [=ArrayBuffer/detached=] and no longer usable, but <code>theChunk</code> will be a
+   new view (of the same type) onto the same backing memory region, with any received bytes written
+   into it.
```
domenic, line 1411: Maybe mention that its `byteLength` will be equal to the number of bytes read?

index.bs
domenic: Boolean seems fine to me. If we can combine them (i.e., if one element in the 2x2 matrix is not currently filled out) that sounds even nicer.

domenic (approved):
Sorry for the delay. This LGTM with the minor potential improvement of combining the boolean with reader type. @ricea do you think we could implement this? Shouldn't take long...

index.bs
MattiasBuelens: A "default read-fully" request wouldn't make much sense, since there's no BYOB view that needs to be filled. The only reasonable thing this could do is keep reading until the stream becomes closed, but then `readable.arrayBuffer()` (#1019) seems like a better solution.

I'll keep it a boolean for now. 🙂

index.bs
```diff
@@ -1382,6 +1392,34 @@ value: newViewOnSameMemory, done: true }</code> for closed streams. If the strea
...
   [=underlying source=].
 
+ <dt><code>{ <var ignore>value</var>, <var ignore>done</var> } = await <var ignore>reader</var>.{{ReadableStreamBYOBReader/readFully()|readFully}}(<var ignore>view</var>)</code>
+ <dd>
+  <p>Attempts to read bytes into |view| until it is completely full, and returns a promise resolved
+  with the result:
+
+  <ul>
+   <li>If enough bytes become available to fill |view|, the promise will be fulfilled with an
+   object  of the form <code highlight="js">{ value: theView, done: false }</code>. In this case,
```
ricea, line 1402: Nit: There's an extra space between "object" and "of".

ricea (commented):
With regards to the spec, this looks good.

Sorry to show up late with a bikeshed, but I like the name `fill()`. `reader.fill(view)` feels good to me. Disadvantages: no precedent for the name, and it would a nuisance to change all the tests and everything now. It may be to harder to find (but the reader doesn't have so many methods that it would be lost in the noise). I'm just suggesting it to see how others feel.

Regarding implementation in Blink, I think this clearly fixes an ergonomic problem. Some people expect `read()` to work this way already. Implementation is stuck behind #1123 and other things that are higher priority.

index.bs
MattiasBuelens: I'm not sure about "theChunk", since in the case of `readFully()` that view may be filled with *multiple* chunks.

I'll replace both "theChunk" and "theView" with "newView" instead, since that also aligns with `respondWithNewView()`. 🙂

MattiasBuelens:
Sure, `fill(view)` also works. Luckily `readFully()` was a very distinct name, so I could easily find/replace it. 😁

If we still don't like it, it's pretty easy to change/revert that last commit. Bikeshed away!

ricea (approved):
lgtm. I like this very much.

MattiasBuelens:
I take it that means Chrome is interested in implementing this? (Gotta tick those checkboxes! 😁)

ricea:
Yes, you can tick the checkbox for Chrome.

MattiasBuelens:
Updated to use `read(view, { atLeast })` instead of `fill(view)`, as suggested in #1175. This is a more general solution: users can now choose to only fulfill when the view is filled with at least 1 byte, or 2 bytes, or `view.length / 2` bytes, or whatever. With `fill(view)`, it's all or nothing.

We can still introduce `fill(view)` as a shorthand for `read(view, { atLeast: view.length })` if we want to. Happy to discuss. 🙂

Currently, the tests only cover the `atLeast == view.length` case (i.e. the `fill()` use case). I still need to write tests for the more general cases, i.e. where `0 < atLeast <= view.length`.

ricea:
Thanks! I am happy not to have `fill()` since it is not too hard for people to use the more general API.

lucacasonato:
Deno is interested in implementing this.

MattiasBuelens:
A little bikeshedding: how do we feel about the name `"atLeast"` for the new option? I'm thinking `"min"` might be better? (Still avoiding `"minBytes"`, because of multi-byte typed arrays.)
```javascript
const { done, value } = byobReader.read(view, { min: 8 });
// value is minimum 8 elements long
```

ricea:
Yes, `min` is good too. It's a bit more elegant. I don't know whether it's easier to remember or not. `minLength` is another option if we want to be clear what it's the min of, but that's probably not necessary.

MattiasBuelens:
Ready for review. I've added a couple of tests where `min < view.length`, and code coverage is looking good. 🙂

Do we need another implementer's interest in this? Both [Deno](https://github.com/whatwg/streams/pull/1145#issuecomment-1023491792) and [Node.js](https://github.com/whatwg/streams/issues/1143#issuecomment-956895141) have already expressed interest, but I believe our current process requires two *browser* implementers to be onboard.

domenic:
Yeah. Any thoughts from @saschanaz @youennf?

index.bs
```diff
@@ -3317,21 +3365,24 @@ The following abstract operations support the implementation of the
...
     |maxBytesToCopy|.
- 1. Let |maxAlignedBytes| be |maxBytesFilled| − (|maxBytesFilled| mod |elementSize|).
  1. Let |totalBytesToCopyRemaining| be |maxBytesToCopy|.
  1. Let |ready| be false.
- 1. If |maxAlignedBytes| > |currentAlignedBytes|,
+ 1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] < |pullIntoDescriptor|'s
+    [=pull-into descriptor/minimum fill=].
+ 1. Let |maxAlignedBytes| be |maxBytesFilled| − (|maxBytesFilled| mod |pullIntoDescriptor|'s
+    [=pull-into descriptor/element size=]).
+ 1. If |maxAlignedBytes| ≥ [=pull-into descriptor/minimum fill=],
```
crowlKats, line 3379:
```suggestion
 1. If |maxAlignedBytes| ≥ |pullIntoDescriptor|'s
    [=pull-into descriptor/minimum fill=],
```

iuioiua:
Gently pinging those involved in this PR.

annevk:
WebKit supports this addition.

saschanaz (commented):
LGTM, some comments:

index.bs
```diff
@@ -1302,10 +1311,14 @@ The Web IDL definition for the {{ReadableStreamBYOBReader}} class is given as fo
...
   constructor(ReadableStream stream);
 
-  Promise<ReadableStreamReadResult> read(ArrayBufferView view);
+  Promise<ReadableStreamReadResult> read(ArrayBufferView view, optional ReadableStreamBYOBReaderReadOptions options = {});
   undefined releaseLock();
 };
 ReadableStreamBYOBReader includes ReadableStreamGenericReader;
+
+dictionary ReadableStreamBYOBReaderReadOptions {
+  [EnforceRange] unsigned long long min;
```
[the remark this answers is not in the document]
MattiasBuelens: Hmm, good point. 🤔 

We currently have slightly different logic between the `min` and non-`min` cases. It would be much nicer to convert all non-`min` cases to have a default `min`, and then just have `min` everywhere.

I'll look into it.

index.bs
```diff
@@ -3151,7 +3197,8 @@ The following abstract operations support the implementation of the
  1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not empty,
   1. Let |firstPendingPullInto| be
      |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
-  1. If |firstPendingPullInto|'s [=pull-into descriptor/bytes filled=] > 0,
+  1. If |firstPendingPullInto|'s [=pull-into descriptor/bytes filled=]
+     mod |firstPendingPullInto|'s [=pull-into descriptor/element size=] is not 0,
```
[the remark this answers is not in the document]
MattiasBuelens: In the non-min case, `minimum fill` is set to `element size`. As soon as the pull-into descriptor is filled with at least one complete element, that descriptor gets popped off the queue and its read-into request is fulfilled. So it will no longer be in the queue by the time we get to `ReadableByteStreamControllerClose()`.

saschanaz:
TBH rebasing doesn't help as one cannot use "Show changes since your last review" 🥲

annevk:
Click on force-pushed. Unfortunately sometimes you need to rebase.

saschanaz:
>Click on force-pushed.

I'm not sure I follow, you mean there's something I can do?

crowlKats:
What are the blockers for this PR?

MattiasBuelens:
@crowlKats I still want to do some refactoring to make `byobReader.read(view)` desugar to `byobReader.read(view, { min: 1 })`, [see comment](https://github.com/whatwg/streams/pull/1145#discussion_r1224441576). But I didn't find the time for that yet...

I've got some time now, so I'll try to get it done. 🙂

index.bs
```diff
@@ -1413,14 +1434,29 @@ value: newViewOnSameMemory, done: true }</code> for closed streams. If the strea
...
+   1. If |options|["{{ReadableStreamBYOBReaderReadOptions/min}}"] &gt; |view|.\[[ArrayLength]],
+      return [=a promise rejected with=] a {{RangeError}} exception.
+   1. Let |elementSize| be the element size specified in [=the typed array constructors table=] for
+      |view|.\[[TypedArrayName]].
+   1. Set |minimumFill| to |options|["{{ReadableStreamBYOBReaderReadOptions/min}}"] *
+      |elementSize|.
+  1. Otherwise (i.e., it is a {{DataView}}),
+   1. If |options|["{{ReadableStreamBYOBReaderReadOptions/min}}"] &gt; |view|.\[[ByteLength]],
+      return [=a promise rejected with=] a {{RangeError}} exception.
+   1. Set |minimumFill| to |options|["{{ReadableStreamBYOBReaderReadOptions/min}}"].
```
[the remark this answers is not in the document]
MattiasBuelens: I added a default value for `min`.

However, I ended up *not* making the `minimumFill` argument required, because [`ReadableByteStreamTee`](https://streams.spec.whatwg.org/commit-snapshots/14b47076896e155898c1ef698fb08a508837b2d0/#abstract-opdef-readablebytestreamtee) still uses `ReadableStreamBYOBReaderRead` without that argument. Moving the computation out of `ReadableStreamBYOBReaderRead` would make `ReadableByteStreamTee` less readable in my opinion, and it would be harder to use `ReadableStreamBYOBReaderRead` in other parts of the spec in the future.

MattiasBuelens:
@domenic I think this is ready. Let me know when I can start merging things and filing implementation bugs.

index.bs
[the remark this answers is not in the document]
MattiasBuelens: Fixed it. I replaced `minimumFill` with `min` in *both* `ReadableStreamBYOBReaderRead` and `ReadableByteStreamControllerPullInto`, and now we only compute the final `minimumFill` value inside `ReadableByteStreamControllerPullInto`. Much better! 🙂

index.bs
```diff
@@ -3222,7 +3280,8 @@ The following abstract operations support the implementation of the
  1. Assert: |pullIntoDescriptor|.[=pull-into descriptor/reader type=] is not "`none`".
  1. Let |done| be false.
  1. If |stream|.[=ReadableStream/[[state]]=] is "`closed`",
-  1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] is 0.
+  1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=]
+     mod |pullIntoDescriptor|'s [=pull-into descriptor/element size=] is 0.
```
domenic, line 3284: Here and elsewhere, I think we should say "the remainder of dividing |x| by |y| is 0". Modulo and remainder are different for negative numbers, and JS's `%` is remainder.

The result doesn't matter for our case, where all the numbers are positive. But it's also just easier to understand at a glance.

I'm opening an Infra issue on this as well.

domenic (approved):
LGTM if you fix the mod vs. remainder issue; I pushed some minor editorial changes.

MattiasBuelens:
@domenic Apologies for the delay. I've fixed the mod vs. remainder issue, PTAL. 🙂

domenic:
Looks great! I merged https://github.com/web-platform-tests/wpt/pull/29723, so do the test roll thing, merge, and file implementation bugs when you have a chance!

domenic:
Oh, and while merging, please change the title to avoid my pet peeve of using `Static.method()` notation for instance methods :)

jasnell:
Woo! super happy to see this land. Thanks for the work all.

PR #1164, opened 2021-09-04
Remove optional from chunk parameter in QueuingStrategySize

MattiasBuelens:
The `QueuingStrategySize` callback type is currently defined as:
```
callback QueuingStrategySize = unrestricted double (optional any chunk);
```

I don't think the `optional` should be there? The stream implementation must always pass a value for `chunk` (although the chunk itself *could* still be `undefined` if the user did `controller.enqueue(undefined)` or `writer.write(undefined)`).

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1164.html" title="Last updated on Sep 4, 2021, 11:11 PM UTC (05898d3)">Preview</a> | <a href="https://whatpr.org/streams/1164/361897c...05898d3.html" title="Last updated on Sep 4, 2021, 11:11 PM UTC (05898d3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5861,7 +5861,7 @@ dictionary QueuingStrategy {
   QueuingStrategySize size;
 };
 
-callback QueuingStrategySize = unrestricted double (optional any chunk);
+callback QueuingStrategySize = unrestricted double (any chunk);
 </xmp>
 
 <dl>
--- a/reference-implementation/lib/QueuingStrategy.webidl
+++ b/reference-implementation/lib/QueuingStrategy.webidl
@@ -3,4 +3,4 @@ dictionary QueuingStrategy {
   QueuingStrategySize size;
 };
 
-callback QueuingStrategySize = unrestricted double (optional any chunk);
+callback QueuingStrategySize = unrestricted double (any chunk);


ricea (approved):
I guess this is okay if it doesn't break any tests.

domenic:
I believe this would have the observable effect of the following:

```js
const s = new CountQueuingStrategy();
s.size(); // now throws; didn't previously
```

Since it seems bad to treat no argument and `undefined` differently, I think the current spec might be better?

domenic:
Hmm, although maybe it wouldn't have that effect since `QueuingStrategySize` is only used as a return type, so I don't think any arity-checking algorithms get invoked. I.e. the type is just documentation.

Still, in that case I think I'd like to document that `size()` is not different than `size(undefined)`?

MattiasBuelens:
Plus, `CountQueuingStrategy.size` is [currently defined](https://streams.spec.whatwg.org/#cqs-class-definition) to have type `Function` (not `QueuingStrategySize`). So there aren't any checks anyway. 😛

domenic:
Oh, dang, I was just confused. This is for the input. In that case I think removing the optional is correct since it is stored internally and invoked by the spec, always with an argument.

Issue #1103, opened 2021-01-25
Make read requests abortable


domenic:
Noted by @reillyeon in https://github.com/WICG/serial/issues/112#issuecomment-763936183 .

For some cases it would be nice to dequeue (and possibly abort?) a given read request, without canceling the stream entirely. That is,

```js
const controller = new AbortController();

// Assume "rs" contains chunks A and B.
const reader = rs.getReader();

const promise1 = reader.read();
const promise2 = reader.read({ signal: controller.signal });
const promise3 = reader.read();

controller.abort();
```

The intent here is that `promise1` gets fulfilled with chunk A, and `promise3` gets fulfilled with chunk B.

Some questions:

- What does this do to `promise2`? Probably reject it with an `"AbortError"` `DOMException` (like in other `AbortSignal` scenarios).
- What happens if abort is signaled after the read request is already fulfilled? Probably nothing (like in other `AbortSignal` scenarios).
- Does this do more than just dequeue the read request? For example, could it be communicated all the way down to the underlying source, I guess via `pull()`? I don't think this helps much, but it's worth checking...

This is looking relatively straightforward to me to spec, and the use case @reillyeon presents is a good one. Should we just go for it?

reillyeon:
The desired behavior seems analogous to the `preventCancel` and `preventAbort` options to `pipeTo()` and `pipeThrough()`.

IIRC calling `reader.cancel()` will cause `promise2` to resolve with `done` set to `true`. Using the `AbortSignal` should probably cause it to reject with an `"AbortError"` `DOMException` instead since the stream isn't really "done".

If we added an `AbortSignal` parameter to `pull()` then it would make sense to signal it in this case, but only if there were no other reads queued.

MattiasBuelens:
> * What does this do to `promise2`? Probably reject it with an `"AbortError"` `DOMException` (like in other `AbortSignal` scenarios).

Yes, that seems most appropriate. We do the same for `rs.pipeTo(dest, { signal })`.

While we're on the topic: I think `rs.pipeTo(dest, { signal })` should also use the given signal for all of its read requests. That will allow the pipe to shutdown faster.

> * What happens if abort is signaled after the read request is already fulfilled? Probably nothing (like in other `AbortSignal` scenarios).

Yup, that makes sense.

[...]

PR #1168, opened 2021-10-02
Reject pending reads when releasing reader

MattiasBuelens:
Currently, `ReadableStream(Default|BYOB)Reader.releaseLock()` throws an error if there are still pending read requests ([spec](https://streams.spec.whatwg.org/commit-snapshots/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/#default-reader-release-lock)):
> The releaseLock() method steps are:
> 
> 1. If this.\[[stream]] is undefined, return.
> 2. If this.\[[readRequests]] is not empty, throw a TypeError exception.
> 3. Perform ! ReadableStreamReaderGenericRelease(this).

However, in #1103 we realized that it would be more useful if we *allowed* releasing a reader while there are still pending reads, and to *reject* those reads instead. The user can then acquire a new reader to receive those chunks.

This PR implements that proposed change:
* `reader.releaseLock()` no longer throws an error when there are still pending reads. Instead, it immediately rejects all pending reads with a `TypeError`.
* For readable byte streams, we also discard pull-into descriptors corresponding to (now rejected) read-into requests. However, *we must retain the first pull-into descriptor*, since the underlying byte source may still be using it through `controller.byobRequest`. Instead, we mark this descriptor as "detached", such that when we invalidate this BYOB request (as a result of `controller.enqueue()` or `byobRequest.respond()`), the bytes from this descriptor are pushed into the stream's queue and used to fulfill read requests from a future reader.

This also allows expressing `pipeTo()`'s behavior more accurately. Currently, it "cheats" by calling `ReadableStreamReaderGenericRelease` directly without checking if `[[readRequests]]` is empty. (And yes, there's [a test](https://github.com/web-platform-tests/wpt/blob/87a4c80598aee5178c385628174f1832f5a28ad6/streams/piping/error-propagation-backward.any.js#L403-L419) that relies on this happening.) With the proposed changes, we can *safely* release the reader when the pipe finishes (even if there's a pending read), and be sure that any unread chunks can be read by a future reader or pipe.

To do:
* [x] Question: should pending reads be rejected with a `TypeError` or a `AbortError` `DOMException`?
  * Answer: `TypeError`, see [comment](https://github.com/whatwg/streams/pull/1168#issuecomment-1005156894)
* [x] Update spec text
* [ ] Write more tests

Originally, this PR proposed to only add some extra asserts to ensure that `[[readRequests]]` is empty before releasing the lock. However, as shown in the `pipeTo` case (above), this is not sufficient. For context, the original description of that proposal is attached below:
<details>
<summary>Original description</summary>

When using `ReadableStream(Default|BYOB)Reader.releaseLock()`, we throw an error if there are still pending read requests ([spec](https://streams.spec.whatwg.org/commit-snapshots/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/#default-reader-release-lock)):

> The releaseLock() method steps are:
> 
> 1. If this.\[[stream]] is undefined, return.
> 2. If this.\[[readRequests]] is not empty, throw a TypeError exception.
> 3. Perform ! ReadableStreamReaderGenericRelease(this).

However, when we use the abstract op `ReadableStreamReaderGenericRelease` directly, we do not assert this. Some places in the spec often have an extra assert specifically for this (e.g. async iterator's return steps or `ReadableByteStreamTee`), but other places fail to perform this check (e.g. async iterator's next steps or [`ReadableStreamPipeTo`](https://streams.spec.whatwg.org/commit-snapshots/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/#ref-for-readable-stream-reader-generic-release%E2%91%A5)).

I added this missing check, but now the following WPT tests start failing as a result:
* [piping/abort.any.js](https://github.com/web-platform-tests/wpt/blob/87a4c80598aee5178c385628174f1832f5a28ad6/streams/piping/abort.any.js#L375): abort should do nothing after the writable is errored
* [piping/error-propagation-backward.any.js](https://github.com/web-platform-tests/wpt/blob/87a4c80598aee5178c385628174f1832f5a28ad6/streams/piping/error-propagation-backward.any.js#L419): Errors must be propagated backward: becomes errored after piping; preventCancel = true
* [readable-streams/async-iterator.any.js](https://github.com/web-platform-tests/wpt/blob/87a4c80598aee5178c385628174f1832f5a28ad6/streams/readable-streams/async-iterator.any.js#L626): return() should unlock the stream synchronously when preventCancel = true *(also when preventCancel = false)*

The async iterator test fails because we [release the lock in the middle of the read request's close steps](https://streams.spec.whatwg.org/commit-snapshots/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/#ref-for-readable-stream-reader-generic-release), but [`ReadableStreamClose`](https://streams.spec.whatwg.org/commit-snapshots/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/#readable-stream-close) only clears the list of read requests *after* calling all the close steps. The fix is quite simple: empty the list *before* calling the close steps or error steps. I'll push a commit for that on this PR.

The pipe tests fail because the pipe loop actually *starts a read request*, but we don't wait for that read to complete before releasing the lock. I *think* it's possible you could lose a chunk this way, but I haven't yet figured out how. I'll keep trying though. 😛 I think we can solve this one by tracking both `currentWrite` *and* `currentRead`, but I'm not sure about the details yet.
</details>

   * Chromium, as expressed below
   * Other implementations have not yet implemented readable byte streams. As such the editors are treating this as a bugfix to readable byte streams, that falls under their previously-expressed general interest in the feature. (Although, explicit review from other engines would be welcome.)
   * https://github.com/web-platform-tests/wpt/pull/32072
   * Chrome: …
   * Firefox: N/A, falls under general readable byte stream implementation work
   * Safari: N/A, falls under general readable byte stream implementation work

***
<a href="https://whatpr.org/streams/1168.html" title="Last updated on Jan 13, 2022, 8:02 PM UTC (e9e78fe)">Preview</a> | <a href="https://whatpr.org/streams/1168/7f7d68b...e9e78fe.html" title="Last updated on Jan 13, 2022, 8:02 PM UTC (e9e78fe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1231,9 +1231,10 @@ to filling the [=readable stream=]'s [=internal queue=] or changing its state. I
   when the lock is released, the reader will appear errored in the same way from now on; otherwise,
   the reader will appear closed.
 
-  <p>A reader's lock cannot be released while it still has a pending read request, i.e., if a
-  promise returned by the reader's {{ReadableStreamDefaultReader/read()}} method has not yet been
-  settled. Attempting to do so will throw a {{TypeError}} and leave the reader locked to the stream.
+  <p>If the reader's lock is released while it still has pending read requests, then the
+  promises returned by the reader's {{ReadableStreamDefaultReader/read()}} method are immediately
+  rejected with a {{TypeError}}. Any unread chunks remain in the stream's [=internal queue=] and can
+  be read later by acquiring a new reader.
 </dl>
 
 <div algorithm>
@@ -1390,9 +1391,10 @@ value: newViewOnSameMemory, done: true }</code> for closed streams. If the strea
   when the lock is released, the reader will appear errored in the same way from now on; otherwise,
   the reader will appear closed.
 
-  <p>A reader's lock cannot be released while it still has a pending read request, i.e., if a
-  promise returned by the reader's {{ReadableStreamBYOBReader/read()}} method has not yet been
-  settled. Attempting to do so will throw a {{TypeError}} and leave the reader locked to the stream.
+  <p>If the reader's lock is released while it still has pending read requests, then the
+  promises returned by the reader's {{ReadableStreamBYOBReader/read()}} method are immediately
+  rejected with a {{TypeError}}. Any unread chunks remain in the stream's [=internal queue=] and can
+  be read later by acquiring a new reader.
 </dl>
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -2773,7 +2773,8 @@ The following abstract operations support the implementation and manipulation of
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableStreamBYOBReaderErrorReadIntoRequests">ReadableStreamBYOBReaderErrorReadIntoRequests(|reader|, |e|)</dfn>
+ <dfn abstract-op
+ lt="ReadableStreamBYOBReaderErrorReadIntoRequests">ReadableStreamBYOBReaderErrorReadIntoRequests(|reader|, |e|)</dfn>
  performs the following steps:
 
  1. Let |readIntoRequests| be |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=].
@@ -2806,7 +2807,8 @@ The following abstract operations support the implementation and manipulation of
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableStreamDefaultReaderErrorReadRequests">ReadableStreamDefaultReaderErrorReadRequests(|reader|, |e|)</dfn>
+ <dfn abstract-op
+ lt="ReadableStreamDefaultReaderErrorReadRequests">ReadableStreamDefaultReaderErrorReadRequests(|reader|, |e|)</dfn>
  performs the following steps:
 
  1. Let |readRequests| be |reader|.[=ReadableStreamDefaultReader/[[readRequests]]=].
@@ -3250,7 +3252,8 @@ The following abstract operations support the implementation of the
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableByteStreamControllerEnqueueClonedChunkToQueue">ReadableByteStreamControllerEnqueueClonedChunkToQueue(|controller|,
+ <dfn abstract-op
+ lt="ReadableByteStreamControllerEnqueueClonedChunkToQueue">ReadableByteStreamControllerEnqueueClonedChunkToQueue(|controller|,
  |buffer|, |byteOffset|, |byteLength|)</dfn> performs the following steps:
 
  1. Let |cloneResult| be [$CloneArrayBuffer$](|buffer|, |byteOffset|, |byteLength|, {{%ArrayBuffer%}}).
@@ -3262,7 +3265,8 @@ The following abstract operations support the implementation of the
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue">ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue(|controller|,
+ <dfn abstract-op
+ lt="ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue">ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue(|controller|,
  |pullIntoDescriptor|)</dfn> performs the following steps:
 
  1. Assert: |pullIntoDescriptor|'s [=pull-into descriptor/reader type=] is "`none`".
@@ -3449,7 +3453,8 @@ The following abstract operations support the implementation of the
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableByteStreamControllerProcessReadRequestsUsingQueue">ReadableByteStreamControllerProcessReadRequestsUsingQueue(|controller|)</dfn>
+ <dfn abstract-op
+ lt="ReadableByteStreamControllerProcessReadRequestsUsingQueue">ReadableByteStreamControllerProcessReadRequestsUsingQueue(|controller|)</dfn>
  performs the following steps:
 
  1. Let |reader| be |controller|.[=ReadableByteStreamController/[[stream]]=].[=ReadableStream/[[reader]]=].
--- a/index.bs
+++ b/index.bs
@@ -3254,16 +3254,26 @@ The following abstract operations support the implementation of the
 </div>
 
 <div algorithm>
+ <dfn abstract-op lt="ReadableByteStreamControllerEnqueueClonedChunkToQueue">ReadableByteStreamControllerEnqueueClonedChunkToQueue(|controller|,
+ |buffer|, |byteOffset|, |byteLength|)</dfn> performs the following steps:
+
+ 1. Let |cloneResult| be [$CloneArrayBuffer$](|buffer|, |byteOffset|, |byteLength|, {{%ArrayBuffer%}}).
+ 1. If |cloneResult| is an abrupt completion,
+  1. Perform ! [$ReadableByteStreamControllerError$](|cloneResult|.\[[Value]]).
+  1. Return |cloneResult|.
+ 1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|,
+    |cloneResult|.\[[Value]], 0, |byteLength|).
+</div>
+
+<div algorithm>
  <dfn abstract-op lt="ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue">ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue(|controller|,
  |firstDescriptor|)</dfn> performs the following steps:
 
  1. Assert: |firstDescriptor|'s [=pull-into descriptor/reader type=] is "`none`".
- 1. If |firstDescriptor|'s [=pull-into descriptor/bytes filled=] > 0,
-  1. Let |chunk| be ? [$CloneArrayBuffer$](|pullIntoDescriptor|'s [=pull-into
-     descriptor/buffer=], |pullIntoDescriptor|'s [=pull-into descriptor/byte offset=],
-     |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=], {{%ArrayBuffer%}}).
-  1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|, |chunk|, 0,
-     |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=]).
+ 1. If |firstDescriptor|'s [=pull-into descriptor/bytes filled=] > 0, perform ?
+    [$ReadableByteStreamControllerEnqueueClonedChunkToQueue$](|controller|, |pullIntoDescriptor|'s
+    [=pull-into descriptor/buffer=], |pullIntoDescriptor|'s [=pull-into  descriptor/byte offset=],
+    |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=]).
  1. Perform ! [$ReadableByteStreamControllerShiftPendingPullInto$](|controller|).
 </div>
 
@@ -3570,10 +3580,9 @@ The following abstract operations support the implementation of the
  1. If |remainderSize| > 0,
    1. Let |end| be |pullIntoDescriptor|'s [=pull-into descriptor/byte offset=] +
       |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=].
-   1. Let |remainder| be ? [$CloneArrayBuffer$](|pullIntoDescriptor|'s [=pull-into
-      descriptor/buffer=], |end| − |remainderSize|, |remainderSize|, {{%ArrayBuffer%}}).
-   1. Perform ! [$ReadableByteStreamControllerEnqueueChunkToQueue$](|controller|, |remainder|, 0,
-      |remainder|.\[[ByteLength]]).
+   1. Perform ? [$ReadableByteStreamControllerEnqueueClonedChunkToQueue$](|controller|,
+      |pullIntoDescriptor|'s [=pull-into descriptor/buffer=], |end| − |remainderSize|,
+      |remainderSize|).
  1. Set |pullIntoDescriptor|'s [=pull-into descriptor/bytes filled=] to |pullIntoDescriptor|'s
     [=pull-into descriptor/bytes filled=] − |remainderSize|.
  1. Perform !
--- a/index.bs
+++ b/index.bs
@@ -3202,13 +3202,9 @@ The following abstract operations support the implementation of the
      |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
   1. If ! [$IsDetachedBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
      is true, throw a {{TypeError}} exception.
+  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
      [$TransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=]).
- 1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
- 1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
-    [=list/is empty|empty=],
-  1. Let |firstPendingPullInto| be
-     |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
   1. If |firstPendingPullInto|'s [=pull-into descriptor/reader type=] is "`none`",
      perform ? [$ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue$](|controller|,
      |firstPendingPullInto|).
--- a/index.bs
+++ b/index.bs
@@ -1619,6 +1619,14 @@ counterparts for BYOB controllers, as discussed in [[#rs-abstract-ops-used-by-co
   1. Perform ! [$ReadableStreamDefaultControllerCallPullIfNeeded$]([=this=]).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="[[ReleaseSteps]]" for="ReadableStreamDefaultController">\[[ReleaseSteps]]()</dfn>
+ implements the [$ReadableStreamController/[[ReleaseSteps]]$] contract. It performs the following
+ steps:
+
+ 1. Return.
+</div>
+
 <h3 id="rbs-controller-class">The {{ReadableByteStreamController}} class</h3>
 
 The {{ReadableByteStreamController}} class has methods that allow control of a {{ReadableStream}}'s
@@ -1758,7 +1766,8 @@ has the following [=struct/items=]:
    used for constructing a view with which to write into the [=pull-into descriptor/buffer=]
 : <dfn for="pull-into descriptor">reader type</dfn>
 :: Either "`default`" or "`byob`", indicating what type of [=readable stream reader=] initiated this
-   request
+   request, or "`none`" if the initiating [=readable stream reader|reader=] was [=release a read
+   lock|released=]
 
 <h4 id="rbs-controller-prototype">Methods and properties</h4>
 
@@ -1884,6 +1893,18 @@ counterparts for default controllers, as discussed in [[#rs-abstract-ops-used-by
      [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=].
  1. Perform ! [$ReadableStreamAddReadRequest$](|stream|, |readRequest|).
  1. Perform ! [$ReadableByteStreamControllerCallPullIfNeeded$]([=this=]).
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="[[ReleaseSteps]]" for="ReadableByteStreamController">\[[ReleaseSteps]]()</dfn>
+ implements the [$ReadableStreamController/[[ReleaseSteps]]$] contract. It performs the following
+ steps:
+
+ 1. If [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] is not [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. Set |firstPendingPullInto|'s [=pull-into descriptor/reader type=] to "`none`".
+  1. Set [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] to the [=list=]
+     « |firstPendingPullInto| ».
 </div>
 
 <h3 id="rs-byob-request-class">The {{ReadableStreamBYOBRequest}} class</h3>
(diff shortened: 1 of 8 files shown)


MattiasBuelens:
I've fixed `ReadableStreamCancel`, `ReadableStreamClose` and `ReadableStreamError` so they empty the list of read requests *before* calling the close/error steps. That does indeed fix the async iterator tests.

Piping was trickier, because there's interaction with "shutdown with an action". Specifically, when `preventCancel = false`, we don't want to wait for the read to complete *before* we call `ReadableStreamCancel`. I suggest we change it to:
1. Wait until all writes have finished.
2. Perform *action*, and wait for it to complete.
3. Wait until all reads and writes have finished.
4. Finalize, passing along any error from step 2.

This works perfectly, except for [this one test](https://github.com/web-platform-tests/wpt/blob/e1e713c842e54ea0a9410ddc988b63d0e1d31973/streams/piping/error-propagation-backward.any.js#L403-L419):
```javascript
promise_test(t => {

  const pipePromise = promise_rejects_exactly(t, error1, rs.pipeTo(ws, { preventCancel: true }),
                                              'pipeTo must reject with the same error');

  t.step_timeout(() => ws.controller.error(error1), 10);

  return pipePromise.then(() => {
    assert_array_equals(rs.eventsWithoutPulls, []);
    assert_array_equals(ws.events, []);
  });

}, 'Errors must be propagated backward: becomes errored after piping; preventCancel = true');
```
Because `preventCancel = true`, the pipe's read request never finishes, so it gets stuck. I believe this is the correct behavior though: we cannot cancel the stream, so we have to wait for the read to finish before we can safely release our reader.

I see two solutions:
* Change the test to somehow unblock the read request, e.g. by enqueuing a chunk on the readable after erroring the writable.
* Abort the read request without cancelling the stream, as suggested in #1103.

domenic:
Interesting. I think I like that idea. And we could add `getReader({ signal })` as additional sugar for this behavior, similar to how `addEventListener(, { signal })` gives you sugar for later calling `removeEventListener()`.

Does this semantic of rejecting the read() promises work well even for the piping case?

luffy2005 (approved):

MattiasBuelens:
I started working on dealing with pending pull-into descriptors after calling `reader.releaseLock()`, following the ideas from https://github.com/whatwg/streams/issues/1103#issuecomment-932763919.

When calling `releaseLock()`, we go through a new controller-specific `[[ReleaseSteps]]` algorithm. For `ReadableStreamDefaultController`, this does nothing. For `ReadableByteStreamController`, this does two things:
* The first pull-into descriptor is marked as "detached" from its read-into request, by setting its `[[readerType]]` to a new special `"none"` value. We cannot discard this descriptor, since the underlying byte source may have already interacted with it through `controller.byobRequest`.
* Any other pull-into descriptors are immediately discarded. (It's impossible for the underlying byte source to have seen these descriptors in its `controller.byobRequest`.)

When the underlying byte source eventually calls `controller.enqueue()` or `byobRequest.respond()`, we finish cleaning up this detached pull-into descriptor. We have to be very careful though:
* The descriptor may have already been partially filled. If so, these bytes need to *first* be copied into the stream's queue. From there, we can try to fill up one or more non-detached pull-into descriptors.

For now, I've only updated the reference implementation. If we're okay with the approach, I can look into updating the spec text.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -257,13 +258,18 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
     setPromiseIsHandledToTrue(pipeLoop());
 
     function waitForWritesToFinish() {
-      // Another write may have started while we were waiting on this currentWrite, so we have to be sure to wait
```
domenic, line 260: Why this change? The previous version seems a bit more understandable, although both are fairly convoluted...

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -745,10 +751,11 @@ function ReadableStreamCancel(stream, reason) {
 
   const reader = stream._reader;
   if (reader !== undefined && ReadableStreamBYOBReader.isImpl(reader)) {
-    for (const readIntoRequest of reader._readIntoRequests) {
+    const readIntoRequests = reader._readIntoRequests;
```
domenic, line 754: Interesting. Does this change prevent a concrete bug? Are there tests for the bug?

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -945,6 +965,16 @@ function ReadableStreamDefaultReaderRead(reader, readRequest) {
   }
 }
 
+function ReadableStreamDefaultReaderRelease(reader) {
+  const e = new TypeError('Reader was released');
+  const readRequests = reader._readRequests;
+  reader._readRequests = [];
```
domenic, line 971: This dance (lines 970-975) happens enough times it might be worth its own AO.

domenic (commented):
Generally looks good, with just a few questions!

domenic:
> Question: should pending reads be rejected with a TypeError or a AbortError DOMException?

I slightly lean TypeError? We use it for a lot of similar situations in this spec, and now that abort reasons are a thing, we don't use AbortError DOMException directly at all.

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Originally, I was fiddling with `pipeTo()` so that it waits for all of its `read()` and `write()` promises to settle. There was a `waitForReadsAndWritesToFinish()` helper that checked and waited for both, and this pattern made that easier. But that's no longer in the current version, so it's no longer needed.

I'll revert it.

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: This is needed for the default reader such that the async iterator's close steps: https://github.com/whatwg/streams/blob/be00afc855d403d91cc8a6335b80713100ca940e/reference-implementation/lib/ReadableStream-impl.js#L131-L132

do not call the error steps on the remaining read requests:
https://github.com/whatwg/streams/blob/be00afc855d403d91cc8a6335b80713100ca940e/reference-implementation/lib/abstract-ops/readable-streams.js#L970-L975

For consistency, I made the same change for BYOB readers. I'm not aware of any concrete bugs related to BYOB readers that this might solve. You'd need to find a case where the `closeSteps` of a read-into request can *synchronously* release the BYOB reader, and I don't think that ever happens.

reference-implementation/lib/abstract-ops/readable-streams.js
MattiasBuelens: Added `ReadableStreamDefaultReaderErrorReadRequests` and `ReadableStreamBYOBReaderErrorReadIntoRequests` helpers.

MattiasBuelens:
> I slightly lean TypeError? We use it for a lot of similar situations in this spec, and now that abort reasons are a thing, we don't use AbortError DOMException directly at all.

Fair enough.

In #1103 we had an alternative API proposal: `readable.getReader({ signal })`. If we went that route, it *would* have made sense to reject with the signal's abort reason instead. But with the current API proposal, there's indeed no real interaction with an `AbortSignal`.

If desired, users could always propagate their abort reason manually:
```javascript
const abortController = new AbortController();
const { signal } = abortController;
const reader = readable.getReader();
signal.addEventListener("abort", () => {
  // Cannot propagate `signal.reason` to `releaseLock()`
  reader.releaseLock();
});
const read = reader.read().catch((e) => {
  // `e` will be a new `TypeError` instead of `signal.reason` here.
  // A user could propagate it manually though, if necessary:
  if (signal.aborted) {
    throw signal.reason;
  }
  throw e;
}).then(/* ... */);
abortController.abort(new Error("my custom abort reason"));
```

index.bs
```diff
@@ -3156,7 +3207,15 @@ The following abstract operations support the implementation of the
...
      [$TransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=]).
  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
- 1. If ! [$ReadableStreamHasDefaultReader$](|stream|) is true
+ 1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
+    [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be
+     |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. If |firstPendingPullInto|'s [=pull-into descriptor/reader type=] is "`none`",
+   1. Perform ? [$ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue$](|controller|,
+      |firstPendingPullInto|).
```
MattiasBuelens, line 3216: This bit is tricky. `controller.enqueue()` may throw because we failed to allocate a new `ArrayBuffer` to hold any remaining bytes from a now-detached BYOB request:
```javascript
let controller;
  type: 'bytes',
  start(c) { controller = c }
});
const reader1 = readable.getReader({ mode: 'byob' });

reader1.read(new Uint16Array(1));
// controller.byobRequest.view.byteLength === 2

controller.byobRequest.view[0] = 0xaa;
controller.byobRequest.respond(1);
// not enough to fill the BYOB request, so read() remains pending
// controller.byobRequest.view.byteOffset === 1
// controller.byobRequest.view.byteLength === 1

reader1.releaseLock();
// BYOB request becomes detached from request, but we cannot invalidate it yet
// since the underlying byte source might still be using it

controller.enqueue(new Uint8Array([0xbb]));
// At this point, we must move the bytes from the BYOB request ([0xaa]) into the stream's queue,
// and then push the new chunk ([0xbb]) onto the queue.
// However, moving those bytes means allocating a new ArrayBuffer containing that sub-slice,
// which might fail... and then what do we do? 🤷‍♀️
```
What should the stream do in that case?
* We must not drop any data, so the `[0xaa]` byte has to somehow end up in the queue. We do have ownership over the original `ArrayBuffer`, so it would be nice if we could just "shrink" that buffer somehow without reallocating...
* If we can't prevent data loss, should we error the stream instead? It's kinda weird though.

MattiasBuelens: We could also make `ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue` *not* copy the buffer and instead do:
```javascript
ReadableByteStreamControllerEnqueueChunkToQueue(
  controller,
  firstDescriptor.buffer,
  firstDescriptor.byteOffset,
  firstDescriptor.bytesFilled
);
```
But then there's the risk that when a default reader receives this chunk, we could "leak" data before `byteOffset` or after `byteOffset + bytesFilled` which came from a previous BYOB reader. 😕

Or perhaps we could fix that leak by doing a copy inside `ReadableByteStreamControllerFillReadRequestFromQueue` instead? But then again, *that* could throw an OOM error too... 😖

index.bs
```diff
@@ -3196,6 +3255,20 @@ The following abstract operations support the implementation of the
     |controller|.[=ReadableByteStreamController/[[queueTotalSize]]=] + |byteLength|.
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue">ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue(|controller|,
```
MattiasBuelens, line 3259: Better name suggestions are welcome. 😛

index.bs
```diff
@@ -1895,6 +1893,18 @@ counterparts for default controllers, as discussed in [[#rs-abstract-ops-used-by
...
+<div algorithm>
+ <dfn abstract-op lt="[[ReleaseSteps]]" for="ReadableByteStreamController">\[[ReleaseSteps]]()</dfn>
+ implements the [$ReadableStreamController/[[ReleaseSteps]]$] contract. It performs the following
+ steps:
+
+ 1. If [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] is not [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. Set |firstPendingPullInto|'s [=pull-into descriptor/reader type=] to "`none`".
+  1. Set [=this=].[=ReadableByteStreamController/[[pendingPullIntos]]=] to the [=list=]
+     « |firstPendingPullInto| ».
```
MattiasBuelens, line 1905: Alternatively, we could write:
> Remove all pull-into descriptors after *firstPendingPullInto* from this.\[[pendingPullIntos]].

or:
> Remove all pull-into descriptors whose index is not 0 from this.\[[pendingPullIntos]].

but I don't know if that's a correct usage of [remove](https://infra.spec.whatwg.org/#list-remove). 🙃

domenic: "remove" is meant to be fairly flexible in the predicate it takes, so either of those would work. But I like what you have here; it's pretty obvious at a glance what the result will be.

index.bs
domenic: > so it would be nice if we could just "shrink" that buffer somehow without reallocating...

In theory, it feels like transferring or some variant could do this. In practice though, if we're in a situation where ArrayBuffer allocations are failing, I'm not confident we can depend on transferring to avoid such failures itself. I.e., in spec land, transfer might be shrink-in-place, but in implementations, it might need temporary buffers or some allocations or something.

So,

> If we can't prevent data loss, should we error the stream instead?

sounds reasonable to me.

In general I assume these out-of-memory cases are either (1) your computer is in trouble, let's just start erroring as soon as possible to avoid further problems; or (2) you are manipulating huge buffers, and you probably shouldn't be doing that, so let's give you an error so that you can tune your application code to do less huge buffers.

index.bs
```diff
@@ -2776,6 +2786,26 @@ The following abstract operations support the implementation and manipulation of
     |view|, |readIntoRequest|).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableStreamBYOBReaderRelease"
+ id="readable-stream-byob-reader-release">ReadableStreamBYOBReaderRelease(|reader|)</dfn>
```
domenic, line 2791: No hand-crafted id=""s for new abstract-ops

index.bs
```diff
@@ -2776,6 +2786,26 @@ The following abstract operations support the implementation and manipulation of
...
+
+<div algorithm>
+ <dfn abstract-op lt="ReadableStreamBYOBReaderErrorReadIntoRequests">ReadableStreamBYOBReaderErrorReadIntoRequests(|reader|, |e|)</dfn>
+ performs the following steps:
+
+ 1. Let |readIntoRequests| be |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=].
+ 1. Set |reader|.[=ReadableStreamBYOBReader/[[readIntoRequests]]=] to a new empty [=list=].
+ 1. [=list/For each=] |readIntoRequest| of |readIntoRequests|,
+  1. Perform |readIntoRequest|'s [=read-into request/error steps=], given |e|.
+</div>
```
domenic, line 2807: Alphabetize abstract ops within their section

index.bs
```diff
@@ -3156,7 +3207,15 @@ The following abstract operations support the implementation of the
   1. Set |firstPendingPullInto|'s [=pull-into descriptor/buffer=] to !
      [$TransferArrayBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=]).
  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
- 1. If ! [$ReadableStreamHasDefaultReader$](|stream|) is true
+ 1. If |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=] is not
```
domenic, line 3210: Can we merge this with the previous "If controller.[[pendingPullIntos]] is not empty" block? I think it should be OK; the intermediate step should not do anything observable...

index.bs
```diff
@@ -3156,7 +3207,15 @@ The following abstract operations support the implementation of the
...
+    [=list/is empty|empty=],
+  1. Let |firstPendingPullInto| be
+     |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
+  1. If |firstPendingPullInto|'s [=pull-into descriptor/reader type=] is "`none`",
+   1. Perform ? [$ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue$](|controller|,
```
domenic, line 3215: Nit: I like to avoid nesting single-step blocks; this can just be `If X, Y` in one line.

index.bs
domenic: Maybe "Enqueue ... to queue" is not necessary? Although hmm, I see it has a sibling following a similar naming scheme. Oh well.

index.bs
```diff
@@ -3435,6 +3539,8 @@ The following abstract operations support the implementation of the
  |firstDescriptor|)</dfn> performs the following steps:
 
  1. Assert: |firstDescriptor|'s [=pull-into descriptor/bytes filled=] is 0.
+ 1. If |firstDescriptor|'s [=pull-into descriptor/reader type=] is "`none`",
+  1. Perform ! [$ReadableByteStreamControllerShiftPendingPullInto$](|controller|).
```
domenic, line 3543: Same single-nested-step nit here.

domenic (commented):
It'd be great if @nidhijaju could take a glance at this, from a perspective of "does this seem reasonable to implement in Chromium at some point" / "does the motivation and behavior make sense". For the latter https://github.com/web-platform-tests/wpt/pull/32072 might be helpful.

index.bs
MattiasBuelens: Dangit, must have copy-pasted this from somewhere. 😛

index.bs
MattiasBuelens: This hinged on what we wanted to do with the errors from `? ReadableByteStreamControllerEnqueueDetachedPullIntoToQueue()`. That is: if that call failed, is it a problem that we did not yet call `ReadableByteStreamControllerInvalidateBYOBRequest()`?

From https://github.com/whatwg/streams/pull/1168#discussion_r779034631, I think the conclusion is: no. If we're going to error the entire stream in response to an allocation failure in `EnqueueDetachedPullIntoToQueue()`, then it doesn't really matter what happens to the BYOB request. So we can do:
1. Transfer `firstPendingPullInto`'s buffer (and throw an error if it's already detached).
2. If `firstPendingPullInto`'s reader type is `"none"`, move its filled bytes to the stream's queue. If this fails, error the stream (and re-throw the error to the caller of `enqueue()`).
3. Invalidate the BYOB request.

MattiasBuelens: Hmm, looks like we have the same problem with *remainder* in step 6.2 of [`ReadableByteStreamControllerRespondInReadableState`](https://streams.spec.whatwg.org/commit-snapshots/53282dde15be5f844768c2167405660da3c29b2d/#readable-byte-stream-controller-respond-in-readable-state). If that fails, we also leave the stream in an inconsistent state. 🤔

index.bs
```diff
@@ -3153,10 +3202,14 @@ The following abstract operations support the implementation of the
      |controller|.[=ReadableByteStreamController/[[pendingPullIntos]]=][0].
   1. If ! [$IsDetachedBuffer$](|firstPendingPullInto|'s [=pull-into descriptor/buffer=])
      is true, throw a {{TypeError}} exception.
+  1. Perform ! [$ReadableByteStreamControllerInvalidateBYOBRequest$](|controller|).
```
MattiasBuelens, line 3205: Merging the two ifs turned out to be a little tricky. `EnqueueDetachedPullIntoToQueue` eventually calls `ShiftPendingPullInto`, which requires `controller.[[byobRequest]]` to be null. So we must ensure that `InvalidateBYOBRequest` happens *before* `EnqueueDetachedPullIntoToQueue`, but *after* we've done the early-return checks (e.g. throwing a `TypeError` for a detached buffer).

Luckily, it's safe to move the `InvalidateBYOBRequest` call *inside* the if too. If a BYOB request exists, then there must be at least one pull-into descriptor too. 😉

index.bs
```diff
@@ -3196,6 +3249,30 @@ The following abstract operations support the implementation of the
     |controller|.[=ReadableByteStreamController/[[queueTotalSize]]=] + |byteLength|.
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="ReadableByteStreamControllerEnqueueClonedChunkToQueue">ReadableByteStreamControllerEnqueueClonedChunkToQueue(|controller|,
```
domenic, line 3253: Linebreak after `abstract-op`

domenic (approved):
LGTM with final token nit. I can work on reviewing the tests if you think they're ready. But it'd be great to have @ricea and/or @nidhijaju confirm this is a good direction.

Given Safari and Firefox haven't yet implemented readable byte streams I'm comfortable calling this a bugfix that falls under their general already-expressed interest in readable byte streams, as long as we get at least a Chromium sanity check. Of course Safari and Firefox engineer review and thoughts would be highly welcome if they have the time!

nidhijaju (commented):
I'm sorry for the delay in my response! I had a look and I definitely think the motivation and intended behavior makes sense to me. It also looks reasonable to implement this in Chromium from my perspective :)

MattiasBuelens:
I've added a few more tests to improve the test coverage on the reference implementation. The only new lines *not* covered are in `ReadableByteStreamControllerEnqueueClonedChunkToQueue`: https://github.com/whatwg/streams/blob/46b29e02c36abd8f7398c4e290527f5d9326c05d/reference-implementation/lib/abstract-ops/readable-streams.js#L1380-L1383

That's mainly because we don't have a good way to safely and predictably trigger an out-of-memory exception for `new ArrayBuffer()` or `buffer.slice()`. But this is also the case for all the other places where we allocate new `ArrayBuffer`s in the spec, so this is probably fine.

ricea (approved):
lgtm

I think this is a good direction. Although it adds some complexity, the new behaviour is easier to work with.

MattiasBuelens:
> Other implementations have not yet implemented readable byte streams. As such the editors are treating this as a bugfix to readable byte streams, that falls under their previously-expressed general interest in the feature. (Although, explicit review from other engines would be welcome.)

This spec change also affects *default readers*. We no longer throw a `TypeError` when `ReadableStreamDefaultReader.releaseLock()` is called while there are pending read requests, instead we now reject all pending `read()` requests.

So technically speaking, we merged this PR a bit too early. 😅 We should have asked for their interest first. [As discussed on Matrix](https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$bw-O_p88-LdEs46F8tvR3zeeH8hfJE3lvAB8huk1BAk?via=mozilla.org&via=matrix.org&via=igalia.com), I'll mention this when filing spec bugs for *all* browsers, and ask them to let us know if they have any objections.

isiahmeadows:
> So technically speaking, we merged this PR a bit too early. 😅 We should have asked for their interest first. As discussed on Matrix, I'll mention this when filing spec bugs for all browsers, and ask them to let us know if they have any objections.

Looked through the bugs and I think you may have forgotten Chrome?

MattiasBuelens:
I filed [#1287273](https://crbug.com/1287273) with Chrome. It's mentioned in the OP.

isiahmeadows:
Sorry, let me emphasize:

> So technically speaking, we merged this PR a bit too early. 😅 We should have asked for their interest first. As discussed on Matrix, I'll mention this when filing spec bugs for all browsers, **and ask them to let us know if they have any objections.**

Was referring to that last part specifically.

MattiasBuelens:
@nidhijaju and @ricea have already expressed interest on behalf of Chrome in https://github.com/whatwg/streams/pull/1168#pullrequestreview-848528614 and https://github.com/whatwg/streams/pull/1168#pullrequestreview-851912448, so I didn't deem it necessary to ask again in the Chrome issue. Unless I misunderstood something?

PR #1172, opened 2021-10-07
Fix tee() incorrectly closing before enqueuing to the second branch

MattiasBuelens:
I found a case where `tee()` inadvertently drops a chunk for the second branch, even though neither of the branches have been canceled. This is obviously very bad, because both branches *should* receive all enqueued chunks.

See web-platform-tests/wpt#31159 for the problematic test case. Here's how it breaks down for `ReadableStreamDefaultTee`:
1. We call `controller.enqueue('a')`. Since there is a pending read request from `tee()` on the original stream, we synchronously run [its `chunkSteps`](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L368). However, this first calls `queueMicrotask()`, so the remainder will happen later.
2. We call `controller.close()`. Since we've already processed all pending read requests, the stream immediately moves to `state = 'closed'`.
3. One microtick passes. We run the remainder of `chunkSteps`.
    1. We set `reading = false`.
    2. [We call `ReadableStreamDefaultControllerEnqueue` for branch1.](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L391) Since there *is* a pending read request for branch1, it synchronously fulfills the request. And since we created both branches with the default HWM = 1, `ReadableStreamDefaultControllerShouldCallPull` is also true. So we call branch1's `pullAlgorithm`.
        1. We're back [at the start of `pullAlgorithm`](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L361). However, we've already set `reading = false`, we *do* start a new read request on the original stream.
        2. But the original stream is already closed, so [it *synchronously* runs `closeSteps`](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L912-L913).
        3. This brings us [back here](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L401-L406). We immediately close both branches.
     3. But wait a second. We're still in `chunkSteps`. So [we call `ReadableStreamDefaultControllerEnqueue` for branch2](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L395)... except, it's already closed, so [the chunk is ignored](https://github.com/whatwg/streams/blob/8bbfbf83ad06a8b9224bbe1cfbcbbbdbdd827a19/reference-implementation/lib/abstract-ops/readable-streams.js#L1026-L1028)! 😱💥

This PR fixes it by keeping `reading = true` until *after* we've enqueued/responded to both branches. Afterwards, we check if one of the two branches tried to call `pull()` (through an extra `readAgain` flag), and then call `pull()` ourselves. When teeing a readable byte stream, we also have to keep track *which* of the two branches called `pull()`, so we can use the appropriate BYOB request if available.

To do:
* [x] Update spec text to match reference implementation

---

   * …
   * …
   * web-platform-tests/wpt#31159
   * Chrome: N/A (already passes all new tests)
   * Firefox: N/A (already passes all new tests)
   * Safari: N/A (already passes all new tests)

***
<a href="https://whatpr.org/streams/1172.html" title="Last updated on Oct 12, 2021, 11:06 PM UTC (4f0f800)">Preview</a> | <a href="https://whatpr.org/streams/1172/02de71a...4f0f800.html" title="Last updated on Oct 12, 2021, 11:06 PM UTC (4f0f800)">Diff</a>


MattiasBuelens:
I'm getting a ton of link errors when building the spec text locally:
```
$ make local
bikeshed spec index.bs index.html --md-Text-Macro="COMMIT-SHA LOCAL COPY"
LINE ~350: No 'dfn' refs found for 'platform objects' that are marked for export.
LINK ERROR: No 'idl' refs found for 'unrestricted double'.
{{unrestricted double}}
LINK ERROR: No 'interface' refs found for 'object'.
<a data-link-type="interface" data-lt="object">object</a>
LINK ERROR: No 'interface' refs found for 'boolean'.
<a data-link-type="interface" data-lt="boolean">boolean</a>
LINK ERROR: No 'interface' refs found for 'Promise' with spec 'webidl'.
<a data-link-spec="webidl" data-link-type="interface" data-lt="Promise">Promise</a>
LINK ERROR: No 'interface' refs found for 'undefined'.
<a data-link-type="interface" data-lt="undefined">undefined</a>
...
LINE ~6947: No 'dfn' refs found for 'this' that are marked for export.
LINE ~6950: No 'dfn' refs found for 'this' that are marked for export.
LINE ~6955: No 'dfn' refs found for 'new'.
LINK ERROR: No 'dfn' refs found for 'constructor operation' that are marked for export.
[=constructor operation=]
 ✔  Successfully generated, with 439 linking errors
```
The [PR build](https://github.com/whatwg/streams/pull/1172/checks?check_run_id=3843446053) (`make deploy`) seems to run into the same problem. 😕

PR #1174, opened 2021-10-12
Fix assertion failure in RespondInternal after transferring array buffer in RespondWithNewView

nidhijaju:
While implementing https://github.com/whatwg/streams/pull/1123 in Chrome, I realised that calling `RespondInternal(controller, view.[[viewByteLength]]` after calling `TransferArrayBuffer(view.[[ViewedArrayBuffer]])` makes the assertion in RespondInternal(), specifically `Assert: bytesWritten > 0`, fail. This is because `view.[[viewByteLength]]` becomes 0 after transferring the array buffer. 

This can either be solved by not using `TransferArrayBuffer()` (which was the case before) or by storing the `view.[[viewByteLength]]` first before transferring the buffer. In this PR, I've taken the latter approach.

@MattiasBuelens I'm wondering if you have any context on why we changed the standard to use `TransferArrayBuffer()` in `RespondWithNewView()`?

***
<a href="https://whatpr.org/streams/1174.html" title="Last updated on Oct 12, 2021, 9:17 AM UTC (6626c80)">Preview</a> | <a href="https://whatpr.org/streams/1174/1147c01...6626c80.html" title="Last updated on Oct 12, 2021, 9:17 AM UTC (6626c80)">Diff</a>

### DIFF
--- a/reference-implementation/lib/abstract-ops/readable-streams.js
+++ b/reference-implementation/lib/abstract-ops/readable-streams.js
@@ -1634,9 +1634,11 @@ function ReadableByteStreamControllerRespondWithNewView(controller, view) {
     throw new RangeError('The region specified by view is larger than byobRequest');
   }
 
+  const viewByteLength = view.byteLength;
+
   firstDescriptor.buffer = TransferArrayBuffer(view.buffer);
 
-  ReadableByteStreamControllerRespondInternal(controller, view.byteLength);
+  ReadableByteStreamControllerRespondInternal(controller, viewByteLength);
 }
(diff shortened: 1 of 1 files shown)


nidhijaju:
> Could you update [the reference implementation](https://github.com/whatwg/streams/blob/1147c01d0e0ca89dec053ddf93b11d2f7734270e/reference-implementation/lib/abstract-ops/readable-streams.js#L1637-L1639) to reflect the spec text changes?

Done :)

Thank you for the explanation!

Issue #1165, opened 2021-09-14
WritableStreamDefaultController.abortReason


plinss:
@atanassov and I looked took a look at this via https://github.com/w3ctag/design-reviews/issues/672. We're wondering if the `abortReason` would be better served as an extension to `AbortController` and `AbortSignal` rather than being carried an a separate property.

yutakahirano:
@ricea @domenic @MattiasBuelens what do you think? This sounds like a good suggestion - we can make [AbortController.abort()](https://dom.spec.whatwg.org/#dom-abortcontroller-abort) take a reason argument, and add AbortSignal.reason.

MattiasBuelens:
I like it! 🙂

Should the `abort` event fired by an `AbortSignal` also carry an `abortReason` property? That is: instead of "fire an event named `abort`" in step 5 of [signal abort](https://dom.spec.whatwg.org/#abortsignal-signal-abort), we introduce an `AbortEvent` that has an `abortReason` and fire that event instead.

domenic:
I think if this is properly integrated into all calling specifications, such that calling `.abort(x)` causes methods like `fetch()` to reject with `x` instead of with an `"AbortError"` `DOMException`, then it would be good.

It would also solve the issue discussed in https://github.com/whatwg/dom/issues/951 about `"TimeoutError"` `DOMException` being most appropriate. /cc @annevk. We would make `AbortSignal.timeout()` call the spec equivalent of `controller.abort(new DOMException("...", "TimeoutError"))`.

The spec changes would probably be something like:

- Easy changes to the base primitives:
    - Each `AbortSignal` gets an **abort reason** JavaScript value, initially undefined (or null?).
    - `AbortSignal` gets a `reason` getter that returns this's abort reason.
    - `AbortController`'s `abort()` gets an `optional any reason` property. If it is not passed, it creates a new `"AbortError"` `DOMException`. It stores the `reason` in the abort reason for the created signal.
- Changes to all the callers:
  - Fetch:
      - all calls to "Abort fetch" which are triggered by an `AbortSignal` need to use the signal's abort reason, instead of creating a new "AbortError"
      - maybe also the stream erroring step?
  - Streams:
      - ReadableStreamPipeTo needs to consult the signal's abort reason
      - Remove `abortReason` stuff
  - Need to also update per https://dontcallmedom.github.io/webidlpedia/names/AbortSignal.html:
    - Web NFC
    - Web Bluetooth
    - App History
    - EyeDropper
[...]

PR #1177, opened 2021-10-18
Remove abortReason property from WritableStream spec

nidhijaju:
Based on https://github.com/whatwg/streams/issues/1165, this change removes the abortReason property from the WritableStream spec, as we are no longer implementing it, and will rather use the reason from the AbortSignal when that is added in the DOM spec.

@yutakahirano Does this look okay?

***
<a href="https://whatpr.org/streams/1177.html" title="Last updated on Oct 25, 2021, 5:37 PM UTC (96e5a0b)">Preview</a> | <a href="https://whatpr.org/streams/1177/ea03a24...96e5a0b.html" title="Last updated on Oct 25, 2021, 5:37 PM UTC (96e5a0b)">Diff</a>


nidhijaju:
Closing and reopening to trigger the CI again as I believe the wpt PR should have fixed the check failure.

nidhijaju:
I'm running into some test failures related to the `WritableStreamDefaultController interface`, however I think those can only be fixed by updating https://github.com/web-platform-tests/wpt/blob/master/interfaces/streams.idl. However, IIUC that file will not be generated until the spec is updated, and the spec cannot be updated until the tests pass. What should be done in this case?

MattiasBuelens:
You're right. This is quite the chicken or the egg situation! 😅

We could *temporarily* disable the Web IDL tests by adding `idlharness.any.html` here:
https://github.com/whatwg/streams/blob/ea03a24deb75928bfa7c285611d4ac2daf332957/reference-implementation/run-web-platform-tests.js#L35-L39

Then we can land the spec change, wait for [`@webref/idl`](https://github.com/w3c/webref) and WPT to pick up the new Web IDL, and finally roll WPT in this repo and re-enable the test.

@domenic Thoughts?

MattiasBuelens:
This is quite scary though. Any spec change that affects the Web IDL won't get tested against the reference implementation until WPT has updated their Web IDL. So proposed additions like `ReadableStream.from(asyncIterable)` or `ReadableStreamBYOBReader.fill(view)` can't be fully tested before we land them. 😕

nidhijaju:
Thank you @domenic! I'll make the relevant WPT PR/changes on Monday. Please let me know by then if you think we should wait more.

PR #1181, opened 2021-10-22
Replace IRC link with Matrix

bakkot:
I gather the IRC room is dead.


domenic:
Sorry for missing this. The participation check doesn't pass... are you still working for Shape Security? If so you need to publicize your membership in the shapesecurity GitHub org. If not you can sign up as an individual.

bakkot:
Done.

domenic (approved):
Thanks!

PR #1182, opened 2021-10-27
Use abort reason in ReadableStreamPipeTo

nidhijaju:
Following on from https://github.com/whatwg/dom/pull/1027, this PR uses the AbortSignal's abort reason in the [ReadableStreamPipeTo](https://streams.spec.whatwg.org/#readable-stream-pipe-to) function. It also updates [WritableStreamAbort](https://streams.spec.whatwg.org/#writable-stream-abort) to use the `reason` to signal abort.

/cc @annevk, @yutakahirano, @domenic 

- [x] At least two implementers are interested (and none opposed):
   * Google
   * Mozilla
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/31400
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1267135
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1739564
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=232747

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/streams/1182.html" title="Last updated on Nov 17, 2021, 4:54 AM UTC (8bd9267)">Preview</a> | <a href="https://whatpr.org/streams/1182/4b6b93c...8bd9267.html" title="Last updated on Nov 17, 2021, 4:54 AM UTC (8bd9267)">Diff</a>

PR #1189, opened 2021-11-24
Roll WPT for abort reason

MattiasBuelens:
It looks like we forgot to roll WPT for the new tests in #1182. We rolled it to https://github.com/web-platform-tests/wpt/commit/4905296d5e2381e2feaadeb7b92a77abe9ebf0f6, but that's *behind* https://github.com/web-platform-tests/wpt/commit/b724cac4269298cfa57064bd5751e7e6e7593727 where we actually landed the new tests. 😬

**Update:** there were some bugs in the new tests, so I opened https://github.com/web-platform-tests/wpt/pull/31738 to fix them.


domenic:
Oh no, the `_pageShowingFlag` jsdom error again... what did we do last time??

MattiasBuelens:
Looks like we also have some failing tests. Fun! 😛 

```
  × (reason: 'null') all the error objects should be the same object
  
    promise_rejects_dom: pipeTo rejects with AbortError function "function() { throw e }" threw null, not an object
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
        at async Test.<anonymous> (http://127.0.0.1:39171/streams/piping/abort.any.js:67:7)

  × (reason: 'null') abort should prevent further reads
  
    promise_rejects_dom: pipeTo rejects with AbortError function "function() { throw e }" threw null, not an object
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
        at async Test.<anonymous> (http://127.0.0.1:39171/streams/piping/abort.any.js:137:7)

  × (reason: 'null') all pending writes should complete on abort
  
    promise_rejects_dom: pipeTo rejects with AbortError function "function() { throw e }" threw null, not an object
        at async Test.<anonymous> (http://127.0.0.1:39171/streams/piping/abort.any.js:176:7)
```
It seems like the tests expect `controller.abort(null)` to behave like `controller.abort(undefined)`, meaning that the `AbortSignal` becomes aborted with an `AbortError`. However, [the spec](https://dom.spec.whatwg.org/commit-snapshots/4ca187f9ccd4b1548d87293307ee58430ef4908b/#abortsignal-signal-abort) says:
> 2. Set _signal_’s abort reason to _reason_ if it is given; otherwise to a new `"AbortError"` `DOMException`.

AFAIK "if it is given" means "if it is not `undefined`", so `null` should be passed through unmodified and `signal.reason` becomes `null`. At least, that's how it's implemented [in jsdom](https://github.com/jsdom/jsdom/blob/a3037212707dfdcfe3d21ec89aea33d4cc674a5f/lib/jsdom/living/aborting/AbortSignal-impl.js#L26) and [in Deno](https://github.com/denoland/deno/blob/ccd730a8b77c188657ec15b5d096f62a24bd0be6/ext/web/03_abort_signal.js#L48). [Node.js](https://github.com/nodejs/node/blob/7633c86eebbf6e381e60e63dc6e3eb36a3cce276/lib/internal/abort_controller.js#L108) doesn't seem to create an `AbortError` yet, instead it always passes `reason` unmodified.

domenic:
Yep, I think those are test bugs.

MattiasBuelens:
> Yep, I think those are test bugs.

Should be easy enough to fix. On it. 👷‍♂️ 

> Oh no, the `_pageShowingFlag` jsdom error again... what did we do last time??

It looks like those happen *as a result* of some of the tests failing (and thus not completing properly). Making the tests pass also fixes those errors.

MattiasBuelens:
Opened https://github.com/web-platform-tests/wpt/pull/31738.

MattiasBuelens:
Hmm, tests are now passing, but the `_pageShowingFlag` thingy still shows up in the logs... 😕

I believe what's happening is that wpt-runner [calls `window.close()`](https://github.com/domenic/wpt-runner/blob/v4.0.0/lib/wpt-runner.js#L222) *before* the `load` event fires on that `window`. That means that jsdom has already [deleted `window._document`](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L773) by the time it tries to [handle the `load` event](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L914).

The easiest solution is probably to add an `if (!window._document) { return }` statement right before [this line](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L912). I *think* that aligns with step 9.2 of [13.2.7 The end](https://html.spec.whatwg.org/multipage/parsing.html#the-end)?
> 2. If the `Document` object's browsing context is null, then abort these steps.

MattiasBuelens:
Never mind, it's *worse*. 😛

1. jsdom fires a `load` event on `window`. ([source](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L912))
2. testharness.js has a `load` event listener, which calls `tests.complete()`. ([source](https://github.com/web-platform-tests/wpt/blob/71c864f8bd7f855f12c81bfc7a1fc3ebf18657f3/resources/testharness.js#L3359-L3363))
3. testharness.js eventually calls `tests.notify_complete()`, which runs all done callbacks. ([source](https://github.com/web-platform-tests/wpt/blob/71c864f8bd7f855f12c81bfc7a1fc3ebf18657f3/resources/testharness.js#L3672))
4. wpt-runner has a done callback, which calls `window.close()`. ([source](https://github.com/domenic/wpt-runner/blob/v4.0.0/lib/wpt-runner.js#L203))
5. jsdom's `Window.close()` deletes the window's document. ([source](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L773))
6. Finally we unwind the stack and come back to jsdom's `load` event listener. We now try to retrieve the wrapper of `window._document`, but it's already `undefined`. ([source](https://github.com/jsdom/jsdom/blob/18.1.1/lib/jsdom/browser/Window.js#L914))

In other words, while dispatching the `load` event, the window is closed *synchronously*. jsdom doesn't expect this, and I'm not entirely sure if the HTML spec even covers this? 😕

domenic (approved):

domenic:
Great diagnosis! Let's move that over to the jsdom repo and consider the fix there, while we merge this PR :)

PR #1190, opened 2021-11-25
Add WritableStreamDefaultController.releaseBackpressure()

### LINKED ISSUE #1158: Piping to writable streams with HWM 0?
Right now, piping to a writable stream with `{ highWaterMark: 0 }` stalls indefinitely:
```javascript
  start(c) {
    c.enqueue("a");
    c.enqueue("b");
    c.enqueue("c");
    c.close();
  }
});
  write(chunk) {
    console.log("wrote:", chunk);
  }
}, { highWaterMark: 0 });
rs.pipeTo(ws); // never resolves, and no messages are logged
```
This makes it impossible to pipe through a `TransformStream` without increasing the total queue size of a pipe chain by at least one chunk:
```javascript
const upperCaseTransform = new TransformStream({
    controller.enqueue(chunk.toUpperCase());
  }
}, { highWaterMark: 0 }, { highWaterMark: 0 });
rs.pipeThrough(upperCaseTransform).pipeTo(ws); // stalls indefinitely

const upperCaseTransform = new TransformStream({
    controller.enqueue(chunk.toUpperCase());
  }
}, { highWaterMark: 1 }, { highWaterMark: 0 }); // same as default strategies
rs.pipeThrough(upperCaseTransform); // works, but already pulls the first chunk from `rs`
```
This is unfortunate, since there are many use cases for synchronous `TransformStream`s that shouldn't need buffering (i.e. every call to `transform()` immediately results in at least one `enqueue()`):
* A generic `mapTransform(fn)`, similar to `array.map(fn)`:
  ```javascript
  function mapTransform(fn) {
    return new TransformStream({
        controller.enqueue(fn(chunk));
      }
    });
  }
  rs.pipeThrough(mapTransform(x => x.toUpperCase()));
  ```
* `TextEncoderStream` and `TextDecoderStream` from [Encoding](https://encoding.spec.whatwg.org/).

Prior discussions on this topic noted that this is not possible. `writer.desiredSize` is always `<= 0`, so `writer.ready` is always pending:
* #777:
  > We can't reduce the HWM of the writableStrategy to 0 because it would have permanent backpressure preventing the pipe from working.
* [#1083](https://github.com/whatwg/streams/pull/1083#discussion_r562292762):
  > Yes. As you observed, a writable stream with a HWM of 0 will always have backpressure. So adding an identity TransformStream to a pipe can't be a complete no-op: it always increases the total queue size by 1.

But that got me thinking. A `ReadableStream`'s source can be `pull()`ed as a result of `reader.read()`, even if `controller.desiredSize <= 0`. Maybe a `WritableStream`'s sink should then also be able to release backpressure even if `writer.desiredSize <= 0`? 🤔

We could add a method on `WritableStreamDefaultController` (`controller.pull()`? `controller.releaseBackpressure()`? `controller.notifyReady()`?) that would have the result of immediately resolving the current `writer.ready` promise. Internally, we would do something like `WritableStreamUpdateBackpressure(stream, false)`. My hope is that we can then use this inside `TransformStreamSetBackpressure()`, so that pulling from the readable end of a transform stream would also resolve `ready` on the writable end.

...Or am I missing something very obvious? 😛

MattiasBuelens:
This adds a new `releaseBackpressure()` method *(name to be decided)* to `WritableStreamDefaultController`, allowing a `WritableStream` with HWM = 0 to resolve the pending `writer.ready` promise manually.

This also integrates with `TransformStream`s: when `pull()` is called on the readable side, the transform stream calls `releaseBackpressure()` on its writable side. This allows piping through `TransformStream`s with HWM = 0 on *both* their readable and writable side, enabling "unbuffered" transformations:
```javascript
  start(c) {
    c.enqueue("a");
    c.enqueue("b");
    c.enqueue("c");
    c.close();
  }
});

const upperCaseTransform = new TransformStream({
    controller.enqueue(chunk.toUpperCase());
  }
}, { highWaterMark: 0 }, { highWaterMark: 0 });

  write(chunk) {
    console.log(chunk);
  }
});

await rs.pipeThrough(upperCaseTransform).pipeTo(ws);
```
Previously, the above snippet would stall.

Fixes #1158.

To do:
* [ ] Write spec text
* [ ] Write more tests
* [ ] Bikeshed on the method name

---

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …


reference-implementation/lib/abstract-ops/writable-streams.js
```diff
@@ -657,6 +659,9 @@ function WritableStreamDefaultControllerErrorIfNeeded(controller, error) {
 }
 
 function WritableStreamDefaultControllerGetBackpressure(controller) {
+  if (controller._releaseBackpressure === true) {
```
MattiasBuelens, line 662: We might have to add `&& controller._queue.length === 0` here. Not sure yet. 🤔

domenic:
Looks reasonable so far. Should we make unbuffered transforms the default? I suspect it's more desirable, and hopefully wouldn't be *too* breaking...

MattiasBuelens:
> Should we make unbuffered transforms the default? I suspect it's more desirable, and hopefully wouldn't be _too_ breaking...

I'd like that *a lot*.
* Identity transform streams would no unnecessarily longer increase the total buffer size of a pipe chain.
* In a pipe chain, we would only start pulling from the source as soon as we've connected our chain to either a buffered transform stream or to the final sink. (Right now, we start pulling from the original source stream as soon as we pipe it to the first transform stream, which might not yet be connected to a sink.)

I tried out the change locally, and it looks like we only need to change a couple of tests around `writer.desiredSize`. So I *think* we can get away with it? 🤞

reference-implementation/lib/abstract-ops/writable-streams.js
MattiasBuelens: This check is indeed necessary. It's covered by this new test: https://github.com/web-platform-tests/wpt/pull/31742/commits/f69d4b733da5cba396b95b7379bffb4f352aacb1

ricea:
> Looks reasonable so far. Should we make unbuffered transforms the default? I suspect it's more desirable, and hopefully wouldn't be too breaking...

I'm concerned that people may be relying on the existing behaviour. With buffering, all the transforms in a pipe can run in parallel, but without it they will run in series. This could cause mysterious performance regressions.

MattiasBuelens:
Hmm, good point. We don't know if or when `transform()` will call `enqueue()`. It might do it synchronously (e.g. for an identity transform or for `TextDecoderStream`), in which case buffering doesn't help and writable HWM = 0 would make more sense. But we cannot assume that. 🤔

If we don't change the default, we should document how authors can construct an unbuffered transform stream, and when they should use this. We should also go through other specifications that use `TransformStream` and check if they should have writable HWM 0 or 1.

MattiasBuelens:
I've updated [set up a `TransformStream`](https://whatpr.org/streams/1190.html#transformstream-set-up) to take optional arguments for the writable/readable HWMs and size algorithms, such that we can use it in [create an identity `TransformStream`](https://whatpr.org/streams/1190.html#create-an-identity-transformstream).

As I said before, we'll want to update other specifications to also set the writable HWM to 0 for synchronous unbuffered transform streams, such as [`CompressionStream`](https://wicg.github.io/compression/#compressionstream) and [`TextDecoderSteam`](https://encoding.spec.whatwg.org/#textdecoderstream).

ricea:
I haven't had a chance to look at the implementation in detail. I think Chrome is interested in this functionality in general. I will leave it to @domenic to make the call on when we are happy with the spec change.

In the spirit of bikeshedding, I generally like the name `releaseBackpressure()` although I'd make it one word if I could. One problem I see with the name is that people might interpret it to mean that it will release backpressure even if `desiredSize` is negative.

Alternative names I've thought of
* `acceptChunk()` - has the same problem
* `ready()` - ambiguous
* `markReady()` - also might be misinterpreted
* `pull()` - confusing?

domenic:
So basically if you're doing a sync transform, you should use HWM = 0, but an async transform should use HWM = 1 (or higher)? That's rough for web developers. I guess the conservative default is to stick with 1 and tell people to customize for sync transforms...

Maybe we could make it better in the future with something like `TransformStream.fromSyncMap(x => y)`, but I dunno, then you might want to have flush logic...

> Alternative names I've thought of

I like `markReady()` since the `ready` promise is the existing developer-facing manifestation of backpressure. I agree it's hard to communicate the conditional nature of the API, without a long name.

index.bs
```diff
@@ -1150,8 +1150,8 @@ interface ReadableStreamDefaultReader {
 ReadableStreamDefaultReader includes ReadableStreamGenericReader;
 
 dictionary ReadableStreamDefaultReadResult {
- any value;
- boolean done;
+  any value;
```
domenic, line 1153: Might be worth pulling the Web IDL and editorial changes out into a separate PR that we can just approve and merge ASAP.

index.bs
```diff
@@ -5147,6 +5179,20 @@ The following abstract operations support the implementation of the
   1. Perform ! [$WritableStreamFinishInFlightWriteWithError$](|stream|, |reason|).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="WritableStreamDefaultControllerReleaseBackpressure"
+ id="writable-stream-default-controller-release-backpressure">WritableStreamDefaultControllerReleaseBackpressure(|controller|)</dfn>
```
domenic, line 5184: No need for custom IDs for new abstract-ops

index.bs
```diff
@@ -5147,6 +5179,20 @@ The following abstract operations support the implementation of the
...
+ performs the following steps:
+
+ 1. Let |stream| be |controller|.[=WritableStreamDefaultController/[[stream]]=].
+ 1. Assert: |stream|.[=WritableStream/[[state]]=] is "`writable`".
+ 1. Set |controller|.[=WritableStreamDefaultController/[[releaseBackpressure]]=] to true.
+ 1. If ! [$WritableStreamHasOperationMarkedInFlight$](|stream|) is false and
+    [$WritableStreamCloseQueuedOrInFlight$](|stream|) is false,
```
domenic, line 5191:
```suggestion
    ! [$WritableStreamCloseQueuedOrInFlight$](|stream|) is false,
```

index.bs
```diff
@@ -6850,14 +6900,22 @@ reason.
...
  newly-[=new|created-via-Web IDL=] {{TransformStream}} |stream| given an algorithm <dfn export
- for="TransformStream/set up"><var>transformAlgorithm</var></dfn> and an optional algorithm <dfn
- export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, perform the following steps.
+ for="TransformStream/set up"><var>transformAlgorithm</var></dfn>, an optional algorithm <dfn
+ export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, an optional number <dfn
+ export for="TransformStream/set up"><var>writableHighWaterMark</var></dfn> (default 1), an optional
+ algorithm <dfn export for="TransformStream/set up"><var>writableSizeAlgorithm</var></dfn>, an
+ optional number <dfn export for="TransformStream/set up"><var>readableHighWaterMark</var></dfn>
+ (default 0), and an optional algorithm <dfn export
+ for="TransformStream/set up"><var>readableSizeAlgorithm</var></dfn>, perform the following steps.
```
domenic, line 6909: We could have separate set up algorithms for sync vs. async transforms, and thus take away the responsibility of picking the right high water mark from the other spec author. So far other specs haven't needed these extra customization points and I think it might be simpler not to give them out.

(On the other hand, the fact that existing byte-stream-transform specs are calculating the size of all chunks as 1 is a bit unfortunate...)

index.bs
MattiasBuelens: Fair point, opened #1192.

MattiasBuelens: Dangit, I probably should have moved https://github.com/whatwg/streams/pull/1190/commits/4fd1b3badcd36bb45b78b7f226a8828002973da0 out to that PR too. 🤦‍♂️

index.bs
MattiasBuelens: Shall I remove the HWM and size algorithm arguments from "set up a `ReadableStream`" and "set up a `WritableStream`" too then?

> (On the other hand, the fact that existing byte-stream-transform specs are calculating the size of all chunks as 1 is a bit unfortunate...)

Hmm, true. Once we have proper byte support for `WritableStream` and `TransformStream`, we'll have "set up with byte reading support" versions for those too, but for now we're in a bit of an awkward position. 😕

Maybe we can already provide such algorithms, even though they won't vend BYOB readers or BYOB writers yet?

jan-ivar:
> I'm concerned that people may be relying on the existing behaviour. With buffering, all the transforms in a pipe can run in parallel, 

@ricea I assume you mean "all the transforms in a pipe can run in parallel _with each other_" here. I.e. they all have data available to work on at the same time.

> but without it they will run in series.

And here your concern is all but one will sit idle waiting for data upstream, as observed from any one point in time?

I don't think that'd be the case. Assuming sufficient input data and transforms taking approximately the same time, all async transforms should still run in parallel with each other, even with HWM = 0. The only difference being that, from the perspective of each transform, the transform upstream of it is working on a newer chunk, at the same time.

* Snapshot at time X: camera → transformA(frame4) → transformB(frame3) → transformC(frame2) → sink(frame1)

These seem like reasonable assumptions for a healthy real-time stream.

But what seems true (and maybe this is what you worry about) is that this ideal never happens, and even tiny transform delays will accumulate (delayABC), which could blow a realtime buffer. Sure.

For example e.g. transformB spikes, taking 50% longer than usual on frame3, but then immediately makes up for it by taking half the normal time on frame4. Here HWM = 1 might improve resilience to variance. But to smooth out any longer delays than that will take a deeper buffer anyway, so I don't think the difference of 0 to 1 is foundational, unless I'm missing something.

> This could cause mysterious performance regressions.

PR #1191, opened 2021-11-28
Fix readable/writable file stream examples

dubiousjim:
These examples weren't closing the filehandle correctly (they called `close` on a nonexistent `fd` variable, rather than calling the `fileHandle.close` method).

Also the ReadableStream example misused the `fileHandle.read` call: it was supplying an ArrayBuffer and v's byteOffset within the ArrayBuffer, but the API requires supplying v itself and an offset relative to v's start.

Also added the file offset `position` to that `fileHandle.read` call. There's no reason otherwise for the example to keep track of this. Alternatively, the argument could be supplied as null, and we just rely on implicit file positioning.

***
<a href="https://whatpr.org/streams/1191.html" title="Last updated on Nov 29, 2021, 1:06 PM UTC (0173abf)">Preview</a> | <a href="https://whatpr.org/streams/1191/5885d94...0173abf.html" title="Last updated on Nov 29, 2021, 1:06 PM UTC (0173abf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7302,7 +7302,7 @@ function makeReadableFileStream(filename) {
     },
 
     async pull(controller) {
-      const buffer = new ArrayBuffer(CHUNK_SIZE);
+      const buffer = new Uint8Array(CHUNK_SIZE);
 
       const { bytesRead } = await fileHandle.read(buffer, 0, CHUNK_SIZE, position);
       if (bytesRead === 0) {
@@ -7310,7 +7310,7 @@ function makeReadableFileStream(filename) {
         controller.close();
       } else {
         position += bytesRead;
-        controller.enqueue(new Uint8Array(buffer, 0, bytesRead));
+        controller.enqueue(buffer.subarray(0, bytesRead));
       }
     },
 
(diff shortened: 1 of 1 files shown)


MattiasBuelens:
Thanks! Looks good. 👍

> Also the ReadableStream example misused the `fileHandle.read` call: it was supplying an ArrayBuffer and v's byteOffset within the ArrayBuffer, but the API requires supplying v itself and an offset relative to v's start.

I noticed a similar mistake in the ["A readable stream with an underlying pull source"](https://streams.spec.whatwg.org/commit-snapshots/5885d94e532e168959b98d27887e7f9f5e6becd3/#example-rs-pull) example. We're passing an `ArrayBuffer` to `fileHandle.read()`, but [it only accepts `Buffer | TypedArray | DataView`](https://nodejs.org/api/fs.html#filehandlereadbuffer-offset-length-position). https://github.com/whatwg/streams/blob/5885d94e532e168959b98d27887e7f9f5e6becd3/index.bs#L7305-L7307

We should probably use `new Uint8Array(CHUNK_SIZE)` instead, and use `.subarray(0, bytesRead)` (or `.slice`) in the `controller.enqueue()` call. Could you fix that too as part of this PR? 🙂

MattiasBuelens (approved):

PR #1193, opened 2021-12-02
Explainer for proposed stream changes related to VideoFrame processing

youennf:
As discussed at last WebRTC/WHATWG Streams meeting, we are converging on a potential solution for VideoFrame handling in streams. This explainer is the first step towards a more formal solution.

### DIFF
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -26,39 +26,37 @@ Transferring ownership should be opt-in. For that purpose, a new streams type, n
 ## Example
 
 Below is an example of JavaScript that shows how this can be used.
-The example creates a processing pipe starting with a camera stream and applying two transforms, one for doing a processing for every 30 frame, and one for doing background blur.
+The example creates a processing pipe starting with a VideoFrame stream and applying two transforms, one for doing a processing like logging every 30 frame, and one for doing background blur.
 
-```javascript
-// worker.js
-let frameCount = 0;
-const frameCountTransform = new TransformStream({
-  transform: async (videoFrame, controller) => {
-    try {
-      // videoFrame is under the responsibility of the script and must be closed when no longer needed
-      controller.enqueue(videoFrame);
-      // At this point, videoFrame has been transfered within controller.enqueue call. frameCountTransform cannot mutate it.
-      if (!(++frameCount % 30) && frameCountTransform.onEach30Frame)
-          frameCountTransform.onEach30Frame(frameCount);
-    } catch {
-      videoFrame.close();
-    }
-  },
-  readableType: 'transfer',
-  writableType: 'transfer'
-});
-frameCountTransform.onEach30Frame = (frameCount) => {
-  // Do something periodically.
-};
-// Native transform is of type 'transfer'
-const backgroundBlurTransform = new BackgroundBlurTransform();
+```worker.js javascript
+function doBackgroundBlurOnVideoFrames(videoFrameStream, doLogging)
+{
+  // JavaScript custom transform.
+  let frameCount = 0;
+  const frameCountTransform = new TransformStream({
+    transform: async (videoFrame, controller) => {
+      try {
+        // videoFrame is under the responsibility of the script and must be closed when no longer needed.
+        controller.enqueue(videoFrame);
+        // controller.enqueue was called, videoFrame is transferred.
+        if (!(++frameCount % 30) && doLogging)
+            doLogging(frameCount);
+      } catch (e) {
+        // In case of exception, let's make sure videoFrame is closed. This is a no-op if videoFrame was previously transferred.
+        videoFrame.close();
+        // If exception is unrecoverable, let's error the pipe.
+        controller.error(e);
+      }
+    },
+    readableType: 'transfer',
+    writableType: 'transfer'
+  });
+  // Native transform is of type 'transfer'
+  const backgroundBlurTransform = new BackgroundBlurTransform();
 
-const cameraStream = await navigator.mediaDevices.getUserMedia({ video : true });
-const videoFrameStream = getReadableStreamFromTrack(cameraStream.getVideoTracks()[0]);
-const blurredVideoFrameStream = videoFrameStream.pipeThrough(backgroundBlurTransform)
-                                                .pipeThrough(frameCountTransform);
-const blurredStream = new MediaStream([getTrackFromReadableStream(blurredVideoFrameStream)]);
-// Make use of blurredStream.
-...
+  return videoFrameStream.pipeThrough(backgroundBlurTransform)
+                         .pipeThrough(frameCountTransform);
+}
 ```
 
 ## Goals
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -30,23 +30,23 @@ The example creates a processing pipe starting with a camera stream and applying
 
 ```javascript
 // JS transform
+let frameCount = 0;
 const frameCountTransform = new TransformStream({
   transform: async (videoFrame, controller) => {
     try {
       // videoFrame is under the responsibility of the script and must be closed when no longer needed
       controller.enqueue(videoFrame);
       // At this point, videoFrame has been transfered within controller.enqueue call. frameCountTransform cannot mutate it.
-      if (!controller.count)
-        controller.count = 0;
-      if (!(++controller.count % 30) && frameCountTransform.onEach30Frame)
-          frameCountTransform.onEach30Frame(controller.count);
+      if (!(++frameCount % 30) && frameCountTransform.onEach30Frame)
+          frameCountTransform.onEach30Frame(frameCount);
     } catch {
       videoFrame.close();
     }
   },
-  type: 'transfer'
+  readableType: 'transfer',
+  writableType: 'transfer'
 });
-frameCountTransform.onEach30Frame = (count) => {
+frameCountTransform.onEach30Frame = (frameCount) => {
   // Do something periodically.
 };
 // Native transform is of type 'transfer'
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -93,8 +93,8 @@ The envisioned changes to the streams specification could look like the followin
 *   Streams of the 'transfer' type can only manipulate Transferable or Serializable objects.
     If a non Transferable or Serializable object is enqueued or written, the object is ignored as if it was never enqueued/written.
 *   If a Transferable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
-    or `WritableStreamDefaultWriter`, create a copy of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
-    Proceed with the regular stream algorithm by using the copy of the object instead of the object itself.
+    or `WritableStreamDefaultWriter`, create a transferred version of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
+    Proceed with the regular stream algorithm by using the transferred object instead of the object itself.
 *   If a Serializable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
     or `WritableStreamDefaultWriter`, create a copy of the object using StructuredSerialize/StructuredDeserialize.
     Proceed with the regular stream algorithm by using the copy of the object instead of the object itself.
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -90,7 +90,7 @@ const blurredStream = new MediaStream([getTrackFromReadableStream(blurredVideoFr
 The envisioned changes to the streams specification could look like the following:
 *   Add a new 'transfer' value that can be passed to `ReadableStream` type, `WritableStream` type and `TransformStream` readableType/writableType.
     For streams that do not use the 'transfer' type, nothing changes.
-*   Streams of the 'transfer' type can only manipulate that are marked both as Transferable and Serializable.
+*   Streams of the 'transfer' type can only manipulate objects that are marked both as Transferable and Serializable.
 *   If an object that is either not Transferable or not Serializable is enqueued or written, the object is ignored as if it was never enqueued/written.
 *   If a Transferable and Serializable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
     or `WritableStreamDefaultWriter`, create a transferred version of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -96,6 +96,7 @@ The envisioned changes to the streams specification could look like the followin
 *   Introduce a WhatWG streams 'close-able' concept. A chunk that is 'close-able' defines closing steps.
     For instance `VideoFrame` closing steps could be defined using https://www.w3.org/TR/webcodecs/#close-videoframe.
     `ArrayBuffer` closing steps could be defined using https://tc39.es/ecma262/#sec-detacharraybuffer.
+    The 'close-able' steps should be a no-op on a transferred chunk.
 *   Execute the closing steps of a 'close-able' chunk for streams with the 'transfer' type when resetting the queue of `ReadableStreamDefaultController`
     or emptying `WritableStream`.[[writeRequests]] in case of abort/error.
 *   When calling tee() on a `ReadableStream` of the 'transfer' type, call ReadableStream with cloneForBranch2 equal to true. 
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -1,4 +1,4 @@
-# Transfering Ownership Streams Explained
+# Transferring Ownership Streams Explained
 
 
 ## Introduction
--- a/streams-for-raw-video-explainer.md
+++ b/streams-for-raw-video-explainer.md
@@ -5,21 +5,21 @@
 
 The streams APIs provide a convenient way to build processing pipelines.
 Using streams with `VideoFrame` objects has known shortcomings, as illustrated by https://github.com/whatwg/streams/issues/1155 or https://github.com/whatwg/streams/issues/1185 for instance.
-This proposal addresses these shortcomings by improving streams support for objects similar but not limited to WebCodec `VideoFrame`.
-It could also be useful for streams making use of `ArrayBuffer` or objects that own `ArrayBuffer` like `RTCEncodedVideoChunk`.
+This proposal addresses these shortcomings by improving streams support for chunks similar but not limited to WebCodec `VideoFrame`.
+It could also be useful for streams making use of `ArrayBuffer` or chunks that own `ArrayBuffer` like `RTCEncodedVideoChunk`.
 
-Streams APIs can create coupling between the processing units of the pipeline when objects in the pipe are mutable:
-a processing unit A might pass an object O to another unit B through a stream (say a `WritableStream`) W.
+Streams APIs can create coupling between the processing units of the pipeline when chunks in the pipe are mutable:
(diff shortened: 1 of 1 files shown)


youennf:
@MattiasBuelens, this is a first draft of the explainer, could you take a look?

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+```javascript
+// JS transform
+const frameCountTransform = new TransformStream({
+  transform: async (videoFrame, controller) => {
+    try {
+      // videoFrame is under the responsibility of the script and must be closed when no longer needed
+      controller.enqueue(videoFrame);
+      // At this point, videoFrame has been transfered within controller.enqueue call. frameCountTransform cannot mutate it.
+      if (!controller.count)
+        controller.count = 0;
```
MattiasBuelens, line 40: It might be a bit confusing to use "expando properties" in these examples. (What if we decide to add `TransformStreamDefaultController.prototype.count` in the future? 😛)

For these examples, it's probably better to use a regular variable:
```javascript
let count = 0;
const frameCountTransform = new TransformStream({
  transform: async (videoFrame, controller) => { /* ... */ }
});
```

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+      if (!(i++controller.count % 30) && frameCountTransform.onEach30Frame)
```
MattiasBuelens, line 41: `i++controller.count` is a `SyntaxError`. 😬

```suggestion
      if (!(++controller.count % 30) && frameCountTransform.onEach30Frame)
```

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+*   `VideoFrame` needs specific management and be closed as quickly as possible, without relying on garbage collection.
+    This is important to not create hangs/stutters in the processing pipeline. By building support for safe patterns
+    directly in streams, this will allow web developers to optimize `VideoFrame` management, and allow user experience
+    to be more consistent accross devices.
+
+## Principles
+
+The envisioned changes to the streams specification could look like the following:
+*   Add a new 'transfer' type to `ReadableStream`, `WritableStream` and `TransformStream`.
+    For streams that do not have the 'transfer' type, nothing changes.
```
MattiasBuelens, line 92: A `TransformStream` may transform one type of chunks to a different type of chunks. For example, a `TextDecoderStream` accepts `ByteSource`s on its writable end and produces `string`s on its readable end. We should try to support such use cases for ownership-transferring streams too. Therefore, we shouldn't "just" have `type: "transfer"` for `TransformStream`.

We have already reserved `readableType` and `writableType` for such use cases, see the [`TransformStream` constructor](https://streams.spec.whatwg.org/#ts-constructor). I suggest we use that:
* If `writableType` is `"transfer"`, then the writable end will accept transferable objects. Ownership is transferred into the stream when calling `writer.write(chunk)`, and transferred out when the stream calls `transformer.transform(chunk, controller)`.
* If `readableType` is `"transfer"`, then the readable end will produce transferable objects. Ownership is transferred into the stream when calling `controller.enqueue(chunk)`, and transferred out when resolving a `reader.reader()` call.

For convenience, we *could* still add `transformer.type`, as a shorthand for setting both `readableType` and `writableType` to the same value. That would make the common case of transforming a `VideoFrame` to another `VideoFrame` easier. Or maybe we want to leave that as a future improvement.

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+*   Streams of the 'transfer' type can only manipulate Transferable or Serializable objects.
```
MattiasBuelens, line 93: I believe we have to *require* all objects to be `Serializable`. Otherwise, we cannot call `StructuredSerialize`/`StructuredDeserialize` inside `tee()` to make a "real" clone. (A transferred version of an object is not a clone, since you can no longer use the original.)

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+    If a non Transferable or Serializable object is enqueued or written, the object is ignored as if it was never enqueued/written.
+*   If a Transferable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
+    or `WritableStreamDefaultWriter`, create a copy of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
+    Proceed with the regular stream algorithm by using the copy of the object instead of the object itself.
```
MattiasBuelens, line 97:
```suggestion
    or `WritableStreamDefaultWriter`, create a transferred version of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
    Proceed with the regular stream algorithm by using the transferred object instead of the object itself.
```

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+*   If a Serializable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
+    or `WritableStreamDefaultWriter`, create a copy of the object using StructuredSerialize/StructuredDeserialize.
+    Proceed with the regular stream algorithm by using the copy of the object instead of the object itself.
```
MattiasBuelens, line 100: We should also close the original object if it supports `Closeable`. (If it doesn't support it, we just let it be garbage collected as normal.)

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+*   Introduce a WhatWG streams 'close-able' concept. An object that is 'close-able' defines closing steps.
+    For instance `VideoFrame` closing steps could be defined using https://www.w3.org/TR/webcodecs/#close-videoframe.
+    `ArrayBuffer` closing steps could be defined using https://tc39.es/ecma262/#sec-detacharraybuffer.
```
MattiasBuelens, line 103: `DetachArrayBuffer` doesn't deallocate the underlying Data Block of the `ArrayBuffer`, since it's also used to *transfer* an `ArrayBuffer` (see [step 5.4 of `StructuredSerializeWithTransfer`](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializewithtransfer)). However, in our case we know that we won't transfer it to a new `ArrayBuffer`, and thus the Data Block *could* be deallocated immediately.

We should make a note that implementations are *allowed* to deallocate the Data Block immediately after closing an `ArrayBuffer` (or they could just wait for garbage collection as normal).

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+    directly in streams, this will allow web developers to optimize `VideoFrame` management, and allow user experience
+    to be more consistent accross devices.
+
+## Principles
+
+The envisioned changes to the streams specification could look like the following:
+*   Add a new 'transfer' type to `ReadableStream`, `WritableStream` and `TransformStream`.
+    For streams that do not have the 'transfer' type, nothing changes.
+*   Streams of the 'transfer' type can only manipulate Transferable or Serializable objects.
+    If a non Transferable or Serializable object is enqueued or written, the object is ignored as if it was never enqueued/written.
```
MattiasBuelens, line 94: Hmm, not sure if I like this "silent error". For readable byte streams, we throw a `TypeError` in `controller.enqueue()` if the given chunk is not an `ArrayBufferView`. Ideally, we'd do something similar for ownership-transferring streams.

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,120 @@
...
+
+*   Permit `ReadableStream`, `WritableStream` and `TransformStream` objects to take ownership of objects they manipulate.
+*   Permit to build a safe and optimal video pipeline using `ReadableStream`, `WritableStream` and `TransformStream` objects that manipulate `VideoFrame` objects.
+*   Permit both native and JavaScript-based streams of type 'transfer'.
+*   Permit to optimize streams pipelines of transferable objects like `ArrayBuffer`, `RTCEncodedVideoFrame` or `RTCEncodedAudioFrame`.
+*   Permit to tee a `ReadableStream` of `VideoFrame` objects without tight coupling between the teed branches.
+
+## Non-goals
+
+*   Add support for supporting transfer and closing of arbitrary JavaScript objects.
```
MattiasBuelens, line 74:
```suggestion
*   Add support for transferring and closing of arbitrary JavaScript objects.
```

MattiasBuelens (changes requested):
Thanks for kicking this off! Looks pretty good already. 👍

streams-for-raw-video-explainer.md
youennf: Will update the example

streams-for-raw-video-explainer.md
youennf: Right, I'll update this line for TransformStream.

streams-for-raw-video-explainer.md
youennf: The two options are only support Serialisable, or throw.
The use case only requires Serializable right now, so let's go with this.

streams-for-raw-video-explainer.md
youennf: This is probably safer but probably unneeded in the case of transferable (& serialisable) since the object would have been transferred then closed. I guess the question remains for serialisable but not transferable objects. Use case only requires Transferable+Serializable.

streams-for-raw-video-explainer.md
youennf: Good point, not sure that it fits the explainer but would definitely be good for a PR.

streams-for-raw-video-explainer.md
youennf: I'll add an open question section to mention what to do when enqueuing/writing not serializable objects.
I prefer rejecting over being silent if we are ok not erroring the stream either.

streams-for-raw-video-explainer.md
youennf: Done

streams-for-raw-video-explainer.md
youennf: Restricted to Transferable+Serializable and making use of transfer steps.

youennf: Plus added an item about serializable only objects

streams-for-raw-video-explainer.md
youennf: Added

streams-for-raw-video-explainer.md
youennf: Done

streams-for-raw-video-explainer.md
youennf: Restricted to Transferable+Serializable for now.
Added open questions for only transferable and only serializable.

streams-for-raw-video-explainer.md
youennf: Added a corresponding item in future work.

streams-for-raw-video-explainer.md
MattiasBuelens: Agreed. If we can make this an early error that doesn't error the entire stream, that'd be great. 🙂

youennf:
@alvestrand, @aboba, @jan-ivar, @ricea, PTAL.

youennf:
@domenic, could you also take a look whether that makes sense to you and whether that is worth including in streams (instead of special casing VideoFrame handling).

domenic:
Generally looks good to me, and makes sense as a concept. Lots of details to work out, but I'm happy for that to be done in future revisions or in a spec PR, instead of having a perfect explainer ahead of time.

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,124 @@
...
+
+Transferring ownership should be opt-in. For that purpose, a new streams type, named 'transfer' in this document,  would be added.
+
+## Example
+
+Below is an example of JavaScript that shows how this can be used.
+The example creates a processing pipe starting with a camera stream and applying two transforms, one for doing a processing for every 30 frame, and one for doing background blur.
+
+```javascript
+// JS transform
```
jan-ivar, line 32: I think it's important we lead with worker examples from the get-go.
```suggestion
// worker.js
```

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,124 @@
...
+};
+// Native transform is of type 'transfer'
+const backgroundBlurTransform = new BackgroundBlurTransform();
+
+const cameraStream = await navigator.mediaDevices.getUserMedia({ video : true });
+const videoFrameStream = getReadableStreamFromTrack(cameraStream.getVideoTracks()[0]);
+const blurredVideoFrameStream = videoFrameStream.pipeThrough(backgroundBlurTransform)
+                                                .pipeThrough(frameCountTransform);
+const blurredStream = new MediaStream([getTrackFromReadableStream(blurredVideoFrameStream)]);
+// Make use of blurredStream.
```
jan-ivar, line 60: Use workers part 2. Also, mediacapture-transform is (almost as of this writing) adopted, so I think we can be specific now:
```suggestion
self.onmessage = async ({data: {cameraTrack}}) => {
  const {readable} = new MediaStreamTrackProcessor({cameraTrack});
  const {writable, track} = new VideoTrackGenerator();
  parent.postMessage({track}, [track]);
  await readable.pipeThrough(backgroundBlurTransform).pipeThrough(frameCountTransform).pipeTo(writable);
}

// main.js
const [cameraTrack] = (await navigator.mediaDevices.getUserMedia({video: true}).getVideoTracks();
const worker = new Worker('worker.js');
worker.postMessage({cameraTrack}, [cameraTrack]);

const {data} = await new Promise(r => worker.onmessage);
const blurredStream = new MediaStream([data.track]);
video.srcObject = blurredStream;
```

jan-ivar (commented):
Content looks great! Just a question on mentioning WebIDL, and nits. I also think we should show workers in the example, since that's what we've spec'ed (see suggestion).

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,124 @@
...
+    This is important to not create hangs/stutters in the processing pipeline. By building support for safe patterns
+    directly in streams, this will allow web developers to optimize `VideoFrame` management, and allow user experience
+    to be more consistent accross devices.
+
+## Principles
+
+The envisioned changes to the streams specification could look like the following:
+*   Add a new 'transfer' value that can be passed to `ReadableStream` type, `WritableStream` type and `TransformStream` readableType/writableType.
+    For streams that do not use the 'transfer' type, nothing changes.
+*   Streams of the 'transfer' type can only manipulate that are marked both as Transferable and Serializable.
```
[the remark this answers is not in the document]
youennf: Moving to chunk is probably more consistent.

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,124 @@
...
+*   If an object that is either not Transferable or not Serializable is enqueued or written, the object is ignored as if it was never enqueued/written.
+*   If a Transferable and Serializable object is enqueueud/written in a 'transfer' type `ReadableStreamDefaultController`, `TransformStreamDefaultController`
+    or `WritableStreamDefaultWriter`, create a transferred version of the object using StructuredSerializeWithTransfer/StructuredDeserializeWithTransfer.
+    Proceed with the regular stream algorithm by using the transferred object instead of the object itself.
+*   Introduce a WhatWG streams 'close-able' concept. An object that is 'close-able' defines closing steps.
```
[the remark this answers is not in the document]
youennf: The idea is to start with a WhatWG stream closable or closable concept, but look at making it a WebIDL concept if there is enough interest. See one of the last item in proposal

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,124 @@
...
+  transform: async (videoFrame, controller) => {
+    try {
+      // videoFrame is under the responsibility of the script and must be closed when no longer needed
+      controller.enqueue(videoFrame);
+      // At this point, videoFrame has been transfered within controller.enqueue call. frameCountTransform cannot mutate it.
+      if (!(++frameCount % 30) && frameCountTransform.onEach30Frame)
+          frameCountTransform.onEach30Frame(frameCount);
+    } catch {
+      videoFrame.close();
+    }
```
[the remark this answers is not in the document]
youennf: > It cannot mutate it, because it cannot even access it, is that right?

It has access to the object, but no longer to the underlying bytes.

> Also, don't we need to rethrow in the catch, to not encourage overriding default error handling?

One of the complaint we heard from streams was that errors propagate very easily, while there may be recoverable errors hence why I did not add throw there. I would prefer calling the error method explicitly in case of unrecoverable errors.

> I also worry this risks calling close on a transferred object, or is that benign? 

I think it is begnin.

streams-for-raw-video-explainer.md
youennf: Right

streams-for-raw-video-explainer.md
youennf: I prefer we keep it abstract from MediaStreamTrackProcessor and we keep the example as small as possible.
I'll remove the call to getUserMedia which is indeed confusing.

streams-for-raw-video-explainer.md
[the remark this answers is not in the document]
alvestrand: If we borrow the "transfer" concept from "transferable", the object left behind after the transfer is "neutered"; usually this means all the methods on it check a "neutered" flag before doing anything. (I may have the name wrong.)

I think we should mandate that all closable objects have a close operation that is a no-op if the object is already closed (or transferred).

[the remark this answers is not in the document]
youennf: Right, the explainer refers to https://www.w3.org/TR/webcodecs/#close-videoframe which will be a no-op in practice if you call it more than once.

streams-for-raw-video-explainer.md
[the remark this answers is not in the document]
youennf: Moved to chunk

streams-for-raw-video-explainer.md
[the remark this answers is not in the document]
youennf: I kept close-able for now but am happy to change to closable if found better.
Interestingly, Java defines closeable as: 'a source or destination of data that can be closed. The close method is invoked to release resources that the object is holding (such as open files).'

streams-for-raw-video-explainer.md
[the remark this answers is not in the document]
youennf: > I think we should mandate that all closable objects have a close operation that is a no-op if the object is already closed (or transferred).

I added a line to explicitly state this.

youennf:
Build error seems unrelated to this PR:
LINK ERROR: Multiple possible 'WebSocket' idl refs.

MattiasBuelens:
Correct, #1198 should fix the spec build.

jan-ivar (commented):
LGTM

youennf:
Discussed at WebRTC editor's meeting today, and was thought as good to go as a first version.
@MattiasBuelens , @domenic, do you think it is ready for merging?

domenic (approved):
LGTM too. @MattiasBuelens has a changes-requested review which we need to make sure it satisfied though.

streams-for-raw-video-explainer.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+## Goals
+
+*   Permit `ReadableStream`, `WritableStream` and `TransformStream` objects to take ownership of chunks they manipulate.
+*   Permit to build a safe and optimal video pipeline using `ReadableStream`, `WritableStream` and `TransformStream` objects that manipulate `VideoFrame` objects.
+*   Permit both native and JavaScript-based streams of type 'transfer'.
+*   Permit to optimize streams pipelines of transferable chunks like `ArrayBuffer`, `RTCEncodedVideoFrame` or `RTCEncodedAudioFrame`.
+*   Permit to tee a `ReadableStream` of `VideoFrame` objectss without tight coupling between the teed branches.
```
MattiasBuelens, line 68:
```suggestion
*   Permit to tee a `ReadableStream` of `VideoFrame` objects without tight coupling between the teed branches.
```

MattiasBuelens (approved):
LGTM! 👍

Issue #1207, opened 2022-01-15
pipeTo: Read fullfilled after shutdown started


evilpie:
The spec is quite clear that after shutting down there should be no new reads. However there is an open question in Firefox's old implementation of pipeTo about what happens when a read is fulfilled after shutdown. I am quoting Waldos' comment in full below:

> If |source| becomes errored not during a pending read, it's clear we must
> react immediately.
> 
> But what if |source| becomes errored *during* a pending read?  Should this
> first error, or the pending-read second error, predominate?  Two semantics
> are possible when |source|/|dest| become closed or errored while there's a
> pending read:
> 
>   1. Wait until the read fulfills or rejects, then respond to the
>      closure/error without regard to the read having fulfilled or rejected.
>      (This will simply not react to the read being rejected, or it will
>      queue up the read chunk to be written during shutdown.)
>   2. React to the closure/error immediately per "Error and close states
>      must be propagated".  Then when the read fulfills or rejects later, do
>      nothing.
> 
> The spec doesn't clearly require either semantics.  It requires that
> *already-read* chunks be written (at least if |dest| didn't become errored
> or closed such that no further writes can occur).  But it's silent as to
> not-fully-read chunks.  (These semantic differences may only be observable
> with very carefully constructed readable/writable streams.)
> 
> It seems best, generally, to react to the temporally-earliest problem that
> arises, so we implement option #2.  (Blink, in contrast, currently
> implements option #1.)
> 
> All specified reactions to a closure/error invoke either the shutdown, or
> shutdown with an action, algorithms.  Those algorithms each abort if either
> shutdown algorithm has already been invoked.  So we don't need to do
> anything special here to deal with a pending read.
https://searchfox.org/mozilla-central/rev/3de56eb5f266f523340e739ae1b53258e0a95dfe/js/src/builtin/streams/PipeToState.cpp#454-483

When working on my [new implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1734241) I actually just ended up doing what is described as option #1 (aka Chrome's behavior). My own intuition here is that it would be better to not just drop an already read chunk onto the floor and at least try to write it. The write might of course still fail anyway, considering that we might cancel the stream during shutdown or release the writer.

Sorry if this has already been discussed somewhere else.

MattiasBuelens:
> But what if |source| becomes errored _during_ a pending read?

[...]

PR #1208, opened 2022-01-16
Release reader immediately when shutting down a pipe

MattiasBuelens:
In #1207, we raised concern about whether it's currently possible for `pipeTo()` to drop chunks when the pipe shuts down *while it still has a pending read request*.

It turns out that this *is* indeed possible. [The new WPT tests](https://github.com/web-platform-tests/wpt/pull/32399) demonstrate at least one way this can happen: by aborting a pipe with an `AbortSignal`, and then enqueuing a chunk immediately after aborting the signal.

This PR fixes this by releasing the pipe's reader immediately when *starting* the shutdown process, rather than when it *finalizes* the shutdown. This ensures that any pending reads that were started are immediately rejected, and the chunks stay in the source's queue.

~This does have the side effect of releasing the reader *before* the pipe promise resolves. I don't know if this is acceptable.~ Fixed in https://github.com/whatwg/streams/pull/1208#issuecomment-1015911238.

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/32399
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/streams/1208.html" title="Last updated on May 25, 2026, 9:25 PM UTC (5e26f17)">Preview</a> | <a href="https://whatpr.org/streams/1208/4c5b332...5e26f17.html" title="Last updated on May 25, 2026, 9:25 PM UTC (5e26f17)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2305,6 +2305,14 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
    1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
       ! [$ReadableStreamBYOBReaderRelease$](|reader|).
    1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|source|).
+      <p class="note">The initial reader is released to ensure that any pending read requests
+      are immediately aborted, and no more chunks are pulled from |source|. A new reader is
+      acquired in order to keep |source| locked until the shutdown is [=finalized=], for example
+      to [=cancel a readable stream|cancel=] |source| if necessary.
+      This exchange of readers is not observable to author code and the user agent is free to
+      implement this differently, for example by keeping the same reader and internally aborting
+      its pending read requests.
    1. If |dest|.[=WritableStream/[[state]]=] is "`writable`" and !
       [$WritableStreamCloseQueuedOrInFlight$](|dest|) is false,
      1. If any [=chunks=] have been read but not yet written, write them to |dest|.
@@ -2320,6 +2328,7 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
    1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
       ! [$ReadableStreamBYOBReaderRelease$](|reader|).
    1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
+   1. Set |reader| to ! [$AcquireReadableStreamDefaultReader$](|source|).
    1. If |dest|.[=WritableStream/[[state]]=] is "`writable`" and !
       [$WritableStreamCloseQueuedOrInFlight$](|dest|) is false,
     1. If any [=chunks=] have been read but not yet written, write them to |dest|.
@@ -2328,8 +2337,10 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
    1. [=Finalize=], passing along |error| if it was given.
   * <dfn id="rs-pipeTo-finalize"><i>Finalize</i></dfn>: both forms of shutdown will eventually ask
     to finalize, optionally with an error |error|, which means to perform the following steps:
-   1. Assert: |reader|.[=ReadableStreamGenericReader/[[stream]]=] is undefined.
    1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
+   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
+      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
+   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
    1. If |signal| is not undefined, [=AbortSignal/remove=] |abortAlgorithm| from |signal|.
    1. If |error| was given, [=reject=] |promise| with |error|.
    1. Otherwise, [=resolve=] |promise| with undefined.
--- a/index.bs
+++ b/index.bs
@@ -2302,6 +2302,9 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
     |originalError|, then:
    1. If |shuttingDown| is true, abort these substeps.
    1. Set |shuttingDown| to true.
+   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
+      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
+   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
    1. If |dest|.[=WritableStream/[[state]]=] is "`writable`" and !
       [$WritableStreamCloseQueuedOrInFlight$](|dest|) is false,
      1. If any [=chunks=] have been read but not yet written, write them to |dest|.
@@ -2314,6 +2317,9 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
     ask to shutdown, optionally with an error |error|, then:
    1. If |shuttingDown| is true, abort these substeps.
    1. Set |shuttingDown| to true.
+   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
+      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
+   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
    1. If |dest|.[=WritableStream/[[state]]=] is "`writable`" and !
       [$WritableStreamCloseQueuedOrInFlight$](|dest|) is false,
     1. If any [=chunks=] have been read but not yet written, write them to |dest|.
@@ -2322,10 +2328,8 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
    1. [=Finalize=], passing along |error| if it was given.
   * <dfn id="rs-pipeTo-finalize"><i>Finalize</i></dfn>: both forms of shutdown will eventually ask
     to finalize, optionally with an error |error|, which means to perform the following steps:
+   1. Assert: |reader|.[=ReadableStreamGenericReader/[[stream]]=] is undefined.
    1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
-   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
-      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
-   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
    1. If |signal| is not undefined, [=AbortSignal/remove=] |abortAlgorithm| from |signal|.
    1. If |error| was given, [=reject=] |promise| with |error|.
    1. Otherwise, [=resolve=] |promise| with undefined.
--- a/index.bs
+++ b/index.bs
@@ -2337,10 +2337,9 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
    1. [=Finalize=], passing along |error| if it was given.
   * <dfn id="rs-pipeTo-finalize"><i>Finalize</i></dfn>: both forms of shutdown will eventually ask
     to finalize, optionally with an error |error|, which means to perform the following steps:
+   1. Assert: |reader| [=implements=] {{ReadableStreamDefaultReader}}.
    1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
-   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
-      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
-   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
+   1. Perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
    1. If |signal| is not undefined, [=AbortSignal/remove=] |abortAlgorithm| from |signal|.
    1. If |error| was given, [=reject=] |promise| with |error|.
    1. Otherwise, [=resolve=] |promise| with undefined.
--- a/index.bs
+++ b/index.bs
@@ -2279,7 +2279,7 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
        |source|.[=ReadableStream/[[storedError]]=].
     1. Otherwise, [=shutdown=] with |source|.[=ReadableStream/[[storedError]]=].
    1. <strong>Errors must be propagated backward:</strong> if |dest|.[=WritableStream/[[state]]=]
-      is or becomes "`errored`", then
+      is or becomes "`erroring`" or "`errored`", then
     1. If |preventCancel| is false, [=shutdown with an action=] of !
        [$ReadableStreamCancel$](|source|, |dest|.[=WritableStream/[[storedError]]=]) and with
        |dest|.[=WritableStream/[[storedError]]=].
--- a/reference-implementation/lib/abstract-ops/readable-streams.js
+++ b/reference-implementation/lib/abstract-ops/readable-streams.js
@@ -227,33 +227,50 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
     }
 
     // Errors must be propagated forward
-    sourceIsOrBecomesErrored(() => {
+    function sourceIsOrBecomesErrored() {
       const storedError = source._storedError;
       if (preventAbort === false) {
         shutdownWithAction(() => WritableStreamAbort(dest, storedError), true, storedError);
       } else {
         shutdown(true, storedError);
       }
-    });
+    }
 
     // Errors must be propagated backward
-    destIsOrBecomesErroringOrErrored(() => {
+    function destIsOrBecomesErroringOrErrored() {
       const storedError = dest._storedError;
       if (preventCancel === false) {
         shutdownWithAction(() => ReadableStreamCancel(source, storedError), true, storedError);
       } else {
         shutdown(true, storedError);
       }
-    });
+    }
 
     // Closing must be propagated forward
-    sourceIsOrBecomesClosed(() => {
+    function sourceIsOrBecomesClosed() {
       if (preventClose === false) {
         shutdownWithAction(() => WritableStreamDefaultWriterCloseWithErrorPropagation(writer));
       } else {
         shutdown();
       }
-    });
+    }
+
+    function checkState() {
+      const sourceState = source._state;
+      const destState = dest._state;
+      if (sourceState === 'errored') {
+        // Errors must be propagated forward
+        sourceIsOrBecomesErrored();
+      } else if (destState === 'erroring' || destState === 'errored') {
+        // Errors must be propagated backward
+        destIsOrBecomesErroringOrErrored();
+      } else if (sourceState === 'closed') {
+        // Closing must be propagated forward
+        sourceIsOrBecomesClosed();
+      }
+    }
+
+    checkState();
 
     // Closing must be propagated backward
     if (WritableStreamCloseQueuedOrInFlight(dest) === true || dest._state === 'closed') {
@@ -266,6 +283,11 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
       }
     }
 
+    if (!shuttingDown) {
+      readerAddStateChangeListener(reader, checkState);
+      defaultWriterAddStateChangeListener(writer, checkState);
+    }
+
     setPromiseIsHandledToTrue(pipeLoop());
 
     function waitForWritesToFinish() {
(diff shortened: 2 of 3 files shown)


index.bs
```diff
@@ -2198,6 +2198,9 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
     |originalError|, then:
    1. If |shuttingDown| is true, abort these substeps.
    1. Set |shuttingDown| to true.
+   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
+      ! [$ReadableStreamBYOBReaderRelease$](|reader|).
+   1. Otherwise, perform ! [$ReadableStreamDefaultReaderRelease$](|reader|).
```
MattiasBuelens, line 2203: Drive-by fix: the user agent is allowed to use a BYOB reader for the pipe (see step 8), but now we have *separate* `Release()` abstract ops for default and BYOB readers. (Before #1168, we had a single `ReadableStreamGenericReaderRelease` that worked for both.)

MattiasBuelens:
> This does have the side effect of releasing the reader _before_ the pipe promise resolves. I don't know if this is acceptable.

Hmm, this means that the pipe could still cancel the readable *after* it has released its reader lock but *before* it settles the pipe promise. That might be quite unexpected: the user might have already acquired a new reader, and only that *new reader* should have the ability to cancel the stream.

Possible solutions:
1. We could try to split up `ReadableStream(Default|BYOB)Release`, such that the read requests are already rejected *but* the reader is still locked to the stream. We would then *actually* release the reader as part of the "finalize" steps (as in the current spec).
2. We could make the pipe releases its reader, and immediately acquire a new reader. It would no longer use that reader to request new chunks (since *shuttingDown* is already true), it would only be used to keep the stream locked (in case we still want to cancel it).

Option 2 is probably the easiest to implement, but it may require a note to explain why this weird "release + acquire" dance is necessary. Then again, we already mention that:
> the exact manner in which this happens is not observable to author code, and so there is flexibility in how this is done

so user agents are free to implement these steps differently if they want. 🤷‍♂️

index.bs
domenic: Mind factoring this out into a separate PR?

domenic (commented):
It sounds like either option (1) or (2) is observably equivalent, right? My instinct would then be to go with whichever is easiest to spec (probably (2)?), and then explain that it's equivalent to the other one.

If we go with (2) and explain it's equivalent to (1), we're explaining the weird dance and saying that perhaps a more natural implementation would be (1), if you're less into code reuse than we are.

If we go with (1) and explain it's equivalent to (2), then we're explaining how despite doing some tricky direct manipulations of the internals, we're not actually doing anything which you couldn't already do with the public API.

MattiasBuelens:
Yes, they're equivalent. I'll go with option 2. 👍

index.bs
MattiasBuelens: Done, see #1209.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -200,6 +200,12 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
       }
 
       return transformPromiseWith(writer._readyPromise, () => {
+        if (shuttingDown === true) {
+          return promiseResolvedWith(true);
+        }
+        if (dest._state !== 'writable' || WritableStreamCloseQueuedOrInFlight(dest) === true) {
+          return promiseResolvedWith(true);
+        }
```
MattiasBuelens, line 208: This implements @domenic's suggestion from https://github.com/whatwg/streams/issues/1207#issuecomment-1015851512.

I don't know if we need to update the spec text for this. It already specifies that these checks *must* happen before performing *any* reads and writes:
> **Shutdown must stop activity:** if _shuttingDown_ becomes true, the user agent must not initiate further reads from _reader_, and must only perform writes of already-read chunks, as described below. In particular, the user agent must check the below conditions before performing any reads or writes, since they might lead to immediate shutdown.

We should still add a test for this particular case (although that might not be easy looking at the discussion in #1207).

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -228,7 +235,8 @@ function ReadableStreamPipeTo(source, dest, preventClose, preventAbort, preventC
     });
 
     // Errors must be propagated backward
-    isOrBecomesErrored(dest, writer._closedPromise, storedError => {
+    WritableStreamDefaultWriterIsOrBecomesErrored(writer, () => {
```
MattiasBuelens, line 238: This new helper allows attaching a *synchronous* callback for when *dest* becomes `"erroring"` or `"errored"`, following the discussion in https://github.com/whatwg/streams/issues/1207#issuecomment-1017947644. I added a test for this in https://github.com/web-platform-tests/wpt/commit/1646d657bf258f850afb5f22860ec02dccbc6b7a.

index.bs
```diff
@@ -2175,7 +2175,7 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
        |source|.[=ReadableStream/[[storedError]]=].
     1. Otherwise, [=shutdown=] with |source|.[=ReadableStream/[[storedError]]=].
    1. <strong>Errors must be propagated backward:</strong> if |dest|.[=WritableStream/[[state]]=]
-      is or becomes "`errored`", then
+      is or becomes "`erroring`" or "`errored`", then
```
MattiasBuelens, line 2178: I don't think it's useful to keep the pipe going when *dest* has already become `"erroring"`? Any new writes will just error immediately, as per [step 9 of WritableStreamDefaultWriterWrite](https://streams.spec.whatwg.org/commit-snapshots/908ceddfb813ec0d9c727393edfb13d810e694f1/#writable-stream-default-writer-write).

MattiasBuelens (commented):
I've updated the reference implementation to fix the issue from https://github.com/whatwg/streams/issues/1207#issuecomment-1017947644. I still need to update the spec text, although this might be tricky because the `pipeTo()` specification is a bit vague on *how* to implement the "if dest becomes errored" check. 😕

Also, is it okay to piggyback this fix onto this PR, or would you prefer I move it to a separate one? (The new fix does *depend* on the initial fix though, we can't land it entirely separately.)

index.bs
ricea: Agreed.

reference-implementation/lib/abstract-ops/readable-streams.js
ricea: I personally don't think we need to update the spec text.

reference-implementation/lib/abstract-ops/writable-streams.js
```diff
@@ -475,6 +483,20 @@ function WritableStreamDefaultWriterGetDesiredSize(writer) {
   return WritableStreamDefaultControllerGetDesiredSize(stream._controller);
 }
 
+function WritableStreamDefaultWriterIsOrBecomesErrored(writer, errorListener) {
```
ricea, line 486: Maybe since this is not part of the standard, it should start with a lower-case letter?

MattiasBuelens: Sure. 👍 

What do you suggest we *do* put in the spec text?
* We can add a note saying that "errors must be propagated backward" must be handled *synchronously* as soon as that condition becomes true, in other words that it's not enough to add an *asynchronous* callback to `writer.closed`.
* We can add a note in `WritableStreamStartErroring` below the "set `stream.[[state]]` to `"erroring"`" step to remind implementers that *this* is where that condition from `pipeTo()` can become true.
* Alternatively, we tell implementers to look at the reference implementation for an example on how to do it... 😛

MattiasBuelens:
I've generalized the implementation a bit. I added two new helpers:
* `readerAddStateChangeListener(reader, stateChangeListener)`
  * Calls `stateChangeListener` the next time `reader._stream._state` changes from `"readable"` to either `"closed"` or `"errored"`.
* `writerAddStateChangeListener(writer, stateChangeListener)`
  * Calls `stateChangeListener` the next time `writer._stream._state` changes from `"writable"` to either `"closed"` or `"erroring"`.

The `pipeTo()` reference implementation now uses these helpers instead of `reader._closedPromise` and `writer._closedPromise` to detect state changes. A `checkState()` helper checks the shutdown conditions when the pipe starts, and when the state of one of the streams changes.

This should handle all possible synchronous state transitions, and check the shutdown conditions in the correct order as specified by the standard's text.

(Perhaps we want to also use these helpers for `tee()`, and possibly eliminate the extra microtask from *readRequest*'s chunk steps there? I'll leave that for another time though. 😛)

index.bs
```diff
@@ -2210,6 +2221,10 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
     ask to shutdown, optionally with an error |error|, then:
    1. If |shuttingDown| is true, abort these substeps.
    1. Set |shuttingDown| to true.
+   1. If |reader| [=implements=] {{ReadableStreamBYOBReader}}, perform
```
domenic, line 2224: Reader will always be a default reader at this point right?

reference-implementation/lib/abstract-ops/writable-streams.js
domenic: Is this very specific to the one change from "writable" to "erroring"? In the reference implementation you've added a much more general listener setup.

I would suggest a general note saying "for all the 'becomes' conditions in the above, they must be processed synchronously as part of the [[state]] update, before any other web developer code can run." And then, if we anticipate that only being impactful in the one transition, we could append the extra note: "NOTE: Currently this requirement only has observable consequences for [the transition for writable stream states from from "writable" to "erroring"], and others could be done as asynchronous listeners". Or, if we think we might expand this listener usage in the future, then we should probably omit that note.

index.bs
MattiasBuelens: Not here, this is the *start* of the shutdown process.

But indeed, at the *end* of the shutdown process (i.e. "finalize"), *reader* will always be a default reader, since we would have gone through either "shutdown" or "shutdown with an action".

reference-implementation/lib/abstract-ops/writable-streams.js
MattiasBuelens: > Is this very specific to the one change from "writable" to "erroring"?

That's the most noticeable case, since it determines whether or not we may drop a chunk (i.e. we accidentally read a chunk that we can no longer write).

I'm not sure whether it matters for the state transitions of the readable end. There might be an edge case where *two* shutdown conditions become true at the same time, and then it matters which condition is handled first. For example:
```javascript
readableController.error(error1); // pipeTo() should immediately call writer.abort(error1)
writableController.error(error2); // should be ignored, since writable is already erroring
// => pipeTo() rejects with error1
```
versus:
```javascript
writableController.error(error2); // pipeTo() should immediately call reader.cancel(error2)
readableController.error(error1); // should be ignored, since readable is already closed
// => pipeTo() rejects with error2
```
If we were to use a synchronous reaction for the `writable -> erroring` transition but an *asynchronous* reaction for the `readable -> errored` transition, then the first snippet would *also* behave like the second one... I think. 😛

I'll try to whip up some more WPTs to double check.

index.bs
MattiasBuelens: Hmm, it would conflict with this note in "shutdown with an action" though:
> This exchange of readers is not observable to author code and the user agent is free to implement this differently, for example by keeping the same reader and internally aborting its pending read requests. 

🤔

domenic: Eh, I think it's clear enough that if you're doing something different from the spec you need to tweak more than just one step.

reference-implementation/lib/abstract-ops/writable-streams.js
MattiasBuelens: Here's two possible tests for this:
```javascript
promise_test(async t => {

  const pipeToPromise = rs.pipeTo(ws);
  await flushAsyncEvents();

  rs.controller.error(error1);
  ws.controller.error(error2);

  await promise_rejects_exactly(t, error1, pipeToPromise, 'pipeTo must reject with readable\'s error');
  assert_array_equals(rs.eventsWithoutPulls, []);
  assert_array_equals(ws.events, []);

  await promise_rejects_exactly(t, error1, rs.getReader().closed);
  await promise_rejects_exactly(t, error2, ws.getWriter().closed);
}, 'Piping: error the readable stream right before erroring the writable stream');

promise_test(async t => {

  const pipeToPromise = rs.pipeTo(ws);
  await flushAsyncEvents();

  ws.controller.error(error1);
  rs.controller.error(error2);

  await promise_rejects_exactly(t, error1, pipeToPromise, 'pipeTo must reject with writable\'s error');
  assert_array_equals(rs.eventsWithoutPulls, ['cancel', error1]);
  assert_array_equals(ws.events, []);

  await promise_rejects_exactly(t, error1, ws.getWriter().closed);
  await rs.getReader().closed;
}, 'Piping: error the writable stream right before erroring the readable stream');
```
The behavior might be a bit surprising though. In the first test, *ws* is still writable when we call `rs.controller.error()`, so we end up in:
https://github.com/whatwg/streams/blob/5afb0e014c08be94e49db7c0305aa211545e3add/reference-implementation/lib/abstract-ops/readable-streams.js#L294

This adds at least one microtask of delay (even if there are no pending writes), so we will *not yet* call `ws.abort(error1)`. Instead, `ws.controller.error(error2)` goes through, and the abort gets ignored later on.

However, in the second test, because *ws* immediately becomes errored, we *don't* wait for pending writes to complete and instead we *synchronously* call `rs.cancel(error1)`. Therefore, `rs.controller.error(error2)` gets ignored, and the stream ends up *cancelled* instead of *errored*.

---

The specification is a bit vague about this. It says:
> Wait until every chunk that has been read has been written (i.e. the corresponding promises have settled).

It doesn't say *how long* this step can take. We may want to require that if there are *no* pending writes (i.e. we've never started any writes, or all writes have already settled), then this step must complete *synchronously*. Then, in the first test, we *would* call `ws.abort(error1)` synchronously and prevent `ws.controller.error(error2)`. However, that might be tricky to actually implement correctly... 🤔

MattiasBuelens: Oh boy, this is a can of worms... 😅

Making `waitForWritesToFinish()` possibly synchronous (https://github.com/MattiasBuelens/streams/commit/7eef3d2ccd447f674b2b2f82b4651201a91cd6d0) allows weird things to happen:
1. `pipeTo()` calls `ReadableStreamDefaultReaderRead()`, reading the *last* chunk of the stream.
2. `ReadableStreamDefaultController[[PullSteps]]` dequeues the chunk, and notices that the queue is now empty *and* close was requested. So it calls `ReadableStreamClose()`.
3. `ReadableStreamClose` changes `[[state]]` to `"closed"`, which fires `pipeTo()`'s state change listener.
4. `pipeTo()` starts the shutdown process. There's no pending write, so it immediately finalizes the pipe and releases both the reader and the writer.
5. The call stack unwinds up to `ReadableStreamDefaultController[[PullSteps]]`. We continue with `readRequest.chunkSteps()`.
6. `chunkSteps()` brings us back inside `pipeTo()`. It now wants to start a new write for this chunk, *but we've already released the writer.* 💥

This wasn't a problem before, since even if we never actually started a write, `currentWrite` would still be initialized to `Promise.resolve(undefined)`. `waitForWritesToFinish()` would always have to wait at least one microtask to detect that this is indeed the last write, which is enough time for `chunkSteps()` to *actually* start a new write on time.

The root of the problem is that the state change listener is called *in the middle* of `ReadableStreamDefaultReaderRead()`, before we call `chunkSteps()`.
* We could try to only call the state change listeners *at the end* of all "top-level" abstract ops. That is, instead of calling them within the internal `ReadableStreamClose()`, we call them at the end of `ReadableStreamDefaultReaderRead()` etc. However, this could quickly become unmaintainable. 😕 
* We could have `pipeTo()` "remember" that it still has a pending read, and *wait* until either its `chunkSteps`, `closeSteps` or `errorSteps` have been called before finalizing the pipe. But then I'm not sure if `waitForWritesToFinish()` would still be synchronous in the cases where it matters... 🤔

ricea: I suggest limiting the sync part to as small as possible to fix the issue. This still leaves the problem of how to spec it. We've tried to give latitude for implementations to optimise in their own way, but we're increasingly constraining their behaviour. Transparent thread offloading etc. may become impossible. I'm worried about it but I don't have an answer.

MattiasBuelens: > I suggest limiting the sync part to as small as possible to fix the issue.

The sync part is already minimal. We have to go from "if source becomes errored" all the way to "perform WritableStreamAbort" in order to avoid `ws.controller.error()` from affecting the result. Thus, the entirety of [step 3 in "shutdown with an action"](https://streams.spec.whatwg.org/commit-snapshots/5afb0e014c08be94e49db7c0305aa211545e3add/#rs-pipeTo-shutdown-with-action) must become synchronous (only if there are no pending writes that need to be awaited).

---

Anyway, I found another way to fix it. We keep track of how many `ReadableStreamDefaultReaderRead()` requests are still outstanding, and we *only* handle the `source.[[state]] == "closed"` transition after all those requests are settled. See https://github.com/MattiasBuelens/streams/commit/3c8b3c28f37c99bf49a58db89ff3baaff28aa503.

However, [this test](https://github.com/web-platform-tests/wpt/blob/99d74f9529e16ec0722ef11136ab29b9e80fff26/streams/piping/abort.any.js#L237) still fails. We do call `dest.abort()` and `source.cancel()` in the correct order, but it seems like `underlyingSink.abort()` and `underlyingSource.cancel()` are being called in the wrong order. When we call [WritableStreamStartErroring](https://github.com/whatwg/streams/blob/main/reference-implementation/lib/abstract-ops/writable-streams.js#L383), the writable controller is not yet started, so we postpone calling `sink.abort()` until after `sink.start()` resolves. Previously, the writable would already have been started while we were *asynchronously* [waiting for the writes to finish](https://github.com/whatwg/streams/blob/5afb0e014c08be94e49db7c0305aa211545e3add/reference-implementation/lib/abstract-ops/readable-streams.js#L294) (even if there were no pending writes).

Adding `await flushAsyncEvents()` before calling `pipeTo()` in that test restores the order and fixes the problem. Good enough? 🤷‍♂️

---

> We've tried to give latitude for implementations to optimise in their own way, but we're increasingly constraining their behaviour. Transparent thread offloading etc. may become impossible. I'm worried about it but I don't have an answer.

I agree, the reference implementation is becoming increasingly complicated in order to deal with these edge cases. I'm wondering if it's even *worth* trying to spec these edge cases, or instead allow some wiggle room in how quickly `pipeTo()` must respond to these state transitions. But then it would become impossible to test the behavior, or we'd have to allow *multiple* valid outcomes... 😕

domenic: I suspect most of the testable constraints we're imposing are in cases where the web developer controls one or both ends of the pipe, right? I'm not sure those are the ones we were planning to feasibly optimize, so starting to constrain them still seems like the right thing to do to me. But I might be missing something so please let me know.

On the larger problem, the root of the issue seems to be how imprecise "[[state]] is or becomes" is. Does that mean: (1) synchronously after the algorithm step which sets [[state]], probably interrupting other streams-implementation code; (2) synchronously after any streams-implementation code runs; (3) synchronously after any browser code runs; (4) asynchronously is OK to some degree?

My preference would be to try to resolve things like so:

- Decide whether we're OK constraining all observable behavior, or want to allow leeway. In particular when one or both ends of a pipe are web-developer-created streams, a good bit more becomes observable.
- Write tests reflecting the result of the previous decision. E.g. if we want to nail down all observable behavior, I think @MattiasBuelens has done a great job capturing as many scenarios as possible. (❤️!) We should analyze them for what reasonable expected behavior is, and assert that. (If we don't have strong feelings on reasonable expected behavior, then we can feel free to change the assertions in the next step.)
- Come up with some more-rigorous formulation of "[[state]] is or becomes" for the spec which meets the expectations of those tests. This probably will make the spec more complex, and more like the reference implementation. E.g. it could be adding promise handlers (probably in combination with something like https://github.com/MattiasBuelens/streams/commit/3c8b3c28f37c99bf49a58db89ff3baaff28aa503), or having separate synchronous state-change steps. Given that this will only be used for pipeTo, we can probably consider spec strategies that aren't as detailed and algorithmic as the rest of the spec, but they do need to be clear and unambigious between the (1)-(4) above.

As an example of how to apply this process,

> Adding await flushAsyncEvents() before calling pipeTo() in that test restores the order and fixes the problem. Good enough? 🤷‍♂️

My preference would be that, if we decide to constrain all observable behavior, we have both variants of the test, with the version without flushAsyncEvents() having the assert for the other order.

MattiasBuelens: > I suspect most of the testable constraints we're imposing are in cases where the web developer controls one or both ends of the pipe, right? I'm not sure those are the ones we were planning to feasibly optimize, so starting to constrain them still seems like the right thing to do to me. But I might be missing something so please let me know.

Correct. Streams created by the user agent will use [the exported algorithms](https://streams.spec.whatwg.org/#other-specs), and I think it's safe to assume that those will be called in a separate task, outside of web author code.

> On the larger problem, the root of the issue seems to be how imprecise "[[state]] is or becomes" is. Does that mean: (1) synchronously after the algorithm step which sets [[state]], probably interrupting other streams-implementation code; (2) synchronously after any streams-implementation code runs; (3) synchronously after any browser code runs; (4) asynchronously is OK to some degree?

(2) may be ill-specified, since there are cases where streams code calls into author code, which can then call back into streams code. We've even had cases in the past where streams code calls back *into itself*, e.g. #1172.

I still prefer (1), and that's what I've been implementing. Yes, we need to be very careful when speccing, but at least any problems that arise can be fixed *within the streams implementation*.

> My preference would be that, if we decide to constrain all observable behavior, we have both variants of the test, with the version without flushAsyncEvents() having the assert for the other order.

That seems reasonable. 👍

MattiasBuelens: Coming back to this:

> Adding `await flushAsyncEvents()` before calling `pipeTo()` in that test restores the order and fixes the problem. Good enough? 🤷‍♂️

It seems the main difference is that, when you call `readable.cancel()`, we *always* synchronously call `source.cancel()` regardless of whether `source.start()` has already settled. On the other hand, when you call `writable.abort()`, we first wait for `sink.start()` to settle before we call `sink.abort()`.

IIRC the reason for this difference is so you can do e.g. an async loop in `source.start()`:
```javascript
new ReadableStream({
  async start(c) {
    for (let i = 0; i < 10; i++) {
      await new Promise(r => setTimeout(r, 1000));
      c.enqueue("chunk");
    }
    c.close();
  }
})
```
whereas for `sink.start()` this doesn't make sense.

I guess, if we *really* wanted to, we could have the test check when `writableController.signal` becomes aborted? That should happen synchronously *regardless* of whether `sink.start()` has already settled.

Issue #1224, opened 2022-03-03
Editorial: update use of, or stop using, completion values?


domenic:
https://github.com/tc39/ecma262/pull/2547 has changed how completion records work. Now, if an operation is infallible, the recommended practice is have it not return a completion record. Also, everything has a header saying what it returns.

We could update to follow, similar to https://github.com/whatwg/html/pull/7661 .

Or, we could try to further harmonize with the rest of the web spec ecosystem, and stop doing completion records entirely. (Except where necessary to interface with ECMAScript.) I think I'd kind of prefer that. Concretely, we'd:

- Move away from ECMAScript abstract ops as much as possible, e.g. using the stuff introduced in https://github.com/whatwg/webidl/pull/987
- Get rid of all ? and !s for internal abstract op calls
- Change all explicit completion processing, e.g. instead of "If result is an abrupt completion, return [a promise rejected with](https://webidl.spec.whatwg.org/#a-promise-rejected-with) result.[[Value]].", say "If this throws an exception, catch it and return a promise rejected with that exception."
- Update any remaining use of ECMAScript abstract ops to appropriately use or not use ?/!

ricea:
I like the `?` and `!` but there's no doubt they're super confusing to people coming to the spec for the first time.

Would we have to explicitly write "Assert: This does not throw an exception" everywhere?

domenic:
> Would we have to explicitly write "Assert: This does not throw an exception" everywhere?

We wouldn't have to. We could if it's helpful. I'd suggest only doing it in cases where the not-throwing is potentially surprising. (Which is of course subjective.)

This is related to [the question of whether/how we should introduce headers](https://github.com/whatwg/streams/pull/1225#issuecomment-1060810692) to the abstract operations, like ECMAScript now has.

If we keep !/? and just follow ECMAScript, then the headers could look something like:

> AcquireWritableStreamDefaultWriter(stream) returns either a [normal completion containing](https://tc39.es/ecma262/#sec-completion-record-specification-type) a `WritableStreamDefaultWriter` or an [abrupt completion](https://tc39.es/ecma262/#sec-completion-record-specification-type). It performs the following steps:

> IsReadableStreamLocked(stream) returns a boolean. It performs the following steps:

> ReadableStreamAddReadRequest(stream, readRequest) returns unused. It performs the following steps:

Or they could be omitted if we think they're not worthwhile.

[...]

PR #1225, opened 2022-03-04
Remove redundant ReturnIfAbrupt shorthands

saschanaz:
See #1224

   * …
   * …
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …
   * Deno: …
   * Node.js: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -813,13 +813,13 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
 <div algorithm>
  The <dfn id="rs-locked" attribute for="ReadableStream">locked</dfn> getter steps are:
 
- 1. Return ! [$IsReadableStreamLocked$]([=this=]).
+ 1. Return [$IsReadableStreamLocked$]([=this=]).
 </div>
 
 <div algorithm>
  The <dfn id="rs-cancel" method for="ReadableStream">cancel(|reason|)</dfn> method steps are:
 
- 1. If ! [$IsReadableStreamLocked$]([=this=]) is true, return [=a promise rejected with=] a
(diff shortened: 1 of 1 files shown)


saschanaz:
I'm doing this since I guess this shouldn't collide with https://github.com/whatwg/streams/issues/1224#issue-1157883735:

>Or, we could try to further harmonize with the rest of the web spec ecosystem, and stop doing completion records entirely. (Except where necessary to interface with ECMAScript.) I think I'd kind of prefer that.

Just want to make sure this is in the right direction since I'm not super familiar with ReturnIfAbrupt. Especially:

>Also, everything has a header saying what it returns.

Is this needed to do this cleanup?

@domenic

domenic:
Yes, I think this is in in the right direction. Basically:

- Find all never-fail operations in the streams spec, and remove all !s from calls to them.

- Remove all !s from calls to ECMAScript operations that do not return abstract completions anymore. I believe the following list is comprehensive: CopyDataBlockBytes, CreateArrayFromList, CreateBuiltinFunction, IsDetachedBuffer, IsIntegralNumber (we need to update we're using IsInteger), OrdinaryObjectCreate, SameValue, Type.

I think we should be sure to do all of these at once, as otherwise it's a bit confusing.

> > Also, everything has a header saying what it returns.
> 
> Is this needed to do this cleanup?

I'm not sure; I welcome your input or that of the other editors.

Without the header, the situation is a bit confusing, because you can only know whether an abstract op returns a completion or a bare value by inspecting the call sites. (I.e., if all call sites use ?/!, it's a completion. If all do not, it's a bare value. If it's a mix, that's a spec bug.) It seems relatively easy to get confused.

However, it might be worth thinking about the world post-#1224, to avoid wasted effort. I'll comment over there with some ideas.

PR #1250, opened 2022-12-12
Editorial: Add a non-promise version of "fully reading" a stream

noamr:
This makes it easier for non-JS consumers of streams, like the Fetch specs, to use streams without a JS
execution context.

See https://github.com/whatwg/fetch/issues/1568

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …

***
<a href="https://whatpr.org/streams/1250.html" title="Last updated on Dec 13, 2022, 7:26 AM UTC (52dc6af)">Preview</a> | <a href="https://whatpr.org/streams/1250/839a5a6...52dc6af.html" title="Last updated on Dec 13, 2022, 7:26 AM UTC (52dc6af)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6821,7 +6821,7 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
   1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
    : [=read request/chunk steps=], given |chunk|
    ::
-    1. If |chunk| is not a {{Uint8Array}} object, [=reject=] |promise| with a {{TypeError}} and
+    1. If |chunk| is not a {{Uint8Array}} object, call |failureSteps| with a {{TypeError}} and
        abort these steps.
     1. Append the bytes represented by |chunk| to |bytes|.
     1. [=Read-loop=] given |reader|, |bytes|, |successSteps|, and |failureSteps|.
--- a/index.bs
+++ b/index.bs
@@ -6816,7 +6816,7 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
 
  <div algorithm="read-loop">
   For the purposes of the above algorithm, to <dfn>read-loop</dfn> given |reader|, |bytes|,
-  |successSteps|, and |failureSteps:
+  |successSteps|, and |failureSteps|:
 
   1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
    : [=read request/chunk steps=], given |chunk|
--- a/index.bs
+++ b/index.bs
@@ -6809,7 +6809,7 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
 
 <div algorithm="read all bytes">
  <p>To <dfn export for="ReadableStreamDefaultReader" lt="read all bytes|reading all bytes">read all
- bytes as a promise</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given |successSteps|,
+ bytes</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given |successSteps|,
  which is an algorithm accepting a [=byte sequence=], and |failureSteps|, which is an algorithm
  accepting a JavaScript value: [=read-loop=] given |reader|, a new [=byte sequence=],
  |successSteps|, and |failureSteps|.
--- a/index.bs
+++ b/index.bs
@@ -6809,30 +6809,21 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
 
 <div algorithm="read all bytes">
  <p>To <dfn export for="ReadableStreamDefaultReader" lt="read all bytes|reading all bytes">read all
- bytes as a promise</dfn> from a {{ReadableStreamDefaultReader}} |reader|, perform the following
- steps. The result will be a {{Promise}} for a [=byte sequence=].
-
- 1. Let |promise| be [=a new promise=].
- 1. Let |success steps| given a [=byte sequence=] |data| be to [=resolve=] |promise| with |data|.
- 1. Let |failure steps| given a JavaScript value |error| be to [=reject=] |promise| with |error|.
- 1. [=Consume all bytes=] from |reader| given |success steps| and |failure steps|.
-
- <p>To <dfn export for="ReadableStreamDefaultReader">consume all bytes</dfn> from
- {{ReadableStreamDefaultReader}} |reader|, given |success steps|, which is an algorithm accepting a
- [=byte sequence=], and |failure steps|, which is an algorithm accepting a JavaScript value,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6809,36 +6809,41 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
+ bytes as a promise</dfn> from a {{ReadableStreamDefaultReader}} |reader|, perform the following
+ steps. The result will be a {{Promise}} for a [=byte sequence=].
 
  1. Let |promise| be [=a new promise=].
- 1. Let |bytes| be an empty [=byte sequence=].
- 1. [=Read-loop=] given |reader|, |bytes|, and |promise|.
- 1. Return |promise|.
+ 1. Let |success steps| given a [=byte sequence=] |data| be to [=resolve=] |promise| with |data|.
+ 1. Let |failure steps| given a JavaScript value |error| be to [=reject=] |promise| with |error|.
+ 1. [=Consume all bytes=] from |reader| given |success steps| and |failure steps|.
```
domenic, line 6818: I'd like to get rid of "read all bytes as a promise". Nobody should use it, for various reasons (e.g. having to deal with `then` tampering). https://dontcallmedom.github.io/webdex/r.html#read%20all%20bytes%40%40ReadableStreamDefaultReader%40dfn implies it wouldn't be so bad to break some consumers; I think we could fix Fetch ourselves, file issues on the rest, then move on.

index.bs
```diff
@@ -6809,36 +6809,41 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
+
+ <p>To <dfn export for="ReadableStreamDefaultReader">consume all bytes</dfn> from
+ {{ReadableStreamDefaultReader}} |reader|, given |success steps|, which is an algorithm accepting a
```
domenic, line 6821: successSteps and failureSteps, per more modern style, here and below.

index.bs
```diff
@@ -6809,36 +6809,41 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
+  For the purposes of the above algorithm, to <dfn>read-loop</dfn> given |reader|, |bytes|,
+  |success steps|, and |failure steps|:
 
   1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
    : [=read request/chunk steps=], given |chunk|
    ::
-    1. If |chunk| is not a {{Uint8Array}} object, [=reject=] |promise| with a {{TypeError}} and
+    1. If |chunk| is not a {{Uint8Array}} object, call |failure steps| with a {{TypeError}} and
        abort these steps.
     1. Append the bytes represented by |chunk| to |bytes|.
```
domenic, line 6834: While here, it might be best to use https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy to more rigorously convert |chunk| to a byte sequence.

domenic (commented):
This works surprisingly well! I think we should just be a bit more aggressive.

index.bs
noamr: sounds good, once this is reviewed I'll post a Fetch patch and issues to the other specs.

index.bs
noamr: Done

index.bs
noamr: Done

ricea:
No problem for Chrome, we don't do "full reading" at the JavaScript layer anyway. I think you can call us "interested" because it clears up any possible difference between our implementation and the spec.

noamr:
> No problem for Chrome, we don't do "full reading" at the JavaScript layer anyway. I think you can call us "interested" because it clears up any possible difference between our implementation and the spec.

I think this is purely editorial anyway..

index.bs
```diff
@@ -6809,36 +6809,32 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
 <div algorithm="read all bytes">
  <p>To <dfn export for="ReadableStreamDefaultReader" lt="read all bytes|reading all bytes">read all
- bytes</dfn> from a {{ReadableStreamDefaultReader}} |reader|, perform the following steps. The
- result will be a {{Promise}} for a [=byte sequence=].
-
- 1. Let |promise| be [=a new promise=].
- 1. Let |bytes| be an empty [=byte sequence=].
- 1. [=Read-loop=] given |reader|, |bytes|, and |promise|.
- 1. Return |promise|.
+ bytes as a promise</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given |successSteps|,
```
domenic, line 6812:
```suggestion
 bytes</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given |successSteps|,
```

index.bs
```diff
@@ -6809,36 +6809,32 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
+ which is an algorithm accepting a [=byte sequence=], and |failureSteps|, which is an algorithm
+ accepting a JavaScript value, [=read-loop=] given |reader|, a new [=byte sequence=],
```
domenic, line 6814:
```suggestion
 accepting a JavaScript value: [=read-loop=] given |reader|, a new [=byte sequence=],
```

index.bs
```diff
@@ -6809,36 +6809,32 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
+ |successSteps|, and |failureSteps|.
 
  <div algorithm="read-loop">
-  For the purposes of the above algorithm, to <dfn>read-loop</dfn> given |reader|, |bytes|, and
-  |promise|:
+  For the purposes of the above algorithm, to <dfn>read-loop</dfn> given |reader|, |bytes|,
+  |successSteps|, and |failureSteps:
```
domenic, line 6819:
```suggestion
  |successSteps|, and |failureSteps|:
```

index.bs
```diff
@@ -6809,36 +6809,32 @@ a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read re
...
 
   1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
    : [=read request/chunk steps=], given |chunk|
    ::
-    1. If |chunk| is not a {{Uint8Array}} object, [=reject=] |promise| with a {{TypeError}} and
```
domenic, line 6827: Hmm, I realized I made this more complicated because "get a copy of the bytes held" expects a BufferSource, but chunk steps get arbitrary values, which given the current state of the spec I'm not sure if they're even IDL versus JavaScript values.

Probably best to just revert this to the previous imprecise wording (including the type check and reject). Eventually we can do a pass through the spec to use the new-ish IDL BufferSource manipulation stuff everywhere. Sorry for thinking that this sub-case would be easier and thus worth doing while you were in the area.

noamr: No worries. Reverted

domenic (approved):

Issue #1243, opened 2022-10-14
Can pipeTo() be synchronous?


ricea:
Consider the following code:
```js
let synchronous;
new ReadableStream({
  pull(controller) {
    synchronous = true;
    controller.enqueue();
    synchronous = false;
  }
}).pipeTo(new WritableStream({
  write() {
    console.log(synchronous);
  }
});
```

1. Is this allowed to print `true`?
2. Should it be allowed to print `true`?
3. Is an implementation which prints `true` web-compatible?

domenic:
I discussed this a bit with Adam in person, but for the record, my opinions are:

1. Yes; per the current spec either true or false is allowed
2. Probably no; allowing browsers to differ in this seems too much of a sharp edge for interop. It's too easy for web developers to write code which assumes either sync or async, and only test in one browser, and get bad results in the other.
3. Unclear.

So my vote would be for ensuring that writes do not occur synchronously, and are at least delayed by a queued microtask.

ricea:
Okay. If there are no conflicting opinions then @nidhijaju or I will write a PR to prohibit synchronous behaviour.

saschanaz:
> Okay. If there are no conflicting opinions then @nidhijaju or I will write a PR to prohibit synchronous behaviour.

Has this happened? I found Gecko is behaving synchronously and intend to write WPT for this.

ricea:
Sorry, no, it slipped off my radar. Some WPT would be great!

saschanaz:
Another issue:

```js
await Promise.resolve().then(async () => {
  let synchronous = false;
  let readable = new ReadableStream({
    pull() {
      synchronous = true;
    }
  }, { highWaterMark: 0 });
  await new Promise(setTimeout); // make sure the stream is started
  readable.pipeTo(new WritableStream());
  console.log(synchronous);
});
```

This logs true on Gecko/Blink but false on WebKit. Maybe it's more natural to be synchronous here as `getReader().read()` pulls synchronously, except I'm not a fan of synchronous JS call...

ricea:
> Another issue:
>    ...
> This logs true on Gecko/Blink but false on WebKit. Maybe it's more natural to be synchronous here as `getReader().read()` pulls synchronously, except I'm not a fan of synchronous JS call...

Previously we considered this kind of thing "acceptable variation". If I recall correctly there was a similar difference between Blink's implementation and the reference implementation.

[...]

PR #1264, opened 2023-04-04
Editorial: update WPT

MattiasBuelens:
It looks like it's been a while since we synced the reference implementation with the latest WPT. This aims to fix that.

The following errors came up:
* `Error: Unexpected URL: /common/gc.js`: I opened https://github.com/domenic/wpt-runner/pull/27 to include this file.
* `structuredClone is not defined`: I expose this function in the test runner, and bumped the CI to Node 18.
* `enqueue() must not synchronously call write algorithm`: This test was added in https://github.com/web-platform-tests/wpt/pull/39103, but was not tested against the reference implementation. I'm not entirely sure what the intention was here: does the spec need to be updated for this? 😕

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,6 @@ jobs:
         submodules: true
     - uses: actions/setup-node@v1
       with:
-        node-version: 14
+        node-version: 18
     - run: npm install
     - run: npm test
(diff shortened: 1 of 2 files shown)


ricea:
> I'm not entirely sure what the intention was here: does the spec need to be updated for this? confused

Yes. This is #1243. We plan to update the spec to require a microtask between `enqueue()` and `write()`, but we haven't done it yet.

youennf:
> Yes. This is #1243. We plan to update the spec to require a microtask between `enqueue()` and `write()`, but we haven't done it yet.

Is #1243 easily resolvable?
Another approach would be to move this specific test in a separate file and disable it in this repo until #1243 can be resolved.

MattiasBuelens:
Closing, because superseded by #1282.

PR #1271, opened 2023-04-09
Add support for ReadableStream "transfer" type

youennf:
Implement part of streams-for-raw-video-explainer.md.

   * Chrome
   * Safari
   * Firefox
   * Not yet


index.bs
```diff
@@ -3065,6 +3084,9 @@ The following abstract operations support the implementation of the
  1. Let |startAlgorithm| be an algorithm that returns undefined.
  1. Let |pullAlgorithm| be an algorithm that returns [=a promise resolved with=] undefined.
  1. Let |cancelAlgorithm| be an algorithm that returns [=a promise resolved with=] undefined.
+ 1. Let |cancelAlgorithm| be an algorithm that returns [=a promise resolved with=] undefined.
```
ricea, line 3087: This line appears to be duplicated.

index.bs
```diff
@@ -3065,6 +3084,9 @@ The following abstract operations support the implementation of the
...
+ 1. Let |isTransferring| be true if |underlyingSourceDict|["{{UnderlyingSource/pull}}"] is
```
ricea, line 3088:
```suggestion
 1. Let |isTransferring| be true if |underlyingSourceDict|["{{UnderlyingSource/type}}"] is
```

index.bs
```diff
@@ -6596,6 +6626,18 @@ The following abstract operations are a grab-bag of utilities.
  1. Return ? [$StructuredDeserialize$](|serialized|, [=the current Realm=]).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="StructuredTransferOrSerialize">StructuredTransferOrSerialize(|value|)</dfn>
```
ricea, line 6630: Maybe "StructuredTransferOrClone" would be a clearer name?

ricea (commented):
General approach looks good. I made some early nitpicks.

We can bikeshed whether `type: 'transfer'` or a new optional flag `transfer: true` is better. `type: 'transfer'` makes it easier to extend TransformStream. `transfer: true` maintains the 1:1 relationship between types and controllers.

index.bs
```diff
@@ -6368,7 +6390,11 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
  1. Assert: |container| has \[[queue]] and \[[queueTotalSize]] internal slots.
  1. If ! [$IsNonNegativeNumber$](|size|) is false, throw a {{RangeError}} exception.
  1. If |size| is +∞, throw a {{RangeError}} exception.
- 1. [=list/Append=] a new [=value-with-size=] with [=value-with-size/value=] |value| and
+ 1. Let |enqueuedValue| be |value|.
+ 1. If |container|.\[[isTransferring]] is true, perform the following steps:
```
MattiasBuelens, line 6394: `EnqueueValueWithSize` is also used by `WritableStreamDefaultControllerWrite` and `WritableStreamDefaultControllerClose`, but `WritableStreamDefaultController` does not have an `[[isTransferring]]` slot.

I think we should move the transferring logic out of `EnqueueValueWithSize`, and instead put it at the call site in `ReadableStreamDefaultControllerEnqueue`.

index.bs
```diff
@@ -6388,6 +6414,10 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
  performs the following steps:
 
  1. Assert: |container| has \[[queue]] and \[[queueTotalSize]] internal slots.
+ 1. If |container|.\[[isTransferring]] is true, perform the following steps until |container|.\[[queue]]
```
MattiasBuelens, line 6417: Same problem as above.

Maybe add a `ReadableStreamDefaultControllerResetQueue` abstract op which "wraps" `ResetQueue`?

index.bs
```diff
@@ -6596,6 +6626,18 @@ The following abstract operations are a grab-bag of utilities.
...
+<div algorithm>
+ <dfn abstract-op lt="StructuredTransferOrSerialize">StructuredTransferOrSerialize(|value|)</dfn>
+    performs the following steps:
+ 1. If |value| is [=transferable objects|transferable=], perform the following steps:
+    // FIXME: We should probably handle other transferable types like ArrayBuffers.
+    1. Let |serialized| be ! [$StructuredSerializeWithTransfer$](|value|, « |value| »).
+    1. Let |deserialized| be ! [$StructuredDeserializeWithTransfer$](|serialized|,
+       [=the current Realm=]).
+    1. return |deserialized|.\[[Deserialized]].
+ 1. return [$StructuredClone$](|value|).
```
MattiasBuelens, line 6638:
```suggestion
 1. Return [$StructuredClone$](|value|).
```

index.bs
youennf, line 3087:
```suggestion
```

youennf: ok

index.bs
youennf: If no slot, then my understanding is that the value is undefined, which is different from true.
That said, the end plan is to use transfer for writable streams as well, which would probably allow reusing the same 1EnqueueValueWithSize`.

index.bs
youennf: ok

youennf:
> We can bikeshed whether type: 'transfer' or a new optional flag transfer: true is better.

I went with type: 'transfer' as transfer does not make sense for byob streams.
I guess a separate flag might be useful if we anticipate other types where transfer:true would be useful.

youennf:
A few thoughts:
- The current algorithm tries to transfer/serialize only if it is sure to be able to enqueue. This seems a good property to me.
- If transfer/serialization fails, we silently terminate without either throwing or erroring the stream. I wonder whether this will be surprising to web developers.
- Cloning a `transfer:true` stream with transfer-only objects would succeed but the second branch would be errored. This seems ok to me.
- The current transfer-or-clone algorithm has surprising and potentially undesired effects:
  - `enqueue(videoFrame)` would make sure to transfer videoFrame but `enqueue({ videoFrame })` would clone it.
  - We could introduce a second optional parameter to controller.enqueue, something like `enqueue(any message, sequence<object> transfer = [])` and update accordingly the `StructuredTransferOrClone` algorithm. This would make it closer to postMessage. In that case, should we keep `enqueue(videoFrame)` as a shortcut to `enqueue(videoFrame, [videoFrame])`?

ricea:
> I went with type: 'transfer' as transfer does not make sense for byob streams.

That makes sense to me.

> I guess a separate flag might be useful if we anticipate other types where transfer:true would be useful.

I'm hoping we won't have to add any more controller types, since they are costly in terms of specification and implementation size.

ricea:
> * The current algorithm tries to transfer/serialize only if it is sure to be able to enqueue. This seems a good property to me.

Agreed.

> * If transfer/serialization fails, we silently terminate without either throwing or erroring the stream. I wonder whether this will be surprising to web developers.

I would lean towards erroring the stream. Silently losing data is something I want to avoid.

> * Cloning a `transfer:true` stream with transfer-only objects would succeed but the second branch would be errored. This seems ok to me.

It makes me a bit uncomfortable but I think it's acceptable.

> * The current transfer-or-clone algorithm has surprising and potentially undesired effects:
>   
>   * `enqueue(videoFrame)` would make sure to transfer videoFrame but `enqueue({ videoFrame })` would clone it.
>   * We could introduce a second optional parameter to controller.enqueue, something like `enqueue(any message, sequence<object> transfer = [])` and update accordingly the `StructuredTransferOrClone` algorithm. This would make it closer to postMessage. In that case, should we keep `enqueue(videoFrame)` as a shortcut to `enqueue(videoFrame, [videoFrame])`?

If we add two-argument `enqueue()` we should also add two-argument `write()`. I would like the two-argument form to work even when not using `type: transfer` as it is extremely useful in combination with transferable streams. When I was thinking about this previously I was concerned about the extra complexity it adds to `pipeTo()`, but maybe it's really not that bad now that we use `ReadRequest` internally.

index.bs
youennf: We will probably want to reuse the same reset queue algorithm for writable stream default controller.
That said, ResetQueue is also used for Byob. Given there will be no isTransferring slot, this should have no impact. But I wonder whether this is in line with this spec style.

youennf:
> I would lean towards erroring the stream. Silently losing data is something I want to avoid.

Done.

> If we add two-argument enqueue() we should also add two-argument write()

Yes, I plan to do a separate PR for WritableStream.
And probably another for transform stream.

> I would like the two-argument form to work even when not using type: transfer as it is extremely useful in combination with transferable streams. 

I did not add this in this PR but it should be fairly easy to extend it as a follow-up.
I am a bit hesitant to do this at this point:
- Adding a transfer list would trigger cloning of the data (observable in non transferable streams case)
- ReadableStreamTee might fail with those transfer-only values.
- Adding `transfer:true` explicitly makes it clear that the web page opts into that behavior. It is not a big adoption step, filling the transferList is anyway needed.

index.bs
youennf: I updated to `if |container| has a \[[isTransferring]] slot whose value is true` to make that clearer

index.bs
youennf: I updated to `if |container| has a \[[isTransferring]] slot whose value is true` to make that clearer

youennf:
I think the current version of the PR is ready for reference implementation prototyping.
Should I update the reference implementation in this PR as well?

ricea:
> I did not add this in this PR but it should be fairly easy to extend it as a follow-up. I am a bit hesitant to do this at this point:
> 
> * Adding a transfer list would trigger cloning of the data (observable in non transferable streams case)
> * ReadableStreamTee might fail with those transfer-only values.
> * Adding `transfer:true` explicitly makes it clear that the web page opts into that behavior. It is not a big adoption step, filling the transferList is anyway needed.

These are good points. Let's do it your way.

MattiasBuelens:
> * We could introduce a second optional parameter to controller.enqueue, something like `enqueue(any message, sequence<object> transfer = [])` and update accordingly the `StructuredTransferOrClone` algorithm. This would make it closer to postMessage.

I like that a lot! 🙂 I understand that it's a bit out-of-scope for this PR, but I'd love to see that in a follow-up PR.

When we do add this, I would suggest we use an options object though, to align with [`postMessage(message, { transfer })`](https://html.spec.whatwg.org/multipage/web-messaging.html#dom-messageport-postmessage-dev). That makes it easier to add extra options for `enqueue()` and `write()` in the future.

`ResetQueue` would then transfer away ownership of all transfer lists still in the queue, without needing a separate `[[isTransferring]]` flag. In essence, `type: "transfer"` would merely force `enqueue(chunk)` to always be handled as `enqueue(chunk, { transfer: [chunk] })`.

> I think the current version of the PR is ready for reference implementation prototyping.
> Should I update the reference implementation in this PR as well?

Yes, please. Each spec change on the Streams standard should be accompanied with new WPT tests, and the reference implementation should pass those new tests.

ricea:
> Should I update the reference implementation in this PR as well?

Yes, please do.

youennf:
OK, I'll start doing it.
Rethinking a bit, I wonder the following:
- Would `type:'serialize'` be a better name than `type:'transfer'` (which conflicts a bit with transform streams, transferable streams and we sometimes serialize and not transfer chunks)?
- Is the syntactic sugar for transferable objects (`enqueue(videoFrame)` instead of `enqueue(videoFrame, [videoFrame])`) a good idea now that we have the additional `transfer` parameter?

youennf:
For testing, it seems we use node.js with WPT and the ref implementation.
Are there any transferable object like MessagePort supported in node.js?
Ditto for serializable, with say VideoFrame?

ricea:
> OK, I'll start doing it. Rethinking a bit, I wonder the following:
> 
> * Would `type:'serialize'` be a better name than `type:'transfer'` (which conflicts a bit with transform streams, transferable streams and we sometimes serialize and not transfer chunks)?

This maybe needs a bit more bikeshedding. Another alternative I can think of is `type: 'owning'`.

> * Is the syntactic sugar for transferable objects (`enqueue(videoFrame)` instead of `enqueue(videoFrame, [videoFrame])`) a good idea now that we have the additional `transfer` parameter?

I really like the syntactic sugar personally, and I think being a different `type` of stream is a good enough excuse to use it. I'd like to get other people's input on this because there may be footguns I'm overlooking.

domenic:
> I really like the syntactic sugar personally, and I think being a different `type` of stream is a good enough excuse to use it. I'd like to get other people's input on this because there may be footguns I'm overlooking.

I would be most comfortable with requiring `enqueue(videoFrame, { transfer: [videoFrame] })`. That is the most like other web platform APIs that do transferring, such as `structuredClone()` and `postMessage()`. (Note: the `{ transfer: [x] }` form is the modern version that replaces `[x]`, so indeed, @MattiasBuelens's suggestion in https://github.com/whatwg/streams/pull/1271#issuecomment-1504867016 is a good one.)

However, I recognize it is verbose. So we might want to blaze a new path here and have `enqueue()` be the first sort of "auto-transfer" API. We should think carefully about how that works, and what pattern we might be setting up for future such APIs.

> For testing, it seems we use node.js with WPT and the ref implementation.
> Are there any transferable object like MessagePort supported in node.js?
> Ditto for serializable, with say VideoFrame?

You might run up against the limits of the reference implementation here, for which I apologize. It appears that for transferrable streams https://github.com/whatwg/streams/pull/1053 we did not make any reference implementation changes and just skipped running the relevant tests against the reference implementation. But that was mostly fine because it didn't change the streams API itself. This change is more intrusive to the streams API so it would be a real shame to let the reference implementation get out of sync.

Your main tools for trying to tackle this are:

- Patching the `window` in which the reference implementation and test runs: https://github.com/whatwg/streams/blob/main/reference-implementation/run-web-platform-tests.js#L48 . In particular you could put things from the Node.js environment in here.
- Node.js does support `globalThis.structuredClone`: https://nodejs.org/api/all.html#all_globals_structuredclonevalue-options . I think that might be enough to implement what you need, with some creativity.
- ArrayBuffers are transferable, and plain JS objects are cloneable. So if you put those in the tests you write, it should work in any environment. You probably also want to add separate WPTs for VideoFrame specifically, since that's an important use case, but keep those in a separate file from the generic clone/transfer ones, so that we can exclude them when running reference implementation tests in Node.

youennf:
> Another alternative I can think of is `type: 'owning'`.

I'll change to this, it is less confusing than `transfer`.

> I would be most comfortable with requiring `enqueue(videoFrame, { transfer: [videoFrame] })`.

Yes, let's start with this and let's do the shortcut as a follow-up.

> * ReadableStreamTee might fail with those transfer-only values.

ReadableStreamTee would error both branches according the current algorithm.
Maybe we should relax this rule to only error the second branch. Let's also look at this in a follow-up.

youennf:
PR is ready for review.
If moving forward, this could be completed by the following tasks:
- Add WritableStream & TransformStream support
- Investigate shorter enqueue-with-transfer syntax
- Investigate transferring chunks when transferring owning streams
- Investigate ReadableStreamTee to only error clone 2 branch if structure cloning.
- Investigate ReadableStreamTee realtime mode

index.bs
```diff
@@ -6596,6 +6629,14 @@ The following abstract operations are a grab-bag of utilities.
  1. Return ? [$StructuredDeserialize$](|serialized|, [=the current Realm=]).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="StructuredTransferOrClone">StructuredTransferOrClone(|value|, |transferList|)</dfn>
+    performs the following steps:
+ 1. Let |serialized| be ! [$StructuredSerializeWithTransfer$](|value|, |transferList|).
+ 1. Let |deserialized| be ! [$StructuredDeserializeWithTransfer$](|serialized|, [=the current Realm=]).
+ 1. return |deserialized|.\[[Deserialized]].
```
ricea, line 6637:
```suggestion
 1. Return |deserialized|.\[[Deserialized]].
```

index.bs
```diff
@@ -2950,13 +2966,19 @@ The following abstract operations support the implementation of the
 <div algorithm>
  <dfn abstract-op lt="ReadableStreamDefaultControllerEnqueue"
  id="readable-stream-default-controller-enqueue">ReadableStreamDefaultControllerEnqueue(|controller|,
- |chunk|)</dfn> performs the following steps:
+ |chunk|, |transferList|)</dfn> performs the following steps:
 
  1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|controller|) is false, return.
```
ricea, line 2971: If `isOwning` is false, then we won't use `transferList`, so we should throw an exception if it is set to let the user know that they're holding it wrong.

ricea (commented):
Looking good. I want this with both my implementer and spec editor hats on.

I think we need at least the pipeTo changes before we can land this, which means we need the WritableStream changes too.

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1169,6 +1177,8 @@ function SetUpReadableStreamDefaultController(
   controller._pullAlgorithm = pullAlgorithm;
   controller._cancelAlgorithm = cancelAlgorithm;
 
+  controller.isOwning = isOwning;
```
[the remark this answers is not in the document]
youennf: OK, will change

index.bs
youennf: There is a theoretical backward compatibility issue, is the potential extra clarity gain worth it?

ricea: Do you mean if existing code calls `controller.enqueue(chunk, { transfer: [ chunk ] })` it will suddenly start throwing an exception? That seems unlikely enough that I wouldn't worry about it. Even if someone was doing that, their code was arguably wrong since the second argument was never doing anything.

youennf: I guess there is already a backward compatibility issue with just introducing the second parameter as a `StructuredSerializeOptions` instead of `any`.

youennf:
> I think we need at least the pipeTo changes before we can land this, which means we need the WritableStream changes too.

https://github.com/whatwg/streams/pull/1272 adds WritableStream support.
It includes guidelines to handle pipeTo read-write in case of the four isOwning cases.
It also includes a draft reference implementation of pipeTo (not tested) that handles dropped chunks and the optimized ownership transfer in case of pipeTo between an owning ReadableStream and an owning WritableStream.

ricea:
Sorry I didn't get to this today. I will do a proper review tomorrow.

index.bs
```diff
@@ -2950,13 +2968,19 @@ The following abstract operations support the implementation of the
...
```
ricea, line 2973: Generally when the public API has a precondition, we put an assert in the private API to enforce the precondition (to provide a hint to implementers and other specs). Something like
> Assert: |transferList| is empty or |controller|.[[isOwning]] is true.

index.bs
```diff
@@ -6362,13 +6389,17 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
- id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|)</dfn> performs the
+ id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|, |transferList|)</dfn> performs the
  following steps:
 
  1. Assert: |container| has \[[queue]] and \[[queueTotalSize]] internal slots.
  1. If ! [$IsNonNegativeNumber$](|size|) is false, throw a {{RangeError}} exception.
  1. If |size| is +∞, throw a {{RangeError}} exception.
- 1. [=list/Append=] a new [=value-with-size=] with [=value-with-size/value=] |value| and
+ 1. Let |enqueuedValue| be |value|.
+ 1. If |container| has a \[[isOwning]] internal slot whose value is true, perform the following steps:
```
ricea, line 6399:
```suggestion
 1. If |container| has an \[[isOwning]] internal slot whose value is true, perform the following steps:
```

index.bs
```diff
@@ -6596,6 +6631,14 @@ The following abstract operations are a grab-bag of utilities.
  1. Return ? [$StructuredDeserialize$](|serialized|, [=the current Realm=]).
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="StructuredTransferOrClone">StructuredTransferOrClone(|value|, |transferList|)</dfn>
+    performs the following steps:
+ 1. Let |serialized| be ! [$StructuredSerializeWithTransfer$](|value|, |transferList|).
```
ricea, line 6637: Both StructuredSerializeWithTransfer and StructuredDeserializeWithTransfer can throw, so they should have **?** in front of them, not **!**.

ricea (commented):
This is looking good.

I thought I saw you had updated the non-normative `domintro` description of `enqueue` to include the new parameter, but I can't see it now. If it's not there, could you add it? I'd put this comment on line 1557 but github won't let me.

We will need an example of using an owning stream (but no need to add one now).

index.bs
youennf: Done

index.bs
youennf: Added

reference-implementation/lib/abstract-ops/queue-with-sizes.js
```diff
@@ -40,6 +42,18 @@ exports.PeekQueueValue = container => {
 exports.ResetQueue = container => {
   assert('_queue' in container && '_queueTotalSize' in container);
 
+  if (container._isOwning) {
+    while (container._queue.length > 0) {
+      const value = exports.DequeueValue(container);
+      if (typeof value.close === 'function') {
```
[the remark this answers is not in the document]
youennf: I switched to a private symbol, it is anyway difficult to test.

index.bs
```diff
@@ -1569,11 +1583,14 @@ the following table:
 
 <div algorithm>
  The <dfn id="rs-default-controller-enqueue" method
- for="ReadableStreamDefaultController">enqueue(|chunk|)</dfn> method steps are:
+ for="ReadableStreamDefaultController">enqueue(|chunk|, |options|)</dfn> method steps are:
 
+ 1. Let |transferList| be |options|["transfer"].
+ 1. If |transferList| is not [=list/is empty|empty=] and [=this=].[=ReadableStreamDefaultController/[[isOwning]]=] is false
```
MattiasBuelens, line 1589:
```suggestion
 1. If |transferList| is not [=list/is empty|empty=] and [=this=].[=ReadableStreamDefaultController/[[isOwning]]=] is false,
```

index.bs
```diff
@@ -2287,10 +2306,10 @@ create them does not matter.
       1. Otherwise, set |chunk2| to |cloneResult|.\[[Value]].
      1. If |canceled1| is false, perform !
         [$ReadableStreamDefaultControllerEnqueue$](|branch1|.[=ReadableStream/[[controller]]=],
-        |chunk1|).
+        |chunk1|, undefined).
```
MattiasBuelens, line 2309: `ReadableStreamDefaultControllerEnqueue` takes a list as third argument, and does not handle `undefined`. (It would fail on the "Assert: *transferList* is empty" check.)

You need to pass an explicit empty list here:
```suggestion
        |chunk1|, « »).
```

index.bs
```diff
@@ -2287,10 +2306,10 @@ create them does not matter.
...
      1. If |canceled2| is false, perform !
         [$ReadableStreamDefaultControllerEnqueue$](|branch2|.[=ReadableStream/[[controller]]=],
-        |chunk2|).
+        |chunk2|, undefined).
```
MattiasBuelens, line 2312:
```suggestion
        |chunk2|, « »).
```

index.bs
```diff
@@ -2950,13 +2969,20 @@ The following abstract operations support the implementation of the
 <div algorithm>
  <dfn abstract-op lt="ReadableStreamDefaultControllerEnqueue"
  id="readable-stream-default-controller-enqueue">ReadableStreamDefaultControllerEnqueue(|controller|,
- |chunk|)</dfn> performs the following steps:
+ |chunk|, |transferList|)</dfn> performs the following steps:
 
+ 1. Assert: |transferList| is empty or |controller|.[=ReadableStreamDefaultController/[[isOwning]]=] is true.
```
MattiasBuelens, line 2974:
```suggestion
 1. Assert: |transferList| [=list/is empty=] or |controller|.[=ReadableStreamDefaultController/[[isOwning]]=] is true.
```

index.bs
```diff
@@ -2950,13 +2969,20 @@ The following abstract operations support the implementation of the
...
  1. If ! [$IsReadableStreamLocked$](|stream|) is true and !
-    [$ReadableStreamGetNumReadRequests$](|stream|) > 0, perform !
-    [$ReadableStreamFulfillReadRequest$](|stream|, |chunk|, false).
+    [$ReadableStreamGetNumReadRequests$](|stream|) > 0, perform the following steps:
+    1. Let |internalChunk| be |chunk|.
+    1. If |controller|.[=ReadableStreamDefaultController/[[isOwning]]=] is true, perform the following steps:
+       1. Set |internalChunk| to [$StructuredTransferOrClone$](|chunk|, |transferList|).
+       1. If |internalChunk| is an abrupt completion,
+          1. Perform ! [$ReadableStreamDefaultControllerError$](|controller|, |internalChunk|.\[[Value]]).
+          1. Return |internalChunk|.
```
MattiasBuelens, line 2984: `internalChunk` starts out as a regular JavaScript value (with value `chunk`), but then changes into a completion record. We then use this record as-is later on in `ReadableStreamFulfillReadRequest`, which is wrong.

We need to manually unwrap the completion:
```suggestion
       1. Let |result| be [$StructuredTransferOrClone$](|chunk|, |transferList|).
       1. If |result| is an abrupt completion,
          1. Perform ! [$ReadableStreamDefaultControllerError$](|controller|, |result|.\[[Value]]).
          1. Return |result|.
       1. Set |internalChunk| to |result|.\[[Value]].
```

index.bs
```diff
@@ -5884,7 +5913,7 @@ The following abstract operations support the implementaiton of the
  1. If ! [$ReadableStreamDefaultControllerCanCloseOrEnqueue$](|readableController|) is false, throw
     a {{TypeError}} exception.
  1. Let |enqueueResult| be [$ReadableStreamDefaultControllerEnqueue$](|readableController|,
-    |chunk|).
+    |chunk|, undefined).
```
MattiasBuelens, line 5916:
```suggestion
    |chunk|, « »).
```

index.bs
```diff
@@ -6362,13 +6391,17 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
- id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|)</dfn> performs the
+ id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|, |transferList|)</dfn> performs the
  following steps:
 
  1. Assert: |container| has \[[queue]] and \[[queueTotalSize]] internal slots.
  1. If ! [$IsNonNegativeNumber$](|size|) is false, throw a {{RangeError}} exception.
  1. If |size| is +∞, throw a {{RangeError}} exception.
- 1. [=list/Append=] a new [=value-with-size=] with [=value-with-size/value=] |value| and
+ 1. If |container| has an \[[isOwning]] internal slot whose value is true, perform the following steps:
+ 1. If |container| has a \[[isOwning]] internal slot whose value is true, perform the following steps:
```
MattiasBuelens, line 6401: Accidentally duplicated line. 😅
```suggestion
 1. If |container| has an \[[isOwning]] internal slot whose value is true, perform the following steps:
```

index.bs
```diff
@@ -6362,13 +6391,17 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
+    1. Set |enqueuedValue| to [$StructuredTransferOrClone$](|value|, |transferList|).
+    1. If |enqueuedValue| is an abrupt completion, return |enqueuedValue|.
```
MattiasBuelens, line 6403: Use `?` to unwrap the completion.
```suggestion
    1. Set |enqueuedValue| to ? [$StructuredTransferOrClone$](|value|, |transferList|).
```

index.bs
```diff
@@ -6447,7 +6484,7 @@ abstract operations are used to implement these "cross-realm transforms".
   1. Let |value| be ! [$Get$](|data|, "`value`").
   1. Assert: [$Type$](|type|) is String.
   1. If |type| is "`chunk`",
-   1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|controller|, |value|).
+   1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|controller|, |value|, undefined).
```
MattiasBuelens, line 6487:
```suggestion
   1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|controller|, |value|, « »).
```

reference-implementation/lib/ReadableStreamDefaultController-impl.js
```diff
@@ -17,12 +17,16 @@ exports.implementation = class ReadableStreamDefaultControllerImpl {
     aos.ReadableStreamDefaultControllerClose(this);
   }
 
-  enqueue(chunk) {
+  enqueue(chunk, options) {
+    const transferList = options ? options.transfer : undefined;
```
MattiasBuelens, line 21:
```suggestion
    const transferList = options ? options.transfer : [];
```
Otherwise `if (transferList.length)` will throw. 😉

reference-implementation/lib/abstract-ops/queue-with-sizes.js
```diff
@@ -37,9 +39,23 @@ exports.PeekQueueValue = container => {
   return pair.value;
 };
 
+const closingStepsSymbol = Symbol('closing-steps');
```
MattiasBuelens, line 42: This symbol isn't exposed anywhere, so I suppose we can't really write tests for this? 😕

MattiasBuelens (changes requested):

reference-implementation/lib/abstract-ops/queue-with-sizes.js
youennf: It is difficult to write tests except when not running closing steps would have a functional effect (this is the case for some VideoFrame sources).
We could add some specific tests, but they would not be WPT tests.

reference-implementation/lib/abstract-ops/transform-streams.js
```diff
@@ -155,7 +155,7 @@ function TransformStreamDefaultControllerEnqueue(controller, chunk) {
   // accept TransformStreamDefaultControllerEnqueue() calls.
 
   try {
-    ReadableStreamDefaultControllerEnqueue(readableController, chunk);
+    ReadableStreamDefaultControllerEnqueue(readableController, chunk, undefined);
```
MattiasBuelens, line 158:
```suggestion
    ReadableStreamDefaultControllerEnqueue(readableController, chunk, []);
```

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -392,10 +392,10 @@ function ReadableStreamDefaultTee(stream, cloneForBranch2) {
           // }
 
           if (canceled1 === false) {
-            ReadableStreamDefaultControllerEnqueue(branch1._controller, chunk1);
+            ReadableStreamDefaultControllerEnqueue(branch1._controller, chunk1, undefined);
```
MattiasBuelens, line 395:
```suggestion
            ReadableStreamDefaultControllerEnqueue(branch1._controller, chunk1, []);
```

reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -392,10 +392,10 @@ function ReadableStreamDefaultTee(stream, cloneForBranch2) {
...
           }
           if (canceled2 === false) {
-            ReadableStreamDefaultControllerEnqueue(branch2._controller, chunk2);
+            ReadableStreamDefaultControllerEnqueue(branch2._controller, chunk2, undefined);
```
MattiasBuelens, line 398:
```suggestion
            ReadableStreamDefaultControllerEnqueue(branch2._controller, chunk2, []);
```

MattiasBuelens (commented):
Apologies, still missed a few `undefined`s in the reference implementation.

youennf:
> Apologies, still missed a few undefineds in the reference implementation.

Ah thanks, I wanted to do that and forgot about it.

reference-implementation/lib/abstract-ops/queue-with-sizes.js
MattiasBuelens: Let's hope the [explicit resource management proposal](https://github.com/tc39/proposal-explicit-resource-management) advances quickly, so we can integrate with something like `Symbol.dispose` instead. 😁

youennf: It might be better to rename closing-steps to dispose-steps, I'll do this.

youennf:
@ricea, any thoughts? Is this good to merge?

ricea:
We need interest from one other browser besides Chrome. I should also review the tests before landing, but that's not strictly a blocker.

jan-ivar (commented):
This looks great to me! I'll ping some Mozilla folks to respond on the standards repo.

jan-ivar:
The examples in the [streams-for-raw-video-explainer.md](https://github.com/whatwg/streams/blob/main/streams-for-raw-video-explainer.md) seem outdated now. Can we update them? That would make wide review easier.

If I follow this PR right, we seem to have:
```js
  new ReadableStream({
    type: "owning",
    pull(controller) {
      const frame = new VideoFrame(offscreenCanvas);
      controller.enqueue(frame, {transfer: [frame]});
    }
  }
```
Did I get that right?

youennf:
> The examples in the [streams-for-raw-video-explainer.md](https://github.com/whatwg/streams/blob/main/streams-for-raw-video-explainer.md) seem outdated now. Can we update them?

Will do.

> Did I get that right?

Yes.
A follow-up may introduce a shorter syntax `controller.enqueue(frame});` but this is out of scope for this first version.

.github/workflows/test.yml
```diff
@@ -19,6 +19,6 @@ jobs:
         submodules: true
     - uses: actions/setup-node@v1
       with:
-        node-version: 14
+        node-version: 19
```
saschanaz, line 22: You sure you want 19, not 18 LTS?

.github/workflows/test.yml
```diff
@@ -19,6 +19,6 @@ jobs:
         submodules: true
     - uses: actions/setup-node@v1
```
saschanaz, line 20: A good chance to upgrade to setup-node@v3.

reference-implementation/lib/ReadableStreamDefaultController.webidl
```diff
@@ -1,8 +1,12 @@
...
+  sequence<object> transfer = [];
+};
+
 [Exposed=(Window,Worker,Worklet)]
 interface ReadableStreamDefaultController {
   readonly attribute unrestricted double? desiredSize;
 
   undefined close();
-  undefined enqueue(optional any chunk);
+  undefined enqueue(optional any chunk, optional StructuredSerializeOptions options = { });
```
saschanaz, line 10:
```suggestion
  undefined enqueue(optional any chunk, optional StructuredSerializeOptions options = {});
```

`{}` is more conventional.

index.bs
```diff
@@ -651,8 +651,16 @@ enum ReadableStreamType { "bytes" };
   <p>For an example of how to set up a readable byte stream, including using the different
   controller interface, see [[#example-rbs-push]].
 
-  <p>Setting any value other than "{{ReadableStreamType/bytes}}" or undefined will cause the
-  {{ReadableStream()}} constructor to throw an exception.
+  <p>Can be set to "<dfn enum-value for="ReadableStreamType">owning</dfn>" to signal that the
+  constructed {{ReadableStream}} will own chunks (via transfer or serialization) before enqueuing them.
+  This ensures that enqueued chunks are not mutable by the source.
+  Transferred or serialized chunks may have <dfn>dispose steps</dfn> which are executed if
```
saschanaz, line 657:
```suggestion
  Transferred or serialized chunks may have <dfn export>dispose steps</dfn> which are executed if
```

saschanaz: BTW, chunks should already have dispose steps before transfer/serialization, so this phrase looks a bit off. Can it be:

>Chunks may have dispose steps which are executed when they are owned by the stream via enqueuing and then dequeued without ...

index.bs
```diff
@@ -651,8 +651,16 @@ enum ReadableStreamType { "bytes" };
...
+  enqueued chunks are dequeued without being provided to the application, for instance when
+  a {{ReadableStream}} is errored.
+  </p>
+
+  <p>Setting any value other than "{{ReadableStreamType/bytes}}", "{{ReadableStreamType/owning}}"
+  or undefined will cause the {{ReadableStream()}} constructor to throw an exception.
```
saschanaz, line 663: Do we need this? This should be clear from the IDL.

index.bs
```diff
@@ -1461,7 +1470,7 @@ interface ReadableStreamDefaultController {
   readonly attribute unrestricted double? desiredSize;
 
   undefined close();
-  undefined enqueue(optional any chunk);
+  undefined enqueue(optional any chunk, optional StructuredSerializeOptions options = { });
```
saschanaz, line 1473:
```suggestion
  undefined enqueue(optional any chunk, optional StructuredSerializeOptions options = {});
```

index.bs
```diff
@@ -1541,9 +1554,10 @@ the following table:
   previously-enqueued [=chunks=] from the stream, but once those are read, the stream will become
   closed.
 
- <dt><code><var ignore>controller</var>.{{ReadableStreamDefaultController/enqueue()|enqueue}}(<var ignore>chunk</var>)</code>
+ <dt><code><var ignore>controller</var>.{{ReadableStreamDefaultController/enqueue()|enqueue}}(<var ignore>chunk</var>, <var ignore>options</var>)</code>
```
saschanaz, line 1557:
```suggestion
 <dt><code><var ignore>controller</var>.{{ReadableStreamDefaultController/enqueue()|enqueue}}(|chunk|, |options|)</code>
```

`<var ignore>` is only needed when it's not referenced, and here `chunk` and `options` have their references.

index.bs
```diff
@@ -1541,9 +1554,10 @@ the following table:
...
  <dd>
-  <p>Enqueues the given [=chunk=] <var ignore>chunk</var> in the controlled readable stream.
+  <p>Enqueues the given [=chunk=] <var ignore>chunk</var> in the controlled readable stream,
```
saschanaz, line 1559:
```suggestion
  <p>Enqueues the given [=chunk=] |chunk| in the controlled readable stream,
```

index.bs
```diff
@@ -1541,9 +1554,10 @@ the following table:
...
+    with <var ignore>options</var>.
```
saschanaz, line 1560:
```suggestion
    with |options|.
```

index.bs
```diff
@@ -6362,13 +6392,15 @@ for="value-with-size">value</dfn> and <dfn for="value-with-size">size</dfn>.
...
  <dfn abstract-op lt="EnqueueValueWithSize"
- id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|)</dfn> performs the
+ id="enqueue-value-with-size">EnqueueValueWithSize(|container|, |value|, |size|, |transferList|)</dfn> performs the
  following steps:
 
  1. Assert: |container| has \[[queue]] and \[[queueTotalSize]] internal slots.
  1. If ! [$IsNonNegativeNumber$](|size|) is false, throw a {{RangeError}} exception.
  1. If |size| is +∞, throw a {{RangeError}} exception.
- 1. [=list/Append=] a new [=value-with-size=] with [=value-with-size/value=] |value| and
+ 1. If |container| has an \[[isOwning]] internal slot whose value is true, perform the following steps:
```
saschanaz, line 6401:
```suggestion
 1. Let |enqueuedValue| be |value|.
 1. If |container| has an \[[isOwning]] internal slot whose value is true, perform the following steps:
```
`enqueuedValue` initialization is missing for isOwning=false branch.

saschanaz (changes requested):
The idea looks fine to me. Some nits:

index.bs
youennf: Will update the sentence and add export.

.github/workflows/test.yml
youennf: ok

index.bs
```diff
@@ -2950,13 +2969,21 @@ The following abstract operations support the implementation of the
 <div algorithm>
  <dfn abstract-op lt="ReadableStreamDefaultControllerEnqueue"
  id="readable-stream-default-controller-enqueue">ReadableStreamDefaultControllerEnqueue(|controller|,
- |chunk|)</dfn> performs the following steps:
+ |chunk|, |transferList|)</dfn> performs the following steps:
```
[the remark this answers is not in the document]
youennf: This is consistent in this spec to do it this way.

index.bs
```diff
@@ -6596,6 +6632,14 @@ The following abstract operations are a grab-bag of utilities.
...
 </div>
 
+<div algorithm>
+ <dfn abstract-op lt="StructuredTransferOrClone">StructuredTransferOrClone(|value|, |transferList|)</dfn>
+    performs the following steps:
+ 1. Let |serialized| be ? [$StructuredSerializeWithTransfer$](|value|, |transferList|).
+ 1. Let |deserialized| be ? [$StructuredDeserializeWithTransfer$](|serialized|, [=the current Realm=]).
+ 1. Return |deserialized|.\[[Deserialized]].
+</div>
+
```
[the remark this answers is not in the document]
youennf: We could do this as a follow-up, the only difference between the two right now is passing transferList or options.

reference-implementation/lib/ReadableStreamDefaultController-impl.js
```diff
@@ -17,12 +17,16 @@ exports.implementation = class ReadableStreamDefaultControllerImpl {
     aos.ReadableStreamDefaultControllerClose(this);
   }
 
-  enqueue(chunk) {
+  enqueue(chunk, options) {
+    const transferList = options ? options.transfer : [];
```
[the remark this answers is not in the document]
youennf: I think you are right, will update this.

index.bs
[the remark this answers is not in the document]
saschanaz: I don't think it is, all uses of `« »` is to call external functions outside Streams.

reference-implementation/lib/ReadableStreamDefaultController-impl.js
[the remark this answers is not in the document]
youennf: Yes, will do today

youennf:
I updated tests in https://github.com/web-platform-tests/wpt/pull/39520.
Should we merge these WPT tests first so that I update this PR to use the WPT commit?

ricea:
> I updated tests in [web-platform-tests/wpt#39520](https://github.com/web-platform-tests/wpt/pull/39520). Should we merge these WPT tests first so that I update this PR to use the WPT commit?

Yes, please.

MattiasBuelens:
And now we're blocked on #1264 to get WPT working again... 😛

youennf:
PR is green and was approved. Is it ready for being merged?

index.bs
```diff
@@ -5188,7 +5217,7 @@ The following abstract operations support the implementation of the
  id="writable-stream-default-controller-close">WritableStreamDefaultControllerClose(|controller|)</dfn>
  performs the following steps:
 
- 1. Perform ! [$EnqueueValueWithSize$](|controller|, [=close sentinel=], 0).
+ 1. Perform ! [$EnqueueValueWithSize$](|controller|, [=close sentinel=], 0, undefined).
```
saschanaz, line 5220: Should this `undefined` be `« »` for consistency, or vise versa?

index.bs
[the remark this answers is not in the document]
saschanaz: (Please file an issue to not keep the duplication too long, or let me know so that I can file one)

saschanaz (commented):
BTW, should we care about [`Symbol.dispose` ECMA262 Stage 3 proposal](https://github.com/tc39/proposal-explicit-resource-management)? It looks reasonable for this API to support that, and if so it should be better sooner than later to prevent webcompat issue.

index.bs
```diff
@@ -651,8 +651,15 @@ enum ReadableStreamType { "bytes" };
   <p>For an example of how to set up a readable byte stream, including using the different
   controller interface, see [[#example-rbs-push]].
 
-  <p>Setting any value other than "{{ReadableStreamType/bytes}}" or undefined will cause the
-  {{ReadableStream()}} constructor to throw an exception.
+  <p>Can be set to "<dfn enum-value for="ReadableStreamType">owning</dfn>" to signal that the
+  constructed {{ReadableStream}} will own chunks (via transfer or serialization) before enqueuing them.
+  This ensures that enqueued chunks are not mutable by the source.
```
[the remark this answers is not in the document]
youennf: Let's change it to `cannot be mutated by the source`

index.bs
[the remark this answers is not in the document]
youennf: OK, will do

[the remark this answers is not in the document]
youennf: https://github.com/whatwg/streams/issues/1285

saschanaz (approved):
LGTM, with a remaining question about Symbol.dispose.

youennf:
> LGTM, with a remaining question about Symbol.dispose.

Oh sorry, I missed the question.
The plan is to support this at some point in a follow-up PR, we will concentrate on ReadableStream and others first.

index.bs
```diff
@@ -1538,6 +1546,10 @@ the following table:
   <tr>
    <td><dfn>\[[stream]]</dfn>
    <td class="non-normative">The {{ReadableStream}} instance controlled
+  <tr>
+   <td><dfn>\[[isOwning]]</dfn>
+   <td class="non-normative">A boolean flag indicating whether to take ownership of enqueued chunks
+   via transfer or serialization.
```
MattiasBuelens, line 1552: Nit-pick: this table is sorted alphabetically, so this should go between `[[closeRequested]]` and `[[pullAgain]]`.

youennf: Done

ricea:
Sorry for the radio silence. I'm trying to catch up on the latest changes, but I've been a bit swamped.

MattiasBuelens:
@ricea Friendly ping. If you could find some time to review this, that'd be appreciated. 😉

jasnell:
Overall I'm happy with the general idea but this feels rather heavyweight for the use case. Perhaps `transfer` can be built into the existing model without introducing a new stream "type".

For instance, `ReadableStreamDefaultController` can have a new method `controller.clone(value, transferList)` that is like `controller.enqueue(...)` but clones/transfers the value when enqueued.

MattiasBuelens:
@jasnell That would imply that a stream can hold both "owned" and "unowned" chunks at the same time. This would mean a bunch more bookkeeping and passing around extra parameters:

* The stream's queue would need to record for each chunk whether it is owned or not, so `ResetQueue` knows to *only* call the dispose steps on owned chunks. We'll need to extend [value-with-size](https://streams.spec.whatwg.org/#value-with-size) to have an extra *isOwned* item.
* `ReadableStreamTee` would no longer unconditionally clone chunks to both branches. Instead, it needs to first check for each chunk received from the original stream whether it was owned or not. We'd need to extend [read request's chunk steps](https://streams.spec.whatwg.org/#read-request-chunk-steps) to have an extra parameter *isOwned* to store that information.
  * Should this information be exposed through `ReadableStreamDefaultReader.read()` too? Something like returning a `{ done, value, owned }` tuple instead? (This is probably a very bad idea... 😅)
* `TransformStreamDefaultController` would need the same treatment, and get a `clone()` method.
* We would need to extend this to `WritableStream` too, e.g. `WritableStreamDefaultWriter.writeCloned(chunk, { transfer: [] })` or `.write(chunk, { clone: true, transfer: [] })`
* Would we need to inform a `WritableStream`'s sink or a `TransformStream`'s transformer that a written chunk is now owned by the stream, in case it wants to forward that chunk to `Readable|TransformStreamDefaultController.clone()`? That would mean passing an *isOwned* parameter to `sink.write()` or `transformer.transform()` somehow... 😕

All in all, I think it's easier to have a separate type for owning streams.

MattiasBuelens:
Hmm, how does this interact with transferable streams?

Sure, we transfer in `controller.enqueue()` to push the chunk onto the stream's queue. But if the stream has been transferred, then:
* the cross-realm writable we create should also be of type `owning`
* the message we post through the internal `MessagePort` containing the chunk should have a proper transfer list

To support this, I think we'll have to store the transfer list alongside each chunk in the queue, so we can re-transfer the chunk later on. 🤔

saschanaz:
Random thought about being heavyweight, what about:

```js
let readable = new ReadableStream({
  start(controller) {
     // Wrap the chunk with a new interface (name chosen randomly)
     controller.enqueue(new Ownership(chunk));
  }
});
// ResetQueue checks whether the chunk is Ownership, and if yes,
// run dispose steps of the inner data of Ownership.
readable.cancel();
// Same for WritableStream, it would handle chunks that are Ownership.
readable.pipeTo(writable)
```

youennf:
I also think a stream that only handles owned chunks is simpler to reason about and as powerful.
Some cases for individually owned chunks can be covered with `structuredClone` though not for disposing in case of cancel.

PR #1277, opened 2023-05-02
Editorial: account for removal of AbortSignal's follow

shaseley:
"Follow" is being removed in https://github.com/whatwg/dom/pull/1152.

cc: @annevk. AFAIK this is the [only other use](https://dontcallmedom.github.io/webdex/f.html#follow%40%40AbortSignal%40dfn) aside from fetch.

***
<a href="https://whatpr.org/streams/1277.html" title="Last updated on May 3, 2023, 6:25 PM UTC (d4a04ff)">Preview</a> | <a href="https://whatpr.org/streams/1277/2942e89...d4a04ff.html" title="Last updated on May 3, 2023, 6:25 PM UTC (d4a04ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6999,7 +6999,7 @@ above [=WritableStream/set up=] algorithm:
 |stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[signal]]=].
 Specifications can [=AbortSignal/add=] or [=AbortSignal/remove=] algorithms to this
 {{AbortSignal}}, or consult whether it is [=AbortSignal/aborted=] and its [=AbortSignal/abort
-reason=]. Specifications should not [=AbortSignal/signal abort=], as that would interfere with the
+reason=]. Specifications must not [=AbortSignal/signal abort=], as that would interfere with the
 normal use of this signal to respond to the stream being [=abort a writable stream|aborted=].
 
 <p class="note">The usual usage is, after [=WritableStream/setting up=] the {{WritableStream}},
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6999,9 +6999,8 @@ above [=WritableStream/set up=] algorithm:
 |stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[signal]]=].
 Specifications can [=AbortSignal/add=] or [=AbortSignal/remove=] algorithms to this
 {{AbortSignal}}, or consult whether it is [=AbortSignal/aborted=] and its [=AbortSignal/abort
-reason=]. Specifications should not [=AbortSignal/signal abort=] or make it [=AbortSignal/follow=]
-another {{AbortSignal}}, as that would interfere with the normal use of this signal to respond to
-the stream being [=abort a writable stream|aborted=].
+reason=]. Specifications should not [=AbortSignal/signal abort=], as that would interfere with the
+normal use of this signal to respond to the stream being [=abort a writable stream|aborted=].
```
annevk, line 7003: Follow-up nit: must not? Or what would a good reason be?

annevk (approved):
Ah, this is pretty straightforward compared to Fetch.

index.bs
MattiasBuelens: Agreed, "must not" would be better. Only the Streams standard itself should signal abort.

However, since https://github.com/whatwg/dom/pull/1152 will make "to signal abort on an `AbortSignal`" private, it will no longer be possible for other specifications to affect this signal. So we can hopefully get rid of this entire footnote. 😁

By the way, should we already refactor Streams to use "to signal abort on an `AbortController`" instead? Or should we wait until https://github.com/whatwg/dom/pull/1152 has landed?

annevk: Oh, I had missed that it no longer exports that definition.

@shaseley if that was intentional this would have to be a larger PR and other specifications would need updating as well: https://dontcallmedom.github.io/webdex/s.html#signal%20abort%40%40AbortSignal%40dfn.

@MattiasBuelens no need to do anything just yet. Perhaps we should scope down 1152 and leave that as a follow-up.

shaseley: +1 to scoping down 1152 and making 'signal abort' private in a follow-up. I'll fix 1152, update the language to 'must', and work on the follow-up after 1152 lands. Sorry about the confusion!

domenic (approved):

PR #1279, opened 2023-05-24
Use an AbortController instead of signaling abort on AbortSignal

shaseley:
For https://github.com/whatwg/dom/issues/1194 (and follow-up from https://github.com/whatwg/streams/pull/1277).

 - Replace WritableStreamDefaultController's [[signal]] with [[abortController]], using this abort controller's signal where [[signal]] was used. Signal abort on [[abortController]] rather than the [[signal]].
 - Remove the note/warning about specs not signaling abort on the WritableStream's signal, since this won't be allowed once AbortSignal's "signal abort" no longer exported.

***
<a href="https://whatpr.org/streams/1279.html" title="Last updated on May 25, 2023, 4:52 PM UTC (0303684)">Preview</a> | <a href="https://whatpr.org/streams/1279/cf7670a...0303684.html" title="Last updated on May 25, 2023, 4:52 PM UTC (0303684)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4446,6 +4446,10 @@ the following table:
    <td class="non-normative">A promise-returning algorithm, taking one argument (the abort reason),
    which communicates a requested abort to the [=underlying sink=]
   <tr>
+   <td><dfn>\[[abortController]]</dfn>
+   <td class="non-normative">An {{AbortController}} that can be used to abort the pending write or
+   close operation when the stream is [=abort a writable stream|aborted=].
+  <tr>
    <td><dfn>\[[closeAlgorithm]]</dfn>
    <td class="non-normative">A promise-returning algorithm which communicates a requested close to
    the [=underlying sink=]
@@ -4456,10 +4460,6 @@ the following table:
    <td><dfn>\[[queueTotalSize]]</dfn>
    <td class="non-normative">The total size of all the chunks stored in
    [=WritableStreamDefaultController/[[queue]]=] (see [[#queue-with-sizes]])
-  <tr>
-   <td><dfn>\[[abortController]]</dfn>
-   <td class="non-normative">An {{AbortController}} that can be used to abort the pending write or
-   close operation when the stream is [=abort a writable stream|aborted=].
   <tr>
    <td><dfn>\[[started]]</dfn>
    <td class="non-normative">A boolean flag indicating whether the [=underlying sink=] has finished
@@ -4508,7 +4508,8 @@ closed. It is only used internally, and is never exposed to web developers.
  The <dfn id="ws-default-controller-signal" attribute
  for="WritableStreamDefaultController">signal</dfn> getter steps are:
 
- 1. Return [=this=].[=WritableStreamDefaultController/[[abortController]]=].{{AbortController/signal}}.
+ 1. Return [=this=].[=WritableStreamDefaultController/[[abortController]]=]'s
+    [=AbortController/signal=].
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


shaseley:
@annevk how does this look?

index.bs
```diff
@@ -4508,7 +4508,7 @@ closed. It is only used internally, and is never exposed to web developers.
  The <dfn id="ws-default-controller-signal" attribute
  for="WritableStreamDefaultController">signal</dfn> getter steps are:
 
- 1. Return [=this=].[=WritableStreamDefaultController/[[signal]]=].
+ 1. Return [=this=].[=WritableStreamDefaultController/[[abortController]]=].{{AbortController/signal}}.
```
annevk, line 4511: You need to return the value of the "internal slot" and not reference the "public" getter.

index.bs
```diff
@@ -6996,11 +6997,9 @@ above [=WritableStream/set up=] algorithm:
 [$WritableStreamDefaultControllerErrorIfNeeded$](|stream|.[=WritableStream/[[controller]]=], |e|).
 
 <p>The <dfn export for="WritableStream">signal</dfn> of a {{WritableStream}} |stream| is
-|stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[signal]]=].
-Specifications can [=AbortSignal/add=] or [=AbortSignal/remove=] algorithms to this
-{{AbortSignal}}, or consult whether it is [=AbortSignal/aborted=] and its [=AbortSignal/abort
-reason=]. Specifications must not [=AbortSignal/signal abort=], as that would interfere with the
-normal use of this signal to respond to the stream being [=abort a writable stream|aborted=].
+|stream|.[=WritableStream/[[controller]]=].[=WritableStreamDefaultController/[[abortController]]=].{{AbortController/signal}}.
```
annevk, line 7000: Same as above, cannot use the getter.

annevk (commented):
If you replace the getter with the appropriate reference I think this works. Thanks!

@MattiasBuelens should do the final review though.

index.bs
```diff
@@ -4457,8 +4457,8 @@ the following table:
    <td class="non-normative">The total size of all the chunks stored in
    [=WritableStreamDefaultController/[[queue]]=] (see [[#queue-with-sizes]])
   <tr>
-   <td><dfn>\[[signal]]</dfn>
-   <td class="non-normative">An {{AbortSignal}} that can be used to abort the pending write or
+   <td><dfn>\[[abortController]]</dfn>
+   <td class="non-normative">An {{AbortController}} that can be used to abort the pending write or
```
MattiasBuelens, line 4461: These slots are sorted alphabetically. So move it up in the table, between `[[abortAlgorithm]]` and `[[closeAlgorithm]]`.

index.bs
MattiasBuelens: If I understand [Bikeshed's autolink syntax](https://speced.github.io/bikeshed/#autolink-wrappers) correctly, this should do the trick:
```suggestion
 1. Return [=this=].[=WritableStreamDefaultController/[[abortController]]=].[=AbortController/signal=].
```

MattiasBuelens (changes requested):

index.bs
annevk: It'd have to be something like
```suggestion
 1. Return [=this=].[=WritableStreamDefaultController/[[abortController]]=]'s [=AbortController/signal=].
```
as we don't define dot syntax for these.

index.bs
shaseley: Done.

index.bs
shaseley: Done.

shaseley (commented):
Thanks for reviewing! I'll put together the dom changes this depends on.

MattiasBuelens (approved):

ricea (approved):
lgtm once it builds.

MattiasBuelens:
[Latest build failure](https://github.com/whatwg/streams/actions/runs/5082538123/jobs/9202930894) looks like a problem with Bikeshed itself. I've opened speced/bikeshed#2561.

MattiasBuelens:
The new Bikeshed version fixes the build issue, so I think we're good to go! 🙂

PR #1283, opened 2023-06-08
TransformStream cleanup using "Transformer.cancel"

### LINKED ISSUE #1212: Handling stream errors in a TransformStream transformer
The background to this is that I found [a bug](https://github.com/denoland/deno/issues/13142) in deno's `TextDecoderStream` implementation that results in it failing to clean up a resource it holds if its stream pipeline aborts with an error. The implementation holds a `TextDecoder` which it uses in streaming mode. The `TextDecoder` holds a resource handle to a native object handling text decoding for it, which it closes when `decode()` is called without `{stream: true}`. When the deno `TextDecoderStream` implementation's transformer gets a `flush()` call, it calls `decode()` to close the `TextDecoder`. However, if the stream aborts, `flush()` is not called, so the native resource handle is not closed, and gets leaked.

I've looked through the Streams spec, and as I understand it there's no built-in way for a transformer to be notified of a stream error.
It is possible to work around this as an API user, as I mention in the deno issue:

> I played around with the Streams API a bit and came up with a fairly straightforward way to implement a TransformStream whose Transformer gets notified of stream aborts. Basically two parts:
> 
> - A WritableStream can be monitored for errors by wrapping it with another WritableStream that opens a reader on the monitored stream, exposes the reader's closed promise (which rejects if the monitored stream is aborted), and forwards start/write/close/abort calls to the monitored stream.
> That looks like this: https://deno.land/x/shutdown_aware_transform_stream@1.0.0/shutdown_monitor_writable_stream.ts
> - Then a TransformStream can react to stream aborts by monitoring its writable side with the monitor stream, and using the closed promise to be notified when the stream aborts.
> That looks like this: https://deno.land/x/shutdown_aware_transform_stream@1.0.0/shutdown_aware_transform_stream.ts#L98

Although I say "fairly straightforward", it's not exactly trivial. And another alternative of not using `TransformStream` and instead tying together a readable and writable stream manually to create a (readable, writable) pair is even more fiddly to do correctly.

As an API user, it seems like there should be an idiomatic way to handle stream errors in a transformer. The underlying sink of a `WritableStream` can do so either with its `abort()` method, or via the `AbortSignal` on `WritableStreamDefaultController`'s `signal` property.

What do you think about giving transformers similar capabilities to handle aborts as underlying sinks? 

Even just giving `TransformStreamDefaultController` an `AbortSignal` would be helpful (I presume that's simpler to spec than a method on transformer, as it can't affect the error propagation behaviour). Although I suppose a method would allow for asynchronous cleanup...

lucacasonato:
This commit adds a `cancel` hook to `Transformer`. This allows users to perform resource cleanup when the readable side of the `TransformStream` is cancelled, or the writable side is aborted.

To preserve existing behavior, when the readable side is cancelled with a reason, the writable side is always immediately aborted with that same reason. The same is true in the reverse case. This means that the status of both sides is always either `closed`, `erroring`, or `erroring` when the `cancel` hook is called.

`flush` and `cancel` are never both called. As per existing behaviour, when the writable side is closed the `flush` hook is called. If the readable side is cancelled while a promise returned from `flush` is still pending, `cancel` is not called. In this scenario the readable side ends up in the "errored" state, while the writable side ends up in the `closed` state.

I have opted for a `cancel` hook instead of a `finally` hook, to mirror the API in `WritableStream` - it has one hook for successful completion (`close`), and one hook for errored completion (`abort`). `Transformer` already has a `flush` hook for successful completion. The logical addition is an `cancel` hook for errored completion.

Closes #1212

---

Open questions:

- Is `cancel` the right name? Mirrors `ReadableStream`, but we use `abort` for `WritableStream`.
- Is the immediate cancellation behaviour correct?
  - Ie should call the `cancel` hook before or after cancelling the underlying streams?
  - Calling it before means that current behaviour is changed (number of microticks from `readable.cancel()` to the `writable` actually being aborted). This is the current behaviour.
  - Calling it after means that you can not modify the reason passed to the other side using the `cancel` hook. This is different to `abort` and `cancel` hooks in `WritableStream` and `ReadableStream` respectively.

---

   * WebKit (https://github.com/whatwg/streams/pull/1283#issuecomment-1738663536)
   * Chromium (https://github.com/whatwg/streams/pull/1283#issuecomment-1738960442)
   * (Deno)
   * https://github.com/web-platform-tests/wpt/pull/40453
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1488142
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1856103
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=262424
   * Deno: https://github.com/denoland/deno/issues/20741
   * Node.js: https://github.com/nodejs/node/issues/49971
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/29375

***
<a href="https://whatpr.org/streams/1283.html" title="Last updated on Sep 30, 2023, 2:07 AM UTC (4d2abe5)">Preview</a> | <a href="https://whatpr.org/streams/1283/8d7a0bf...4d2abe5.html" title="Last updated on Sep 30, 2023, 2:07 AM UTC (4d2abe5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7197,8 +7197,7 @@ reason.
  for="TransformStream/set up"><var>transformAlgorithm</var></dfn>, an optional algorithm <dfn
  export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, and an optional algorithm <dfn
  export for="TransformStream/set up"><var>cancelAlgorithm</var></dfn>, perform the following steps.
- |transformAlgorithm|, if given, |flushAlgorithm|, and, if given, |cancelAlgorithm|, may return a
- promise.
+ |transformAlgorithm|, if given, |flushAlgorithm| and |cancelAlgorithm|, may return a promise.
 
  1. Let |writableHighWaterMark| be 1.
  1. Let |writableSizeAlgorithm| be an algorithm that returns 1.
--- a/index.bs
+++ b/index.bs
@@ -5497,7 +5497,7 @@ dictionary Transformer {
 callback TransformerStartCallback = any (TransformStreamDefaultController controller);
 callback TransformerFlushCallback = Promise<undefined> (TransformStreamDefaultController controller);
 callback TransformerTransformCallback = Promise<undefined> (any chunk, TransformStreamDefaultController controller);
-callback TransformerCancelCallback = Promise<undefined> (any reason, TransformStreamDefaultController controller);
+callback TransformerCancelCallback = Promise<undefined> (any reason);
 </xmp>
 
 <dl>
@@ -5560,7 +5560,7 @@ callback TransformerCancelCallback = Promise<undefined> (any reason, TransformSt
    {{Transformer/flush|flush()}}; the stream is already in the process of successfully closing down,
    and terminating it would be counterproductive.)
 
-  <dt><dfn dict-member for="Transformer" lt="cancel">cancel(<var ignore>reason</var>, <var ignore>controller</var>)</dfn></dt>
+  <dt><dfn dict-member for="Transformer" lt="cancel">cancel(<var ignore>reason</var>)</dfn></dt>
   <dd>
    <p>A function called when the [=writable side=] is aborted, or when the [=readable side=] is
    cancelled.
@@ -5948,8 +5948,8 @@ The following abstract operations support the implementaiton of the
     with argument list «&nbsp;|controller|&nbsp;» and [=callback this value=] |transformer|.
  1. If |transformerDict|["{{Transformer/cancel}}"] [=map/exists=], set |cancelAlgorithm| to an
     algorithm which takes an argument |reason| and returns the result of [=invoking=]
-    |transformerDict|["{{Transformer/cancel}}"] with argument list «&nbsp;|reason|,
-    |controller|&nbsp;» and [=callback this value=] |transformer|.
+    |transformerDict|["{{Transformer/cancel}}"] with argument list «&nbsp;|reason|&nbsp;» and
+    [=callback this value=] |transformer|.
  1. Perform ! [$SetUpTransformStreamDefaultController$](|stream|, |controller|,
     |transformAlgorithm|, |flushAlgorithm|, |cancelAlgorithm|).
 </div>
--- a/index.bs
+++ b/index.bs
@@ -5888,6 +5888,18 @@ The following abstract operations operate on {{TransformStream}} instances at a 
 </div>
 
 <div algorithm>
+ <dfn abstract-op lt="TransformStreamSetBackpressure"
+ id="transform-stream-set-backpressure">TransformStreamSetBackpressure(|stream|,
+ |backpressure|)</dfn> performs the following steps:
+
+ 1. Assert: |stream|.[=TransformStream/[[backpressure]]=] is not |backpressure|.
+ 1. If |stream|.[=TransformStream/[[backpressureChangePromise]]=] is not undefined, [=resolve=]
+    stream.[=TransformStream/[[backpressureChangePromise]]=] with undefined.
+ 1. Set |stream|.[=TransformStream/[[backpressureChangePromise]]=] to [=a new promise=].
+ 1. Set |stream|.[=TransformStream/[[backpressure]]=] to |backpressure|.
+</div>
+
+<div algorithm>
  <dfn abstract-op lt="TransformStreamUnblockWrite"
  id="transform-stream-unblock-write">TransformStreamUnblockWrite(|stream|)</dfn> performs the
  following steps:
@@ -5898,18 +5910,6 @@ The following abstract operations operate on {{TransformStream}} instances at a 
  <p class="note">The [$TransformStreamDefaultSinkWriteAlgorithm$] abstract operation could be
  waiting for the promise stored in the [=TransformStream/[[backpressureChangePromise]]=] slot to
  resolve. The call to [$TransformStreamSetBackpressure$] ensures that the promise always resolves.
-</div>
-
-<div algorithm>
- <dfn abstract-op lt="TransformStreamSetBackpressure"
- id="transform-stream-set-backpressure">TransformStreamSetBackpressure(|stream|,
- |backpressure|)</dfn> performs the following steps:
-
- 1. Assert: |stream|.[=TransformStream/[[backpressure]]=] is not |backpressure|.
- 1. If |stream|.[=TransformStream/[[backpressureChangePromise]]=] is not undefined, [=resolve=]
-    stream.[=TransformStream/[[backpressureChangePromise]]=] with undefined.
- 1. Set |stream|.[=TransformStream/[[backpressureChangePromise]]=] to [=a new promise=].
- 1. Set |stream|.[=TransformStream/[[backpressure]]=] to |backpressure|.
 </div>
 
 <h4 id="ts-default-controller-abstract-ops">Default controllers</h4>
--- a/index.bs
+++ b/index.bs
@@ -5749,7 +5749,7 @@ the following table:
  <tbody>
   <tr>
    <td><dfn>\[[cancelAlgorithm]]</dfn>
-   <td class="non-normative">A promise-returning algorithm, taking one argument (the [=reason=] for
+   <td class="non-normative">A promise-returning algorithm, taking one argument (the reason for
    cancellation), which communicates a requested cancellation to the [=transformer=]
   <tr>
    <td><dfn>\[[finishPromise]]</dfn>
@@ -6122,18 +6122,6 @@ The following abstract operation is used to implement the [=underlying source=] 
 side=] of [=transform streams=].
 
 <div algorithm>
- <dfn abstract-op lt="TransformStreamDefaultSourcePullAlgorithm"
- id="transform-stream-default-source-pull">TransformStreamDefaultSourcePullAlgorithm(|stream|)</dfn>
- performs the following steps:
-
- 1. Assert: |stream|.[=TransformStream/[[backpressure]]=] is true.
- 1. Assert: |stream|.[=TransformStream/[[backpressureChangePromise]]=] is not undefined.
- 1. Perform ! [$TransformStreamSetBackpressure$](|stream|, false).
- 1. Return |stream|.[=TransformStream/[[backpressureChangePromise]]=].
-</div>
-
-
-<div algorithm>
  <dfn abstract-op lt="TransformStreamDefaultSourceCancelAlgorithm"
  id="transform-stream-default-source-cancel">TransformStreamDefaultSourceCancelAlgorithm(|stream|,
  |reason|)</dfn> performs the following steps:
@@ -6160,6 +6148,17 @@ side=] of [=transform streams=].
    1. Perform ! [$TransformStreamUnblockWrite$](|stream|).
    1. [=Reject=] |controller|.[=TransformStreamDefaultController/[[finishPromise]]=] with |r|.
  1. Return |controller|.[=TransformStreamDefaultController/[[finishPromise]]=].
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="TransformStreamDefaultSourcePullAlgorithm"
+ id="transform-stream-default-source-pull">TransformStreamDefaultSourcePullAlgorithm(|stream|)</dfn>
+ performs the following steps:
+
+ 1. Assert: |stream|.[=TransformStream/[[backpressure]]=] is true.
+ 1. Assert: |stream|.[=TransformStream/[[backpressureChangePromise]]=] is not undefined.
+ 1. Perform ! [$TransformStreamSetBackpressure$](|stream|, false).
+ 1. Return |stream|.[=TransformStream/[[backpressureChangePromise]]=].
 </div>
 
 <h2 id="qs">Queuing strategies</h2>
@@ -7209,7 +7208,7 @@ reason.
  for="TransformStream/set up"><var>transformAlgorithm</var></dfn>, an optional algorithm <dfn
  export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, and an optional algorithm <dfn
  export for="TransformStream/set up"><var>cancelAlgorithm</var></dfn>, perform the following steps.
- |transformAlgorithm|, if given, |flushAlgorithm| and |cancelAlgorithm|, may return a promise.
+ |transformAlgorithm| and, if given, |flushAlgorithm| and |cancelAlgorithm|, may return a promise.
 
  1. Let |writableHighWaterMark| be 1.
  1. Let |writableSizeAlgorithm| be an algorithm that returns 1.
--- a/index.bs
+++ b/index.bs
@@ -5562,11 +5562,11 @@ callback TransformerCancelCallback = Promise<undefined> (any reason);
 
   <dt><dfn dict-member for="Transformer" lt="cancel">cancel(<var ignore>reason</var>)</dfn></dt>
   <dd>
-   <p>A function called when the [=writable side=] is aborted, or when the [=readable side=] is
-   cancelled.
+   <p>A function called when the [=readable side=] is cancelled, or when the [=writable side=] is
+   aborted.
 
-   <p>Typically this is used to clean up underlying transformer resources when the  stream is
-   aborted or cancelled.
+   <p>Typically this is used to clean up underlying transformer resources when the stream is aborted
+   or cancelled.
 
    <p>If the cancellation process is asynchronous, the function can return a promise to signal
    success or failure; the result will be communicated to the caller of
@@ -6076,6 +6076,8 @@ side=] of [=transform streams=].
  performs the following steps:
 
  1. Let |readable| be |stream|.[=TransformStream/[[readable]]=].
+ 1. If |readable|.[=ReadableStream/[[state]]=] is not "`readable`", return
+    [=a promise resolved with=] undefined.
  1. Let |controller| be |stream|.[=TransformStream/[[controller]]=].
  1. Let |flushPromise| be the result of performing
     |controller|.[=TransformStreamDefaultController/[[flushAlgorithm]]=].
--- a/index.bs
+++ b/index.bs
@@ -5591,9 +5591,9 @@ callback TransformerCancelCallback = Promise<undefined> (any reason);
 </dl>
 
 The <code>controller</code> object passed to {{Transformer/start|start()}},
-{{Transformer/transform|transform()}}, {{Transformer/flush|flush()}}, and
-{{Transformer/cancel|cancel()}} is an instance of {{TransformStreamDefaultController}}, and has the
-ability to enqueue [=chunks=] to the [=readable side=], or to terminate or error the stream.
+{{Transformer/transform|transform()}}, and {{Transformer/flush|flush()}} is an instance of
+{{TransformStreamDefaultController}}, and has the ability to enqueue [=chunks=] to the
+[=readable side=], or to terminate or error the stream.
 
 <h4 id="ts-prototype">Constructor and properties</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -5489,6 +5489,7 @@ dictionary Transformer {
   TransformerStartCallback start;
   TransformerTransformCallback transform;
   TransformerFlushCallback flush;
+  TransformerCancelCallback cancel;
   any readableType;
   any writableType;
 };
@@ -5496,6 +5497,7 @@ dictionary Transformer {
 callback TransformerStartCallback = any (TransformStreamDefaultController controller);
 callback TransformerFlushCallback = Promise<undefined> (TransformStreamDefaultController controller);
 callback TransformerTransformCallback = Promise<undefined> (any chunk, TransformStreamDefaultController controller);
+callback TransformerCancelCallback = Promise<undefined> (any reason, TransformStreamDefaultController controller);
 </xmp>
 
 <dl>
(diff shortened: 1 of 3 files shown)


ricea:
> * Is `cancel` the right name? Mirrors `ReadableStream`, but we use `abort` for `WritableStream`.

Since it's used for both the `cancel` and `abort` cases, it might be good to come up with a third name to avoid confusion. `stop` is one possibility.

> * Is the immediate cancellation behaviour correct?
>   
>   * Ie should call the `cancel` hook before or after cancelling the underlying streams?
>   * Calling it before means that current behaviour is changed (number of microticks from `readable.cancel()` to the `writable` actually being aborted). This is the current behaviour.
>   * Calling it after means that you can not modify the reason passed to the other side using the `cancel` hook. This is different to `abort` and `cancel` hooks in `WritableStream` and `ReadableStream` respectively.

Interesting. Calling it before would be better in terms of consistency with the other APIs. I personally think that changing the number of microticks is an acceptable breakage, but others may disagree.

lucacasonato:
> Interesting. Calling it before would be better in terms of consistency with the other APIs. I personally think that changing the number of microticks is an acceptable breakage, but others may disagree.

For reference, exactly one WPT is broken by this.

lucacasonato:
> Since it's used for both the cancel and abort cases, it might be good to come up with a third name to avoid confusion. stop is one possibility.

What about `terminate`? This aligns with `controller.terminate`

ricea:
> What about `terminate`? This aligns with `controller.terminate`

`controller.terminate` is rather a niche and subtle operation, which is why we gave it a long name.

This operation, however, is something we'd like people to implement if they need to do cleanup, so a short friendly name seems appropriate.

index.bs
```diff
@@ -5489,13 +5489,15 @@ dictionary Transformer {
...
   TransformerFlushCallback flush;
+  TransformerCancelCallback cancel;
   any readableType;
   any writableType;
 };
 
 callback TransformerStartCallback = any (TransformStreamDefaultController controller);
 callback TransformerFlushCallback = Promise<undefined> (TransformStreamDefaultController controller);
 callback TransformerTransformCallback = Promise<undefined> (any chunk, TransformStreamDefaultController controller);
+callback TransformerCancelCallback = Promise<undefined> (any reason, TransformStreamDefaultController controller);
```
MattiasBuelens, line 5500: Remove the `controller` argument. Both readable and writable sides have already transitioned to a final state by the time we call `cancel()`, so you can no longer use the controller for anything useful.

Also, this aligns with ReadableStream's `cancel()` algorithm and WritableStream's `abort()` algorithm.

lucacasonato:
I don't particularly like `dispose`, as it implies it is called both on successful and errored close (because that is what `Symbol.dispose` does. This however is only called on the error case, because successful disposal is already handled through `flush`.

I think we need something that implies early _unsuccessful_ termination. `abort` and `cancel` fit the bill here, except that we already use them elsewhere so using the same word may be confusing. I'm not super strongly concerned with confusion here tho, as the use case is very similar to those hooks in RS and WS.

index.bs
lucacasonato: Done

lucacasonato:
@domenic @MattiasBuelens @saschanaz Do you have any thoughts on the "Is the immediate cancellation behaviour correct?" point? I'm leaning towards the currently implemented behaviour because it minimizes chance of breakage, but I do also sympathize with @ricea's points above.

domenic:
What about `cancelOrAbort` for the name? Since it's literally called when either the readable side is canceled, or the writable side is aborted? It's not short though.

I think calling it before is better. I agree that changing the number of microtasks is acceptable.

index.bs
```diff
@@ -5558,6 +5560,25 @@ callback TransformerTransformCallback = Promise<undefined> (any chunk, Transform
    {{Transformer/flush|flush()}}; the stream is already in the process of successfully closing down,
    and terminating it would be counterproductive.)
 
+  <dt><dfn dict-member for="Transformer" lt="cancel">cancel(<var ignore>reason</var>)</dfn></dt>
+  <dd>
+   <p>A function called when the [=writable side=] is aborted, or when the [=readable side=] is
```
saschanaz, line 5565: Maybe mention the readable side first as that's the general order in the spec.

index.bs
```diff
@@ -5558,6 +5560,25 @@ callback TransformerTransformCallback = Promise<undefined> (any chunk, Transform
...
+   success or failure; the result will be communicated to the caller of
+   {{WritableStream/abort()|stream.writable.abort()}} or
+   {{ReadableStream/cancel()|stream.readable.cancel()}}. Throwing an exception is treated the same
+   as returning a rejected promise.
+
+   <p>(Note that there is no need to call
+   {{TransformStreamDefaultController/terminate()|controller.terminate()}} inside
+   {{Transformer/cancel|cancel()}}; the stream is already in the process of cancelling/aborting, and
+   terminating it would be counterproductive.)
+
```
saschanaz, line 5581:
```suggestion
```

This part is redundant after 1c65d61bfed17c117f9c34a577a713b4d1a1c640.

lucacasonato:
> I think calling it before is better.

Ok, I'll change that. This opens up a question on promise handling:

```js
const ts = new TransformStream({
  async cancel(reason) {
    console.log("t.cancel start with", reason);
    await delay(100);
    console.log("t.cancel end");
    throw "t.cancel error";
  },
});
console.log("ts constructed");
ts.readable.cancel("rs.cancel error")
  .then(() => console.log("rs.cancel fulfilled"))
  .catch((err) => console.log("rs.cancel rejected with", err));
ts.writable.abort("ws.abort error")
  .then(() => console.log("ws.abort fulfilled"))
  .catch((err) => console.log("ws.abort rejected with", err));
console.log("started rs.cancel and ws.abort");
```

I think the logical ordering of logs is:

```
# a-1
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.abort
t.cancel end
rs.cancel rejected with t.cancel error
ws.abort rejected with t.cancel error
```

But the following are also possible:

```
# a-2
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.abort
ws.abort fulfilled
t.cancel end
rs.cancel rejected with t.cancel error
```


```
# a-3
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.abort
t.cancel end
rs.cancel fulfilled
ws.abort fulfilled
```

Also consider:

```js
const ts = new TransformStream({
  flush() { console.log("t.flush") },
  async cancel(reason) {
    console.log("t.cancel start with", reason);
    await delay(100);
    console.log("t.cancel end");
    throw "t.cancel error";
  },
});
console.log("ts constructed");
ts.readable.cancel("rs.cancel error")
  .then(() => console.log("rs.cancel fulfilled"))
  .catch((err) => console.log("rs.cancel rejected with", err));
ts.writable.close()
  .then(() => console.log("ws.close fulfilled"))
  .catch((err) => console.log("ws.close rejected with", err));
console.log("started rs.cancel and ws.close");
```

Here similar options are possible:

```
# b-1
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.close
t.cancel end
rs.cancel rejected with t.cancel error
ws.close rejected with t.cancel error
```

```
# b-2
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.close
ws.close fulfilled
t.cancel end
rs.cancel rejected with t.cancel error
```

```
# b-3
ts constructed
t.cancel start with rs.cancel error
started rs.cancel and ws.close
t.cancel end
rs.cancel fulfilled
ws.close fulfilled
```

These each also have effect on whether the error is visible in `writer.closed` or not. In a-1 and b-1 examples `writer.closed` would reject, in a-2, a-3, b-2, and b-3 the it resolves to `undefined`.

index.bs
```diff
@@ -7106,9 +7175,11 @@ reason.
 <div algorithm="create a TransformStream">
  To <dfn export for="TransformStream" lt="set up|setting up">set up</dfn> a
  newly-[=new|created-via-Web IDL=] {{TransformStream}} |stream| given an algorithm <dfn export
- for="TransformStream/set up"><var>transformAlgorithm</var></dfn> and an optional algorithm <dfn
- export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, perform the following steps.
- |transformAlgorithm| and, if given, |flushAlgorithm|, may return a promise.
+ for="TransformStream/set up"><var>transformAlgorithm</var></dfn>, an optional algorithm <dfn
+ export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, and an optional algorithm <dfn
+ export for="TransformStream/set up"><var>cancelAlgorithm</var></dfn>, perform the following steps.
+ |transformAlgorithm|, if given, |flushAlgorithm|, and, if given, |cancelAlgorithm|, may return a
```
saschanaz, line 7181: Can we say "if given, |flushAlgorithm| and |cancelAlgorithm|"?

saschanaz (changes requested):

saschanaz:
>Does anyone have preferences here? I think a-1/b-1 make the most sense.

I fail to see why any of the resulting promises should not reject, so I also prefer a-1/b-1 (and I also think it's okay to have extra microtask).

lucacasonato:
What about:

```js
const ts = new TransformStream({
  flush() { console.log("t.flush") },
  async cancel(reason) {
    console.log("t.cancel start with", reason);
    await delay(100);
    console.log("t.cancel end");
    throw "t.cancel error";
  },
});
console.log("ts constructed");
ts.writable.close()
  .then(() => console.log("ws.close fulfilled"))
  .catch((err) => console.log("ws.close rejected with", err));
ts.readable.cancel("rs.cancel error")
  .then(() => console.log("rs.cancel fulfilled"))
  .catch((err) => console.log("rs.cancel rejected with", err));
console.log("started rs.cancel and ws.close");
```

```
ts constructed
t.flush
started ws.close and rs.cancel
ws.close fulfilled
rs.cancel fulfilled
```

I'm trying to figure out if it's observable from the outside that the RS was closed through a cancellation rather than a "clean close" (ie is `"rs.cancel error"` exposed through a `reader.read()` promise rejection)?

index.bs
lucacasonato: Well it's each of them individually are given. Ie `Option<FlushAlgorithm>` and `Option<CancelAlgorithm>` not `Option<(FlushAlgorithm, CancelAlgorithm)>`.

I'm not sure if they can just be coalesced like this. @domenic WDYT?

index.bs
lucacasonato: My train of thought is that you could extract the `controller` in `start` and assign it to a local variable.

lucacasonato: Previously it was impossible because of the immediate close semantics. I just pushed the new semantics, and will add tests for this case.

lucacasonato:
I have now implemented a v2 that runs the cancel algorithm prior to cancel / abort of the other side. I have opted for the approach where all of `rs.cancel` `ws.close` and `ws.abort` will all either resolve or reject (never one resolves one rejects), and that the first one always wins.

https://github.com/whatwg/streams/pull/1283/commits/5b00d087808627517d5b35f765eaf5a1ecb2dc5b

This required updating some test expectations (the ones making the assumption that `ws.abort` and `rs.cancel` are sync, which due to micotask changes is not the case anymore).

https://github.com/web-platform-tests/wpt/pull/40453/commits/41567169c2658d99482a75214a60c575808e7ef2

PTAL!

index.bs
domenic: I think @saschanaz's phrasing is correct English and implies `Option<FlushAlgorithm>` and `Option<CancelAlgorithm>`. Consider:

> Dogs, and if present, cats and birds, may bite you in the park.

saschanaz (approved):
Nothing looks obviously off, which means, LGTM.

For the test part, could we replace `await delay(0);` with something else?

lucacasonato:
> For the test part, could we replace `await delay(0);` with something else?

I could replace them with a fixed set of `queueMicrotask`, but I instead went with the approach used in other streams tests. It exists only to drain the current microtask queue before the next statement, not because the tests are timing sensitive.

lucacasonato:
@ricea @domenic Could you take another look?

index.bs
```diff
@@ -5854,6 +5884,14 @@ The following abstract operations operate on {{TransformStream}} instances at a
  1. Perform ! [$TransformStreamDefaultControllerClearAlgorithms$](|stream|.[=TransformStream/[[controller]]=]).
  1. Perform !
     [$WritableStreamDefaultControllerErrorIfNeeded$](|stream|.[=TransformStream/[[writable]]=].[=WritableStream/[[controller]]=], |e|).
+ 1. Perform ! [$TransformStreamUnblockWrite$](|stream|).
+</div>
+
+<div algorithm>
+ <dfn abstract-op lt="TransformStreamUnblockWrite"
```
MattiasBuelens, line 5891: Nit-pick: these abstract ops are ordered alphabetically, so this new one should go after `TransformStreamSetBackpressure`.

index.bs
```diff
@@ -7124,12 +7213,16 @@ reason.
      null otherwise. If this throws an exception |e|, return [=a promise rejected with=] |e|.
   1. If |result| is a {{Promise}}, then return |result|.
   1. Return [=a promise resolved with=] undefined.
+ 1. Let |cancelAlgorithmWrapper| be an algorithm that runs these steps given a value |reason|:
+  1. Let |result| be the result of running |cancelAlgorithm| given |reason|, if |cancelAlgorithm|
+     was given, or null otherwise. If this throws an exception |e|, return
+     [=a promise rejected with=] |e|.
```
MattiasBuelens, line 7219: The handling of `result` is missing, like we do in `transformAlgorithmWrapper` and `flushAlgorithmWrapper`:
```suggestion
 1. Let |cancelAlgorithmWrapper| be an algorithm that runs these steps given a value |reason|:
  1. Let |result| be the result of running |cancelAlgorithm| given |reason|, if |cancelAlgorithm|
     was given, or null otherwise. If this throws an exception |e|, return
     [=a promise rejected with=] |e|.
  1. If |result| is a {{Promise}}, then return |result|.
  1. Return [=a promise resolved with=] undefined.
```

reference-implementation/lib/abstract-ops/transform-streams.js
```diff
@@ -221,32 +231,61 @@ function TransformStreamDefaultSinkWriteAlgorithm(stream, chunk) {
...
 
 function TransformStreamDefaultSinkAbortAlgorithm(stream, reason) {
-  // abort() is not called synchronously, so it is possible for abort() to be called when the stream is already
-  // errored.
-  TransformStreamError(stream, reason);
-  return promiseResolvedWith(undefined);
+  verbose('TransformStreamDefaultSinkAbortAlgorithm()');
+
+  const controller = stream._controller;
+  if (controller._finishPromise) {
```
MattiasBuelens, line 237:
```suggestion
  if (controller._finishPromise !== undefined) {
```
We try to be very explicit in the reference implementation. 😉

reference-implementation/lib/abstract-ops/transform-streams.js
```diff
@@ -221,32 +231,61 @@ function TransformStreamDefaultSinkWriteAlgorithm(stream, chunk) {
...
+  });
+
+  return controller._finishPromise;
 }
 
 function TransformStreamDefaultSinkCloseAlgorithm(stream) {
   verbose('TransformStreamDefaultSinkCloseAlgorithm()');
 
+  const controller = stream._controller;
+  if (controller._finishPromise) {
```
MattiasBuelens, line 266:
```suggestion
  if (controller._finishPromise !== undefined) {
```

reference-implementation/lib/abstract-ops/transform-streams.js
```diff
@@ -264,3 +303,35 @@ function TransformStreamDefaultSourcePullAlgorithm(stream) {
   // Prevent the next pull() call until there is backpressure.
   return stream._backpressureChangePromise;
 }
+
+function TransformStreamDefaultSourceCancelAlgorithm(stream, reason) {
+  verbose('TransformStreamDefaultSourceCancelAlgorithm()');
+
+  const controller = stream._controller;
+  if (controller._finishPromise) {
```
MattiasBuelens, line 311:
```suggestion
  if (controller._finishPromise !== undefined) {
```

index.bs
```diff
@@ -6021,29 +6067,45 @@ side=] of [=transform streams=].
...
+ 1. Let |readable| be |stream|.[=TransformStream/[[readable]]=].
+ 1. Let |controller|.[=TransformStreamDefaultController/[[finishPromise]]=] be a new promise.
  1. Let |flushPromise| be the result of performing
     |controller|.[=TransformStreamDefaultController/[[flushAlgorithm]]=].
  1. Perform ! [$TransformStreamDefaultControllerClearAlgorithms$](|controller|).
- 1. Return the result of [=reacting=] to |flushPromise|:
+ 1. [=React=] to |flushPromise|:
   1. If |flushPromise| was fulfilled, then:
-   1. If |readable|.[=ReadableStream/[[state]]=] is "`errored`", throw
-      |readable|.[=ReadableStream/[[storedError]]=].
```
MattiasBuelens, line 6041: Why was this check removed?

MattiasBuelens (changes requested):

index.bs
lucacasonato: Because _I think_, it is not possible for the readable to get into the errored state while this processing is happening.

Let me explain:

1. User calls `writable.close()`, returning promise 1
2. `WS#close` hook is called, invoking `TransformStreamDefaultSinkCloseAlgorithm`. The close algorithm returns a pending promise
3. User calls `readable.cancel()`, returning promise 2
4. `RS#cancel` hook is called, returning `finishPromise` (which is pending)
5. close algorithm pending promise is resolved
6. `finishPromise` gets resolved because of this (in `TransformStreamDefaultSinkCloseAlgorithm`)
7. promise 1 resolves
8. promise 2 resolves

The error passed to `readable.cancel()` is swallowed.

Because the `controller` of the RS is not exposed to the user, it is not possible to close the RS without going thru `cancelAlgorithm` in the TS.

This is inline with my comments above that all of `rs.cancel`, `ws.close` and `ws.abort` will all either resolve or reject (never one resolves one rejects), and that the first one always wins.

MattiasBuelens: > Because the `controller` of the RS is not exposed to the user, it is not possible to close the RS without going thru `cancelAlgorithm` in the TS.

You can error the readable side by calling `TransformStreamDefaultController.error()`. So I believe this check is trying to handle some sort of race condition between `writer.close()` and `controller.error()`.

Perhaps it's to handle the case where `flush()` called `controller.error()` instead of rejecting?
```javascript
new TransformStream({
  flush(controller) {
    controller.error(new Error("Boom!"));
  }
})
```

lucacasonato: I see! That's a fun one. I hadn't considered this case. I'll update and add a test

lucacasonato: Fixed in https://github.com/whatwg/streams/pull/1283/commits/40eadec8a733451e3819a16c31934fcb2a7987e9. Updated tests in https://github.com/web-platform-tests/wpt/pull/40453/commits/1f2f6c8912f414763e9ec177c8dfe316809a48c5

MattiasBuelens (approved):

lucacasonato:
I have one editor approval - does anyone else still want to take a look, or is this good to merge? @MattiasBuelens do we usually do a "final comment period" for spec PRs to help keep things moving?

domenic:
This needs implementer interest before it's ready to merge. Although I trust @MattiasBuelens , I and other editors might be holding off on further review until we're clear on whether this has a cross-browser future.

lucacasonato:
Ok, would be great to start getting some explicit feedback on interest from Chrome / Firefox / Safari then.

I would urge folks to look at this with some urgency - it's a very serious omission in `TransformStream` that makes implementations of many transformer algorithms impossible unless cleanup is performed exclusively via GC. There is significant user interest, both from us at Deno, from folks using Cloudflare Workers (cc @jasnell), and from other interested parties (see the original issue #1212).

@domenic is Chrome interested in shipping this? It sounded in this issue as if you may be interested? Looking it over, it seems like a relatively trivial implementation for Chromium :)

@saschanaz can I note down your LGTM as "interested in shipping" for Gecko?

@annevk is WebKit interested? (if you are the wrong person to ping, please forward :) )

saschanaz:
LGTM spec-wise, but not actively pursuing implementation-wise. If others are interested to ship it then Mozilla will too.

domenic:
@ricea is the relevant Chromium implementer

lucacasonato:
@ricea Is Chrome interested in shipping this?

youennf:
This addition makes sense to me as we have more and more web platform objects that benefit from being explicitly closed.

ricea:
Yes, Chrome is interested. Sorry for the delay.

lucacasonato:
Fantastic, thank you folks! I'll open implementation bugs shortly so we can get this landed 😃

index.bs
```diff
@@ -5726,6 +5747,16 @@ the following table:
    <th>Internal Slot</th>
    <th>Description (<em>non-normative</em>)</th>
  <tbody>
+  <tr>
+   <td><dfn>\[[cancelAlgorithm]]</dfn>
+   <td class="non-normative">A promise-returning algorithm, taking one argument (the [=reason=] for
```
domenic, line 5752:
```suggestion
   <td class="non-normative">A promise-returning algorithm, taking one argument (the reason for
```

index.bs
```diff
@@ -6062,6 +6132,36 @@ side=] of [=transform streams=].
  1. Return |stream|.[=TransformStream/[[backpressureChangePromise]]=].
 </div>
 
+
+<div algorithm>
+ <dfn abstract-op lt="TransformStreamDefaultSourceCancelAlgorithm"
+ id="transform-stream-default-source-cancel">TransformStreamDefaultSourceCancelAlgorithm(|stream|,
+ |reason|)</dfn> performs the following steps:
```
domenic, line 6139: Reorder to be alphabetically before the previous one

index.bs
```diff
@@ -7106,9 +7206,10 @@ reason.
 <div algorithm="create a TransformStream">
  To <dfn export for="TransformStream" lt="set up|setting up">set up</dfn> a
  newly-[=new|created-via-Web IDL=] {{TransformStream}} |stream| given an algorithm <dfn export
- for="TransformStream/set up"><var>transformAlgorithm</var></dfn> and an optional algorithm <dfn
- export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, perform the following steps.
- |transformAlgorithm| and, if given, |flushAlgorithm|, may return a promise.
+ for="TransformStream/set up"><var>transformAlgorithm</var></dfn>, an optional algorithm <dfn
+ export for="TransformStream/set up"><var>flushAlgorithm</var></dfn>, and an optional algorithm <dfn
+ export for="TransformStream/set up"><var>cancelAlgorithm</var></dfn>, perform the following steps.
+ |transformAlgorithm|, if given, |flushAlgorithm| and |cancelAlgorithm|, may return a promise.
```
domenic, line 7212:
```suggestion
 |transformAlgorithm| and, if given, |flushAlgorithm| and |cancelAlgorithm|, may return a promise.
```

domenic (approved):
Found a few editorial nits, but looks good! Thanks so much for working on this.

As you may or may not be aware, because of the reference implementation landing things in the streams repo is a bit annoying. First we have to land the tests PR, then we roll the submodule, then we land this PR. I'll go merge the tests PR now.

lucacasonato:
All implementation bugs and the MDN issue have been filed. This PR is now ready to land :)

domenic:
I think all your issues have the wrong titles, using "cleanup" instead of "cancel" :). I'll merge this though.

lucacasonato:
@domenic The cancel hook allows for cleanup of transformers backing TransformStreams, so I think it makes some sense?

domenic:
There is no "transformer.cleanup" property though.

lucacasonato:
Oh I realize my mistake now. Will fix. Thx

PR #1306, opened 2024-02-19
Add explainer for ReadableStream async iteration

MattiasBuelens:
Although the spec for `ReadableStream` async iteration has already been finalized four years ago in #980 and #1035, we never wrote a proper explainer for it. And as part of implementing [`ReadableStream` async iteration in Chromium](https://chromestatus.com/feature/5143121161879552), I need a link to an explainer. So here it is! 😁


ricea (approved):
lgtm, thanks!

PR #1310, opened 2024-03-25
Use `async iterable` webidl type in ReadableStream.from

lucacasonato:
Ref https://github.com/whatwg/webidl/pull/1397

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
- [ ] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/streams/1310.html" title="Last updated on Jul 18, 2024, 8:51 AM UTC (28809cb)">Preview</a> | <a href="https://whatpr.org/streams/1310/007d729...28809cb.html" title="Last updated on Jul 18, 2024, 8:51 AM UTC (28809cb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2123,7 +2123,7 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ <dfn abstract-op export lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
  ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
 
  1. Let |stream| be undefined.
@@ -2137,8 +2137,8 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
     1. Otherwise, perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |v|).
    - If |next| was rejected with reason |r|, perform ! [$ReadableStreamDefaultControllerError$](|stream|.[=ReadableStream/[[controller]]=], |r|).
  1. Let |cancelAlgorithm| be the following steps, given |reason|:
-  1. Let |finishPromise| be the result of [=finishing iteration/finish iterating an async iterable=]
-     of |asyncIterable|.
+  1. Let |finishPromise| be the result of [=closing/close an async iterable=] |asyncIterable| with
+     reason |reason|.
   1. Return the result of [=reacting=] to |finishPromise|:
     - If |finishPromise| was fulfilled, return undefined.
     - If |finishPromise| was rejected with reason |r|, throw |r|.
--- a/index.bs
+++ b/index.bs
@@ -2123,7 +2123,7 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
 </div>
 
 <div algorithm>
- <dfn abstract-op export lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
  ReadableStreamFromIterable(|asyncIterable|)</dfn> performs the following steps:
 
  1. Let |stream| be undefined.
--- a/index.bs
+++ b/index.bs
@@ -488,7 +488,7 @@ The Web IDL definition for the {{ReadableStream}} class is given as follows:
 interface ReadableStream {
   constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
-  static ReadableStream from(any asyncIterable);
+  static ReadableStream from(async iterable<any> asyncIterable);
 
   readonly attribute boolean locked;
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2127,39 +2127,21 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
...
+    1. If |v| is [=end of iteration=], perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
+    1. Otherwise, perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=], |v|).
+   - If |next| was rejected with reason |r|, perform ! [$ReadableStreamDefaultControllerError$](|stream|.[=ReadableStream/[[controller]]=], |r|).
  1. Let |cancelAlgorithm| be the following steps, given |reason|:
-  1. Let |iterator| be |iteratorRecord|.\[[Iterator]].
-  1. Let |returnMethod| be [$GetMethod$](|iterator|, "`return`").
-  1. If |returnMethod| is an abrupt completion, return [=a promise rejected with=]
-     |returnMethod|.\[[Value]].
-  1. If |returnMethod|.\[[Value]] is undefined, return [=a promise resolved with=] undefined.
-  1. Let |returnResult| be [$Call$](|returnMethod|.\[[Value]], |iterator|, « |reason| »).
```
MattiasBuelens, line 2155: The `reason` argument for the `return()` call is missing in the new spec text.

I propose we add an optional argument to "finish iterating an async iterable" to support this use case, see [my comment on the other PR](https://github.com/whatwg/webidl/pull/1397#discussion_r1538889449).

index.bs
```diff
@@ -2127,39 +2127,21 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
...
-   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
-   1. Let |done| be ? [$IteratorComplete$](|iterResult|).
-   1. If |done| is true:
-    1. Perform ! [$ReadableStreamDefaultControllerClose$](|stream|.[=ReadableStream/[[controller]]=]).
-   1. Otherwise:
-    1. Let |value| be ? [$IteratorValue$](|iterResult|).
-    1. Perform ! [$ReadableStreamDefaultControllerEnqueue$](|stream|.[=ReadableStream/[[controller]]=],
-       |value|).
-       <!-- TODO (future): If we allow changing the queuing strategy, this Enqueue might throw.
-             We'll then need to catch the error and close the async iterator. -->
```
MattiasBuelens, line 2148: Could you restore this comment in the updated spec text? It's still relevant.

MattiasBuelens (changes requested):
Looks good so far! 👍 

Once the Web IDL spec change lands, we'll need to update [webidl2js](https://github.com/jsdom/webidl2js) to implement the new algorithms for async iterables, and then we can update our reference implementation. Let's keep this PR as a draft until that's sorted out.

lucacasonato:
Moving over the question of whether this API should support `async iterable<T> or DOMString`, or just `async iterable<T>` as an argument (from https://github.com/whatwg/webidl/pull/1397#issuecomment-2020386670). In the current streams spec, strings are implicitly allowed because they implement the async iterable protocol. The new `async iterable<T>` Web IDL type however, only accepts _objects_ that implement the async iterable protocol. The reason for this is because anything else would realistically not be possible due to overloading behaviour in Web IDL. That point is not the point of discussion here.

The question posed is, "Should `ReadableStream.from("foobar")` be valid, and if so, what should it do?"

I think there are three possible outcomes here:
1. We disallow this behaviour and throw in this case. This mirrors the behaviour of `Array.prototype.flatMap` of not iterating over strings implicitly.
2. We allow this behaviour, and call `[Symbol.iterator]` on the string. Each chunk returned from the iterator is enqueued individually into the stream. This mirrors the behaviour of `Iterator.from()`.
3. We allow this behaviour, and enqueue the entire string as one chunk into the stream. This mirrors the behaviour of `new Response("foobar").body`.

The current behaviour of the spec is option 2. Option 1 would be a more significant breaking change than option 3, but both are unlikely to have web compat issues. I say this because a) `ReadableStream.from()` is very new, and b) I have not seen any use cases where anyone actually wants to get a stream of individual string chunks.

If we want to support passing strings, option 3 would likely be faster than option 2, because of the lower number of enqueue operations, string allocations, and object allocations involved. It would also mirror the behaviour of `new Reponse("foobar").body`.

On the other hand, if we do not support strings at all (option 1), users would never be suprised by the choice we made between 1 and 2. Instead, they always get an explicit error that they can then deal with themselves. Both options are still trivially expressible using minimal changes by the developer. If they wanted to emulate the behaviour of option 2, they can do `ReadableStream.from(Iterator.from("foobar"))`, and if they want to emulate option 3, they can do `ReadableStream.from(["foobar"])`.

My preference is to option 1, followed by option 2 (if we want to support strings at all). I acknowledge that there would then be divergence between `new Response("string").body` and `ReadableStream.from("string")`, which _is_ unfortunate, however I think a divergence between the `from` method of the streaming primitive `Iterator`/`AsyncIterator`, and the other streaming primitive `ReadableStream` would be unfortunate.

cc @domenic @bakkot

saschanaz:
Mozilla has no use counter for string case. I can't imagine the use case either, but might be worth checking first.

annevk:
@ricea @saschanaz can we take your comments to count as implementer interest for this change? That would also unblock the IDL change.

ricea:
Yes, you can consider Blink interested.

saschanaz:
Also LGTM for stream change.

index.bs
```diff
@@ -2123,43 +2123,27 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
...
-  1. Let |returnPromise| be [=a promise resolved with=] |returnResult|.\[[Value]].
-  1. Return the result of [=reacting=] to |returnPromise| with the following fulfillment steps,
-     given |iterResult|:
-   1. If [$Type$](|iterResult|) is not Object, throw a {{TypeError}}.
-   1. Return undefined.
+  1. Let |finishPromise| be the result of [=closing/close an async iterable=] |asyncIterable| with
+     reason |reason|.
+  1. Return the result of [=reacting=] to |finishPromise|:
+    - If |finishPromise| was fulfilled, return undefined.
+    - If |finishPromise| was rejected with reason |r|, throw |r|.
```
MattiasBuelens, line 2146: Nit-pick: can we return `finishPromise` as-is? It looks like [the close algorithm in IDL](https://whatpr.org/webidl/1397.html#async-iterable-close) already does the right thing.

MattiasBuelens (commented):
Still LGTM, with one small nit.

index.bs
```diff
@@ -2123,43 +2123,27 @@ The following abstract operations operate on {{ReadableStream}} instances at a h
 </div>
 
 <div algorithm>
- <dfn abstract-op lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
+ <dfn abstract-op export lt="ReadableStreamFromIterable" id="readable-stream-from-iterable">
```
saschanaz, line 2126: Why export, btw?

lucacasonato: Will be needed to resolve https://github.com/whatwg/fetch/issues/1291, which is my next project :)

saschanaz: In that case you would want a separate op in section 9, this one is in section 4 and not supposed to be directly used by other specs.

lucacasonato: Ok, will revert this here

lucacasonato:
@annevk I did not have time yet to finish the WebIDL PR - I am doing that now though :)

MattiasBuelens:
@lucacasonato Do you mind if I pick this up again?

I noticed that the specification changed a little bit: `async iterable<T>` was renamed to `async_sequence<T>` in https://github.com/whatwg/webidl/pull/1500. This is mostly a syntactic change, the behavior should remain the same.

I've opened https://github.com/jsdom/webidl2js/pull/278. Once that lands, we can update the reference implementation to let webidl2js handle the conversion.

MattiasBuelens:
Fixed in https://github.com/whatwg/streams/pull/1372.

Issue #1317, opened 2024-07-04
WebRTC endpoint pair example is out of date


jan-ivar:
### What is the issue with the Streams Standard?

The https://streams.spec.whatwg.org/#example-endpoint-pair-webrtc example is out of date. The API it mentions was removed in https://github.com/w3c/webrtc-encoded-transform/pull/64. The name of the spec has also changed.

A direct update would be something like:
> [WebRTC Encoded Transform](https://w3c.github.io/webrtc-encoded-transform/) is an example of this technique, with its [RTCRtpScriptTransformer](https://w3c.github.io/webrtc-encoded-transform/#rtcrtpscripttransformer) interface which has both `readable` and `writable` as members.

PR #1318, opened 2024-07-09
Update endpoint pair example and fix citation

jan-ivar:
Fixes https://github.com/whatwg/streams/issues/1317 (editorial).

***
<a href="https://whatpr.org/streams/1318.html" title="Last updated on Jul 11, 2024, 9:49 PM UTC (cecef0e)">Preview</a> | <a href="https://whatpr.org/streams/1318/53f7c7d...cecef0e.html" title="Last updated on Jul 11, 2024, 9:49 PM UTC (cecef0e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7459,19 +7459,19 @@ readable and writable streams represent the two ends of a longer pipeline, with 
 web developer code insert [=transform streams=] into the middle of them.
 
 <div class="example" id="example-endpoint-pair-usage">
- Assuming we had a web-platform-provided function `createEndpointPair()`, web developers would write
- code like so:
+ Assuming we had a web-platform-provided interface `Transformer` that had `readable` and
+ `writable` attributes, web developers would write code like so:
 
  <xmp highlight="js">
- const { readable, writable } = createEndpointPair();
- readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
+ async function handleTransformer({transformer: {readable, writable}}) {
+   await readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
+ }
  </xmp>
 </div>
 
-<p class="example" id="example-endpoint-pair-webrtc"><cite>WebRTC Insertable Media using
-Streams</cite> is an example of this technique, with its `sender.createEncodedStreams()` and
-`receiver.createEncodedStreams()` methods.
-<!-- TODO cite it and cross-link to it https://github.com/tobie/specref/issues/620 -->
+<p class="example" id="example-endpoint-pair-webrtc"><cite>WebRTC Encoded Transform</cite>
+is an example of this technique, with its {{RTCRtpScriptTransformer}} interface which has
+both `readable` and `writable` attributes.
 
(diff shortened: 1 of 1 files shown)


ricea (approved):
Thanks. lgtm.

index.bs
```diff
@@ -7459,19 +7459,19 @@ readable and writable streams represent the two ends of a longer pipeline, with
...
+ Assuming we had a web-platform-provided interface `Transformer` that had `readable` and
+ `writable` attributes, web developers would write code like so:
 
  <xmp highlight="js">
- const { readable, writable } = createEndpointPair();
- readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
+ async function handleTransformer({readable, writable}) {
+   await readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
+ }
  </xmp>
```
domenic, line 7469: This example has been simplified a bit too much I think. It talks about a platform-provided `Transformer` interface but that interface is now super-implicit, because of the destructuring of function arguments. I think reverting to something like

```js
const { readable, writable } = createTransformer();
readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
```

would work better.

Additionally, the point of `createEndpointPair()` was to match the official name given in the standard, "endpoint pair". So I think that that was even better.

Can you say more about why you changed this example?

domenic (changes requested):

index.bs
jan-ivar: I read _"web-platform-provided function `createEndpointPair()`"_ as a callout to the same-shaped `sender.createEncodedStreams()` method removed in https://github.com/w3c/webrtc-encoded-transform/pull/64.

I was trying to maintain this similarity to the API it's referencing. But the [spec](https://w3c.github.io/webrtc-encoded-transform/#rtcrtpscripttransformer)'s [worker-only](https://w3ctag.github.io/design-principles/#worker-only) API is different. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_Encoded_Transforms#basic_webrtc_encoded_transform) and [blog post](https://blog.mozilla.org/webrtc/end-to-end-encrypt-webrtc-in-all-browsers/) examples:
```js
// main.html
const worker = new Worker("worker.js");
sender.transform = new RTCRtpScriptTransform(worker);

// worker.js
onrtctransform = async ({transformer: {readable, writable}}) => {
  await readable.pipeThrough(new TransformStream(...)).pipeTo(writable);
};
```
The spec fires an event with the transformer which has no constructor. The event part seemed unnecessary but I can try to add it if the context is useful.

domenic: Sorry, this seems to have gone in the opposite direction of what I'm requesting. The point of this example is not to reference any other API. It's to give an example of the simplest possible concept of an endpoint pair, i.e. a platform-provided API that returns a `{ readable, writable }` object.

Can you revert it to its previous version?

domenic (approved):
Thank you very much!

PR #1171, opened 2021-10-05
Discard auto-allocated BYOB request on enqueue

MattiasBuelens:
As discovered in [#1170 (comment)](https://github.com/whatwg/streams/issues/1170#issuecomment-933822324): if a BYOB request was auto-allocated, but then you call `controller.enqueue()` instead of `byobRequest.respond()`, then the auto-allocated pull-into descriptor stays in the queue. If you later on switch from a default reader to a BYOB reader, you run into trouble because the list of read-into requests no longer matches up with the list of pending pull-into descriptors.

This PR fixes it by making `enqueue()` discard that auto-allocated BYOB request.

- [ ] At least two implementers are interested (and none opposed):
   * Chrome (`@ricea`)
   * …
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * web-platform-tests/wpt#31128
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/1257465
   * Firefox: N/A (does not yet support readable byte streams)
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=231628

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/streams/1171.html" title="Last updated on Oct 12, 2021, 10:02 PM UTC (1b5dd22)">Preview</a> | <a href="https://whatpr.org/streams/1171/8bbfbf8...1b5dd22.html" title="Last updated on Oct 12, 2021, 10:02 PM UTC (1b5dd22)">Diff</a>

PR #1326, opened 2024-09-10
Commit pull-into descriptors after filling from queue

MattiasBuelens:
In [Chromium bug #339877167](https://issues.chromium.org/issues/339877167), it was discovered that a user could run JavaScript code *synchronously* during `ReadableStreamFulfillReadIntoRequest` by patching `Object.prototype.then`, and use this gadget to break some invariants within `ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue`.

To prevent this, we now postpone all calls to `ReadableByteStreamControllerCommitPullIntoDescriptor` until *after* all pull-into descriptors have been filled up by `ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue`. This way, we won't trigger any patched `then()` method until the stream is in a stable state.

Fixes GHSA-p5g2-876g-95h9.

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/48085
   * Chromium: https://issues.chromium.org/issues/381211736
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1933826
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=283771
   * Deno: https://github.com/denoland/deno/issues/27117
   * Node.js: https://github.com/nodejs/node/issues/56044
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/streams/1326.html" title="Last updated on Nov 27, 2024, 9:55 PM UTC (1391f21)">Preview</a> | <a href="https://whatpr.org/streams/1326/86d07e9...1391f21.html" title="Last updated on Nov 27, 2024, 9:55 PM UTC (1391f21)">Diff</a>

### DIFF
--- a/reference-implementation/lib/abstract-ops/readable-streams.js
+++ b/reference-implementation/lib/abstract-ops/readable-streams.js
@@ -1676,9 +1676,13 @@ function ReadableByteStreamControllerRespondInClosedState(controller, firstDescr
 
   const stream = controller._stream;
   if (ReadableStreamHasBYOBReader(stream) === true) {
+    const filledPullIntos = [];
     while (ReadableStreamGetNumReadIntoRequests(stream) > 0) {
       const pullIntoDescriptor = ReadableByteStreamControllerShiftPendingPullInto(controller);
-      ReadableByteStreamControllerCommitPullIntoDescriptor(stream, pullIntoDescriptor);
+      filledPullIntos.push(pullIntoDescriptor);
+    }
+    for (const filledPullInto of filledPullIntos) {
+      ReadableByteStreamControllerCommitPullIntoDescriptor(stream, filledPullInto);
     }
   }
 }
--- a/reference-implementation/lib/abstract-ops/readable-streams.js
+++ b/reference-implementation/lib/abstract-ops/readable-streams.js
@@ -1533,9 +1533,10 @@ function ReadableByteStreamControllerInvalidateBYOBRequest(controller) {
 function ReadableByteStreamControllerProcessPullIntoDescriptorsUsingQueue(controller) {
   assert(controller._closeRequested === false);
 
+  const filledPullIntos = [];
   while (controller._pendingPullIntos.length > 0) {
     if (controller._queueTotalSize === 0) {
-      return;
+      break;
     }
 
     const pullIntoDescriptor = controller._pendingPullIntos[0];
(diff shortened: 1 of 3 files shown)


reference-implementation/lib/abstract-ops/readable-streams.js
```diff
@@ -1443,7 +1447,14 @@ function ReadableByteStreamControllerFillPullIntoDescriptorFromQueue(controller,
...
     const destStart = pullIntoDescriptor.byteOffset + pullIntoDescriptor.bytesFilled;
-    CopyDataBlockBytes(pullIntoDescriptor.buffer, destStart, headOfQueue.buffer, headOfQueue.byteOffset, bytesToCopy);
+
+    if (SafeCopyDataBlockBytes(pullIntoDescriptor.buffer, destStart, headOfQueue.buffer, headOfQueue.byteOffset,
+                               bytesToCopy) === false) {
+      // This should never happen. Please report an issue if it does! https://github.com/whatwg/streams/issues
+      const e = new TypeError('Invalid buffer');
+      ReadableByteStreamControllerError(controller, e);
+      return false;
+    }
```
MattiasBuelens, line 1457: This check should *always* return `true`, so we can't write a test to check if an implementation correctly handles the case where it's `false`. That said, this check is our last chance to prevent a potential memory corruption bug.

Is it better to have a bit of untestable code here, or to only keep it as an assertion and hope that we don't break this again in the future? @domenic @saschanaz

domenic: My preference is an assertion. Perhaps in the spec it's an assertion with a big `<p class="warning">` suggesting that implementers really might want to consider crashing if the assertions somehow gets violated?

MattiasBuelens: I changed it to an assertion with a big warning.

domenic (approved):
Not sure how this fell off my queue. Stellar work as always. Let's work on getting those checkboxes in the OP checked and tests rolled...

MattiasBuelens:
I'll start filing implementation bugs.

@domenic Do we need implementer's interest for this? In https://github.com/whatwg/streams/pull/1171#issuecomment-941635222 we ruled it wasn't necessary for a bugfix.

domenic:
I guess I would feel better if we could get a signoff from at least one of Gecko (@saschanaz? @evilpie?) and Chromium (@ricea). We can assume that WebKit is OK with this since they still don't implement readable byte streams, so they should be happy to write their eventual implementation with bugs fixed. But since Chromium and Gecko would be changing their mitigation strategy slightly to pass the new tests, let's get at least one of them to sign off on it. @ricea is usually pretty responsive anyway.

ricea:
Chromium supports.

nidhijaju:
+1, thank you for fixing this in the spec!

PR #1339, opened 2025-01-23
Add WritableStream.write convenience method

lucacasonato:
Add a WriteableStream.write convenience method to write a single chunk. It locks, writes, then unlocks all in one call.

   * Deno
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4064,6 +4064,7 @@ interface WritableStream {
 
   readonly attribute boolean locked;
 
+  Promise<undefined> write(optional any chunk);
   Promise<undefined> abort(optional any reason);
   Promise<undefined> close();
   WritableStreamDefaultWriter getWriter();
@@ -4270,6 +4271,17 @@ as seen for example in [[#example-ws-no-backpressure]].
  <dd>
   <p>Returns whether or not the writable stream is [=locked to a writer=].
 
+ <dt><code>await <var ignore>stream</var>.{{WritableStream/write(chunk)|write}}([ <var ignore>chunk</var> ])</code>
+ <dd>
+  <p>Writes a [=chunk=] of data to the stream just like the {{WritableStreamDefaultWriter/write()}}
+   method.
+
+  <p>This is a convenience method that first acquires a writer, then writes the chunk, and releases
+   the writer (regardless of the outcome of the write). The returned promise will fulfill if the
+   write was successful, or reject if the write failed. Additionally, it will reject with a
+   {{TypeError}} if the stream is currently [=locked to a writer|locked=] (without attempting to
+   write).
+
  <dt><code>await <var ignore>stream</var>.{{WritableStream/abort(reason)|abort}}([ <var ignore>reason</var> ])</code>
  <dd>
   <p>[=abort a writable stream|Aborts=] the stream, signaling that the producer can no longer
@@ -4330,6 +4342,22 @@ as seen for example in [[#example-ws-no-backpressure]].
  The <dfn id="ws-locked" attribute for="WritableStream">locked</dfn> getter steps are:
 
  1. Return ! [$IsWritableStreamLocked$]([=this=]).
+</div>
+
+<div algorithm>
+ The <dfn id="ws-write" method for="WritableStream">write(|chunk|)</dfn> method steps are:
+
+ 1. Let |result| be ? [$AcquireWritableStreamDefaultWriter$]([=this=]).
+ 1. If |result| is an abrupt completion, return [=a promise rejected with=] |result|.\[[Value]].
+ 1. Let |writer| be |result|.\[[Value]].
+ 1. Let |writePromise| be ! [$WritableStreamDefaultWriterWrite$]([=this=], |chunk|).
+ 1. Return the result of [=reacting=] to |writePromise|:
+  1. If |writePromise| is fulfilled with a value |v|, then:
+   1. Perform ! [$WritableStreamDefaultWriterRelease$]([=this=]).
+   1. Return |v|.
+  1. If |writePromise| is rejected with a value |r|, then:
+   1. Perform ! [$WritableStreamDefaultWriterRelease$]([=this=]).
+   1. Return [=a promise rejected with=] |r|.
 </div>
 
 <div algorithm>


index.bs
```diff
@@ -4332,6 +4344,22 @@ as seen for example in [[#example-ws-no-backpressure]].
  1. Return ! [$IsWritableStreamLocked$]([=this=]).
 </div>
 
+<div algorithm>
+ The <dfn id="ws-write" method for="WritableStream">write(|chunk|)</dfn> method steps are:
```
saschanaz, line 4348: Would we want to imply from the name that each write would be exclusive and extra parallel calls would reject?

lucacasonato: I just realized the behavior is is wrong. The writer should be released immediately, not once the write promise resolves. This mirrors the behavior of the write method on https://fs.spec.whatwg.org/#filesystemwritablefilestream

jasnell:
The fact that parallel writes wouldn't be allowed (since each write locks the writable) is a bit of a challenge here. I'm not entirely convinced that this is the right thing to add. Can you expand on the motivating use cases a bit more?

index.bs
saschanaz: That sounds better!

jasnell: Releasing the writer works so long as releasing it does not lead the write being canceled. I believe it's fine now but I believe a [change](https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreadererrorreadrequests) was recently made to `ReadableStream` that caused pending reads to be canceled. Are we certain a similar type of change won't be wanted with writers and would this interfere with that?

MattiasBuelens: That's very unlikely to change.

For readers, we previously did not allow you to `releaseLock()` while there were pending read requests. For writers, we never had that restriction: you're always allowed to call `releaseLock()` even when there are pending write requests. In fact, [we even recommend](https://streams.spec.whatwg.org/#example-manual-write-dont-await) not awaiting the `writer.write()` promise so you can better respect backpressure (as long as you await `writer.ready` though).

index.bs
```diff
@@ -4332,6 +4344,22 @@ as seen for example in [[#example-ws-no-backpressure]].
...
+ 1. If |result| is an abrupt completion, return [=a promise rejected with=] |result|.\[[Value]].
+ 1. Let |writer| be |result|.\[[Value]].
+ 1. Let |writePromise| be ! [$WritableStreamDefaultWriterWrite$](|writer|, |chunk|).
+ 1. Return the result of [=reacting=] to |writePromise|:
+  1. If |writePromise| is fulfilled with a value |v|, then:
+   1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
+   1. Return |v|.
+  1. If |writePromise| is rejected with a value |r|, then:
+   1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
+   1. Return [=a promise rejected with=] |r|.
```
MattiasBuelens, line 4360: We don't need to wait for the `writePromise` to resolve or reject before releasing the lock. Instead, we should release it *immediately*.
```suggestion
 1. Let |writePromise| be ! [$WritableStreamDefaultWriterWrite$](|writer|, |chunk|).
 1. Perform ! [$WritableStreamDefaultWriterRelease$](|writer|).
 1. Return |writePromise|.
```
This works, because [WritableStreamDefaultWriterRelease](https://streams.spec.whatwg.org/#writable-stream-default-writer-release) does not reject any pending write requests.

This also resolves https://github.com/whatwg/streams/pull/1339#issuecomment-2611228239.

index.bs
```diff
@@ -4332,6 +4344,22 @@ as seen for example in [[#example-ws-no-backpressure]].
  1. Return ! [$IsWritableStreamLocked$]([=this=]).
 </div>
 
+<div algorithm>
+ The <dfn id="ws-write" method for="WritableStream">write(|chunk|)</dfn> method steps are:
+
+ 1. Let |result| be ? [$AcquireWritableStreamDefaultWriter$]([=this=]).
```
MattiasBuelens, line 4350: It might be worth introducing a new abstract op `WritableStreamWrite(stream, chunk)` for doing a "writer-less write", similar to [`WritableStreamAbort`](https://streams.spec.whatwg.org/#writable-stream-abort) and [`WritableStreamClose`](https://streams.spec.whatwg.org/#writable-stream-close). We could then rework [`WritableStreamDefaultWriterWrite`](https://streams.spec.whatwg.org/#writable-stream-default-writer-write) to use the same abstract op, similar to how [`WritableStreamDefaultWriterAbort`](https://streams.spec.whatwg.org/#writable-stream-default-writer-abort) delegates to `WritableStreamAbort`.

(We may have to take `chunkSize` as an extra argument though, since [step 5 of `WritableStreamDefaultWriterWrite`](https://streams.spec.whatwg.org/commit-snapshots/2811932a34b5cfdb0a6e33ad204f747ee9c71292/#writable-stream-default-writer-write) might be difficult to implement otherwise.)

lucacasonato: I will do that if I can get implementer interest :)

lucacasonato:
I'm looking for implementer interest for this change. How are Chromium, Gecko, WebKit, and Cloudflare folks feeling about this? @saschanaz @domenic @jasnell @annevk

lucacasonato:
@jasnell Yeah see the first thread on this PR. I changed it to do `acquire(); write(); release(); waitForWrite();` now.

annevk:
Nit: please change the commit/PR title to make this not look like it's adding a static method.

youennf:
Is it the same as https://fs.spec.whatwg.org/#dom-filesystemwritablefilestream-write?

lucacasonato:
@youennf Yes

saschanaz:
I'm interested to have this to replace Bucket FS side equivalent. cc @janvarga

janvarga:
delegating to @jjjalkanen

jasnell:
@lucacasonato Cloudflare will implement if this moves forward. I'm not super convinced of the use cases yet as I haven't had anyone ask for this but I certainly don't have a reason not to support it.

annevk:
If this is happening we should probably remove `FileSystemWritableFileStream#write` at the same time.

domenic:
I'm in favor of this but for Chromium it's @ricea and @nidhijaju who would be doing the work, so please wait for one of them to confirm.

index.bs
```diff
@@ -4270,6 +4271,17 @@ as seen for example in [[#example-ws-no-backpressure]].
...
   <p>Returns whether or not the writable stream is [=locked to a writer=].
 
+ <dt><code>await <var ignore>stream</var>.{{WritableStream/write(chunk)|write}}([ <var ignore>chunk</var> ])</code>
+ <dd>
+  <p>Writes a [=chunk=] of data to the stream just like the {{WritableStreamDefaultWriter/write()}}
+   method.
+
+  <p>This is a convenience method that first acquires a writer, queues the chunk to be written,
+   releases the writer (without waiting on the write to complete), and returns a promise that
+   fulfills if the write was successful, or reject if the write failed. Additionally, if the stream
```
nidhijaju, line 4281: minor nit:
```suggestion
   fulfills if the write was successful, or rejects if the write failed. Additionally, if the stream
```

nidhijaju (commented):
This seems okay, and I have no objections, but I haven't heard a ton of requests for this, which may affect when Chrome implements this.

jjjalkanen:
It seems to me that no major changes to the Gecko implementation are required, and I don't have any objections to this proposal.

ricea:
I weakly support. We have no resources to commit to this.

PR #1355, opened 2025-09-01
Meta: move explainers to subdirectory

MattiasBuelens:
I think new explainers should go in an `explainers/` subdirectory instead of the root of the repository.

~~To avoid breaking deeplinks, all existing explainers have a link to the new location at their old location. These are just Markdown files with a link in them. (I tried creating symlinks instead [in this branch](https://github.com/MattiasBuelens/streams/tree/move-explainers-symlinks), but [GitHub doesn't even render a link for those](https://github.com/MattiasBuelens/streams/blob/move-explainers-symlinks/byte-streams-explainer.md)...)~~


domenic:
If we're going to do this, I think it's OK to break the old links. They're relatively easy to find from the 404 page.

I'm happy to be overruled though if people would really prefer.

MattiasBuelens:
@domenic That also works for me. PR updated.

PR #1368, opened 2026-04-21
Remove abstract-op

### LINKED ISSUE #1365: Do not mark abstract operations with `data-export`
### What is the issue with the Streams Standard?

Those are intended to be internal and should not be marked as data-export. But things like https://streams.spec.whatwg.org/#readable-stream-cancel gets:

```html
<dfn class="dfn-paneled has-dfn-panel" data-dfn-type="abstract-op" data-export="" data-lt="ReadableStreamCancel" id="readable-stream-cancel" role="button" aria-expanded="true" tabindex="100">ReadableStreamCancel(<var>stream</var>, <var>reason</var>)<button class="specmonkey-box-button"><img src="moz-extension://19959b45-e51e-4132-a315-badf5b67dba2/searchfox.png" alt="SpecMonkey" class="specmonkey-box-icon"></button></dfn>
```

Not quite sure where the `data-export` is coming from, it's not in the source:

https://github.com/whatwg/streams/blob/76113627313e3e81e0d2c80141b9da6f84881e5f/index.bs#L2711-L2712

Perhaps the explicit ID is somehow making it exported?

saschanaz:
Fixes #1365

One alternative way is to add `no-export` to all abstract-ops, but given web specs don't actively use abstract-op, I think we can just remove it.

***
<a href="https://whatpr.org/streams-1/1368.html" title="Last updated on Apr 21, 2026, 12:03 PM UTC (4d69cb5)">Preview</a> | <a href="https://whatpr.org/streams/1368/7611362...4d69cb5.html" title="Last updated on Apr 21, 2026, 12:03 PM UTC (4d69cb5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -810,17 +810,17 @@ option. If {{UnderlyingSource/type}} is set to undefined (including via omission
     <p class="note">We cannot declare the |underlyingSource| argument as having the
     {{UnderlyingSource}} type directly, because doing so would lose the reference to the original
     object. We need to retain the object so we can [=invoke=] the various methods on it.
- 1. Perform ! [$InitializeReadableStream$]([=this=]).
+ 1. Perform ! [=InitializeReadableStream=]([=this=]).
  1. If |underlyingSourceDict|["{{UnderlyingSource/type}}"] is "{{ReadableStreamType/bytes}}":
   1. If |strategy|["{{QueuingStrategy/size}}"] [=map/exists=], throw a {{RangeError}} exception.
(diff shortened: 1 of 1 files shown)


annevk:
But these follow all the conventions of abstract operations so making them the same as definitions which are "I am a definition" and not "IAmADefinition" seems wrong.

saschanaz:
But changing the names sounds also not super great, the reference impl uses them, right?

saschanaz:
And calling algorithms with parentheses also sound not right, maybe adding no-export is easier.

saschanaz:
Alright, let's do that instead.

PR #1369, opened 2026-04-21
Add `noexport` for abstract operations

### LINKED ISSUE #1365: Do not mark abstract operations with `data-export`

saschanaz:
Fixes #1365

***
<a href="https://whatpr.org/streams-1/1369.html" title="Last updated on Apr 21, 2026, 3:21 PM UTC (8ed9962)">Preview</a> | <a href="https://whatpr.org/streams/1369/7611362...8ed9962.html" title="Last updated on Apr 21, 2026, 3:21 PM UTC (8ed9962)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1633,7 +1633,7 @@ The readable stream implementation will polymorphically call to either these, or
 counterparts for BYOB controllers, as discussed in [[#rs-abstract-ops-used-by-controllers]].
 
 <div algorithm>
- <dfn abstract-op no-export lt="[[CancelSteps]]" for="ReadableStreamDefaultController"
+ <dfn abstract-op noexport lt="[[CancelSteps]]" for="ReadableStreamDefaultController"
  id="rs-default-controller-private-cancel">\[[CancelSteps]](|reason|)</dfn> implements the
  [$ReadableStreamController/[[CancelSteps]]$] contract. It performs the following steps:
 
@@ -1645,7 +1645,7 @@ counterparts for BYOB controllers, as discussed in [[#rs-abstract-ops-used-by-co
 </div>
 
 <div algorithm>
- <dfn abstract-op no-export lt="[[PullSteps]]" for="ReadableStreamDefaultController"
+ <dfn abstract-op noexport lt="[[PullSteps]]" for="ReadableStreamDefaultController"
  id="rs-default-controller-private-pull">\[[PullSteps]](|readRequest|)</dfn> implements the
  [$ReadableStreamController/[[PullSteps]]$] contract. It performs the following steps:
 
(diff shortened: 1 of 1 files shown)


annevk (approved):
Assuming you checked no spec depends on these, seems fine. Encoding and Fetch both come to mind.

MattiasBuelens:
Encoding uses the following terms [(source)](https://encoding.spec.whatwg.org/#index-defined-elsewhere):
- GenericTransformStream
- ReadableStream
- TransformStream
- chunk
- enqueue
- flushAlgorithm
- pipeThrough(transform)
- readable
- readable stream
- set up
- transform
- transformAlgorithm
- writable
- writable stream 

Fetch uses the following terms [(source)](https://fetch.spec.whatwg.org/#index-defined-elsewhere):
- ReadableStream
- ReadableStreamDefaultReader
- TransformStream
- cancel
- cancelAlgorithm
- chunk steps
- close
- close steps
- creating a proxy
- disturbed
- enqueue (for ReadableStream)
- enqueue (for TransformStream)
- error
- error steps
- errored
- flushAlgorithm
- get a reader
- getReader()
- getting a reader
- identity transform stream
- locked
- piped through
- pull from bytes
- pullAlgorithm
- read a chunk
- read all bytes
- read request
- readable
- set up
- set up with byte reading support
- teeing
- transformAlgorithm 

Those all look fine to me.

Do we have tools to find which specifications use definitions from a specific standard? I know about [WebDex](https://dontcallmedom.github.io/webdex/r.html#ReadableStream%40%40%40%40interface) to find which specifications reference a *specific concept*, but there doesn't seem to be a way to just list all concepts from a single specification and then show which other specifications use those concepts.

MattiasBuelens (approved):

saschanaz:
Do we have other merge blockers? 👀

PR #1370, opened 2026-04-21
Export "set up reader/writer" algorithms for other specs

MattiasBuelens:
These new algorithms allow other specifications to subclass `ReadableStreamDefaultReader` and/or `WritableStreamDefaultWriter`, without depending on the internal `SetUpReadableStreamDefaultReader` and `SetUpWritableStreamDefaultWriter` abstract ops.

Blocks https://github.com/w3c/webtransport/issues/753.

***
<a href="https://whatpr.org/streams/1370.html" title="Last updated on Apr 22, 2026, 9:10 PM UTC (ff88178)">Preview</a> | <a href="https://whatpr.org/streams/1370/7611362...ff88178.html" title="Last updated on Apr 22, 2026, 9:10 PM UTC (ff88178)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7133,6 +7133,16 @@ failures should be handled by the calling specification.
  <p class="note">This will throw an exception if |stream| is already [=ReadableStream/locked=].
 </div>
 
+<div algorithm>
+ <p>To <dfn export for="ReadableStreamDefaultReader">set up</dfn> a newly-[=new|created-via-Web IDL=]
+ {{ReadableStreamDefaultReader}} |reader| for a {{ReadableStream}} |stream|,
+ perform ? [$SetUpReadableStreamDefaultReader$](|reader|, |stream|).
+
+ <p class="note">Subclasses of {{ReadableStreamDefaultReader}} will use the
+ [=ReadableStreamDefaultReader/set up=] operation directly on the [=this=] value inside their
+ constructor steps.</p>
+</div>
+
 <p algorithm>To <dfn export for="ReadableStreamDefaultReader" lt="read a chunk|reading a chunk">read
 a chunk</dfn> from a {{ReadableStreamDefaultReader}} |reader|, given a [=read request=]
 |readRequest|, perform ! [$ReadableStreamDefaultReaderRead$](|reader|, |readRequest|).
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7133,6 +7133,16 @@ failures should be handled by the calling specification.
  <p class="note">This will throw an exception if |stream| is already [=ReadableStream/locked=].
 </div>
 
+<div algorithm>
+ <p>To <dfn export for="ReadableStreamDefaultReader">set up</dfn> a newly-[=new|created-via-Web IDL=]
```
annevk, line 7137: I don't really understand the new bit here and below. If _reader_ is an argument, why is this relevant?

MattiasBuelens: This is following the format for e.g. [set up a `ReadableStream`](https://streams.spec.whatwg.org/#readablestream-set-up). The intention is that you created this object without running the [`new ReadableStreamDefaultReader()` constructor](https://streams.spec.whatwg.org/#default-reader-constructor).

This is needed because (if I understand correctly) the IDL machinery does not automatically call the "super constructor" when you define a constructor for a subclass. If we're wrong about that, then we may want to revise all of our "set up" algorithms. 😅

annevk: I see. I think that's correct although Web IDL construction isn't defined in much detail. I think what would be clearer however is if we can assert something about _reader_ instead that requires someone to use IDL new.

annevk (approved):
I'll approve given that this pattern is pre-existing, but if we can move to an assert I think that would be better as it is a bit confusing to have this prose where you'd only expect a type.

index.bs
saschanaz: I know it's also just an existing phrase but "created-via-Web IDL" is confusing - if you construct via IDL you would run the constructor steps, right? 🤔 But that steps is more like "create an object that implements the interface".

saschanaz (approved):

MattiasBuelens:
I'm not aware of a way to assert that an object was just created with [Web IDL new](https://webidl.spec.whatwg.org/#new) versus one that already had its constructor steps ran. For now, we're just going to have to trust other specifications that they know what they're doing when they want to use these "set up" algorithms.

index.bs
MattiasBuelens: I think the distinction is between "created via Web IDL's [new](https://webidl.spec.whatwg.org/#new)" and "created via a [constructor operation](https://webidl.spec.whatwg.org/#idl-constructors)". As far as I know, specifications usually do not call constructor operations themselves, those are only called by JavaScript code. For example, ["create a `Response` object"](https://fetch.spec.whatwg.org/#response-create) creates a new `Headers` object by setting its fields directly, rather than going through the `new Headers()` constructor.

If you have a better suggestion for making this less confusion, I'd love to hear it. I'm bad at naming things. 😅

